PR #11, opened 2025-11-24
Handle multiple nvm bin paths

mcvasquez:
**Summary**
CodexBar couldn’t find the codex binary when it was installed via nvm because PATH seeding added a literal `~/.nvm/versions/node/*/bin` segment, which never globbed to real versioned `bin` folders. As a result, the RPC probe failed with "Codex RPC probe failed: Malformed response: codex app-server closed stdout" when the CLI wasn’t on PATH.

**What changed**
- `Sources/CodexBar/UsageFetcher.swift`: Enumerate real `~/.nvm/versions/node/*/bin` directories when seeding PATH for RPC and TTY probes, replacing the non-expanding wildcard entry

**Before**

**After**

**Testing**
- swift build
- swift test


steipete:
@mcvasquez Thank you! I wonder if my approach to collecting path's here is right at all, since that likely won't be the last edge case. Let me take a closer look.

PR #14, opened 2025-12-07
[DRAFT] Add status view

antons:
Spent a few hours with Codex this morning setting this up. I haven’t reviewed it in detail yet, so marking as draft.

Claude
Codex


steipete:
I was considering adding a separate view but I'm not 100% convinced - kinda like the current minimalism?

steipete:
Took a closer look and I'm a fan, thank you! Will take over and get this shipped!

steipete:
Also  SwiftUI card used inside the NSMenu to mirror Apple's rich menu panels is clever. I wasn't aware that this is so easy!

PR #21, opened 2025-12-11
Add mise binary resolution

philipp-spiess:
Figured I finally look into why this thing is not working for me:

CleanShot 2025-12-11 at 12 21 08@2x

This PR adds support for managing node via mise:

CleanShot 2025-12-11 at 12 47 27@2x

## Summary

 - Add mise/rtx support to binary resolution so Codex CLI installed via mise is detected (checks shims and installs by newest semver).

## Testing

 - swift test
 - ./Scripts/compile_and_run.sh (with dev signing identity) — app builds, packages, launches, stays running.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ## 0.7.2 — Unreleased
 - Claude “Open Dashboard” now routes subscription accounts (Max/Pro/Ultra/Team) to the usage page instead of the API console billing page. Thanks @auroraflux!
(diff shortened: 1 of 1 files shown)


philipp-spiess:
> Gosh, recreating all of this logic seems so wrong, but I can’t find a way around it. Will merge! Hopefully we slowly have our pokemon-collect-them-all :)

I haven't tested this but _I think_ starting the command via a new login shell should be better since it includes the bash profiles. Something like: `bash -lc codex` (which IIRC is what Codex does too).

Edit: oh well there already is a login shell this is confusing then 🤔

PR #30, opened 2025-12-18
feat: consolidate status items into single unified icon

ibehnam:
## Summary

This PR consolidates multiple status items into a single unified icon when both OpenAI and Claude providers are enabled.

## Improvements

- **Reduced clutter**: Single status bar icon instead of multiple icons
- **Simplified UI**: One menu showing cards for all enabled providers with separators
- **New combined icon style**: When both providers enabled, displays combined status
- **Backward compatible**: Existing behavior preserved when only one provider enabled

## Technical Changes

- Changed from multiple `NSStatusItem` instances to single shared instance
- Added `combined` icon style for unified display
- Refactored menu generation to show cards for each active provider
- Status indicator now shows from any provider with issues
- Maintains all existing animations and display behaviors

## Testing

Verify behavior with:
- Single provider (OpenAI or Claude) - should behave identically to before
- Both providers enabled - shows single icon with combined menu
- Provider failures - status indicators work correctly
- Animation states - all existing animation patterns supported

Note: This is based on changes in the fork ibehnam/_CodexBar

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.10.0 — Unreleased
 - Gemini provider support (usage, plan detection, login flow). Thanks @381181295!
+- Unified menu bar icon mode with a provider switcher and Merge Icons toggle (default on when multiple providers are enabled). Thanks @ibehnam!
 
 ## 0.9.1 — 2025-12-19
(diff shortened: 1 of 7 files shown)


steipete:
Hmm - not sure I want this, plus the menu gained quite a few more features so it would be fairly long in the latest version.
If so it would have to be an option and we likely could do more clever things with the menubar in that mode.

PR #54, opened 2025-12-25
Bugfix/padding tabs

Blueemi:
Made it so the tabs in tabs in the menu bar pop-up now take the whole space, no matter if there's 1/2/3/4 of them, it dynamically changes now for better aesthetics.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,10 +55,17 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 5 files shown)


steipete:
Ah dang. I've spent all day reworking that. Thanks for the effort! We can prevent conflicts like this if you reach out via Twitter, I'm always on there.

steipete:
Sounds great, just @ me on X!

I'm finishing polish today and will craft a release.

PR #60, opened 2025-12-26
fix(antigravity): add -a flag to lsof for correct PID filtering

shaw-baobao:
Without the -a (AND) flag, lsof treats -iTCP and -p as OR conditions, returning listening ports from ALL processes instead of only the specified PID. This caused Antigravity port detection to potentially connect to wrong endpoints.

The fix adds -a between -sTCP:LISTEN and -p to properly AND the conditions, ensuring only the Antigravity language server's ports are returned.


steipete:
thank you!

PR #70, opened 2025-12-27
fix: prioritize OAuth for Claude usage detection

ibehnam:
## Summary

Fixes Claude usage detection failing with "Could not parse Claude usage: Missing Current session" error when OAuth credentials are available but the debug menu is disabled.

## Problem

The current implementation has a problematic fallback chain:
1. If `debugMenuEnabled` is false (default), it ignores the `claudeUsageDataSource` setting
2. Falls back to `.web` (requires browser cookies) or `.cli` (runs `claude /status`)
3. The `.cli` probe hangs in non-interactive/PTY contexts (e.g., ad-hoc signed builds, CI/CD)
4. OAuth mode only works when debug menu is explicitly enabled

This particularly affects:
- Users building with ad-hoc signing (no Apple Developer account)
- Automated builds or non-interactive environments  
- Anyone with OAuth credentials but without browser cookies

## Solution

Improved auto-detection logic that prioritizes OAuth:

1. **OAuth first** - Check for credentials in keychain or `~/.claude/.credentials.json`
2. **Web fallback** - Use browser cookies if OAuth unavailable
3. **CLI last resort** - Only use CLI probe when no other options exist

## Changes

- Modified `ClaudeProviderImplementation.usageStrategy()` to check for OAuth credentials before falling back to CLI
- OAuth now works without requiring debug menu to be enabled
- Added clear comments explaining the fallback chain
- No breaking changes - existing behavior preserved when debug menu is enabled

## Testing

- ✅ Compiles successfully
- ✅ Built and tested with ad-hoc signing
- ✅ OAuth detection works with credentials file
- ✅ No regression for web/CLI modes

## Additional Context

For users with ad-hoc signed builds who can't access keychain entries, they can export credentials:
```bash
security find-generic-password -s "Claude Code-credentials" -w > ~/.claude/.credentials.json
```

### DIFF
--- a/Sources/CodexBar/Providers/Claude/ClaudeProviderImplementation.swift
+++ b/Sources/CodexBar/Providers/Claude/ClaudeProviderImplementation.swift
@@ -46,6 +46,7 @@ struct ClaudeProviderImplementation: ProviderImplementation {
         settings: SettingsStore,
         hasWebSession: () -> Bool = { ClaudeWebAPIFetcher.hasSessionKey() }) -> ClaudeUsageStrategy
     {
+        // When debug menu is enabled, respect the explicit user preference
         if settings.debugMenuEnabled {
             let dataSource = settings.claudeUsageDataSource
             if dataSource == .oauth {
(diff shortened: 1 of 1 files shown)


steipete:
Thanks @ibehnam! This is fixed on main now (81e120d), docs updated (657c2ef), and you’re credited in the changelog (ba6d9bb). I’m closing the PR since it’s been merged via direct commits.

PR #74, opened 2025-12-27
feat: implement Copilot provider integration with token storage and login flow

roshan-c:
This PR adds full support for tracking GitHub Copilot usage statistics in CodexBar. It implements a secure authentication flow and fetches quota data from the internal GitHub Copilot API.

### DIFF
--- a/Sources/CodexBar/CopilotTokenStore.swift
+++ b/Sources/CodexBar/CopilotTokenStore.swift
@@ -1,0 +1,110 @@
+import CodexBarCore
+import Foundation
+import Security
+
+protocol CopilotTokenStoring: Sendable {
+    func loadToken() throws -> String?
+    func storeToken(_ token: String?) throws
+}
+
+enum CopilotTokenStoreError: LocalizedError {
+    case keychainStatus(OSStatus)
+    case invalidData
+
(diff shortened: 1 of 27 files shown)


steipete:
codex analysis:

• Yes — not closing is expected with current flow; we have two real bugs:

  - waitingAlert.runModal() blocks, and we never close the alert window; NSApp.stopModal alone
    doesn’t reliably dismiss the NSAlert.
  - We start two polling tasks; cancel only stops the unused one. The real poll keeps running
    (or failing), so the modal never ends.

  Likely also failing to ever get a token: the device‑flow endpoints expect form‑urlencoded
  bodies, but we send JSON. That can leave polling stuck even though the web page says
  success.

  Fix plan (small, safe):

  1. CopilotDeviceFlow: send application/x-www-form-urlencoded bodies for both device code +
     token. Keep Accept: application/json.
  2. CopilotLoginFlow: single polling task; on success call NSApp.stopModal and
     waitingAlert.window.close() (or switch to beginSheetModal).
  3. Cancel should cancel the same task.

steipete:
Thanks @roshan-c! We merged this into main and added a credit in the CHANGELOG.

PR #85, opened 2025-12-29
fix: Claude token dedup and OAuth scope handling

nicosuave:
## Summary

- Deduplicate streaming chunks in Claude token counting by tracking seen `message.id + requestId` pairs (fixes 2-3x overcounting)
- Check for `user:profile` scope before attempting OAuth usage API
- Fall back to web strategy when OAuth scope is missing (instead of showing error)
- Add tests for scope-based strategy resolution and deduplication behavior

## Details

**Token deduplication**: Claude emits multiple JSONL lines per message during streaming, each with the same cumulative usage values. Without deduplication, tokens were being counted 2-3x or more. Now tracks seen combinations and skips duplicates, matching ccusage's approach.

**OAuth scope handling**: The `/api/oauth/usage` endpoint requires `user:profile` scope, but tokens from `claude` CLI only have `user:inference`. Previously this showed "OAuth request unauthorized" error. Now it checks scopes upfront and falls back to web strategy gracefully.

**Test coverage**: Added tests for:
- Streaming chunk deduplication (3 chunks with same IDs → counted once)
- Backwards compatibility (entries without IDs still counted separately)
- Scope-based strategy resolution

### DIFF
--- a/Tests/CodexBarTests/CostUsageScannerTests.swift
+++ b/Tests/CodexBarTests/CostUsageScannerTests.swift
@@ -352,6 +352,156 @@ struct CostUsageScannerTests {
             #expect(fast == expected)
         }
     }
+
+    @Test
+    func claudeDeduplicatesStreamingChunks() throws {
+        let env = try CostUsageTestEnvironment()
+        defer { env.cleanup() }
+
+        let day = try env.makeLocalNoon(year: 2025, month: 12, day: 20)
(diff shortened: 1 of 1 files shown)


steipete:
Fixed on main. Local cost usage now matches ccusage (streaming dedupe, cache tokens, CLI JSON). Thanks for the patch!

PR #93, opened 2025-12-29
Universal binary support (Intel + Apple Silicon)

ValentineDD:
Makes universal binary (Intel + Apple Silicon) the default for release builds.

The creator mentioned they couldn't test on Intel, so this change defaults the release build to create universal binaries that work on both architectures. This ensures Intel Mac users can use CodexBar without needing to manually specify architectures.

## Changes
- Updated `Scripts/sign-and-notarize.sh` to default to `arm64 x86_64` instead of `arm64` only
- Updated `docs/RELEASING.md` to document universal binary support
- Updated `AGENTS.md` release flow notes
- Added `test_universal.sh` script to verify universal binary builds (can be run on Apple Silicon to verify both architectures are present without needing Intel hardware)

## Testing
- Built universal binary on Intel Mac (x86_64)
- Verified both architectures present: `lipo -archs CodexBar.app/Contents/MacOS/CodexBar` shows `arm64 x86_64`
- Tested app execution on Intel Mac
- Test script can be run on Apple Silicon to verify build without needing Intel hardware

## Commands Run
ARCHES="arm64 x86_64" ./Scripts/package_app.sh release
lipo -archs CodexBar.app/Contents/MacOS/CodexBar
./test_universal.sh


steipete:
Thank you! When I said I can't test it I meant executing and clicking - the script won't help here. The arch is already there for Intel - I'll flip the switch and hope it works out.

PR #96, opened 2025-12-29
perf(openai-web): reduce idle CPU by suspending cached WebViews

douglascamata:
## The issue

I was experiencing high CPU usage, not sure exactly why, by WebView that loads the OpenAI dashboard since I upgrade to v0.16.0. 

I tinkered a bit with LLMs and was able to get to a solution. Note that I have 0 (zero) experience with macOS apps. This is 100% LLM generated. 

I was able to compile and run the app and I can confirm that I do not experience that high long-lasting high CPU usage anymore. It only spikes for a short while, during the parsing, and then the CPU takes a nap. It's like 100x better.

## The solution (LLM generated)

The offscreen WKWebView host window used alphaValue=0.001 to prevent WebKit from throttling during scraping. However, after scraping completed, the window remained 'visible' to WebKit, allowing the ChatGPT SPA to continue running JavaScript timers and animations at full power.

Changes:
- Set alphaValue=0.0 in hide() so WebKit recognizes the page as inactive
- Restore alphaValue in show() to enable hydration during active scraping
- Add inactiveSchedulingPolicy=.suspend on macOS 14+ for explicit control

This allows WebKit to fully suspend JavaScript execution when the WebView is idle, reducing CPU usage from 10-30% to near zero between refreshes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 0.16.1 — Unreleased
 - Menu: reduce layout thrash when opening menus and sizing charts. Thanks @ibehnam!
 - Packaging: default release notarization builds universal (arm64 + x86_64) zip.
-- OpenAI web: reduce idle CPU by suspending cached WebViews when not scraping.
+- OpenAI web: reduce idle CPU by suspending cached WebViews when not scraping. Thanks @douglascamata!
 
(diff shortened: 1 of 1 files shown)


steipete:
Thanks for the CPU fix, @douglascamata — much appreciated!

PR #105, opened 2025-12-30
fix: update website link in About panel

felipeorlando:
A simple one, cause I noticed when I was about to share it with my friends :)


steipete:
Thanks  Felipe! Every little helps!

PR #131, opened 2026-01-05
Optimize startup performance and eliminate UI lag

ibehnam:
## Summary
This PR addresses UI lag and responsiveness issues during app startup by moving several blocking operations from the main thread to background tasks.

### Key Changes
- **Async Binary Resolution**: `UsageStore` now resolves binary paths (which involves spawning interactive shells) in a detached background task.
- **Background Keychain Migration**: Moved `KeychainMigration.migrateIfNeeded()` to an async task in `HiddenWindowView` to prevent blocking the app's `init`.
- **Lazy Status Items**: Refactored `StatusItemController` to create `NSStatusItem`s only when needed (lazily), reducing IPC overhead with the system status bar during launch.
- **Thread Safety**: Ensured UI updates from background tasks are properly dispatched to the `@MainActor`.

## Test plan
- Verified that the app launches instantly and the menu bar icons appear without freezing the main thread.
- Confirmed that binary resolution and version detection still work as expected in the background.
- Ran `./Scripts/compile_and_run.sh` with `CODEXBAR_SIGNING=adhoc` to verify local execution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@
 - Dev: move Chromium profile discovery into SweetCookieKit (adds Helium net.imput.helium). Thanks @hhushhas!
 - Dev: bump SweetCookieKit to 0.2.0.
 - Dev: migrate stored Keychain items to reduce rebuild prompts.
+- Dev: move path debug snapshot off the main thread to avoid startup hitches (#131). Thanks @ibehnam!
 - Tests: expand Kiro CLI coverage.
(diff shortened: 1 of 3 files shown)


steipete:
Landed via temp rebase onto main.\n\n- Gate: pnpm lint && pnpm build && pnpm test\n- Land commit: 3b5628d1f6be11e8e43b00559835a3b895e12ec1\n- Merge commit: 45d6997ff844cac0a3c30fcc31abc09b0828d89c\n\nThanks @ibehnam!

PR #135, opened 2026-01-05
fix(parsing): handle European decimal separator in number parsing

rvben:
## Summary

- Fix parsing of European-format numbers where comma is the decimal separator
- Numbers like `20,00` were incorrectly parsed as `2000.00`, breaking usage display for European locale users
- Prevents hearts skipping a beat by showing that 10 euro's were spent instead of 1000.

Untitled

## Changes

- Add `parseLocalizedNumber()` function with heuristic-based format detection:
  - Both separators present → last one is the decimal separator
  - Only comma with ≤2 digits after → European decimal (`20,00` → `20.0`)
  - Only comma with 3+ digits after → US thousands (`1,000` → `1000`)
- Update `firstNumber()` and `parseCreditsUsed()` to use the new function
- Add 13 comprehensive tests covering US, European, and mixed formats

## Test plan

- [x] All 358 existing tests pass
- [x] New tests cover: US decimal, European decimal, US thousands, European thousands with decimal, plain integers, whitespace handling, empty/invalid inputs
- [x] Regression tests verify `"Credits remaining: 20,00"` parses as `20.0` (not `2000.0`)

### DIFF
--- a/Sources/CodexBarCore/OpenAIWeb/OpenAIDashboardParser.swift
+++ b/Sources/CodexBarCore/OpenAIWeb/OpenAIDashboardParser.swift
@@ -140,10 +140,9 @@ public enum OpenAIDashboardParser {
 
     private static func parseCreditsUsed(_ text: String) -> Double {
         let cleaned = text
-            .replacingOccurrences(of: ",", with: "")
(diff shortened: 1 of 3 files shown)


rvben:
I see https://github.com/steipete/CodexBar/pull/136 got merged and handles the same issue.

Issue #150, opened 2026-01-08
Status bar icon issue


bestwnh:
When I turn on `Check provider status` and `Merge Icons`, A white circle always show in the status menu.
This is the screenshot when turn off provider status.
<img width="86" height="64" alt="Image" src="https://github.com/user-attachments/assets/b7b3689e-e8ca-44d8-87ad-b7f9a6da60c5" />
This is the turn on screenshot
<img width="70" height="54" alt="Image" src="https://github.com/user-attachments/assets/06851a3c-9cfc-4536-a2d6-fd1734888729" />

ngutman:
Thanks for the report. I’m closing this as fixed by #291.

If you still see the merged-icons status overlay issue on the latest release, please open a fresh issue with a current screenshot and version details.

PR #152, opened 2026-01-08
fix: auto pick Claude organization by capabilities

Leechael:
When logging in with multiple Claude accounts, if the first one only has API capability, the Claude Web API Fetcher returns a 403 error.

A better approach would be using an allowlist, but I can't get the full list of available values. So the approach is to skip records with `api` capability.

### DIFF
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeWeb/ClaudeWebAPIFetcher.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeWeb/ClaudeWebAPIFetcher.swift
@@ -567,16 +567,35 @@ public enum ClaudeWebAPIFetcher {
     private struct OrganizationResponse: Decodable {
         let uuid: String
         let name: String?
+        let capabilities: [String]?
+
+        var normalizedCapabilities: Set<String> {
+            Set((self.capabilities ?? []).map { $0.lowercased() })
+        }
+
+        var hasChatCapability: Bool {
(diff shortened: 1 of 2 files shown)


Finesssee:
Thanks for splitting this down from #150. This is much closer, but I still would not merge it yet.

Blockers:

1. `rust/src/cost_scanner.rs` still crosses the 1k-line threshold in this PR (`936 -> 1026`). `codex_sessions.rs` helps, but the new Codex date/range/day-cost helpers and their tests should move out too so `cost_scanner.rs` does not keep becoming the catch-all parser/cache module.

2. The TrayPanel refresh hook is dead/magic wiring. It listens for `flyout-opened`, but this PR does not emit that event anywhere in production; the test fakes an event production never sends. Please either remove this change from this PR or wire it to the real canonical tray-open event.

3. The FloatBar cost UI adds hardcoded user-facing text (`Today`, `30d`, tooltip text). This app is on Fluent locales now, so please route those strings through the existing locale flow before merging.

4. The FloatBar cost-fetch effect depends on `visible`, which is rebuilt as objects and can refetch even when the actual provider key did not change. Please depend on a stable key / memoized target list so this does not create avoidable polling churn.

Good progress: PowerToys/topmost are gone, WSL discovery is split out, and the missing `#[test]` issue is fixed. The remaining ask is to keep this PR from turning `cost_scanner.rs` and FloatBar into the next pile-on points.

PR #159, opened 2026-01-09
Add feature to show highest-usage provider in menu bar

phillco:
Rather than having to pick which service shows in the menubar (if you don't have space to show them all), I thought it would be nice if you could have a feature to just always display the one that's closest to its rate limit.

nb: Claude wrote this


phillco:
I'm going to keep tinkering with this locally, so turning back into a draft. (I didn't expect Claude to submit it already. 😅)

steipete:
Landed via temp rebase onto main.\n\n- Gate: pnpm lint && pnpm build && pnpm test\n- Land commit: de1d6cbf83d259bd7924d7bff01d049636afbec1\n- Merge commit: bc64a6e06a578d7564d4d52ca88ef1410343e6c8\n\nThanks @phillco!

PR #172, opened 2026-01-11
Fix Gemini usage fetching by aligning project discovery with official CLI

lolwierd:
## WHAT
  Refactored the Gemini provider’s initialization and usage fetching logic in GeminiStatusProbe.swift to strictly follow the sequence used by the official Gemini CLI:
   * Sequential API Calls: The probe now calls the loadCodeAssist endpoint before calling retrieveUserQuota.
   * Authoritative Project ID: It extracts the cloudaicompanionProject field from the API response and uses it as the primary project ID for all quota requests.
   * Unified Tier Detection: Plan detection now uses the currentTier data returned directly from the Code Assist API.
   * Robust Fallback: Keeps the original cloud resource manager scanning logic as a secondary fallback.

  ## WHY
  The previous implementation was unreliable because it attempted to "guess" the active Project ID by scanning the user's own Google Cloud project list for specific prefixes.
   * Free Tier Support: Users on the Gemini Free Tier often use a Google-managed project (cloudaicompanionProject) that does not appear in their personal project list, causing the old discovery logic to fail.
   * CLI Parity: This change aligns CodexBar's logic with the official @google/gemini-cli implementation, ensuring that if the CLI can see usage data, CodexBar will as well.

  References:
  This implementation mirrors the user setup and server initialization flow found in the official `@google/gemini-cli-core` package (specifically the setupUser and retrieveUserQuota logic).

  ---
  Generated by Gemini CLI

### DIFF
--- a/Scripts/install.sh
+++ b/Scripts/install.sh
@@ -41,10 +41,16 @@ else
     
     NEEDS_SUDO=false
     
-    if [[ ! -w "$TARGET_DIR" && ! -d "$TARGET_DIR" ]]; then
-        NEEDS_SUDO=true
-    elif [[ ! -w "$TARGET_DIR" ]]; then
-        NEEDS_SUDO=true
+    if [[ -d "$TARGET_DIR" ]]; then
+        if [[ ! -w "$TARGET_DIR" ]]; then
+            NEEDS_SUDO=true
+        fi
+    else
+        # If directory doesn't exist, check if we can create it (parent writable?)
+        PARENT_DIR="$(dirname "$TARGET_DIR")"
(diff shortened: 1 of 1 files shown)


lolwierd:
@codex pls review

steipete:
Landed via temp rebase onto main.\n\n- Gate: pnpm lint && pnpm build && pnpm test\n- Land commit: e4db9fe7b40669f7aab3d1f1d4ab420fbd498b1d\n- Merge commit: 36cf453a421eb05db959e9433e2b526af705aa47\n\nThanks @lolwierd!

PR #204, opened 2026-01-19
fix: Remove libxml2 dependency for Linux CLI builds

oorestisime:
## Problem

The Linux CLI binary fails to run on distros with libxml2 2.13+ (Arch Linux, Fedora 40+, openSUSE Tumbleweed):

```
codexbar: error while loading shared libraries: libxml2.so.2: cannot open shared object file
```

This happens because the binary is linked against `libxml2.so.2`, but newer rolling-release distros ship `libxml2.so.16` (the soname changed in libxml2 2.13).

### When did this start?

The issue started with **v0.18.0-beta.1** when the JetBrains AI provider was added. The JetBrains provider uses `XMLParser` for parsing quota files, which depends on libxml2 on Linux. Previous versions (v0.17.0 and earlier) worked fine because they didn't have this dependency.

### Musl static build
We explored building with Swift's Static Linux SDK (musl) to produce a fully static binary with no shared library dependencies. However this approach had several blockers:

- Requires code changes to support `import Musl` alongside `import Glibc`
- Foundation on musl still requires `libsqlite3.a` which isn't bundled in the Static SDK
- Arch Linux doesn't ship static libraries by default, making local testing difficult
- Would require CI changes to install the Static SDK and additional static libraries

### Building on older Ubuntu
Building on Ubuntu 22.04 instead of 24.04 would help some distros but wouldn't solve the issue for Arch/Fedora 40+ since they've moved to libxml2 2.13+ regardless.

## Solution

Replace the `XMLParser`-based implementation in `JetBrainsStatusProbe.swift` with a simple regex-based parser on Linux. This eliminates the libxml2 dependency entirely.

The XML structure being parsed is simple - just extracting two attribute values from a known element structure - so regex is sufficient and avoids the external dependency.

## Changes

- **JetBrainsStatusProbe.swift**: Replace `XMLParser` delegate class with regex-based `JetBrainsXMLParser` enum (Linux only, macOS still uses `XMLDocument`)
- **CLIIO.swift**: Add missing Glibc import for `platformExit()`

## Testing

- Built and tested on Arch Linux (libxml2 2.13+)
- `.build/release/CodexBarCLI usage --provider codex --source oauth --format json`
- `.build/release/CodexBarCLI usage --provider claude --source oauth --format json`
- `KIMI_K2_API_KEY="your-key" .build/release/CodexBarCLI usage --provider kimik2 --source api --format json`

### DIFF
--- a/TestsLinux/JetBrainsParserLinuxTests.swift
+++ b/TestsLinux/JetBrainsParserLinuxTests.swift
@@ -1,0 +1,190 @@
+import CodexBarCore
+import Foundation
+import Testing
+
+/// Tests for the regex-based JetBrains XML parser used on Linux.
+/// These tests verify that the non-libxml2 implementation correctly parses
+/// JetBrains AI Assistant quota files.
+@Suite
+struct JetBrainsParserLinuxTests {
+    @Test
+    func parsesQuotaXMLWithBothOptions() throws {
+        let quotaInfo = [
+            "{&#10;  &quot;type&quot;: &quot;Available&quot;,",
+            "&#10;  &quot;current&quot;: &quot;7478.3&quot;,",
+            "&#10;  &quot;maximum&quot;: &quot;1000000&quot;,",
+            "&#10;  &quot;until&quot;: &quot;2026-11-09T21:00:00Z&quot;,",
+            "&#10;  &quot;tariffQuota&quot;: {",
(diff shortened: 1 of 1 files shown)


steipete:
God damn. Linux doesn't make things simple. Not a fan but also don't have a better idea. For what we use, this is acceotable. Thanks for digging into this, Orestis!

PR #210, opened 2026-01-19
fix: use classic icon on macOS versions before 26

### LINKED ISSUE #178: [bug] Shouldn't use Liquid Glass icon on mac with version before 26
I stayed on macOS 15 instead of upgrading to macOS 26 because I really don’t want to see any Liquid Glass elements.
So please don’t use the Liquid Glass app icon on macOS versions before 26.
Alternatively, could you provide the previous icon (.icns) file so I can replace it locally?
Thanks!

zerone0x:
## Summary
Fixes #178

On macOS 15 and earlier, the app now displays the classic (non-Liquid Glass) app icon instead of the new Liquid Glass style icon. This addresses user preference to avoid seeing macOS 26 visual elements on older systems.

## Changes
- Added `Icon-classic.icns` (extracted from initial commit) to Resources
- Added `configureAppIconForMacOSVersion()` method in AppDelegate
- At app launch, checks macOS version and swaps to classic icon if running on macOS < 26

## Test plan
- [x] Build compiles without errors related to these changes
- [ ] On macOS 15 or earlier: verify classic icon appears in Dock and About window
- [ ] On macOS 26: verify Liquid Glass icon appears (no change from current behavior)

---

### DIFF
--- a/Sources/CodexBar/CodexbarApp.swift
+++ b/Sources/CodexBar/CodexbarApp.swift
@@ -261,14 +261,21 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
     /// Use the classic (non-Liquid Glass) app icon on macOS versions before 26.
     private func configureAppIconForMacOSVersion() {
-        let osVersion = ProcessInfo.processInfo.operatingSystemVersion
-        guard osVersion.majorVersion < 26 else { return }
-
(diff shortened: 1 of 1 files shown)


steipete:
Better way:

  - Use resource bundle (Bundle.module or shared resourceBundle) not Bundle.main for Icon-classic.icns.
  - Use #unavailable(macOS 26) instead of manual major version compare.

PR #226, opened 2026-01-21
Fix Factory usage percentage by preferring API usedRatio

edlsh:
**Problem**: Factory usage shows incorrect percentages when `totalAllowance` is missing/zero/sentinel in the API response, causing 0% used to display regardless of actual usage.

**Fix**: Use API-provided `usedRatio` field when valid (0-1 range), falling back to `used/allowance` calculation otherwise.

**Changes**:
- Add `standardUsedRatio`/`premiumUsedRatio` to `FactoryStatusSnapshot`
- Pass through from API in `buildSnapshot`  
- Prefer API ratio in `calculateUsagePercent` with floating-point tolerance

**Tests**: 8 Factory tests pass (added 4 new)


steipete:
Superseded by merged changes on main.

steipete:
Thanks @edlsh! These changes are now merged on main (via direct commits that incorporate your work), so I’m closing the PR as superseded. Appreciate the contribution.

PR #233, opened 2026-01-23
fix: use tier-aware effective budget for Cursor usage percentage

### LINKED ISSUE #154: Incorrect usage count for Cursor
The widget shows incorrect usage for cursor. Here is shows 84% usage:

Image

while in reality it is much lower:

Image

The issue could be because I am on old plan of cursor where we used to get requests not rate limits.

KalnuErelis:
## Summary
- Fixes usage percentage display for Pro+/Ultra Cursor plans
- Higher-tier plans now show percentage against effective capacity, not nominal limit

## Problem
Cursor Pro+/Ultra users saw misleading usage percentages. For example, an Ultra user ($200/mo) with $300 usage would see the Plan bar at 0% left, even though Ultra provides ~$800 effective capacity.

Reference: https://x.com/steipete/status/2014620261715226833

## Solution
Introduced plan-tier-aware effective budgets:

| Plan | Nominal | Effective Budget |
|------|---------|-----------------|
| Pro | $20/mo | $40 |
| Pro+ | $60/mo | $120 (3x) |
| Ultra | $200/mo | $800 (20x) |

The Plan progress bar now calculates percentage as `totalUsed / (tierEffectiveBudget + onDemandLimit)`.

## Changes
- Add `CursorPlanTier` enum with effective budget values
- Add `CursorEffectiveUsage` struct for tier-aware usage data
- Update `CursorStatusSnapshot` with plan tier detection and effective percentage calculation
- Update progress bar to use effective percentage for non-legacy plans
- Add comprehensive tests

## Testing
- All 24 `CursorStatusProbeTests` pass
- Manual verification with Ultra account shows correct percentage

Fixes #154

### DIFF
--- a/Sources/CodexBar/Providers/Cursor/CursorProviderImplementation.swift
+++ b/Sources/CodexBar/Providers/Cursor/CursorProviderImplementation.swift
@@ -90,13 +90,18 @@ struct CursorProviderImplementation: ProviderImplementation {
 
     @MainActor
     func appendUsageMenuEntries(context: ProviderMenuUsageContext, entries: inout [ProviderMenuEntry]) {
+        // Show on-demand usage separately
         guard let cost = context.snapshot?.providerCost, cost.currencyCode != "Quota" else { return }
-        let used = UsageFormatter.currencyString(cost.used, currencyCode: cost.currencyCode)
-        if cost.limit > 0 {
-            let limitStr = UsageFormatter.currencyString(cost.limit, currencyCode: cost.currencyCode)
-            entries.append(.text("On-Demand: \(used) / \(limitStr)", .primary))
-        } else {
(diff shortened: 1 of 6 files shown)


ratulsarna:
Thanks for the PR and the tests! We merged #240, which fixes the specific `% used` issue caused by treating `breakdown.total` as the limit. This PR goes further by introducing tier-based ‘effective budgets’ and using those to compute the plan percentage.

Before we merge a change like that, could you share a couple of redacted `/api/usage-summary` examples (and what Cursor shows in their UI for the same account) where the current behavior after #240 still doesn’t match what users expect to see? That’ll help us validate the calculation across plans and avoid relying on assumptions that might vary by account.

PR #248, opened 2026-01-26
Add Qwen Code provider

### LINKED ISSUE #247: Cannot find Qwen in Providers list
Does it support Qwen CLI?

brenorb:
## Summary
- Add Qwen Code provider (JSONL usage parsing, settings, registration).
- Integrate across core/app/widget; add icon + docs.
- Add tests for JSONL usage aggregation.

## Verification
- macOS: Qwen Code shows up in Providers and works locally.
- Vibe-coded from Codex.

## Testing
- swiftformat Sources Tests
- swiftlint --strict
- pnpm check

Closes #247

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # CodexBar 🎚️ - May your tokens never run out.
 
-Tiny macOS 14+ menu bar app that keeps your Codex, Claude, Cursor, Gemini, Antigravity, Droid (Factory), Copilot, z.ai, Kiro, Vertex AI, Augment, Amp, and JetBrains AI limits visible (session + weekly where available) and shows when each window resets. One status item per provider (or Merge Icons mode); enable what you use from Settings. No Dock icon, minimal UI, dynamic bar icons in the menu bar.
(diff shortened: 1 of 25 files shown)


ratulsarna (changes requested):
Thanks for adding Qwen support - the local log parsing itself looks reasonable.

Requesting changes for one core correctness issue: we currently present Qwen as if quota/reset semantics are authoritative ("% left" + "Resets in ..."), but we do not have provider-confirmed window/reset metadata for Qwen OAuth free tier. Public docs expose quota amounts, but not reset timezone/window semantics.

That means current behavior can be materially wrong (both reset countdown and percent-left math), depending on backend window definition.

Please adjust this PR to avoid false precision:
1. Remove/hide "Resets in ..." for Qwen, or explicitly mark it as unknown.
2. Remove/hide quota percent-left for Qwen (or clearly label as estimate).
3. Add an explicit docs/UI note that Qwen values are inferred from local logs and may not match backend quota windows.

Once representation is honest about uncertainty, I'm comfortable with keeping Qwen provider support.

surajmandalcell:
Can't find qwen code in providers list. Did ya all abandon it?

PR #255, opened 2026-01-27
Add Antigravity OAuth support with manual token and local import

validatedev:
## Summary

This pull request adds support for managing Antigravity provider accounts, including OAuth and manual token entry, in the Providers preferences pane. It introduces new UI options for adding, importing, and managing Antigravity credentials, updates the provider descriptor logic, and refines the token account entry experience for providers that require multiple fields. Additionally, it includes dependency and CI updates to support these features.

**Antigravity Provider Account Management:**

* Added support for Antigravity accounts in the Providers preferences pane, including segmented manual entry (access token and optional refresh token), Google OAuth sign-in, and importing credentials from the Antigravity app database. The UI dynamically adapts based on provider and system configuration. [[1]](diffhunk://#diff-beaef700ee049c6561ca976ee3e28f467e63fe2b64180e65c0d66daa4a5f2ef2R167-R196) [[2]](diffhunk://#diff-beaef700ee049c6561ca976ee3e28f467e63fe2b64180e65c0d66daa4a5f2ef2L190-R233) [[3]](diffhunk://#diff-9d17991881f281522bbae36dedd0604991624faf3961491769d4a433c2527d17R247-R280) [[4]](diffhunk://#diff-9d17991881f281522bbae36dedd0604991624faf3961491769d4a433c2527d17L254-R326)
* Implemented error handling and user feedback for credential import, including alerts for missing credentials, database not found, and full disk access requirements.

**UI/UX Improvements:**

* Refined the token account entry row to support two-field entry for providers like Antigravity, and adjusted the layout and controls for better usability. [[1]](diffhunk://#diff-9d17991881f281522bbae36dedd0604991624faf3961491769d4a433c2527d17R247-R280) [[2]](diffhunk://#diff-9d17991881f281522bbae36dedd0604991624faf3961491769d4a433c2527d17L254-R326)
* Added "Antigravity" to the provider picker in the debug pane for log viewing.

**Dependency and Build Updates:**

* Added `swift-protobuf` as a package dependency and linked it to the main target. [[1]](diffhunk://#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677eR25) [[2]](diffhunk://#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677eR36-R38)
* Updated CI workflow to install SQLite3, required for reading the Antigravity database.

**Security and Configuration:**

* Added a `.gitguardian.yml` configuration to ignore known public Antigravity OAuth credentials, preventing false positives in secret scanning.

## Screenshot

CleanShot 2026-01-29 at 15 09 34@2x


validatedev:
Saw some edge case bug, fixing it. Until then it is draft

validatedev:
@steipete should i fix these linting thingies to pass ci? at least for my part of the pr or are you planning something else? it's currently ready to be merged

Sources/CodexBar/PreferencesProvidersPane.swift
```diff
@@ -166,11 +166,36 @@ struct ProvidersPane: View {
...
         guard let support = TokenAccountSupportCatalog.support(for: provider) else { return nil }
         let context = self.makeSettingsContext(provider: provider)
+
+        let isAntigravity = provider == .antigravity
+        let keychainEnabled = !KeychainAccessGate.isDisabled
+
+        let supportsTwoFieldEntry = isAntigravity
+        let supportsManualEntry = !isAntigravity || !keychainEnabled
+        let addActionTitle = isAntigravity ? "Sign in with Google" : nil
+        let addAction: (() async -> Void)? = isAntigravity
```
ratulsarna, line 358: Looks like manual token entry gets hidden whenever Keychain is enabled (`supportsManualEntry = !isAntigravity || !keychainEnabled`). That means most users won’t see the manual paste flow even though storage logic supports it. Can we keep manual entry visible and only switch storage backend internally?

Sources/CodexBarCore/Providers/ProviderFetchPlan.swift
```diff
@@ -29,6 +29,7 @@ public struct ProviderFetchContext: Sendable {
     public let fetcher: UsageFetcher
     public let claudeFetcher: any ClaudeUsageFetching
     public let browserDetection: BrowserDetection
+    public let onAntigravityCredentialsRefreshed: (@Sendable (String, AntigravityOAuthCredentials) -> Void)?
```
ratulsarna, line 32: Nice addition of this callback hook. One thing to double-check: it looks wired in token-account fetches, but not on the normal provider refresh path. In keychain-disabled mode that can leave refreshed Antigravity tokens/expiry unpersisted during regular refreshes. Could we make sure all app fetch contexts pass this consistently?

Sources/CodexBar/PreferencesProvidersPane.swift
```diff
@@ -212,6 +248,114 @@ struct ProvidersPane: View {
...
+            log.debug("Import failed with AntigravityOAuthCredentialsError: \(error)")
+            switch error {
+            case .notFound:
+                if AntigravityLocalImporter.isAvailable() {
+                    self.presentAlert(
+                        title: "No Credentials Found",
+                        message: """
+                        Antigravity database found, but no credentials were found inside.
+
+                        Please ensure:
```
ratulsarna, line 498: I think Full Disk Access failures may get swallowed by this typed catch before we reach the POSIX error branch below. Could we add a dedicated permission-denied case and route it directly to `presentFullDiskAccessAlert()`?

Sources/CodexBarCore/Providers/Antigravity/AntigravityOAuth/AntigravityOAuthFlow.swift
```diff
@@ -0,0 +1,318 @@
...
+        let params = [
+            "client_id": AntigravityOAuthConfig.clientID,
+            "client_secret": AntigravityOAuthConfig.clientSecret,
+            "code": code,
+            "redirect_uri": redirectUri,
+            "grant_type": "authorization_code",
+        ]
+
+        let body = params
+            .map { "\($0.key)=\($0.value.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed) ?? $0.value)" }
```
ratulsarna: We’re building an `application/x-www-form-urlencoded` body using `.urlQueryAllowed`. That can be unsafe for form payloads (`+`, `&`, `=` edge cases). Can we switch to a strict form encoder here (and in the token refresher path too)?

ratulsarna (commented):
Sharing focused follow-ups from deep review.

Sources/CodexBarCore/Providers/Antigravity/AntigravityOAuth/AntigravityAuthorizedFetchStrategy.swift
```diff
@@ -0,0 +1,182 @@
...
+            credentials = try await self.refreshAndSave(credentials, accountLabel: accountLabel, context: context)
+            return try await self.fetchQuotaAndMakeResult(credentials: credentials, sourceLabel: sourceLabel)
+        }
+    }
+
+    private func fetchQuotaAndMakeResult(
+        credentials: AntigravityOAuthCredentials,
+        sourceLabel: String) async throws -> ProviderFetchResult
+    {
+        let quota = try await AntigravityCloudCodeClient.fetchQuota(accessToken: credentials.accessToken)
```
ratulsarna: Should we add a fallback when `fetchAvailableModels` doesn’t return usable quota data? Upstream Code Assist flows usually bootstrap project info and can fall back to `retrieveUserQuota`, which would make this path more resilient.

validatedev:
@ratulsarna I’ve addressed all of your feedback, thank you. I noticed a regression while testing though, so I’ll fix that and then re-request a review asap. It's not currently mergeable, just fyi.

ratulsarna:

validatedev:
@ratulsarna sure! I want this to be landed. Let me do the things

validatedev:
@ratulsarna rebased + fixed conflicts + fixed a bug as a db schema has changed for local import. ready to be merged.

validatedev:
fyi @ratulsarna and @steipete, the first implementation of this pr (before google changed their db schema) has already been in https://github.com/robinebers/openusage, which i'm a maintainer of also. i'd assume it's fairly battle-tested and safe to merge in that sense.

ratulsarna:
Opened replacement PR #790

Thanks again @validatedev for the original Antigravity OAuth work. I kept your branch untouched and carried the audited implementation forward there.

ratulsarna:
@validatedev Ran into leaked antigravity creds when auditing/reviewing this PR. I don't think we should be exposing such secrets in the app especially since they can be revoked/rotated any time by Google and thus requiring a hotfix. what is your take on this?

validatedev:
@ratulsarna Thanks for the comment. You're right, I didn't want to include those secrets either, but I couldn't find a better way to handle it. That said, I'd appreciate it if you could keep this PR open while I look into a feasible alternative. I'm not sure one exists, but I'll update you here within 3 days 🤞

ratulsarna:
Great, PR is open. No issues.

steipete:
Thanks for digging into Antigravity OAuth. Main now has the maintained Antigravity OAuth-backed remote usage path via #635, including credential storage, refresh, and provider settings wiring. This branch is old/dirty and also carries dependency/workflow churn, so I’m closing it as superseded. If there is a specific remaining piece such as manual token import that is still useful, please reopen it as a focused PR on current main.

PR #261, opened 2026-01-28
Fix launch crash from missing SwiftPM resource bundle

ryanchen01:
## Summary
- Avoid SwiftPM-generated resource bundle accessor for CodexBar to prevent launch-time crash.
- Load provider icons from the main app bundle resources.

## Commands
- brew install swiftformat swiftlint
- swiftformat Sources Tests
- swiftlint --strict

## Screenshots
- N/A (no UI changes)

## Issue
- N/A

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -86,9 +86,6 @@ let package = Package(
                     "CodexBarCore",
                 ],
                 path: "Sources/CodexBar",
-                resources: [
-                    .process("Resources"),
-                ],
                 swiftSettings: [
                     // Opt into Swift 6 strict concurrency (approachable migration path).
                     .enableUpcomingFeature("StrictConcurrency"),
--- a/Sources/CodexBar/ProviderBrandIcon.swift
+++ b/Sources/CodexBar/ProviderBrandIcon.swift
@@ -4,26 +4,11 @@ import CodexBarCore
 enum ProviderBrandIcon {
     private static let size = NSSize(width: 16, height: 16)
 
-    /// Lazy-loaded resource bundle for provider icons.
-    private static let resourceBundle: Bundle? = {
-        // SwiftPM creates a CodexBar_CodexBar.bundle for resources in the CodexBar target.
-        if let bundleURL = Bundle.main.url(forResource: "CodexBar_CodexBar", withExtension: "bundle"),
-           let bundle = Bundle(url: bundleURL)
-        {
-            return bundle
-        }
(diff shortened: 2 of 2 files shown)


ryanchen01:
> Have you tried the latest commit and built with it? There’s no crash for me right now, so I’m not sure what this PR is actually fixing.

Yes, I did try the latest commit. When I launched the app, it crashed immediately during startup. (it only crashed on my Mac Mini though, works fine on my MacBook Pro) The system log showed a Swift fatal error coming from resource_bundle_accessor.swift:

Fatal error: could not load resource bundle: CodexBar.app/CodexBar_CodexBar.bundle

That means SwiftPM generated an accessor (Bundle.module) for the CodexBar target’s resources, and that accessor expects the SwiftPM resource bundle to live at the root of the app bundle. But the packaging flow copies resources into Contents/Resources, not the app bundle root. So the accessor looked in the wrong place, didn’t find the bundle, and called fatalError before the app could show anything.

What the fix does:

- I removed SwiftPM resources from the CodexBar executable target in Package.swift. That prevents SwiftPM from generating the crashing resource_bundle_accessor.swift in the first place.
- I updated ProviderBrandIcon to load SVGs directly from Bundle.main (which resolves to Contents/Resources, where the packaging script already puts the icons).

Why this fixes it:

Because the crash was caused by the auto-generated SwiftPM resource bundle accessor failing to find a bundle in the app’s root. By removing that resource bundle entirely and reading directly from Bundle.main, it stops SwiftPM from generating the fatal accessor and instead load the assets from the location we already use.

ryanchen01:
Yes, I'm also on macOS Sequoia. v0.18.0-beta.2 couldn't launch at all, but building from the latest commit would launch fine the first time, but if I exit the app, it wouldn't launch again.

validatedev:
If you:

1. Completely delete `/Applications/CodexBar.app`.
2. Run: `CODEXBAR_SIGNING=adhoc ./Scripts/package_app.sh && ditto CodexBar.app /Applications/CodexBar.app` on repo root.
3. Launch `CodexBar.app` from `/Applications/CodexBar.app`.
4. Quit the app.
5. Launch `CodexBar.app` again from `/Applications/CodexBar.app`.

and you still see the same error (just to be sure you’re not running an older build), then it seems the issue has regressed. That’s interesting, because when I follow the same steps on Sequoia 15.7.3 the app opens correctly without this issue.

Issue #139, opened 2026-01-06
consuming too much power on my macbook


ahmedivy:
Device Details: Macbook Pro 2021
Chip: M1 Pro (base model)
Software: macOS Tahoe 26.2

<img width="764" height="346" alt="Image" src="https://github.com/user-attachments/assets/3d496805-1422-485e-9c5e-a4a11784b5b7" />

matteoveglia:
yeah way too much CPU usage, without the menu bar open (just Codex/Claude/Gemini providers on) I'm seeing 90-100% CPU usage (M4 Pro) for over a minute. Surely it should be a quick check and done, I can't see how it should take that long to query an API.
Great app but yeah I can't use it right now, i'll come back to this app later

wdkang123:
Me too!

in0vik:
## Root Cause Analysis: High CPU from `codex` CLI processes

I investigated the source code and found the likely root cause of the high CPU usage. The issue is in `CodexCLISession.swift` — specifically in how `readChunk()` interacts with the non-blocking PTY.

### The bug

**`Sources/CodexBarCore/Providers/Codex/CodexCLISession.swift:263`** — the PTY is set to non-blocking:
```swift
_ = fcntl(primaryFD, F_SETFL, O_NONBLOCK)
```

**Lines 341-354** — `readChunk()` doesn't handle `read()` returning `-1`:
```swift
private func readChunk() -> Data {
    guard self.primaryFD >= 0 else { return Data() }
    var appended = Data()
    while true {
        var tmp = [UInt8](repeating: 0, count: 8192)
        let n = read(self.primaryFD, &tmp, tmp.count)
        if n > 0 {
            appended.append(contentsOf: tmp.prefix(n))
            continue
        }
        break  // n == -1 (EAGAIN) is treated the same as n == 0 (EOF)
    }
    return appended
}
```

When the PTY has no data, `read()` returns `-1` and sets `errno = EAGAIN`. But the code doesn't check `errno` — it just breaks and returns empty `Data`.

### Why this causes 89%+ CPU per `codex` process

1. CodexBar spawns `codex -s read-only -a untrusted` via PTY
2. The main loop (line 126) calls `readChunk()` → sleeps 120ms → repeats
3. `codex` is a Node.js TUI app connected to the PTY — it continuously renders its interface
4. If `codex` doesn't output the expected status markers (e.g. due to an update prompt, auth issue, or changed output format), the loop spins until the deadline
[...]

PR #283, opened 2026-02-01
fix: prevent animation loop when fallback provider shown

### LINKED ISSUE #269: Significant usage of battery
M2 Pro, 2023, 32gb, Tahoe 26.0.1

CodexBar drained 9% of battery in less then 15 minutes after update.

vignesh07:
## Summary

When all providers are disabled, the fallback provider (`.codex`) would still trigger animation because:
- `isVisible(.codex)` returned `true` (fallback)
- But `hasData` was `false` (no fetch ever ran)
- So `shouldAnimate()` returned `true` → 60fps animation loop forever

This caused unnecessary CPU usage and battery drain.

## The Fix (5 lines)

```swift
// Don't animate for fallback provider - it's only shown as a placeholder
// when nothing is enabled. Animating the fallback causes unnecessary
// CPU usage (battery drain). See #269, #139.
let isEnabled = self.isEnabled(provider)
let isFallbackOnly = !isEnabled && self.fallbackProvider == provider
if isFallbackOnly { return false }
```

## Testing

Added `BatteryDrainDiagnosticTests.swift` with tests that verify:
- ✅ Animation driver should be nil when all providers disabled
- ✅ Animation driver should be nil when provider has data  
- ✅ Idle state should have no active animation or refresh

Fixes #269, #139

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 ### Highlights
 - OpenAI web dashboard refresh cadence now follows 5× the base refresh interval.
 - OpenAI web dashboard WebView is kept warm between scrapes to avoid repeated SPA downloads while idle CPU stays low (#284). Thanks @vignesh07!
+- Menu bar: avoid fallback animation loop when all providers are disabled (#283). Thanks @vignesh07!
 - Codex settings now include a toggle to disable OpenAI web extras.
 
 ### Providers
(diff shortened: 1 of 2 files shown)


steipete:
Landed via temp rebase onto main.\n\n- Gate: pnpm lint && pnpm build && pnpm test\n- Land commit: f7ba0130f91b452334140e98e6d3a87eea1432cc\n- Merge commit: b1b42f09f19f233df948ccc3422b8c6bdf680b01\n\nThanks @vignesh07!

Issue #251, opened 2026-01-26
Why so traffic heavy?


vkartaviy:
Hm, just for 24h it downloaded around 3Gb and uploaded around 2Gb.

Mostly to and from chatgpt.com

<img width="1304" height="1194" alt="Image" src="https://github.com/user-attachments/assets/fae55e6a-a0de-42c2-ba61-39a40cf9664b" />

Flohhhhh:
I assume this is closely related: I noticed that the `https://chatgpt.com` process spawned by CodexBar ("CodexBar Web Content") takes a ton of CPU usage while totally idle.

I've seen multiple other issues both closed and open that at some point mention codex/gpt causing battery drain, but haven't yet seen someone connect this to the CPU usage. You can also see that this process is responsible for most of the usage.

I am on the latest beta release

<img width="548" height="231" alt="Image" src="https://github.com/user-attachments/assets/09be4954-1083-4969-a83e-8cc8caedad77" />

<img width="423" height="392" alt="Image" src="https://github.com/user-attachments/assets/1a9ebe7c-f2bc-41dc-b674-7e874a32a43f" />

<img width="800" height="134" alt="Image" src="https://github.com/user-attachments/assets/f6b9ab89-ac84-484d-bc96-eba18d64dfe2" />

thezelium:
Same issue here, I was wondering what's happening there.

<img width="421" height="877" alt="Image" src="https://github.com/user-attachments/assets/55b172fe-b2c7-4a3d-827c-02afe5bc45b2" />

steipete:
Closing as resolved on current main. The traffic shown here was from the old chatgpt.com WebView extras behavior; current main makes OpenAI web extras opt-in for fresh installs, adds battery-saver gating, and slows/limits background dashboard refreshes. If a current build still transfers this much data with OpenAI web extras disabled or battery saver enabled, please open a fresh issue with the current version and network sample.

PR #284, opened 2026-02-01
fix: keep WebView cached to reduce network traffic

### LINKED ISSUE #269: Significant usage of battery

vignesh07:
## Summary

The OpenAI dashboard WebView was being destroyed after every fetch, causing the full ChatGPT SPA bundle to be re-downloaded on each refresh. Users reported 15+ GB of network traffic from `chatgpt.com` over a two-week period.

## Root Cause

Commit 1c55dc7 ("perf: reduce OpenAI web idle CPU") changed the release behavior:

```diff
-                    entry.host.hide()
-                    self.prune(now: Date())
+                    entry.host.close()
+                    self.entries.removeValue(forKey: key)
```

This was intended to reduce idle CPU usage by destroying cached WebViews. However, it caused every refresh to create a new WebView and download all assets fresh.

With the default 5-minute refresh interval (multiplied by 5 for OpenAI web = 25 minutes), this resulted in approximately 57 page loads per day. If each load downloads 20-50 MB of assets, that adds up to 1-3 GB per day.

## The Fix

On release, call `hide()` instead of `close()` to keep the WebView in cache:

```swift
// Hide instead of close - keep WebView cached for reuse.
// This avoids re-downloading the ChatGPT SPA bundle on every refresh,
// saving significant network bandwidth. See GitHub issues #269, #251.
entry.host.hide()
self.prune(now: Date())
```

This gives us the best of both worlds:
- **Low idle CPU**: WebView is suspended when hidden (`inactiveSchedulingPolicy=.suspend` + `alphaValue=0`)
- **Low network usage**: WebView is reused across refreshes, assets stay cached
- **Memory cleanup**: The existing 10-minute idle timeout prunes unused WebViews

## Testing

Added 7 unit tests to verify cache behavior:

| Test | Verifies |
|------|----------|
| `webViewCachedAfterRelease` | WebView stays in cache after release |
| `separateCachesPerDataStore` | Different accounts get separate WebViews |
| `webViewPrunedAfterIdleTimeout` | 10-minute idle timeout works |
| `recentlyUsedWebViewNotPruned` | Active WebViews are not pruned |
| `evictRemovesSpecificWebView` | Explicit eviction works |
| `busyWebViewCreatesTemporary` | Concurrent access creates temporary WebView |
| `sequentialFetchesReuseWebView` | 5 sequential fetches reuse the same WebView |

All tests pass.

## Impact

- Network traffic reduction: ~1-3 GB/day to ~50 MB/day (first load only)
- No increase in idle CPU (WebView is suspended when hidden)
- No increase in memory (10-minute prune timeout unchanged)

Fixes #269, #251

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 0.18.0-beta.2 — 2026-01-21
 ### Highlights
 - OpenAI web dashboard refresh cadence now follows 5× the base refresh interval.
-- OpenAI web dashboard WebView is torn down after each scrape to reduce idle CPU.
+- OpenAI web dashboard WebView is kept warm between scrapes to avoid repeated SPA downloads while idle CPU stays low (#284). Thanks @vignesh07!
 - Codex settings now include a toggle to disable OpenAI web extras.
 
 ### Providers
--- a/Tests/CodexBarTests/OpenAIDashboardWebViewCacheTests.swift
+++ b/Tests/CodexBarTests/OpenAIDashboardWebViewCacheTests.swift
@@ -39,7 +39,7 @@ struct OpenAIDashboardWebViewCacheTests {
     func webViewCachedAfterRelease() async throws {
         let cache = OpenAIDashboardWebViewCache()
         let store = WKWebsiteDataStore.nonPersistent()
-        let url = URL(string: "https://example.com")!
+        let url = URL(string: "about:blank")!
         
         // First acquire
         let lease1 = try await cache.acquire(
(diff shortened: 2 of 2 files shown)


steipete:

- Gate: pnpm lint && pnpm build && pnpm test
- Land commit: 1975e54beefdf2f091af55988ed95c2c73a37269
- Merge commit: 41dc7815ccd15b95999f4b59d72f70238b70bf77

Thanks @vignesh07!

PR #293, opened 2026-02-02
docs: brew official cask

zeldrisho:
https://formulae.brew.sh/cask/codexbar


ratulsarna:
Thanks for the contribution. I’m closing this one because the same documentation change now appears to be reflected in current main, so there is no longer a distinct change left for this PR to land.

PR #307, opened 2026-02-04
Allow --source auto on Linux

Viou:
## What
- Treat `--source auto` as non-web so Linux doesn’t error out by default.
- Keep `--source web` macOS-only behavior unchanged.
- Add a small non-macOS test for `ProviderSourceMode.usesWeb`.
- Document Kimi CLI usage with `KIMI_AUTH_TOKEN` (auto mode).

## Why
On Linux, the CLI defaults to `--source auto`, but it was gated as a web-only mode and aborted before providers could use env/api/cli strategies (e.g., Kimi via `KIMI_AUTH_TOKEN`).

## Testing
- `swift test`

### DIFF
--- a/Sources/CodexBarCLI/CLIUsageCommand.swift
+++ b/Sources/CodexBarCLI/CLIUsageCommand.swift
@@ -95,7 +95,7 @@ extension CodexBarCLI {
         }
 
         #if !os(macOS)
-        if parsedSourceMode?.usesWeb == true {
+        if parsedSourceMode == .web {
             Self.exit(
                 code: .failure,
                 message: "Error: --source web/auto is only supported on macOS.",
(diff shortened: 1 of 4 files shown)


Viou:
Follow-up: reverted `ProviderSourceMode.usesWeb` to include `auto` again (avoids potential macOS regressions where auto should consider web strategies), and changed the Linux guard to block only explicit `--source web` (both at argument-parse time and after provider source resolution).\n\nTested: `swift test`.

ratulsarna:
Thank you for the Linux source-selection fix. The source-gating code has moved on since this PR was opened, and this older diff still has unresolved review feedback, so it is no longer the best vehicle for the change. I am closing it as stale; if this Linux auto-source edge case still reproduces on current main, a fresh focused PR would be appreciated.

PR #317, opened 2026-02-06
Support reading cookies from multiple Safari profiles

### LINKED ISSUE #316: Support for multiple Safari profiles
Dear maintainers,
whenever I used CodexBar I couldn't read my Cursor data.

Image

So I cloned repo and started debugging - tt turns out that SweetCookieKit doesn't support multiple Safari profiles. Created fix in SweetCookieKit. In next message I'll link it.

przemyslaw-szurmak:
Use updated version of SweetCookieKit. Tmp pin to updated version.

This fixes https://github.com/steipete/CodexBar/issues/316

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "74bd6f3ab6e0b0cb0c2cddb00f2167c2ab0a1c00cd54ffc1a2899c7ef8c56367",
+  "originHash" : "0ae4d8873b6d8f26b9fa88a74ac54a3101f7e8771076c80c488e035812f5d80d",
   "pins" : [
     {
       "identity" : "commander",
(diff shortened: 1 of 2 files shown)


Package.resolved
```diff
@@ -31,10 +31,9 @@
     {
       "identity" : "sweetcookiekit",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/steipete/SweetCookieKit",
+      "location" : "https://github.com/przemyslaw-szurmak/SweetCookieKit",
```
przemyslaw-szurmak, line 34: this is just temporal, once new version is released this can be reverted

steipete:
Thanks for digging into the Safari profile limitation and for tracing it down to SweetCookieKit.

I’m going to close this draft because the CodexBar-side change is only a temporary dependency pin to a personal SweetCookieKit fork/revision. That is not something we should merge into the app: it bypasses the maintained upstream package, makes dependency provenance/release tracking harder, and does not leave any CodexBar-side behavior or regression test to review here.

The right path is:

- land Safari multi-profile support in `steipete/SweetCookieKit`
- cut or tag an upstream SweetCookieKit version
- open a small CodexBar PR that bumps from the current upstream version and adds/updates the relevant cookie import coverage

That keeps the app on the maintained dependency path while still preserving the useful work you found.

przemyslaw-szurmak:
okay, I understand. Thanks for checking.

PR #326, opened 2026-02-07
Fix Claude OAuth refresh handling

ratulsarna:
### Problem
CodexBar refreshed Claude OAuth using the refresh token read from the macOS Keychain item `Claude Code-credentials`. If the server rotates refresh tokens, CodexBar updated only its own cache, leaving Claude CLI holding an invalid refresh token and later hitting 401 / forced re-login.

### Fix
- Treat credentials loaded from Claude’s keychain / credentials file as Claude CLI-owned.
- If expired, delegate refresh to `claude` (no direct token-endpoint refresh for Claude-owned creds).
- After delegated refresh, re-read creds and retry OAuth once.
- CodexBar-owned creds (if present) still refresh via the token endpoint.

### Notes
- On some systems, reading `Claude Code-credentials` can still trigger Keychain UI even after “Always Allow”, so OAuth-on-expiry may remain prompty.

### Verification
- `swift test`
- `pnpm check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,23 +13,15 @@
 - CI/tooling reliability improved via pinned lint tools, deterministic macOS test execution, and PTY timing test
   stabilization plus Node 24-ready GitHub Actions upgrades (#292, #312, #290).
 
-### Claude OAuth & Keychain (net result)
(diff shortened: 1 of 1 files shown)


ratulsarna:
@codex review

PR #331, opened 2026-02-08
Build universal binary by default for Intel Mac support

JimLiuxinghai:
Body:
 Summary
Changes the default build behavior to create universal binaries (arm64 + x86_64) instead of host-architecture-only builds, enabling Intel Mac support out of the box.
 Changes
- Modified `Scripts/package_app.sh`: Default `ARCH_LIST` to `arm64 x86_64` instead of host architecture detection
- Modified `Scripts/compile_and_run.sh`: Default `ARCHES_VALUE` to "arm64 x86_64" instead of `uname -m`
 Why
Previously, the build scripts only compiled for the host architecture, which meant:
- Building on Apple Silicon Macs produced arm64-only binaries
- Building on Intel Macs produced x86_64-only binaries
This change ensures that both architectures are always included in the build, making the app work on both Intel and Apple Silicon Macs regardless of where it was built.
 Backward Compatibility
Users can still override this behavior via the `ARCHES` environment variable:
ARCHES="arm64" ./Scripts/package_app.sh      # Apple Silicon only
ARCHES="x86_64" ./Scripts/package_app.sh     # Intel only
Testing
- Verified that lipo -archs CodexBar.app/Contents/MacOS/CodexBar outputs: x86_64 arm64
- Build completes successfully with ./Scripts/compile_and_run.sh

### DIFF
--- a/Scripts/compile_and_run.sh
+++ b/Scripts/compile_and_run.sh
@@ -195,8 +195,8 @@ fi
 if [[ "${DEBUG_LLDB}" == "1" && -n "${RELEASE_ARCHES}" ]]; then
   fail "--release-arches is only supported for release packaging"
 fi
-HOST_ARCH="$(uname -m)"
-ARCHES_VALUE="${HOST_ARCH}"
+# Build universal binary by default (arm64 + x86_64)
+ARCHES_VALUE="arm64 x86_64"
 if [[ -n "${RELEASE_ARCHES}" ]]; then
   ARCHES_VALUE="${RELEASE_ARCHES}"
 fi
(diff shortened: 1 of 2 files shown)


steipete:
Closing after re-checking the current release/dev scripts.

Release notarization already builds a universal app by default via `Scripts/sign-and-notarize.sh`, while `Scripts/package_app.sh` and `Scripts/compile_and_run.sh` intentionally default to the host arch for the fast local dev loop and expose `ARCHES` / `--release-universal` when a universal local package is needed.

So the shipped Intel support path is already covered, and making every dev/package build universal by default would slow the common path. Thanks @JimLiuxinghai!

PR #338, opened 2026-02-09
Use authoritative usageUnitsAvailable for Augment credits limit

bcharleson:
## Summary

- The `creditsLimit` computed property on `AugmentCreditsResponse` now prefers the API-provided `usageUnitsAvailable` field as the total allocation instead of always deriving it from `usageUnitsRemaining + usageUnitsConsumedThisBillingCycle`.
- Falls back to the existing derived calculation when `usageUnitsAvailable` is absent or zero, so the change is fully backward-compatible.

## Motivation

The `usageUnitsAvailable` field was already being parsed from the Augment API response but was never used. Relying on the derived sum (`remaining + consumed`) could theoretically produce an inaccurate total if Augment adjusts credits mid-cycle, applies bonus allocations, or changes how remaining/consumed are reported. Using the authoritative API value ensures the usage progress bar percentage aligns with Augment's own dashboard.

## Test plan

- [x] Verified the progress bar percentage matches Augment dashboard after rebuild
- [x] Confirmed fallback path still works when `usageUnitsAvailable` is nil
- [x] `swift build` and `swift test` pass


steipete:
Landed the narrow version on main in 0966118e.

What changed: `AugmentCreditsResponse.creditsLimit` now prefers positive `usageUnitsAvailable`, with fallback to remaining + consumed for older payloads. Added regression coverage in `AugmentStatusProbeTests`.

Verified with `swift test`, `pnpm check`, `./Scripts/compile_and_run.sh`, and green CI after rerun. Thanks @bcharleson!

bcharleson:
Anytime @steipete!

PR #354, opened 2026-02-12
Add week-over-week usage projection feature

johnlarkin1:
## Summary

- Adds a **weekly projection submenu** that shows estimated end-of-week usage based on current consumption pace vs. the same point last week
- Tracks weekly usage history in a local JSON store (`WeeklyUsageHistoryStore`) with automatic weekly rotation
- Renders a small inline chart (`WeeklyProjectionChartMenuView`) comparing this week's trajectory to last week's actual usage
- Includes projection math (`WeeklyProjection`) that computes linear extrapolation with week-over-week delta

## How it works

1. `UsageStore` snapshots daily totals into `WeeklyUsageHistory` on each fetch cycle
2. `WeeklyProjection` compares current cumulative usage at the same fractional point in the week to last week's data
3. The submenu shows: projected total, delta vs last week (↑/↓), and a mini sparkline chart
4. Submenu appears in both compact and full menu layouts

## New files

| File | Purpose |
|------|---------|
| `WeeklyProjection.swift` | Projection math + week-over-week comparison |
| `WeeklyUsageHistory.swift` | Data model for weekly snapshots |
| `WeeklyUsageHistoryStore.swift` | JSON persistence with auto-rotation |
| `UsageStore+WeeklyHistory.swift` | Integration with main usage store |
| `WeeklyProjectionChartMenuView.swift` | SwiftUI chart submenu view |

## Test plan

- [x] Builds cleanly on upstream/main (`swift build`)
- [ ] Verify submenu appears after first usage fetch
- [ ] Verify projection updates correctly across days
- [ ] Verify chart renders with both single-week and multi-week data

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/Sources/CodexBar/StatusItemController+Menu.swift
+++ b/Sources/CodexBar/StatusItemController+Menu.swift
@@ -322,6 +322,7 @@ extension StatusItemController {
                 provider: context.currentProvider,
                 width: context.menuWidth,
                 webItems: webItems)
+            _ = self.addWeeklyProjectionSubmenu(to: menu, provider: context.currentProvider)
             return true
         }
 
@@ -332,6 +333,7 @@ extension StatusItemController {
         if context.currentProvider == .codex, model.creditsText != nil {
             menu.addItem(self.makeBuyCreditsItem())
         }
+        _ = self.addWeeklyProjectionSubmenu(to: menu, provider: context.currentProvider)
         menu.addItem(.separator())
         return false
     }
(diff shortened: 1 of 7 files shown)


johnlarkin1:
Addressed both review comments:

**P1 — Percentage projection math:** The `avgDailyRate` computation now branches by metric type. For `.percentage` (cumulative values), it uses `latestValue / daysWithData` instead of the arithmetic mean of all values. `projectedValue` for future days also now shows the projected cumulative total (`latestValue + dailyRate * daysAhead`) rather than a flat rate. Example: Mon=10%, Tue=20% → dailyRate=10, projects ~70% end-of-week (not 95%).

**P2 — Timezone mismatch:** Both `UsageStore+WeeklyHistory.swift` and `WeeklyProjection.swift` now set `cal.timeZone = TimeZone.current` on the ISO 8601 calendar, ensuring `weekNumber`, `weekYear`, and `weekday` are computed in the same local timezone as `dayKey`.

PR #369, opened 2026-02-13
Copilot Provider Update

amoranio:
Added clear information of the automatic copy/paste of device codes which maybe missed due to window load. This should make things slightly clearer for those onboarding for the first time. 

Footer Added: Device code is copied to your clipboard — paste with ⌘V.


ratulsarna:
Rebased on to main directly. Thanks @amoranio for the contribution!

PR #375, opened 2026-02-14
Reduce OpenAI WebView idle CPU

jperezrealini:
This reduces idle CPU usage from CodexBar’s offscreen OpenAI dashboard WKWebView cache.

- Stop loading and detach the cached WebView from its host window when hidden so WebKit can fully suspend background activity.
- Reattach the WebView when showing it for the next scrape.


ratulsarna:
Thanks for the quick update. I ran some tests and benchmarks locally but there is too much noise in the data from other apps to accurately gather the impact on WebKit from CodexBar. Do you have any such data to show measurable difference? The PR itself looks good. I'm just trying to find the measurable benefit.

PR #385, opened 2026-02-16
Add provider credential store abstractions

Specter242:
## Summary
- add provider credential store protocols in `CodexBarCore`
- include environment-backed and file-backed implementations
- add composite credential store to support fallback chains

## Why
This adds a platform-neutral interface for API key resolution that Linux and other non-AppKit hosts can share without depending on app-specific storage mechanisms.

## Test plan
- [x] `swift build -c debug`
- [x] `swift test`


ratulsarna:
Thanks for the contribution. I’m closing this one because the credential-store abstraction work appears to be covered by a newer overlapping PR, and I want to keep the queue from splitting across overlapping implementations.

Issue #162, opened 2026-01-09
Copilot error: The data couldn’t be read because it is missing.


emanuelst:
Copilot error: The data couldn’t be read because it is missing.

jornae2w32:
Can we please get this fixed!

ratulsarna:
Hi, I'm looking for help here. Does #325 solve the issue for everyone on this thread?

PR #393, opened 2026-02-18
Fix Copilot free-plan usage parsing

emanuelst:
## Summary
This PR fixes Copilot usage handling for free-plan accounts where the API response shape differs from paid-plan payloads.

### Related Issue
possibly fixes https://github.com/steipete/CodexBar/issues/162

### What changed
- Added parsing support for Copilot quota data from:
  - `monthly_quotas`
  - `limited_user_quotas`
- Updated Copilot fetch behavior to:
  - use chat quota as primary when premium/completions quota is unavailable
  - avoid defaulting to a fake 0%-used primary window

## Why
Free-plan Copilot responses can omit usable `quota_snapshots`, which previously caused incorrect UI output (for example: `Copilot error: The data couldn’t be read because it is missing`).

## Note
AI-assisted, tested locally.

### DIFF
--- a/Sources/CodexBar/UsageStore.swift
+++ b/Sources/CodexBar/UsageStore.swift
@@ -560,8 +560,21 @@ final class UsageStore {
     }
 
     func handleSessionQuotaTransition(provider: UsageProvider, snapshot: UsageSnapshot) {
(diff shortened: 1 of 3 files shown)


emanuelst:
@codex review

PR #422, opened 2026-02-24
Fix Gecko cookie profile detection for uppercase .Default directories

bald-ai:
## Summary
- match Gecko profile directory names case-insensitively when checking for `*.default*`
- apply this to both profile availability checks and `cookies.sqlite` detection
- add regression test for Zen profile directory names like `abc.Default (release)`

## Why
Some Firefox/Zen installs use uppercase `Default` in profile directory names. Current matching only checked lowercase `.default`, which can skip valid cookie stores and block cookie import.

## Validation
- `pnpm check`
- manual local verification: `CodexBarCLI usage --provider cursor --source auto ...` imported cookies successfully from `Zen e6wtnxlb.Default (release)`
- `swift test --filter BrowserDetectionTests`

## Note
`swift test` (full suite) currently crashes in existing `ProviderVersionDetector` teardown path (`NSConcreteTask terminationStatus: task still running`) on current main; unrelated to this change.

### DIFF
--- a/Sources/CodexBarCore/BrowserDetection.swift
+++ b/Sources/CodexBarCore/BrowserDetection.swift
@@ -214,7 +214,7 @@ public final class BrowserDetection: Sendable {
 
         if browser.usesGeckoProfileStore {
             return contents.contains { name in
-                name.contains(".default")
+                name.range(of: ".default", options: [.caseInsensitive]) != nil
             }
         }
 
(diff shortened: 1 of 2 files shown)


bald-ai:
Hi, the issue was that auto cookie were failing in Zen browser due to case sensitive .default vs .Default. Should be fixed now.

PR #423, opened 2026-02-25
Show Cursor plan usage in dollars

### LINKED ISSUE #398: Feature request: Add a display of the total amount spent on tokens.
For example, when using Cursor, it shows 100% progress, but the tokens haven’t run out yet and I can still use it because Cursor offers a certain amount of $ as a bonus. It would be more convenient to see the total amount spent. Thank you.

bald-ai:
## Summary
- show Cursor primary usage labels in dollars (used/left and over-limit) instead of percentage text in menu cards
- show dollar text for Cursor in the menu bar brand+text mode, with no percent fallback
- persist cursorPlanCost in UsageSnapshot so dollar labels survive serialization
- keep Cursor parser dollar-first by avoiding percent-only fallback when plan limit is missing

## Tests
- swift test --filter MenuCardModelTests --filter CursorStatusProbeTests --filter StatusItemAnimationTests
- pnpm check

Refs: https://github.com/steipete/CodexBar/issues/398#issuecomment-3935146412
Fixes #398

### DIFF
--- a/Sources/CodexBar/MenuCardView.swift
+++ b/Sources/CodexBar/MenuCardView.swift
@@ -35,9 +35,13 @@ struct UsageMenuCardView: View {
             let detailRightText: String?
             let pacePercent: Double?
             let paceOnTop: Bool
+            let customLabel: String?
 
             var percentLabel: String {
-                String(format: "%.0f%% %@", self.percent, self.percentStyle.labelSuffix)
+                if let customLabel {
+                    return customLabel
+                }
+                return String(format: "%.0f%% %@", self.percent, self.percentStyle.labelSuffix)
             }
         }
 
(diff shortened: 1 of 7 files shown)


bald-ai:
Hi, displaying dollars offers way better info for Cursor specifically than % as it can go over 100.

bald-ai:
Does it still make sense to support legacy Cursor accounts? I would just cut it, it has been a long time since they changed the pricing (4. 7. 2025). If lots of people somehow still have old accounts, we can put it back. I find it highly unlikely that people who use Codexbar will have legacy accounts.

ratulsarna:
@bald-ai The cursor way of handling usage has changed drastically. I think we'll have to re-look at this from scratch instead of showing dollars.

image

Issue #232, opened 2026-01-23
Copilot doesn't show pace (Broken in "Both" display mode for pace)


jeankvd:
<img width="257" height="26" alt="Image" src="https://github.com/user-attachments/assets/1078641e-1b61-4cb1-bf5c-a881c9dc1d08" />

Copilot doesn't have pace percent or a pace indicator, this causes it to not even show the usage percent on Display Mode.

Copilot resets "Monthly" not weekly and doesn't have the usual 5 hour time limit window

According to copilot "Allowances reset on the 1st of each month at 00:00:00 UTC." https://docs.github.com/en/billing/concepts/product-billing/github-copilot-premium-requests

Context:
PR that added Pace display option (Broken for copilot): https://github.com/steipete/CodexBar/pull/169
PR that added copilot support: https://github.com/steipete/CodexBar/pull/74
PR that added Pace: https://github.com/steipete/CodexBar/pull/40 
PR for Pace indicator: https://github.com/steipete/CodexBar/pull/155

Issue #84, opened 2025-12-29
CodexBar wants to use confidential information from "Chrome Safe storage" stored in your keychain.


socekin:
After updating to Version 0.15.3 (45), it occasionally prompts for permission for "Chrome Safe storage" stored in keychain, which seems very insecure. Why is this permission needed?

<img width="414" height="167" alt="Image" src="https://github.com/user-attachments/assets/67588fde-8828-45bd-90d6-1edfb3d0767f" />

nicosuave:
It's to fetch usage data. I agree it's a little much to default to `on`

steipete:
I use cookies to automatically get oauth data from various providers. Seems better than asking for user/password in the app.
Open for suggestions how this can be improved.

guilleigmu:
I'm running with the same issue @socekin describes. Denying non used permissions works fine, problem is it will keep asking the non granted permissions each time I open the app. Quite annoying :(

steipete:
CodeBar's whole model is based on getting cookies so I don't need to deal with user/password.
Nothing will work if you deny permissions.

guilleigmu:
@steipete Everything works actually. No need to grant Chrome safe storage permissions if I only need Claude. The only annoying thing is the app asking for permissions I don't need EVERY time I open the app. 

In other words, I'm not discussing your model. Just make it the permissions are only asked once or maybe just when enabling a new provider.

steipete:
I explain this more here: https://github.com/steipete/CodexBar#macos-permissions-why-theyre-needed

Previously, I used the cli, but that's slow, brittle and inefficient. Now we use the oauth token from the keychain, which is much faster and more reliable. It just needs permission to access the key. The app is open source so feel free to check out the code and compare it to your own if you are worried that I'll steal your tokens :)

I will add a pre-alert to explain this to new users so onboarding is easier.

steipete:
I changed main to show an alert to explain why permissions are needed, and moved z.ai/copilot to only reqeust permissions when enabled.

steipete:
Tokens expire, that'd be a tedious thing to do. Everything is open source and verifiable if you have any concerns.

imchenway:
我就是因为这个拒绝并且卸载了，从没见过如此夸张的要求

luizv:
> Seems better than asking for user/password in the app.
> Open for suggestions how this can be improved.

@steipete, it's open source and verifiable, but for some it ways feels weird to give broader than needed permissions like that.
[...]

PR #427, opened 2026-02-25
Cache reset times in-memory to preserve pace/deficit/reserve display

Whoaa512:
## Problem

When Claude usage is fetched via a non-cookie auth path (e.g. OAuth), `resetsAt` can come back as `nil`. This causes the pace/deficit/reserve display to disappear until the next cookie-based fetch, which may prompt the user for keychain access.

## Fix

Add a dedicated in-memory reset time cache (`lastKnownResetTimes`) on `UsageStore`, separate from the snapshots dict (which gets cleared on fetch failures and provider disable).

When a fresh fetch returns `nil` for `resetsAt`:
- Backfill from the cached value if it hasn't expired (still in the future)
- Also backfills `windowMinutes` and `resetDescription` when missing
- **Identity-aware**: skips backfill when account email changes between cached and fresh snapshots
- Applied in both the main refresh path and the token-account refresh path
- Cache cleared on credential file changes and provider disable/re-enable

## Changes

- `RateWindow.backfillingResetTime(from:)` — backfill a single rate window
- `UsageSnapshot.backfillingResetTimes(from:)` — backfill primary/secondary/tertiary with identity check
- `UsageStore.lastKnownResetTimes` — dedicated cache dict that survives snapshot clears
- `UsageStore+Refresh.swift` — apply backfill in main refresh path; clear cache on credential changes and provider disable
- `UsageStore+TokenAccounts.swift` — apply backfill in token-account refresh path
- 8 new tests covering backfill, expiry, account switch, same-account, and no-op cases

## Testing

- `swift build` ✅
- `swift test` — all 803+ tests pass ✅
- New `ResetTimeBackfillTests` — 8/8 pass ✅

Related: #232, #84

### DIFF
--- a/Sources/CodexBarCore/UsageFetcher.swift
+++ b/Sources/CodexBarCore/UsageFetcher.swift
@@ -17,6 +17,16 @@ public struct RateWindow: Codable, Equatable, Sendable {
     public var remainingPercent: Double {
         max(0, 100 - self.usedPercent)
     }
+
+    public func backfillingResetTime(from cached: RateWindow?, now: Date = .init()) -> RateWindow {
+        if self.resetsAt != nil { return self }
+        guard let cachedReset = cached?.resetsAt, cachedReset > now else { return self }
+        return RateWindow(
+            usedPercent: self.usedPercent,
+            windowMinutes: self.windowMinutes ?? cached?.windowMinutes,
+            resetsAt: cachedReset,
+            resetDescription: self.resetDescription ?? cached?.resetDescription)
+    }
 }
 
 public struct ProviderIdentitySnapshot: Codable, Sendable {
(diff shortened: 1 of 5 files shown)


Sources/CodexBarCore/UsageFetcher.swift
```diff
@@ -199,6 +209,38 @@ public struct UsageSnapshot: Codable, Sendable {
...
+            minimaxUsage: self.minimaxUsage,
+            openRouterUsage: self.openRouterUsage,
+            cursorRequests: self.cursorRequests,
+            updatedAt: self.updatedAt,
+            identity: self.identity)
+    }
+
+    private static func identitiesMatch(_ a: ProviderIdentitySnapshot?, _ b: ProviderIdentitySnapshot?) -> Bool {
+        if a == nil, b == nil { return true }
+        if a == nil || b == nil { return true }
```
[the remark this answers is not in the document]
Whoaa512: fixed

Sources/CodexBarCore/UsageFetcher.swift
```diff
@@ -17,6 +17,16 @@ public struct RateWindow: Codable, Equatable, Sendable {
...
         max(0, 100 - self.usedPercent)
     }
+
+    public func backfillingResetTime(from cached: RateWindow?, now: Date = .init()) -> RateWindow {
+        if self.resetsAt != nil { return self }
+        guard let cachedReset = cached?.resetsAt, cachedReset > now else { return self }
+        return RateWindow(
+            usedPercent: self.usedPercent,
+            windowMinutes: self.windowMinutes ?? cached?.windowMinutes,
+            resetsAt: cachedReset,
```
ratulsarna, line 27: Could we ensure this reconstructed `RateWindow` preserves the existing metadata from the fresh window? In particular, providers that expose rolling regeneration details may already populate additional fields that should remain intact after backfilling the reset time.

Sources/CodexBarCore/UsageFetcher.swift
```diff
@@ -237,6 +247,38 @@ public struct UsageSnapshot: Codable, Sendable {
...
+        guard let cached else { return self }
+        if !Self.identitiesMatch(self.identity, cached.identity) { return self }
+        let bp = self.primary?.backfillingResetTime(from: cached.primary, now: now)
+        let bs = self.secondary?.backfillingResetTime(from: cached.secondary, now: now)
+        let bt = self.tertiary?.backfillingResetTime(from: cached.tertiary, now: now)
+        if bp?.resetsAt == self.primary?.resetsAt,
+           bs?.resetsAt == self.secondary?.resetsAt,
+           bt?.resetsAt == self.tertiary?.resetsAt
+        { return self }
+        return UsageSnapshot(
```
ratulsarna, line 261: Could we ensure this copy path preserves the full `UsageSnapshot` payload? Claude snapshots may include `extraRateWindows`, and it looks worth checking whether those are retained when reset-time backfill constructs a new snapshot.

steipete:
Thanks @Whoaa512 — fixed this bug class on current `main` in 9b466375.

I kept the reset-time backfill idea, but wired it against the current `UsageSnapshot` model so sparse refreshes keep known reset countdowns without dropping newer fields like extra windows, provider-specific usage payloads, cursor request data, or account identity. I also added focused regression coverage in `ResetTimeBackfillTests`.

Closing this PR as superseded by the mainline fix.

PR #433, opened 2026-02-25
Add configurable quota warning notifications

DaddioTime:
Notifies users before quota depletion at configurable thresholds (default 50% and 20% remaining). Tracks primary (session) and secondary (weekly) windows independently per provider. Warnings fire once per threshold crossing and auto-reset when quota recovers.


steipete:
Thanks, this feature direction still makes sense, but this older PR has been overtaken by #852. The newer branch covers the same configurable quota warning idea with provider-specific controls, warning markers, flash feedback, and much broader tests. Closing this one so review can focus on the newer implementation.

PR #435, opened 2026-02-25
Add time-until-reset countdown to menu bar

guglielmofonda:
## Summary

Adds a **time-until-reset countdown** directly in the menu bar, so you can see how long until your quota resets without having to click the icon.

**Before:** click the icon → scroll down to find "Resets in 28m"
**After:** menu bar already shows `72% · 28m`

## Settings

New toggle in **Settings → Display → Menu bar**:

> **Show time until reset** — Display a countdown (e.g. "28m") next to the percentage in the menu bar.

The toggle is disabled/faded when "Menu bar shows percent" is off, since the countdown only appears in text mode.

## Countdown format

| Time remaining | Display |
|---|---|
| Hours + minutes | `2h 5m` |
| Under an hour | `28m` |
| Under a minute | `<1m` |
| Past or unknown | *(hidden)* |

Picks the nearest upcoming reset across both the primary and secondary quota windows.

## Implementation

- `MenuBarDisplayText.timeUntilResetText()` — pure formatter, testable in isolation
- `StatusItemController` starts a 60-second `Timer` when the feature is on, so the label ticks down every minute without waiting for a data refresh; timer is torn down when disabled
- Persisted via `UserDefaults` key `menuBarShowsTimeUntilReset`, default `false`

## Test plan

- [ ] Enable "Menu bar shows percent" + "Show time until reset" → see e.g. `72% · 28m`
- [ ] Wait ~1 minute → countdown decrements automatically
- [ ] Disable toggle → only `72%` shown
- [ ] Disable "Menu bar shows percent" → countdown toggle is greyed out

### DIFF
--- a/Sources/CodexBar/MenuBarDisplayText.swift
+++ b/Sources/CodexBar/MenuBarDisplayText.swift
@@ -2,6 +2,20 @@ import CodexBarCore
 import Foundation
 
 enum MenuBarDisplayText {
+    /// Returns a compact countdown string (e.g. "28m" or "2h 5m") until the next quota reset.
+    /// Returns nil if `resetsAt` is nil or already in the past.
+    static func timeUntilResetText(resetsAt: Date?, now: Date = .init()) -> String? {
+        guard let resetsAt, resetsAt > now else { return nil }
+        let totalSeconds = Int(resetsAt.timeIntervalSince(now))
+        let hours = totalSeconds / 3600
+        let minutes = (totalSeconds % 3600) / 60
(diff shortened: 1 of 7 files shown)


ratulsarna:
Thank you for adding the reset countdown idea. It is a nice usability improvement, but this PR has gone stale and still has unresolved review feedback around settings observation, which would make the toggle behave inconsistently. I am going to close it for now so we can keep the backlog focused on changes that are ready to move forward.

PR #457, opened 2026-03-02
Publish macOS CLI tarballs in release workflow

androidshu:
## Summary

- Expand `.github/workflows/release-cli.yml` from Linux-only to a cross-platform CLI release workflow.
- Keep existing Linux assets and add two macOS CLI tarballs on release:
  - `CodexBarCLI-v<tag>-macos-arm64.tar.gz`
  - `CodexBarCLI-v<tag>-macos-x86_64.tar.gz`
- Preserve static Swift stdlib packaging for Linux only.
- Update CLI-related docs (`README.md`, `docs/cli.md`, `docs/releasing-homebrew.md`) to reflect the new macOS + Linux CLI asset matrix.

## Why

Intel macOS users can run `CodexBarCLI`, but release assets currently only publish Linux CLI tarballs. Publishing macOS CLI tarballs closes that distribution gap and makes downstream package automation (including Homebrew formula updates) straightforward.

## Validation

- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/release-cli.yml")'`
- `swift build -c release --product CodexBarCLI`
- `./.build/release/CodexBarCLI --version`

## Follow-up

Homebrew tap formula updates are still required to consume the new macOS CLI tarballs.

### DIFF
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -1,4 +1,4 @@
-name: Release Linux CLI
+name: Release CLI
 
 on:
   release:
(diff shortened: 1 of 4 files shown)


androidshu:
Addressed the Codex review feedback: Swift 6.2.1 setup now runs for macOS jobs as well (not Linux-only), so both macOS matrix entries use an explicit toolchain before build/package.

steipete:
Landed a consolidated release-workflow version on main in 0966118e.

What changed: the existing CLI release workflow now publishes macOS arm64/x86_64 tarballs alongside Linux x86_64/aarch64, with per-arch smoke tests and docs/changelog updates.

Verified local macOS arm64 CLI build/smoke, `swift test`, `pnpm check`, `./Scripts/compile_and_run.sh`, Linux CLI CI, and green CI after rerun. Thanks @androidshu!

PR #461, opened 2026-03-02
Add multi-account token support for Codex and API-key providers

Minoo7:
## Summary
- extend token-account support catalog for Codex plus API-key providers (Copilot, Kimi K2, Synthetic, Warp, OpenRouter)
- enable Codex token-account settings visibility when cookie source is manual so users can add accounts before any token exists
- add Codex manual-cookie actions in Preferences: Test cookie (validates immediately) and Save as account (appends token accounts)
- surface Codex cookie import/test status directly in the Codex settings pane subtitle
- keep token-account injection consistent so selected accounts override configured API keys in app and CLI environments
- add regression tests for Codex catalog support, Codex manual cookie actions, >2 account appends, Codex CLI snapshot cookie-source forcing, Codex settings visibility, and API-provider precedence

## Verification
- swift test --filter ProviderSettingsDescriptorTests/codexManualCookieFieldExposesTestAndSaveActions
- swift test --filter ProviderSettingsDescriptorTests/codexSaveCookieActionAppendsAccountsBeyondTwo
- swift test --filter TokenAccountEnvironmentPrecedenceTests
- swift test --filter ProviderSettingsDescriptorTests/codex
- swiftformat Sources Tests && swiftlint --strict
- pnpm check

### DIFF
--- a/Sources/CodexBar/UsageStore.swift
+++ b/Sources/CodexBar/UsageStore.swift
@@ -717,12 +717,13 @@ extension UsageStore {
                 self.failureGates[.codex]?.recordSuccess()
                 self.lastSourceLabels[.codex] = "openai-web"
             }
-            if self.credits == nil, let credits = dash.toCreditsSnapshot() {
+            if let credits = dash.toCreditsSnapshot() {
                 self.credits = credits
                 self.lastCreditsSnapshot = credits
                 self.lastCreditsError = nil
                 self.creditsFailureStreak = 0
             }
+            self.updateSelectedCodexTokenAccountSupplementaryData(dashboard: dash)
         }
 
         if let email = targetEmail, !email.isEmpty {
@@ -753,6 +754,7 @@ extension UsageStore {
         let targetEmail = self.codexAccountEmailForOpenAIDashboard()
         let tokenAccountID = self.settings.selectedTokenAccount(for: .codex)?.id
         self.handleOpenAIWebTargetEmailChangeIfNeeded(targetEmail: targetEmail, tokenAccountID: tokenAccountID)
+        let preferredStoreEmail = self.codexOpenAIWebStoreEmail()
 
         let now = Date()
         let minInterval = self.openAIWebRefreshIntervalSeconds()
@@ -780,7 +782,7 @@ extension UsageStore {
             let normalized = targetEmail?
                 .trimmingCharacters(in: .whitespacesAndNewlines)
                 .lowercased()
-            var effectiveEmail = targetEmail
+            var effectiveEmail = targetEmail ?? preferredStoreEmail
 
             // Use a per-email persistent `WKWebsiteDataStore` so multiple dashboard sessions can coexist.
             // Strategy:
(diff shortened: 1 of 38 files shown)


Sources/CodexBar/UsageStore.swift
```diff
@@ -1109,6 +1150,11 @@ extension UsageStore {
     }
 
     func codexAccountEmailForOpenAIDashboard() -> String? {
+        if self.settings.codexCookieSource == .manual,
+           !self.settings.tokenAccounts(for: .codex).isEmpty
+        {
+            return nil
```
ratulsarna, line 1231: Could this `targetEmail` becoming `nil` in manual token-account mode interact with the account-switch path in a way that keeps stale dashboard state after switching accounts?

Sources/CodexBar/UsageStore.swift
```diff
@@ -891,23 +893,39 @@ extension UsageStore {
...
-        self.lastOpenAIDashboardTargetEmail = normalized
+        let previousTokenAccountID = self.lastOpenAIDashboardTargetTokenAccountID
+        self.lastOpenAIDashboardTargetEmail = normalizedOrNil
+        self.lastOpenAIDashboardTargetTokenAccountID = tokenAccountID
 
-        if let previous,
-           !previous.isEmpty,
-           previous != normalized
-        {
+        let emailChanged = if let previous, !previous.isEmpty, let normalizedOrNil {
```
ratulsarna, line 913: Should transitions like `previous email -> nil` or `previous tokenAccountID nil -> non-nil` be treated as account changes here, or is there a reason we want to keep existing dashboard state in that first switch case?

Sources/CodexBarCore/CookieHeaderNormalizer.swift
```diff
@@ -2,14 +2,14 @@ import Foundation
 
 public enum CookieHeaderNormalizer {
     private static let headerPatterns: [String] = [
-        #"(?i)-H\s*'Cookie:\s*([^']+)'"#,
-        #"(?i)-H\s*\"Cookie:\s*([^\"]+)\""#,
+        #"(?i)(?:^|\s)-H\s+'Cookie:\s*([^']+)'"#,
```
ratulsarna, line 5: Are we intentionally requiring whitespace after -H now, or should compact forms like -H'Cookie: ...' still be accepted?

Sources/CodexBarCore/CookieHeaderNormalizer.swift
```diff
@@ -2,14 +2,14 @@ import Foundation
...
+        #"(?i)(?:^|\s)-H\s+\"Cookie:\s*([^\"]+)\""#,
         #"(?i)\bcookie:\s*'([^']+)'"#,
         #"(?i)\bcookie:\s*\"([^\"]+)\""#,
         #"(?i)\bcookie:\s*([^\r\n]+)"#,
-        #"(?i)(?:--cookie|-b)\s*'([^']+)'"#,
-        #"(?i)(?:--cookie|-b)\s*\"([^\"]+)\""#,
-        #"(?i)(?:--cookie|-b)\s*([^\s]+)"#,
+        #"(?i)(?:^|\s)(?:--cookie|-b)\s+'([^']+)'"#,
```
ratulsarna, line 10: For --cookie / -b, do we still want to accept compact input like -b'...', or should it only match when there is whitespace?

PR #494, opened 2026-03-07
Refactor Claude provider planning foundations

ratulsarna:
- locked down current behavior with baseline docs/tests
- centralized plan inference, credential routing, and source planning
- split the fetcher and Auth store into clearer internal responsibilities while preserving existing behavior


ratulsarna:
@codex review

PR #497, opened 2026-03-09
Fix Gemini OAuth: resolve multi-level symlink chain

LeoLin990405:
## Summary

- **Bug**: Homebrew-installed `gemini` CLI has a 4-level symlink chain: `/usr/local/bin/gemini` → `/opt/homebrew/bin/gemini` → `../Cellar/gemini-cli/0.32.1/bin/gemini` → `../libexec/bin/gemini`. CodexBar only resolves one level using `destinationOfSymbolicLink`, so it lands at `/opt/homebrew/bin/` which lacks the `libexec/` tree containing `oauth2.js`. This causes Gemini OAuth credential detection to fail silently.
- **Fix**: Recursive symlink resolution that walks the entire chain, collecting all intermediate paths. It then tries each resolved directory (deepest first) to locate the OAuth credentials file, ensuring the correct `libexec/` path is found regardless of how many symlink levels Homebrew uses.

## Test plan

- [x] Tested with Homebrew gemini-cli 0.32.1 on macOS (4-level symlink chain)
- [x] Verified OAuth credentials are correctly located at the deepest resolved path
- [x] Confirmed backward compatibility with direct (non-symlinked) gemini installations

### DIFF
--- a/Sources/CodexBarCore/Providers/Gemini/GeminiStatusProbe.swift
+++ b/Sources/CodexBarCore/Providers/Gemini/GeminiStatusProbe.swift
@@ -451,11 +451,22 @@ public struct GeminiStatusProbe: Sendable {
             return nil
         }
 
+        guard let content = Self.resolveOAuthFileContent(from: geminiPath) else {
+            return nil
+        }
+        return self.parseOAuthCredentials(from: content)
+    }
+
+    /// Resolve the full symlink chain starting from `binaryPath`, then search each
+    /// intermediate directory for the Gemini CLI oauth2.js file.
+    /// Returns the file content if found, or `nil`.
+    /// Visible to tests via `@testable import`.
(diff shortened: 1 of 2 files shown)


ratulsarna:

ratulsarna:
Thanks again for digging into this, @LeoLin990405. The extra `/usr/local/bin/gemini -> /opt/homebrew/bin/gemini` symlink case was a valid edge case, and the regression-test work here was useful.

Since this PR was opened, `main` has changed the Gemini OAuth discovery path to fully resolve the Gemini binary with `resolvingSymlinksInPath()` and then discover credentials from the resolved package root / legacy layouts. I re-checked the current behavior against a Homebrew Gemini install, and the runtime issue this PR targeted is now covered by the newer path.

This branch is also stale/conflicted now, so I’m going to close it as functionally superseded to keep the queue tidy. Appreciate the clear repro and thoughtful follow-through.

Issue #340, opened 2026-02-09
Keychain password prompt repeats every few hours despite 'Always Allow'


lechatlisse:
## Description

CodexBar repeatedly prompts for the macOS login keychain password every few hours, despite clicking "Always Allow" each time. The ACL setting does not persist.

## Environment

- **CodexBar version**: 0.18.0-beta.2 (Build 50)
- **macOS**: 15.3 (Darwin 25.2.0)
- **Architecture**: Apple Silicon

## Affected keychain entries

Both entries under service `com.steipete.codexbar.cache`:
- Account: `oauth.claude`
- Account: `cookie.claude`

## Root cause analysis

The `oauth.claude` entry is being modified frequently (token refresh cycle). Based on the keychain dump, the entry appears to be **deleted and recreated** rather than updated in place. This wipes the ACL on every token refresh, causing macOS to prompt again on the next read.

Evidence:
- CodexBar is already listed in the Access Control tab for the entry
- "Always Allow" is selected each time but does not persist
- The app binary has not changed (signed Jan 21, codesign verification passes)
- The `oauth.claude` entry's `mdat` (modification date) changes frequently, suggesting regular token refresh writes

## Expected behavior

Clicking "Always Allow" should permanently grant CodexBar access without further prompts.

## Suggested fix

Use `SecKeychainItemModifyAttributesAndData` (or the Swift equivalent `SecItemUpdate`) to update the existing keychain entry in place, rather than deleting and recreating it. Updating in place preserves the user-configured ACL.

philliptiongson:
How do you get this fix to appear? I'm running v 0.18.0-beta.3 (51). Do I have to build the app myself?

TheAIMogul:
July 2026 this still persists. On version 0.41.0 (100) 
Mac OS 26.0 (25A354)

gwsiny:
July 15 2026 this still persists. On version 0.43.0 (104)

TheAIMogul:
As end users anything we can do manually?

jamesblackwell:
This is still an issue on latest version

GARMTECH:
We have the same problem

pyLancelle:
Hi ! Still an issue.

BenjaminGittins:
Still an issue here, its pushing me towards uninstalling :(

Altamimi-Dev:
is there a fix for this it is so annoying, pushing me towards uninstalling it

urda:
[...]

PR #499, opened 2026-03-09
Fix missing task cancellation and legacy keychain fingerprint mismatch

nikhil8182:
## Summary

- **Cancel `pathDebugRefreshTask` in `UsageStore.deinit`** — this task was missing from the deinit cleanup, causing it to continue running after deallocation (the `[weak self]` prevents crashes but the task does unnecessary background work)
- **Fix legacy keychain fingerprint mismatch in `ClaudeOAuthCredentials`** — two call sites (`loadFromClaudeKeychainUsingSecurityFramework` and `syncFromClaudeKeychainWithoutPrompt`) used `currentClaudeKeychainFingerprintWithoutPrompt()` after a successful legacy data read. This function prefers the newest *candidate* and may return a fingerprint for a different keychain item than the one actually read via the legacy service-level query. Now builds the fingerprint from `claudeKeychainLegacyCandidateWithoutPrompt()` directly, matching the pattern already used in the candidate path

## Context

The fingerprint mismatch caused `syncWithClaudeKeychainIfChanged` to see a "changed" fingerprint on subsequent checks, triggering unnecessary keychain re-reads. This is related to #340 — while not the root cause (which is Claude Code's delete+recreate ACL wipe), it reduced the effectiveness of the cache layer that mitigates the repeated prompts.

## Test plan

- [x] `swift build` passes cleanly
- [ ] Verify `pathDebugRefreshTask` is properly canceled on store deallocation
- [ ] Verify legacy keychain recovery path stores correct fingerprint, preventing unnecessary background sync re-reads

### DIFF
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeOAuth/ClaudeOAuthCredentials.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeOAuth/ClaudeOAuthCredentials.swift
@@ -1067,8 +1067,17 @@ public enum ClaudeOAuthCredentialsStore {
                 promptMode: promptMode),
                 !data.isEmpty
             {
-                // Same as above: store fingerprint after interactive read to avoid background "sync" reads.
-                self.saveClaudeKeychainFingerprint(self.currentClaudeKeychainFingerprintWithoutPrompt())
+                // Build fingerprint from the legacy candidate directly rather than using
+                // currentClaudeKeychainFingerprintWithoutPrompt(), which prefers the newest
+                // candidate and may return a fingerprint for a different keychain item.
+                if let legacyCandidate = self.claudeKeychainLegacyCandidateWithoutPrompt(
+                    promptMode: promptMode)
+                {
+                    self.saveClaudeKeychainFingerprint(ClaudeKeychainFingerprint(
+                        modifiedAt: legacyCandidate.modifiedAt.map { Int($0.timeIntervalSince1970) },
+                        createdAt: legacyCandidate.createdAt.map { Int($0.timeIntervalSince1970) },
+                        persistentRefHash: Self.sha256Prefix(legacyCandidate.persistentRef)))
+                }
                 return data
             }
         } catch let error as ClaudeOAuthCredentialsError {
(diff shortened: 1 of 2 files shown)


Sources/CodexBarCore/Providers/Claude/ClaudeOAuth/ClaudeOAuthCredentials.swift
```diff
@@ -1067,8 +1067,17 @@ public enum ClaudeOAuthCredentialsStore {
                 promptMode: promptMode),
                 !data.isEmpty
             {
-                // Same as above: store fingerprint after interactive read to avoid background "sync" reads.
-                self.saveClaudeKeychainFingerprint(self.currentClaudeKeychainFingerprintWithoutPrompt())
+                // Build fingerprint from the legacy candidate directly rather than using
```
ratulsarna, line 1070: Could saving the legacy candidate's fingerprint here get out of sync with the fingerprint that later change detection treats as current when duplicate keychain entries exist?

Sources/CodexBarCore/Providers/Claude/ClaudeOAuth/ClaudeOAuthCredentials.swift
```diff
@@ -1592,7 +1601,19 @@ extension ClaudeOAuthCredentialsStore {
            let creds = try? ClaudeOAuthCredentials.parse(data: legacyData),
            !creds.isExpired
         {
-            self.saveClaudeKeychainFingerprint(self.currentClaudeKeychainFingerprintWithoutPrompt())
+            // Build fingerprint from the legacy candidate directly, matching the candidate
```
ratulsarna, line 1604: Could this fallback path end up with the same mismatch too, where we save one item's fingerprint but later background checks compare against another?

Sources/CodexBarCore/Providers/Claude/ClaudeOAuth/ClaudeOAuthCredentials.swift
```diff
@@ -1592,7 +1601,19 @@ extension ClaudeOAuthCredentialsStore {
...
+            // path above. currentClaudeKeychainFingerprintWithoutPrompt() prefers the newest
+            // candidate and may return a fingerprint for a different keychain item, causing
+            // stale comparisons on subsequent background sync checks.
+            if let legacyCandidate = self.claudeKeychainLegacyCandidateWithoutPrompt(
+                promptMode: fallbackPromptMode)
+            {
+                let legacyFingerprint = ClaudeKeychainFingerprint(
```
ratulsarna, line 1611: Would it be worth pulling this fingerprint construction into a small helper? I'm wondering if that would also help with the file-length lint failure on this file.

ratulsarna (commented):
A few questions to sanity-check this change.

ratulsarna:
Thanks for the contribution. I’m doing a backlog cleanup, and this PR still has unresolved review feedback around the cancellation and fallback behavior plus substantial drift from current main, so I’m going to close it for now. If you want to revive it, please rebase onto current main, address the outstanding comments, and reopen or submit a fresh PR.

PR #508, opened 2026-03-11
Avoid false signed-out detection on MiniMax coding-plan shell pages

qipihen:
## Summary
- fix the existing MiniMax web fetch path so shell pages are not misclassified as signed out just because login strings appear inside script/style payloads
- strip non-visible HTML content before checking for `sign in` / `log in` / `登录` / `登入`
- add a MiniMax web fetch regression test that exercises the fallback from a `__NEXT_DATA__` shell page to the remains API

## Scope
This PR does not add MiniMax support as a new provider.
It fixes the MiniMax web-fetch path that already exists in the current upstream codebase.

## Why
MiniMax's current Coding Plan page can return an SSR shell where login-related translation strings live inside `__NEXT_DATA__`, even though the page is not actually the sign-in screen. The previous raw-string check saw `登录` in the script payload and returned `invalidCredentials` before the fetcher could fall back to `coding_plan/remains`.

## Validation
- reproduced locally with a live MiniMax web session and confirmed the old logic returned `MiniMax credentials are invalid or expired`
- verified the patched local build successfully reports MiniMax usage again from the web source
- attempted `swift test --disable-experimental-prebuilts --filter MiniMax` on a clean local checkout; that run is currently blocked here by an unrelated `KeyboardShortcuts` `#Preview` macro build issue, not by this diff

### DIFF
--- a/Sources/CodexBarCore/Providers/MiniMax/MiniMaxUsageFetcher.swift
+++ b/Sources/CodexBarCore/Providers/MiniMax/MiniMaxUsageFetcher.swift
@@ -320,8 +320,27 @@ public struct MiniMaxUsageFetcher: Sendable {
     }
 
     private static func looksSignedOut(html: String) -> Bool {
-        let lower = html.lowercased()
+        let lower = self.visibleText(from: html).lowercased()
         return lower.contains("sign in") || lower.contains("log in") || lower.contains("登录") || lower.contains("登入")
+    }
+
+    private static func visibleText(from html: String) -> String {
+        let patterns = [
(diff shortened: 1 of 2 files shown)


ratulsarna:

steipete:
Landed the narrow fix on main in 0966118e.

What changed: MiniMax signed-out detection now strips scripts/styles/comments/tags before looking for login text, so i18n strings inside `__NEXT_DATA__` no longer produce false signed-out results. Added regression coverage for hidden script login text and visible login text.

Verified with `swift test`, `pnpm check`, `./Scripts/compile_and_run.sh`, and green CI after rerun. Thanks @qipihen!

PR #511, opened 2026-03-12
Preserve exact GPT-5 Codex pricing keys

iam-brain:
## Summary
- Stop collapsing Codex/OpenAI GPT-5 model IDs onto base GPT-5 entries during pricing lookup. We were previously stripping `-codex` and effectively pricing persisted usage as a different model than the one the scanner recorded, which made local cost reports harder to trust and hid model-specific behavior in breakdowns.
- Expand the Codex pricing catalog to include the missing non-chat GPT-5 family entries that now appear in real persisted usage data. This keeps pricing aligned with the exact model IDs we ingest instead of relying on partial aliases or nonexistent base-model fallbacks.
- Treat `gpt-5.3-codex-spark` as a zero-cost Research Preview case. The goal is to keep preview usage visible in scanner output, CLI payloads, and chart details without incorrectly charging it or dropping it from breakdowns just because the cost is zero.
- Preserve exact model IDs end to end through normalization, scanner aggregation, JSON payloads, and cost-history detail rendering. We took the explicit-table approach here so the supported pricing surface is obvious in code, dated suffixes are only stripped when the exact undated model exists, and we do not silently map future GPT-5 variants onto the wrong model.

## Validation
- `swift test --filter 'CostUsagePricingTests|CostUsageScannerTests|CostUsageDecodingTests|CLICostTests|UsageFormatterTests'`
- `pnpm check`

### DIFF
--- a/Sources/CodexBarCore/Vendored/CostUsage/CostUsageCache.swift
+++ b/Sources/CodexBarCore/Vendored/CostUsage/CostUsageCache.swift
@@ -1,6 +1,15 @@
 import Foundation
 
 enum CostUsageCacheIO {
+    private static func artifactVersion(for provider: UsageProvider) -> Int {
+        switch provider {
+        case .codex:
+            2
+        default:
+            1
+        }
+    }
+
     private static func defaultCacheRoot() -> URL {
         let root = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first!
         return root.appendingPathComponent("CodexBar", isDirectory: true)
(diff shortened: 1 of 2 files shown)


Sources/CodexBar/CostHistoryChartMenuView.swift
```diff
@@ -151,7 +151,7 @@ struct CostHistoryChartMenuView: View {
         var peak: (key: String, costUSD: Double)?
         var maxCostUSD: Double = 0
         for entry in sorted {
-            guard let costUSD = entry.costUSD, costUSD > 0 else { continue }
+            guard let costUSD = entry.costUSD, costUSD >= 0 else { continue }
```
ratulsarna, line 154: Now that zero-cost days are included, the first all-zero day becomes the chart “peak” and gets the yellow cap highlight. That’s misleading for preview-only usage. I’d suppress peakKey when maxCostUSD == 0.

iam-brain: Good catch, small fix incoming!

Sources/CodexBarCore/Vendored/CostUsage/CostUsagePricing.swift
```diff
@@ -169,13 +222,25 @@ enum CostUsagePricing {
         if trimmed.hasPrefix("openai/") {
             trimmed = String(trimmed.dropFirst("openai/".count))
         }
-        if let codexRange = trimmed.range(of: "-codex") {
-            let base = String(trimmed[..<codexRange.lowerBound])
-            if self.codex[base] != nil { return base }
+
+        if self.codex[trimmed] != nil {
+            return trimmed
```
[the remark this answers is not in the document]
iam-brain: Whoops, missed bringing that over from #510.

PR #513, opened 2026-03-13
Introduce vertical cost detail list

iam-brain:
## Summary
This PR now focuses only on the cost chart hover UI change: it replaces the single secondary detail line with a vertical per-model list in the chart detail area.

The branch intentionally keeps `main`'s current data behavior. There are no scanner, decoding, CLI payload, or model-shape changes in this version.

## What changed
- `CostHistoryChartMenuView` now shows a stacked detail block under the chart.
- The selected day still renders the existing primary summary line for date, cost, and tokens when available.
- Model breakdowns render as a vertical list with accent bars and per-row cost detail.
- The list is capped at four visible rows.
- The menu reserves detail height based on the busiest loaded day so hover movement does not resize the menu.

## What did not change
- No changes to `CostUsageModels`.
- No changes to the cost scanners.
- No changes to decoding or CLI payloads.
- No pricing or total-cost behavior changes.

## Verification
- [x] `pnpm check`
- [x] `swift test`
- [x] `./Scripts/compile_and_run.sh`

Screenshot 2026-03-17 at 3 13 45 PM

### DIFF
--- a/Sources/CodexBar/CostHistoryChartMenuView.swift
+++ b/Sources/CodexBar/CostHistoryChartMenuView.swift
@@ -371,7 +371,12 @@ struct CostHistoryChartMenuView: View {
                 let lCost = lhs.costUSD ?? -1
                 let rCost = rhs.costUSD ?? -1
                 if lCost != rCost { return lCost > rCost }
-                return lhs.modelName < rhs.modelName
+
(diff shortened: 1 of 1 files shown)


ratulsarna:
Thanks @iam-brain for the contribution!

PR #515, opened 2026-03-13
[Feature] Add separate progress bars status item

johnyoonh:
## Summary

Adds a new display option to show dual progress bars as a separate menu bar status item, independent from the provider icon.

## Changes

- New toggle 'Show separate progress bars' in Display settings (only visible when 'Merge Icons' is enabled)
- Creates a second status item showing only the dual progress bars (no icon, no text)
- Both status items update with the same provider data
- Users can drag to reposition the bars independently in macOS menu bar
- Fixed rendering artifacts when provider changes

## Use Cases

- Keep progress bars visible at all times while the icon changes
- Minimalist view: show only bars in one area, only icon in another
- Redundancy: quick glance at bars from one position, detailed info from clicking the icon

## Testing

- Verified bars render correctly without artifacts when provider changes
- Tested toggle on/off behavior
- Confirmed independent positioning in menu bar

## Note

This is a feature request implementation, not a bug fix.

### DIFF
--- a/Sources/CodexBar/StatusItemController+Animation.swift
+++ b/Sources/CodexBar/StatusItemController+Animation.swift
@@ -161,6 +161,7 @@ extension StatusItemController {
         if mergeIcons {
             let phase: Double? = self.needsMenuBarIconAnimation() ? self.animationPhase : nil
             self.applyIcon(phase: phase)
+            self.applySeparateBarsIcon(phase: phase)
         }
     }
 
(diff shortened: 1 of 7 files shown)


Sources/CodexBar/StatusItemController.swift
```diff
@@ -362,6 +366,19 @@ final class StatusItemController: NSObject, NSMenuDelegate, StatusItemControllin
         let anyEnabled = !self.store.enabledProviders().isEmpty
         let force = self.store.debugForceAnimation
         let mergeIcons = self.shouldMergeIcons
+        let showsSeparateBars = self.settings.menuBarShowsSeparateBars && mergeIcons
+        
+        // Update separate bars status item visibility
+        if showsSeparateBars {
+            if self.separateBarsStatusItem == nil {
+                self.separateBarsStatusItem = self.statusBar.statusItem(withLength: NSStatusItem.variableLength)
```
ratulsarna, line 374: Could we verify how interactions are wired for this new status item? I'm not seeing a `menu` or action assigned to `separateBarsStatusItem`, so I'm wondering whether it may appear in the menu bar but not respond when clicked.

Sources/CodexBar/StatusItemController.swift
```diff
@@ -338,6 +341,7 @@ final class StatusItemController: NSObject, NSMenuDelegate, StatusItemControllin
         let phase: Double? = self.needsMenuBarIconAnimation() ? self.animationPhase : nil
         if self.shouldMergeIcons {
             self.applyIcon(phase: phase)
+            self.applySeparateBarsIcon(phase: phase)
```
ratulsarna, line 344: Could we double-check the startup path for this item when the setting is already enabled? `updateIcons()` appears to run before `updateVisibility()` in `init`, and `applySeparateBarsIcon()` returns early until the item exists, so I'm wondering whether this status item could be visible but initially blank until a later refresh occurs.

Sources/CodexBar/StatusItemController+Animation.swift
```diff
@@ -322,6 +323,99 @@ extension StatusItemController {
...
+            } else {
+                primary = max(pattern.value(phase: phase), Self.loadingPercentEpsilon)
+                weekly = max(pattern.value(phase: phase + pattern.secondaryOffset), Self.loadingPercentEpsilon)
+                stale = false
+            }
+        }
+        
+        let style: IconStyle = self.store.style(for: primaryProvider)
+        let isLoading = phase != nil && self.shouldAnimate(provider: primaryProvider)
+        let blink: CGFloat = {
```
ratulsarna, line 383: Should we also review the code path that clears blink and motion state for this new item? Since this renderer now depends on `blink`, `wiggle`, and `tilt`, I'm wondering whether the separate bars icon could remain on its last animated frame if blinking stops without an immediate redraw.

ratulsarna (commented):
A few questions to double-check the new separate bars status item behavior.

ratulsarna:
Thanks for the contribution. I’m doing a backlog cleanup, and this PR still has unresolved review feedback around startup/rendering behavior and interaction details, plus substantial drift from current main, so I’m going to close it for now. If you want to revive it, please rebase onto current main, address the outstanding comments, and reopen or submit a fresh PR.

Issue #485, opened 2026-03-05
Keychain prompt for Claude Code-credentials reappears repeatedly despite Always Allow


Korkyzer:
After granting 'Always Allow' access to the 'Claude Code-credentials' keychain item, the prompt reappears on every refresh/restart. Running latest CodexBar via Homebrew on macOS, Claude Max plan.

ngutman:
Thanks for the report. I’m closing this as a duplicate/superseded issue in the broader Claude keychain prompt family. The repeated “Always Allow” / prompt reappears behavior was already tracked and investigated in #340, and the current canonical open issue for related Claude keychain prompt behavior is #458.

If you still hit this on the latest release, please continue there or open a fresh issue with current version details.

PR #516, opened 2026-03-13
Improve development build experience

johnyoonh:
## Summary

Improves development build scripts for smoother workflow.

### compile_and_run.sh
- Automatically source `.envrc` for `APP_IDENTITY` environment variable
- Enables consistent code signing with development certificate
- Eliminates need to manually export APP_IDENTITY before each build

### package_app.sh
- Add privacy usage descriptions to Info.plist:
  - **Apple Events**: Interact with other apps for URLs/sessions
  - **Desktop/Documents/Downloads**: Open configuration and log files
  - **Photo Library/Apple Music**: Explicitly state not used (framework requirement)

## Benefits

- Consistent code signing across rebuilds
- Clear privacy descriptions instead of generic dialogs
- Streamlined development workflow

## Note

This PR improves the development experience but does NOT directly fix issue #485 (Claude keychain prompt cycle). That issue requires changes to Claude OAuth keychain handling logic.

### DIFF
--- a/Scripts/compile_and_run.sh
+++ b/Scripts/compile_and_run.sh
@@ -4,6 +4,16 @@
 set -euo pipefail
 
 ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+
+# Source .envrc if it exists to load APP_IDENTITY and other environment variables
+# Note: This executes all commands in .envrc, but direnv requires explicit user approval
+# via 'direnv allow', so this is safe for development use.
+if [[ -f "${ROOT_DIR}/.envrc" ]]; then
+    # shellcheck disable=SC1091
+    source "${ROOT_DIR}/.envrc"
+    export APP_IDENTITY
+fi
+
 APP_BUNDLE="${ROOT_DIR}/CodexBar.app"
 APP_PROCESS_PATTERN="CodexBar.app/Contents/MacOS/CodexBar"
 DEBUG_PROCESS_PATTERN="${ROOT_DIR}/.build/debug/CodexBar"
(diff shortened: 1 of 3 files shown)


Scripts/compile_and_run.sh
```diff
@@ -4,6 +4,16 @@
 set -euo pipefail
 
 ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+
+# Source .envrc if it exists to load APP_IDENTITY and other environment variables
+# Note: This executes all commands in .envrc, but direnv requires explicit user approval
+# via 'direnv allow', so this is safe for development use.
+if [[ -f "${ROOT_DIR}/.envrc" ]]; then
+    # shellcheck disable=SC1091
+    source "${ROOT_DIR}/.envrc"
```
ratulsarna, line 13: Could we verify whether sourcing `.envrc` directly here bypasses direnv's approval model, so a branch-specific change to that file would be executed whenever this script runs?

ratulsarna, line 13: Could we confirm that the `.envrc` files we expect contributors to use are valid to source in plain `bash`, rather than relying on direnv-only helpers such as `use` or `PATH_add` that would cause this script to fail under `set -e`?

ratulsarna:
Thanks for the contribution. I’m doing a backlog cleanup, and this PR still has unresolved review feedback around the development environment and shell-loading behavior, plus substantial drift from current main, so I’m going to close it for now. If you want to revive it, please rebase onto current main, address the outstanding comments, and reopen or submit a fresh PR.

PR #517, opened 2026-03-13
Add Claude CLI fallback for data load failures

johnyoonh:
## Summary

Improves Claude provider reliability by automatically falling back to web/oauth authentication when CLI fails with data load errors.

## Changes

- Detect 'could not load usage data' errors from Claude CLI
- Automatically fallback to web/oauth authentication methods
- Improves reliability when CLI has transient issues
- Preserves existing fallback behavior for other error types

## Testing

- Verified fallback triggers on 'Claude CLI could not load usage data' error
- Confirmed other error types still show appropriate messages
- Tested web/oauth fallback successfully retrieves usage data

## Note

This is a general reliability improvement. Not tied to a specific issue, but helps when users encounter 'Could not parse Claude usage: Claude CLI could not load usage data. Open the CLI and retry `/usage`.' error.

### DIFF
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeProviderDescriptor.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeProviderDescriptor.swift
@@ -334,9 +334,22 @@ struct ClaudeCLIFetchStrategy: ProviderFetchStrategy {
             sourceLabel: "claude")
     }
 
-    func shouldFallback(on _: Error, context: ProviderFetchContext) -> Bool {
+    func shouldFallback(on error: Error, context: ProviderFetchContext) -> Bool {
+        // For CLI data load failures, fallback only when alternate strategies are available
+        let errorDesc = error.localizedDescription.lowercased()
+        if errorDesc.contains("could not load usage data")
+            || errorDesc.contains("failed to load usage data")
+        {
+            // Only fallback in app+auto mode where web/oauth strategies exist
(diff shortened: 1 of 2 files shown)


Sources/CodexBarCore/Providers/Claude/ClaudeProviderDescriptor.swift
```diff
@@ -334,9 +334,22 @@ struct ClaudeCLIFetchStrategy: ProviderFetchStrategy {
             sourceLabel: "claude")
     }
 
-    func shouldFallback(on _: Error, context: ProviderFetchContext) -> Bool {
+    func shouldFallback(on error: Error, context: ProviderFetchContext) -> Bool {
+        // For CLI data load failures, fallback only when alternate strategies are available
+        let errorDesc = error.localizedDescription.lowercased()
+        if errorDesc.contains("could not load usage data")
```
ratulsarna, line 340: Could we walk through how this branch changes the fallback behavior from before? I’m reading both this special case and the default path as still falling back only in app + auto when web is available, so I may be missing what user-visible case now behaves differently.

ratulsarna:

ratulsarna:
Thanks for the contribution. I rechecked this against current main, and the Claude CLI strategy now already falls back to web auth in app + auto mode when web auth is available, so this PR no longer adds distinct behavior for the usage-data-load failure case. I’m going to close it as superseded by the current implementation.

PR #518, opened 2026-03-13
fix: persist refresh interval to UserDefaults on first launch

### LINKED ISSUE #506: Bug: Refresh interval setting not persisted to UserDefaults (resets to default on every launch)
## Bug Report

**CodexBar version**: 0.17.0
**macOS**: 15.3.2 (Sequoia), Apple Silicon (arm64)

## Summary

The refresh interval setting in Settings is not written to `com.steipete.codexbar` UserDefaults. Every time CodexBar restarts, the interval resets to the default value, requiring the user to manually go into Settings to set it again.

## Steps to Reproduce

1. Open CodexBar Settings
2. Change the refresh interval (e.g., to 1 minute)
3. Close Settings — data refreshes correctly while app is running
4. Quit and relaunch CodexBar
5. Refresh interval has reset to default

## Verification

Monitored `com.steipete.codexbar` plist before and after changing the setting:

```bash
defaults read com.steipete.codexbar > before.txt
# (changed setting in UI)
defaults read com.steipete.codexbar > after.txt
diff before.txt after.txt
# → no output (plist unchanged)
```

The plist file's modification timestamp does update on Settings changes, but the refresh interval key is absent from the output both before and after.

## Expected Behavior

Refresh interval should be persisted to UserDefaults (like other settings such as `menuBarShowsBrandIconWithPercent`, `launchAtLogin`, etc.) so it survives app restarts.

## Workaround

None available without knowing the internal key name to write manually via `defaults write`.

Astro-Han:
## Summary

- Write the resolved `refreshFrequency` value to UserDefaults when the key is absent or contains an unrecognized raw value, matching the pattern used by `sessionQuotaNotificationsEnabled` and other settings
- Add test coverage for both the first-launch write-back and invalid raw value repair scenarios

Closes #506

## Test plan

- [x] `swift build` passes
- [x] `swift test` — all SettingsStore tests pass (923 total, 7 pre-existing OAuth/keychain environment failures unrelated to this change)
- [ ] Manual: launch CodexBar, verify `defaults read com.steipete.codexbar | grep refreshFrequency` shows the key
- [ ] Manual: change refresh cadence in Settings, quit and relaunch, verify the setting persists

### DIFF
--- a/Sources/CodexBar/SettingsStore.swift
+++ b/Sources/CodexBar/SettingsStore.swift
@@ -157,9 +157,10 @@ final class SettingsStore {
 
 extension SettingsStore {
     private static func loadDefaultsState(userDefaults: UserDefaults) -> SettingsDefaultsState {
-        let refreshRaw = userDefaults.string(forKey: "refreshFrequency")
-        let refreshFrequency = refreshRaw.flatMap(RefreshFrequency.init(rawValue:)) ?? .fiveMinutes
-        if refreshRaw == nil || refreshRaw.flatMap(RefreshFrequency.init(rawValue:)) == nil {
+        let refreshDefault = userDefaults.string(forKey: "refreshFrequency")
+            .flatMap(RefreshFrequency.init(rawValue:))
+        let refreshFrequency = refreshDefault ?? .fiveMinutes
+        if refreshDefault == nil {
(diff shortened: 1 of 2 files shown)


Astro-Han:
Recreating without attribution markers.

PR #525, opened 2026-03-13
fix: show all enabled providers in merged switcher

### LINKED ISSUE #203: Max Provider View Issue
Installed the Beta and the tool only shows a max of 6 providers.

Image

I've added a 7th but it doesn't show in the tool.

Image

If I remove a 7th the provider view doesn't update. I removed Droid as a provider but it still shows in the tool

Image

Astro-Han:
## Summary

- Adds `enabledProvidersForDisplay()` to `UsageStore` that returns enabled providers **without** filtering by credential availability
- Switcher data source (`populateMenu`) and merge-icons logic (`shouldMergeIcons`) now use the display list, so all user-enabled providers appear in the merged menu regardless of whether their credentials are configured
- Providers without credentials show "No usage yet" when selected (existing behavior)

## Context

The merged menu switcher previously used `enabledProviders()`, which filters out providers whose credentials aren't configured via `isProviderAvailable()`. This silently hid user-enabled providers from the switcher UI, making it appear that adding a provider had no effect.

This is a minimal, defensive fix — data fetch, animations, widgets, and separate-mode logic continue to use the availability-filtered list.

Closes #203

## Test plan

- [x] `swift build` passes
- [x] `swift test` passes (8 pre-existing keychain-related failures on CI/local, unrelated)
- [ ] Manual: enable N providers (including some without credentials), verify all appear in merged switcher
- [ ] Manual: select a provider without credentials, verify "No usage yet" appears

### DIFF
--- a/Sources/CodexBar/StatusItemController+Animation.swift
+++ b/Sources/CodexBar/StatusItemController+Animation.swift
@@ -499,7 +499,9 @@ extension StatusItemController {
                 return provider
             }
         }
-        if let enabled = self.store.enabledProvidersForDisplay().first {
+        // Use availability-filtered list: fallback must pick a provider that can
+        // actually animate, otherwise shouldAnimate() fails on credential-less providers.
+        if let enabled = self.store.enabledProviders().first {
             return enabled
         }
         return .codex
--- a/Sources/CodexBar/StatusItemController+Menu.swift
+++ b/Sources/CodexBar/StatusItemController+Menu.swift
@@ -623,7 +623,9 @@ extension StatusItemController {
         if let selected = self.selectedMenuProvider, enabled.contains(selected) {
             return selected
         }
-        return enabled.first
+        // Prefer an available provider so the default menu content matches the status icon.
+        // Falls back to first display provider when all lack credentials.
+        return enabled.first(where: { self.store.isProviderAvailable($0) }) ?? enabled.first
     }
 
     private func includesOverviewTab(enabledProviders: [UsageProvider]) -> Bool {


Sources/CodexBar/StatusItemController+Menu.swift
```diff
@@ -618,7 +618,7 @@ extension StatusItemController {
     }
 
     private func resolvedMenuProvider(enabledProviders: [UsageProvider]? = nil) -> UsageProvider? {
-        let enabled = enabledProviders ?? self.store.enabledProviders()
+        let enabled = enabledProviders ?? self.store.enabledProvidersForDisplay()
```
ratulsarna, line 621: Could this fallback select a provider that's user-enabled but currently unavailable? If so, is there a path where the merged menu opens on that empty provider even though a different provider is the one reflected in the menu bar icon?

Sources/CodexBar/StatusItemController+Animation.swift
```diff
@@ -499,7 +499,7 @@ extension StatusItemController {
                 return provider
             }
         }
-        if let enabled = self.store.enabledProviders().first {
+        if let enabled = self.store.enabledProvidersForDisplay().first {
```
ratulsarna: Could this end up choosing an unavailable display provider as the unified icon's primary provider? If that happens, is it possible for the merged icon to stop animating even while a later available provider is still loading?

Astro-Han:
@codex analyze the two review comments from ratulsarna:

1. **StatusItemController+Menu.swift:621** — `resolvedMenuProvider` fallback uses `enabledProvidersForDisplay()` which includes credential-less providers. Can this cause a mismatch where the merged menu shows an empty provider while the icon reflects a different available provider?

2. **StatusItemController+Animation.swift:502** — `primaryProviderForUnifiedIcon()` fallback uses `enabledProvidersForDisplay().first`. If an unavailable provider is selected, `shouldAnimate()` returns false because `isEnabled()` fails. Could this stop the merged icon animation while an available provider is still loading?

Please trace through the call chains and confirm whether these are actual bugs introduced by PR #525.

Sources/CodexBar/StatusItemController+Menu.swift
Astro-Han: Good catch — yes, when no `selectedMenuProvider` is set and a credential-less provider is first in order, the menu would show that empty provider while the icon reflects a different available one.

Fixed in 5d7e865: `resolvedMenuProvider` now prefers `enabled.first(where: { store.isProviderAvailable($0) })` before falling back to `enabled.first`. This way:
- Switcher tabs still list all user-enabled providers (the #203 fix intent)
- Default menu content aligns with the status bar icon
- If the user explicitly selects a credential-less tab, that choice is respected (`selectedMenuProvider` takes priority)
- All providers lacking credentials → falls back to `enabled.first` (same as before)

Sources/CodexBar/StatusItemController+Animation.swift
Astro-Han: Yes, this was a regression. The fallback could select an unavailable provider, causing `shouldAnimate()` to fail on `isEnabled()` and stop the merged icon animation while an available provider is still loading.

Fixed in 5d7e865: reverted the fallback to `enabledProviders().first` (availability-filtered). This line's purpose is to find a provider that can actually drive the icon — credential filtering is correct here. When no provider has credentials, it returns nil and falls through to the `.codex` default on the next line.

PR #527, opened 2026-03-13
fix: fall back to percent in Both mode when pace is unavailable

### LINKED ISSUE #232: Copilot doesn't show pace (Broken in "Both" display mode for pace)

Astro-Han:
## Summary
- Fixes #232. When a provider (e.g. Copilot) has no pace data, the "Both" display mode now gracefully degrades to show percent-only instead of hiding the menu bar text entirely.
- This aligns with the original intent from PR #169: "Falls back to percent display for providers that don't support pace."

## Changes
- `MenuBarDisplayText.swift`: `.both` branch returns `percent` instead of `nil` when pace is unavailable (one-line change)
- `StatusItemAnimationTests.swift`: Updated assertions and renamed test functions to match the new fallback behavior

## Test plan
- [x] `swift build` passes
- [x] All 6 `menuBarDisplayText*` tests pass with updated expectations
- [x] Full test suite shows no regressions from this change

### DIFF
--- a/Tests/CodexBarTests/StatusItemAnimationTests.swift
+++ b/Tests/CodexBarTests/StatusItemAnimationTests.swift
@@ -392,7 +392,7 @@ struct StatusItemAnimationTests {
     }
 
     @Test
-    func `menu bar display text hides when pace unavailable`() {
+    func `menu bar display text falls back to percent when pace unavailable`() {
         let percentWindow = RateWindow(usedPercent: 40, windowMinutes: nil, resetsAt: nil, resetDescription: nil)
 
         let pace = MenuBarDisplayText.displayText(
@@ -410,7 +410,7 @@ struct StatusItemAnimationTests {
     }
 
     @Test
-    func `menu bar display text requires provided pace for codex`() {
+    func `menu bar display text falls back to percent when pace nil for codex`() {
         let percentWindow = RateWindow(usedPercent: 40, windowMinutes: nil, resetsAt: nil, resetDescription: nil)
 
         let pace = MenuBarDisplayText.displayText(


Sources/CodexBar/MenuBarDisplayText.swift
```diff
@@ -29,8 +29,8 @@ enum MenuBarDisplayText {
             return self.paceText(pace: pace)
         case .both:
             guard let percent = percentText(window: percentWindow, showUsed: showUsed) else { return nil }
-            let paceText: String? = Self.paceText(pace: pace)
-            guard let paceText else { return nil }
+            // Fall back to percent-only when pace is unavailable (e.g. Copilot)
+            guard let paceText = Self.paceText(pace: pace) else { return percent }
```
ratulsarna, line 33: Could this fallback change what Codex shows when pace is unavailable? I’m wondering if returning percent here means .both no longer falls through to the credits display path in cases where the weekly/session window is exhausted, so it might be worth tracing how menuBarDisplayText(for:) behaves for exhausted Codex snapshots.

Astro-Han: Good catch — I traced this through and the credits fallback path is guarded by `mode == .percent` at `StatusItemController+Animation.swift:465`:

```swift
if provider == .codex,
   mode == .percent,   // ← hard gate
   ...
```

So when `mode == .both`, the credits branch is never entered regardless of what `displayText` returns. The two paths are fully isolated by the mode enum check.

The existing tests (`menuBarDisplayTextUsesCreditsWhenCodexWeeklyIsExhausted` / `...SessionIsExhausted`) also confirm this — they both explicitly set `menuBarDisplayMode = .percent`.

PR #529, opened 2026-03-14
Reduce battery drain from OpenAI web extras

### LINKED ISSUE #139: consuming too much power on my macbook

cbrane:
Closes #139

## Summary
This PR reduces the severe battery drain caused by CodexBar's optional OpenAI web extras.

The root issue was that CodexBar could keep loading and retrying the full `chatgpt.com/codex/settings/usage` single-page app in a hidden `WKWebView`. That is far too expensive for a menu bar app background feature, and in real usage it showed up in Activity Monitor as extreme energy impact.

This change keeps the normal Codex card working while making the expensive dashboard-only path safer, less likely to run unintentionally, and clearer in the UI.

## Root Cause
The OpenAI web extras feature was implemented by loading `https://chatgpt.com/codex/settings/usage` in a hidden `WKWebView` and scraping the hydrated DOM.

That caused three problems:

1. Failed dashboard refresh attempts were not throttled the same way successful attempts were, so stale cookies or login-required states could keep triggering repeated expensive work.
2. Hidden WebViews could stay alive with `chatgpt.com` still loaded, which kept an offscreen ChatGPT tab around longer than necessary.
3. The feature and its refresh behavior were coupled together, which made the product shape unclear once the battery-saving behavior was introduced.

We also found a separate background-refresh issue where scheduled work still walked broader provider lists than necessary.

### OpenAI web refresh and WebView lifecycle
- Added a refresh gate so recent failed OpenAI dashboard attempts are throttled too, not only successful snapshots.
- Evict cached OpenAI dashboard WebViews on login-required, account-mismatch, and reset/error paths.
- Blank released cached WebViews to `about:blank` so a hidden `chatgpt.com` page is not left active between uses.
- Reduced the hidden WebView cache lifetime from 10 minutes to 60 seconds.

### Product/default behavior
- `OpenAI web extras` is now off by default for new installs.
- Existing users with explicit Codex cookie/web configuration are preserved so upgrades do not silently break them.
- Added a nested `Battery Saver` option that only appears when OpenAI web extras are enabled.
- `Battery Saver` defaults to on, so users who enable OpenAI web extras get the safer low-power refresh path by default.
- When `Battery Saver` is on, the dashboard scrape stays on the reduced-refresh path from this PR: explicit/manual refreshes and submenu-driven stale checks.
- When `Battery Saver` is off, OpenAI web extras can refresh normally again.
- The settings copy now makes the battery/network tradeoff explicit.
- OpenAI web submenus are hidden when the feature is disabled, even if stale dashboard data exists in memory.

### Provider refresh scope
- Background refresh/status work now only runs for enabled providers.
- Token-cost refresh sequencing now only runs for enabled providers.
- Disabled-provider state is actively cleared so stale provider data does not keep hanging around.

### Manual refresh discoverability
- Added a visible `Refresh` action back to the main menu so the explicit refresh path is actually reachable from the UI.
- Routed the Codex Providers-pane refresh button through the full explicit refresh path when OpenAI web extras are enabled, so users can intentionally repopulate dashboard-only data.
- Kept disabled-provider refreshes scoped to the provider-only path.

### Documentation
- Updated Codex/provider docs to describe OpenAI web extras as optional and off by default.
- Added a troubleshooting/writeup document capturing the product rationale for the safer default.

## Why This Product Shape
I think there are two separate product questions here:

1. should OpenAI web extras exist at all?
2. if they exist, should they refresh normally or in a battery-saving mode?

This PR now separates those concerns explicitly:
- `OpenAI Web Extras` controls whether the optional dashboard-only data is available at all.
- `Battery Saver` controls whether those extras stay on the safer reduced-refresh path or refresh normally in the background.

That means users still keep the optional feature, but they are protected by default when they turn it on.

## What Users Keep Without OpenAI Web Extras
Users still keep the normal Codex data they check every day without the hidden ChatGPT dashboard:
- session usage
- weekly usage
- reset timers
- account email
- plan label
- normal credits remaining

What the optional web extras add is dashboard-only enrichment:
- code review remaining
- usage breakdown
- credits history
- dashboard purchase link / fallback enrichment

## Measured Before/After
These were real Activity Monitor observations while validating this branch:

### Before
- `CodexBar 12 hr Power`: `305.16`
- `CodexBar Energy Impact`: `83.2`
- Hidden `https://chatgpt.com` child process Energy Impact: `4608.6`

### After
After running the fixed branch for more than 48 hours:
- `CodexBar 12 hr Power`: `41.84`
- `CodexBar Energy Impact`: `0.0`

That is roughly an `86.3%` reduction versus the original `12 hr Power` reading.

### Passed
- `pnpm check`
- `swift test --filter OpenAIWebRefreshGateTests --filter OpenAIDashboardWebViewCacheTests --filter OpenAIWebAccountSwitchTests --filter UsageStoreCoverageTests --filter SettingsStoreTests --filter ProviderSettingsDescriptorTests --filter StatusMenuTests`
  - result: `86 tests in 7 suites passed`
- `swift test --filter StatusMenuTests --filter ProvidersPaneCoverageTests`
  - result: `26 tests in 2 suites passed`
- `swift test --filter SettingsStoreTests --filter ProviderSettingsDescriptorTests --filter OpenAIWebRefreshGateTests --filter ProvidersPaneCoverageTests --filter StatusMenuTests`
  - result: `77 tests in 5 suites passed`

### Broad suite note
A broad `swift test` run in this environment still ends with unrelated failures outside the menu/OpenAI web refresh path touched here.

## Files Of Interest
- `Sources/CodexBar/UsageStore+OpenAIWeb.swift`
- `Sources/CodexBar/UsageStore+BackgroundRefresh.swift`
- `Sources/CodexBar/MenuDescriptor.swift`
- `Sources/CodexBar/PreferencesProvidersPane.swift`
- `Sources/CodexBarCore/OpenAIWeb/OpenAIDashboardFetcher.swift`
- `Sources/CodexBarCore/OpenAIWeb/OpenAIDashboardWebViewCache.swift`
- `Sources/CodexBar/SettingsStore.swift`
- `Sources/CodexBar/Providers/Codex/CodexProviderImplementation.swift`
- `Tests/CodexBarTests/OpenAIWebRefreshGateTests.swift`
- `Tests/CodexBarTests/OpenAIDashboardWebViewCacheTests.swift`
- `Tests/CodexBarTests/UsageStoreCoverageTests.swift`
- `Tests/CodexBarTests/StatusMenuTests.swift`
- `Tests/CodexBarTests/ProvidersPaneCoverageTests.swift`
- `docs/solutions/performance-issues/openai-web-extras-default-off-codexbar-20260307.md`

## Follow-Up
If OpenAI exposes a stable lightweight endpoint for the dashboard-only extras, the long-term fix would be to replace the hidden WebView scrape with direct HTTP requests using imported cookies, similar to the Claude web path.


Sources/CodexBar/UsageStore.swift
```diff
@@ -409,11 +416,15 @@ final class UsageStore {
             // Token-cost usage can be slow; run it outside the refresh group so we don't block menu updates.
             self.scheduleTokenRefresh(force: forceTokenUsage)
 
-            // OpenAI web scrape depends on the current Codex account email (which can change after login/account
-            // switch). Run this after Codex usage refresh so we don't accidentally scrape with stale credentials.
-            await self.refreshOpenAIDashboardIfNeeded(force: forceTokenUsage)
+            // Keep the OpenAI web account state in sync with the current Codex identity, but avoid loading the
+            // full ChatGPT dashboard on the background timer. That scrape is expensive enough to show up in
+            // Activity Monitor, so only run it on explicit/manual refreshes and submenu-driven stale checks.
+            self.syncOpenAIWebState()
```
ratulsarna, line 465: If OpenAI web extras are enabled, what path repopulates the dashboard data after a normal refresh() from app launch, account changes, or relogin? I only see syncOpenAIWebState() here unless this is a forced refresh.

cbrane: Intentional: a normal background `refresh()` no longer repopulates the OpenAI dashboard data on app launch/account-change/relogin by itself. `syncOpenAIWebState()` only keeps the OpenAI-web state aligned with the current Codex identity (for example, clearing stale dashboard data and marking account changes) so we do not silently spin up the hidden ChatGPT WebView on the timer/startup path.

The repopulation paths are all explicit/foreground now:

- manual Refresh: `refreshNow()` -> `refresh(forceTokenUsage: true)` -> `refreshOpenAIDashboardIfNeeded(force: true)`
- manual cookie import: `importOpenAIDashboardBrowserCookiesNow()` -> `refreshOpenAIDashboardIfNeeded(force: true)`
- dashboard submenu open when stale: `menuWillOpen` -> `requestOpenAIDashboardRefreshIfStale(reason: "submenu open")` -> `refreshOpenAIDashboardIfNeeded(force: true)`

So after app launch, account change, or relogin, we intentionally wait until the user explicitly refreshes or opens a dashboard-backed submenu before repopulating the dashboard snapshot. That tradeoff is what removes the hidden background ChatGPT tab from the steady-state refresh path. Happy to expand the inline comment if that would make the intent clearer in the code too.

ratulsarna: Thanks, that clarifies the intended tradeoff. I still may be missing the intended entry point, though: I can see refreshNow() in code, but I’m not finding a visible UI action that calls it. In manual testing, the Providers refresh button only hits refreshProvider(...), so the explicit repopulation path doesn’t seem discoverable from the current UI.

cbrane: Good catch, you were right that the explicit refresh path was not discoverable enough from the current UI.

I pushed a small follow-up in `b6f4adc` to close that gap:

- added a visible `Refresh` action back to the main menu, which calls the existing explicit refresh path
- routed the Codex Providers-pane refresh button through the full explicit refresh path when OpenAI web extras are enabled
- kept disabled-provider refreshes on the provider-only path

That keeps the battery fix intact by avoiding timer-driven background dashboard scraping, but still gives users an obvious manual way to repopulate the dashboard-only OpenAI web data.

I also added coverage for both pieces:

- `StatusMenuTests` checks that `Refresh` is actually present in the menu
- `ProvidersPaneCoverageTests` checks that Codex uses the full explicit refresh path when OpenAI web extras are enabled

Validation:

- `pnpm check`
- `swift test --filter StatusMenuTests --filter ProvidersPaneCoverageTests`

cbrane:
@ratulsarna in addition to the small follow-up above, I also took care of all the merge conflicts as well. Let me know if this is merge ready, and if not, let me know anything else that I need to adjust! I can get that taken care of.

Sources/CodexBar/SettingsStore.swift
```diff
@@ -212,8 +223,11 @@ extension SettingsStore {
         let showOptionalCreditsAndExtraUsage = creditsExtrasDefault ?? true
         if creditsExtrasDefault == nil { userDefaults.set(true, forKey: "showOptionalCreditsAndExtraUsage") }
         let openAIWebAccessDefault = userDefaults.object(forKey: "openAIWebAccessEnabled") as? Bool
-        let openAIWebAccessEnabled = openAIWebAccessDefault ?? true
-        if openAIWebAccessDefault == nil { userDefaults.set(true, forKey: "openAIWebAccessEnabled") }
+        let openAIWebAccessEnabled = openAIWebAccessDefault ?? false
+        if openAIWebAccessDefault == nil { userDefaults.set(false, forKey: "openAIWebAccessEnabled") }
+        let openAIWebBatterySaverDefault = userDefaults.object(forKey: "openAIWebBatterySaverEnabled") as? Bool
+        let openAIWebBatterySaverEnabled = openAIWebBatterySaverDefault ?? true
```
ratulsarna, line 247: Quick question: for someone who already had OpenAI Web Extras enabled before this setting existed, do we want them to land in Battery Saver automatically on upgrade, or is it worth checking whether their current refresh behavior should be preserved until they choose otherwise?

Sources/CodexBar/UsageStore.swift
```diff
@@ -411,11 +416,21 @@ final class UsageStore {
...
+            // scraping stays opt-in behind the Battery Saver setting because the full ChatGPT page is expensive
+            // enough to show up in Activity Monitor.
+            self.syncOpenAIWebState()
+            let refreshPolicy = OpenAIWebRefreshPolicyContext(
+                accessEnabled: self.isEnabled(.codex) &&
+                    self.settings.openAIWebAccessEnabled &&
+                    self.settings.codexCookieSource.isEnabled,
+                batterySaverEnabled: self.settings.openAIWebBatterySaverEnabled,
+                force: forceTokenUsage)
+            if Self.shouldRunOpenAIWebRefresh(refreshPolicy) {
```
ratulsarna, line 472: In Battery Saver mode, do we still end up going through a forced dashboard refresh from the stale-submenu path elsewhere? I’m wondering if it’s worth tracing whether the reduced-refresh path still fully benefits from the cooldown after a failure.

Sources/CodexBar/SettingsStore.swift
cbrane: I intentionally kept existing users on the safe path here. Given how severe the battery regression was, I think defaulting `Battery Saver` to on for upgrades is the right migration, even for users who previously had OpenAI Web Extras enabled.

Sources/CodexBar/UsageStore.swift
cbrane: Good catch, you were right. In Battery Saver mode, the stale-submenu path was still forcing refreshes and bypassing the cooldown after failures.

I fixed that in `381498f` so stale submenu refreshes are non-forced when Battery Saver is enabled, which means they now respect the cooldown. When Battery Saver is off, they keep the more aggressive behavior.

I also added regression coverage in `OpenAIWebRefreshGateTests`.

ratulsarna:
Hi @cbrane , sorry for the delay in response here. This PR should be part of next release. Cleaning up some codex code before we change it further. Post that will get to this PR.

cbrane:
@ratulsarna @steipete I just wanted to drop a quick note here! I hope this doesn’t come off as rude (not at all like, "Hey, look at my PR and merge it!"). This is one of my first real contributions to open source, and I have no ego about it. However, I genuinely believe this PR addresses a crucial issue that should be a top priority, especially considering the significant battery impact the application has without this fix. I would love for users to benefit from this improvement, as it was one of the reasons I paused using the application for a while!

It’s been about 3 weeks to a month now, and while I understand there has been some back and forth and discussions about UX, everything has been ready for around 2 weeks. It feels like this is being overlooked to some degree. Given how urgent this fix is, I’m a bit puzzled as to why it isn’t receiving more attention. There have been multiple versions released since this was completed, and I feel like users are really missing out because we aren’t prioritizing getting this merged as soon as possible. Really appreciate all the work you've both done (especially you @ratulsarna with all the back and forth on this!) and would just love to see this merged! If there's anything else that has to be done or dealt with, please let me know.

ratulsarna:
Hey @cbrane, I'm looking at this right now. THanks for your patience.

cbrane:
Hey @ratulsarna really appreciate it! Hope that didn't come off poorly, I've had a great time with this. Just fixed conflicts as well so it's easily mergeable for you once you complete your review. Thanks :)

ratulsarna:
Superseded by #684

Thanks again, @cbrane, for the original work here.

PR #532, opened 2026-03-14
feat: default to /usr/bin/security CLI for Claude keychain reads

johnlarkin1:
## Summary
- Makes `/usr/bin/security` CLI the default keychain read strategy for Claude OAuth, eliminating recurring macOS keychain password prompts that appear after every rebuild/update
- Renames `.securityCLIExperimental` → `.securityCLI` since this is now the production default
- Security.framework remains available as a fallback and user override

## Context
CodexBar triggers a macOS keychain prompt ("CodexBar wants to access key 'Claude Code-credentials'") even after clicking "Always Allow" because `SecItemCopyMatching` requires the calling binary to be in the keychain item's ACL — every rebuild invalidates it. The `/usr/bin/security` CLI is permanently in the ACL (added by Claude Code during login), so it never prompts.

## Test plan
- [x] `swift build` compiles cleanly
- [x] `swift test` — all tests pass (only pre-existing failures remain)
- [x] `./Scripts/lint.sh lint` — 0 violations
- [ ] Manual: verify no keychain password prompt when app reads Claude credentials
- [ ] Check console logs for `"reader": "securityCLI"` confirming CLI path is used

### DIFF
--- a/Sources/CodexBar/SettingsStore+Defaults.swift
+++ b/Sources/CodexBar/SettingsStore+Defaults.swift
@@ -263,7 +263,7 @@ extension SettingsStore {
     var claudeOAuthKeychainReadStrategy: ClaudeOAuthKeychainReadStrategy {
         get {
             let raw = self.defaultsState.claudeOAuthKeychainReadStrategyRaw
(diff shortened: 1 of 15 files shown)


johnlarkin1:
whoops yeah wrong repo (i wanted it on my specific one)

PR #543, opened 2026-03-16
Wrap long status blurbs in menu

zkforge:
## Summary
- Wrap secondary status blurbs in the menu actions section using a view-backed text row so long error messages can wrap instead of widening the menu.
- Add a StatusMenu test that verifies the status blurb is rendered as a fixed-width custom view.

## Verification
- `pnpm check`
- `swift test` (hit an existing `swiftpm-testing-helper` / AppKit `SIGSEGV` during runner drawing; crash report captured at `~/Library/Logs/DiagnosticReports/swiftpm-testing-helper-2026-03-16-233445.ips`)

### DIFF
--- a/Sources/CodexBar/StatusItemController+Menu.swift
+++ b/Sources/CodexBar/StatusItemController+Menu.swift
@@ -543,7 +543,7 @@ extension StatusItemController {
     }
 
     private func makeWrappedSecondaryTextItem(text: String, width: CGFloat) -> NSMenuItem {
-        let item = NSMenuItem()
+        let item = NSMenuItem(title: text, action: nil, keyEquivalent: "")
         let view = self.makeWrappedSecondaryTextView(text: text)
         let height = self.menuTextItemHeight(for: view, width: width)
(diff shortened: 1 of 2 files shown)


Sources/CodexBar/StatusItemController+Menu.swift
```diff
@@ -538,6 +542,46 @@ extension StatusItemController {
         }
     }
 
+    private func makeWrappedSecondaryTextItem(text: String, width: CGFloat) -> NSMenuItem {
+        let item = NSMenuItem()
```
ratulsarna: Should this wrapped menu item also preserve the real blurb in item.title? Right now it’s created with NSMenuItem(), which seems to leave the title as NSMenuItem, so it may be worth checking what code paths or accessibility behavior still rely on the menu item’s title rather than the custom view.

zkforge: Good point. I agree the wrapped item should still preserve the real blurb in `item.title`, not just in the custom view. I'll update it so we keep the original string on the `NSMenuItem` and use the view only for wrapping/layout.

PR #553, opened 2026-03-17
feat(cursor): show Auto and API pool usage separately

AldikYO:
Cursor Pro plans split quota into two buckets:
- Auto pool: base + bonus credits (autoPercentUsed from API)
- API pool: base quota only for named models (apiPercentUsed)

Previously planPercentUsed (7000/7000 = 100%) was used as primary, making it appear all usage was exhausted even when bonus credits remained. Now primary shows Auto pool and secondary shows API pool.


ratulsarna:
Thank you for putting this together. We have since added Cursor Auto/API pool handling on the current main branch, so this PR is now obsolete and still has unresolved review feedback against the older implementation. I am going to close it to keep the queue focused, but we appreciate the contribution and the direction here.

Issue #128, opened 2026-01-04
Windsurf support


hongbo-miao:
Thanks for the convenient tool!
It would be great to support Windsurf as well, thanks! ☺️

mphassani:
Hey guys! I opened a PR to add windsurf support. Been using it locally and works great.
PR: [https://github.com/steipete/CodexBar/pull/328](https://github.com/steipete/CodexBar/pull/328)

demyxco:
Was just wondering about this. Thanks for adding support for Windsurf! I see that the PR is nearing completion and hopefully released Soon™.

demyxco:
@Coooolfan love the support for windsurf but stats aren't working properly.

CodexBar
<img width="302" height="182" alt="Image" src="https://github.com/user-attachments/assets/a5226d66-ae83-45f6-93b7-fada50f59486" />

Windsurf
<img width="345" height="210" alt="Image" src="https://github.com/user-attachments/assets/f80ca399-ab60-4880-b288-dddca9a31c75" />

Coooolfan:
@demyxco hi! 
Please check whether the **Usage source** value in CodexBar is set to **Web API**. In my local testing, data from **Local (sqlite cache)** was severely outdated.

Using **Web API** requires cookies from the website. If you don’t like frequent Keychain pop-ups, you can refer to <https://github.com/steipete/CodexBar/blob/main/docs/windsurf.md#how-to-get-a-manual-session-bundle> to manually fill in the pop-up.

<img width="1162" height="889" alt="Image" src="https://github.com/user-attachments/assets/4a286e7d-057b-4e04-a50d-187d2a7f920b" />

steipete:
Windsurf support is now implemented on `main`.

It includes Web API usage fetching, Chromium session import/manual session bundle support, and local SQLite-cache fallback. The changelog entry is under the 0.25/0.26 history as `Windsurf: add provider support with web-session usage fetching and local SQLite-cache fallback (#583). Thanks @Coooolfan!`

Verified with:
- `swift test --filter 'WindsurfProviderTests|WindsurfStatusProbeTests|WindsurfWebFetcherTests|WindsurfDevinSessionImporterTests'`

For mismatched stats, use Usage source = Web API. The local SQLite cache can be stale; if Web API still mismatches the Windsurf dashboard, please open a separate focused bug with the source mode and screenshots.

PR #583, opened 2026-03-21
add Windsurf support

Coooolfan:
Read cached plan info from Windsurf's local SQLite database (state.vscdb) and surface daily/weekly quota usage.

for https://github.com/steipete/CodexBar/issues/128

image

image


Coooolfan:
used it for a day. there is a clear delay fetching the value. may need a different way to get the quota...

Coooolfan:
image
 
image

now adds a web-based usage fetching option for the Windsurf provider (via Firebase tokens from browser IndexedDB). It respects the Keychain toggle in Advanced settings.
also, you can manually paste their own token if preferred. Docs for how to grab the token are included as well.

ratulsarna:
@codex review

Ssiswent:
@steipete @ratulsarna Could you please merge this?

Tests/CodexBarTests/SettingsStoreTests.swift
```diff
@@ -936,6 +936,7 @@ struct SettingsStoreTests {
             .synthetic,
             .warp,
             .openrouter,
+            .windsurf,
```
ratulsarna, line 939: Should Windsurf be expected after Mistral here, matching the new `UsageProvider.allCases` order? This older-config path appends missing providers in enum order, so the current expectation looks like it will fail the SettingsStore test.

Sources/CodexBarCore/Providers/Windsurf/WindsurfWebFetcher.swift
```diff
@@ -0,0 +1,632 @@
...
+        let log: (String) -> Void = { msg in logger?("[windsurf-web] \(msg)") }
+
+        if let manualSessionInput, !manualSessionInput.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty {
+            log("Using manual Windsurf session bundle")
+            let auth = try self.parseManualSessionInput(manualSessionInput)
+            let response = try await self.fetchPlanStatus(auth: auth, timeout: timeout, session: session)
+            return response.toUsageSnapshot()
+        }
+
+        guard cookieSource != .off else {
```
ratulsarna, line 158: Should manual mode avoid falling through to browser localStorage import when the manual session bundle is empty? Otherwise selecting `Manual` can still fetch from an automatically imported browser session.

Sources/CodexBarCore/Providers/Windsurf/WindsurfProviderDescriptor.swift
```diff
@@ -0,0 +1,100 @@
...
+
+    func fetch(_ context: ProviderFetchContext) async throws -> ProviderFetchResult {
+        #if os(macOS)
+        let cookieSource = context.settings?.windsurf?.cookieSource ?? .auto
+        let manualToken = Self.manualToken(from: context)
+        let usage = try await WindsurfWebFetcher.fetchUsage(
+            browserDetection: context.browserDetection,
+            cookieSource: cookieSource,
+            manualSessionInput: manualToken,
+            logger: context.verbose ? { print($0) } : nil)
```
ratulsarna, line 63: Should this pass `context.webTimeout` through to `WindsurfWebFetcher.fetchUsage` so Windsurf follows the same configured timeout behavior as the other web-backed providers?

Sources/CodexBarCore/Providers/Windsurf/WindsurfStatusProbe.swift
```diff
@@ -0,0 +1,236 @@
...
+        switch sqlite3_column_type(stmt, index) {
+        case SQLITE_TEXT:
+            guard let c = sqlite3_column_text(stmt, index) else { return nil }
+            return String(cString: c)
+        case SQLITE_BLOB:
+            guard let bytes = sqlite3_column_blob(stmt, index) else { return nil }
+            let data = Data(bytes: bytes, count: Int(sqlite3_column_bytes(stmt, index)))
+            // VSCode/Windsurf state.vscdb schema declares value as BLOB;
+            // try UTF-16LE first (common for VSCode derivatives), then UTF-8.
+            if let decoded = String(data: data, encoding: .utf16LittleEndian) {
```
ratulsarna: Could this accept a UTF-16 decode only after confirming it looks like JSON, or try UTF-8 first? A UTF-8 JSON blob can decode as non-nil UTF-16 mojibake here, which would skip the UTF-8 fallback and fail parsing later.

Tests/CodexBarTests/SettingsStoreTests.swift
Coooolfan: the test models a much older config with only `gemini` and `codex`, so all missing providers are appended in the current `UsageProvider.allCases` order; in real upgrades from newer configs that already have `perplexity`/`abacus`/`mistral`, only `windsurf` would be missing and it would be appended after the existing order. The test passes as-is. We can also move `windsurf` to the end if that ordering is preferred.

Sources/CodexBarCore/Providers/Windsurf/WindsurfStatusProbe.swift
Coooolfan: Fixed. Covered by [`WindsurfStatusProbeTests`](https://github.com/steipete/CodexBar/pull/583/changes#diff-e8855b39e40102447fd9e05d933aae65ed95346ce1f27c4ed2fd9464ac7a1af6R186)

Sources/CodexBarCore/Providers/Windsurf/WindsurfProviderDescriptor.swift
Coooolfan: fixed at  [WindsurfProviderDescriptor](https://github.com/steipete/CodexBar/pull/583/changes#diff-b1bc30077345d1272815ea1085ec06408ecfeb6cc5f787c86ee04e1d52c60a02R62)

Sources/CodexBarCore/Providers/Windsurf/WindsurfWebFetcher.swift
Coooolfan: Fixed. Covered by [`WindsurfWebFetcherTests`](https://github.com/steipete/CodexBar/pull/583/changes#diff-7616d7858a66b1985c61bcfa111bbf20da5138309cf45cbf2378affa48451a97R151)

ratulsarna:
Thanks again @Coooolfan for the Windsurf provider work. I opened a replacement PR at #812 that carries this forward on current main: https://github.com/steipete/CodexBar/pull/812

The only extra tweaks there are minor: resolving the provider-list conflicts from the rebase and fixing the SettingsStore provider-order test expectation.

ratulsarna:
Thanks again for this. I did a live smoke test against a currently installed/running Windsurf app, and I think there may be a storage-shape/version gap here.

On my machine, Windsurf is installed and running, and `state.vscdb` has Windsurf auth/status-shaped data such as:

- `windsurfAuthStatus`
- `windsurfConfigurations`
- `secret://{"extensionId":"codeium.windsurf","key":"windsurf_auth.sessions"}`
- `windsurf_auth-…-usages`

But it does not have `windsurf.settings.cachedPlanInfo`, so the local fetch path reports:

> No plan data found in Windsurf database. Sign in to Windsurf first.

The web-token path also did not validate for me because there was no `windsurf.com` Chromium localStorage session on this machine.

So I don’t think this invalidates your implementation, especially since you clearly had it working and shared screenshots. But it does mean the local DB path may only work for some Windsurf versions/account states/actions, not for every installed/running Windsurf app with existing Windsurf auth/status data.

Before merging as general Windsurf support, I think we should either:

- identify what causes `windsurf.settings.cachedPlanInfo` to be written, or
- support the newer/current `windsurfAuthStatus` storage shape too, if that contains enough usage data, or
- mark this provider path as explicitly best-effort/experimental in the PR notes.

ratulsarna:
Hi @Coooolfan , any update here?

Coooolfan:
> Hi @Coooolfan , any update here?

@ratulsarna sry,  i am currently on vacation. i will follow up on this PR over the next few days. It may take a bit more time.

steipete:
Deep review + live recheck from current `main`:

The PR-head Windsurf tests pass for me: `swift test --filter Windsurf --no-parallel` ran 27 Windsurf tests successfully in an isolated checkout.

Still not mergeable as-is:
- The branch is dirty against current `main`; merge-tree shows conflicts in provider registration/order, `UsageStore`, `TokenAccountCLI`, `ProviderSettingsSnapshot`, widget provider/view mappings, cost scanner, and settings tests.
- Live local storage on my machine now does contain `windsurf.settings.cachedPlanInfo`, but the value shape has `usage` fields (`messages`, `usedMessages`, `remainingMessages`, `flowActions`, etc.) and **no** `quotaUsage` object.
- The local conversion path in `WindsurfCachedPlanInfo.toUsageSnapshot()` only creates primary/secondary windows from `quotaUsage`, so this live shape would decode but produce no usage windows. That means local mode can still look unavailable/misleading even when Windsurf has plan usage data present.
- Running the PR CLI against my current `~/.codexbar/config.json` fails config decode because the branch is stale relative to current provider/config schema; that should go away after a proper rebase, but it blocked a clean CLI live smoke of the PR branch.

Best next shape: rebase first, then map the current `usage` count shape into a useful snapshot when `quotaUsage` is absent, or make web mode the primary path and mark the local DB path as best-effort/experimental with explicit fallback behavior.

steipete:
Extra live proof for the local DB shape:

I added a temporary isolated test against my live Windsurf `state.vscdb` on this PR head. It verified:

- `windsurf.settings.cachedPlanInfo` decodes successfully
- `usage.messages` and `usage.usedMessages` are present
- `quotaUsage` is absent
- `toUsageSnapshot()` returns `primary == nil` and `secondary == nil`

So the current local path can read the live record but still renders no usage windows for this storage shape. The count fields are enough to derive at least one useful window on my machine (`usedMessages/messages` is 2.4%), so I think the next revision should map those counts when `quotaUsage` is missing.

PR #594, opened 2026-03-24
fix: handle errSecInteractionNotAllowed in KeychainCacheStore to prevent cache self-destruction on wake

josepe98:
## Problem

When the Mac wakes from sleep, the keychain is briefly locked. During this window, `KeychainCacheStore.load` calls `SecItemCopyMatching` and gets back `errSecInteractionNotAllowed` (-25308). This falls into the `default` case, returns `.invalid`, and the caller deletes the cache entry:

```swift
// In Repository.loadRecord
case .invalid:
    KeychainCacheStore.clear(key: ClaudeOAuthCredentialsStore.cacheKey)  // cache wiped
```

The cache entry was perfectly valid — it was just temporarily inaccessible. Deleting it forces a fresh read of `"Claude Code-credentials"` on the next access, which triggers a keychain prompt. This is why users see repeated password prompts after wake from sleep even after clicking "Always Allow".

## Fix

Two small changes to `KeychainCacheStore.load`:

1. Apply `KeychainNoUIQuery` to the load query — consistent with how other no-UI reads are performed elsewhere in the codebase, and ensures the call fails fast with `errSecInteractionNotAllowed` rather than blocking.

2. Add an explicit `case errSecInteractionNotAllowed` that returns `.missing` instead of `.invalid`. The entry is valid but temporarily inaccessible — the caller should fall through gracefully and retry on next access, not destroy the cache.

## Context

`errSecInteractionNotAllowed` is already handled explicitly in `ClaudeOAuthCredentials.swift` and `KeychainAccessPreflight.swift`, but was missed in `KeychainCacheStore` itself. The `KeychainNoUIQuery` helper exists precisely for this purpose but wasn't applied to cache reads.

Note: I'm a user who hit this bug (repeated prompts after wake) and traced it to this specific code path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 ### Providers & Usage
 - Gemini: discover OAuth config in fnm/Homebrew/bundled CLI layouts so expired-token refresh keeps working (#723). Thanks @Leechael!
 
+### Fixes
+- Keychain cache: preserve cached credentials when macOS temporarily denies keychain UI after wake, avoiding repeated prompts (#594). Thanks @josepe98!
+
 ## 0.21 — 2026-04-18
 
 ### Highlights
--- a/Sources/CodexBarCore/KeychainCacheStore.swift
+++ b/Sources/CodexBarCore/KeychainCacheStore.swift
@@ -46,13 +46,14 @@ public enum KeychainCacheStore {
             return testResult
         }
         #if os(macOS)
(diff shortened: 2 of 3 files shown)


steipete:

- Land commit: 6d7b4aa6
- Merge commit: aa26573b35fb354d16647bb1d8e92804050f9f8d

Thanks @josepe98!

PR #595, opened 2026-03-24
feat(providers): adapt MiniMax usage UI for Token Plan quotas

shixy96:
## Summary

MiniMax has expanded its former Coding Plan into the new Token Plan, a single subscription that covers text, speech, image, music, and video models with per-model quotas.

This PR updates CodexBar's MiniMax integration to better match that model:

- parse MiniMax remains data into ordered per-model usage entries and prefer the primary `MiniMax-M*` text model for the top-level summary
- enrich the HTML snapshot with remains API data so MiniMax usage survives partial HTML changes and can still expose model-level quota details
- show up to three MiniMax model usage bars in the menu card, append overflow model reset details to usage notes, and add per-model reset lines to the menu
- hide the generic menu bar metric picker for MiniMax because the Token Plan quota is model-specific rather than a single global metric
- route dashboard actions through provider-specific URL resolution so MiniMax opens the correct region-specific subscription page, while Claude subscription users land on the usage page that matches their account type

## Why

MiniMax's official Token Plan now describes:
- text usage as request-based quotas with a 5-hour rolling reset
- non-text usage as per-model daily quotas
- one subscription spanning text, speech, image, music, and video models

The previous CodexBar presentation was too coarse for that structure. This change makes the MiniMax menu and dashboard actions align more closely with the official plan model without changing other providers' usage behavior.

## Screenshots
image
image

## Testing

- `swiftlint --strict`
- `pnpm check`

## References

- https://platform.minimax.io/subscribe/token-plan
- https://platform.minimax.io/docs/token-plan/intro
- https://platform.minimax.io/docs/token-plan/faq

### DIFF
--- a/Sources/CodexBar/MenuCardView.swift
+++ b/Sources/CodexBar/MenuCardView.swift
@@ -649,6 +649,8 @@ struct UsageMenuCardExtraUsageSectionView: View {
 // MARK: - Model factory
 
 extension UsageMenuCardView.Model {
+    private static let miniMaxMetricLimit = 3
+
     struct Input {
         let provider: UsageProvider
         let metadata: ProviderMetadata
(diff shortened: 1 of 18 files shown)


ratulsarna:
Thanks for the contribution. I’m closing this one because the same MiniMax token-plan lane is now better covered by a newer PR, and I want to keep the queue from splitting across overlapping implementations.

PR #622, opened 2026-03-30
Add local Codex session analytics drilldowns

Undertone0809:
## What changed
- add local Codex session analytics backed by rollout JSONL instead of remote dashboard data
- add persisted analytics window presets (10 / 20 / 50 / 100), token usage parsing, and summary/tool diagnostics
- restructure the Session Analytics UI into a compact overview with nested drilldown submenus, add a top-tools donut chart, and remove row chevrons from analytics items

image
image

## Why
- this gives CodexBar a local process-oriented analytics view for Codex sessions without depending on OpenAI dashboard cookies
- the earlier inline-detail layout made the analytics submenu too tall and clipped in practice; nested drilldowns keep the primary menu scannable and stable

## User impact
- Codex users can inspect recent sessions, tool behavior, failures, long-running calls, and token breakdowns from local data
- users can change the analytics window size from the menu and keep that preference across launches

## Validation
- `pnpm check`
- `swift test --filter CodexSessionAnalyticsLoaderTests`
- `swift test --filter SettingsStoreCoverageTests`
- `swift test --filter "codex menu includes session analytics submenu when snapshot is present"`
- `swift test --filter "codex session analytics submenu exposes empty state without local data"`


Undertone0809:
@steipete What do you think of this feature?

steipete:
Thanks for the session analytics work. I ran the focused tests locally and they pass:

`swift test --filter 'CodexSessionAnalyticsIndexerTests|CodexSessionAnalyticsLoaderTests|UsageStoreCodexSessionAnalyticsTests|StatusMenuSessionAnalyticsTests'`

I found one blocker before this is safe to merge: the persisted analytics cache is global while the indexed data is `CODEX_HOME`-scoped. `CodexSessionAnalyticsIndexIO.cacheFileURL` always writes `Caches/CodexBar/session-analytics/codex-v1.json` (`Sources/CodexBarCore/Providers/Codex/CodexSessionAnalyticsIndex.swift:6`-`:12`), and `CodexSessionAnalyticsIndexer.loadPersistedIndex()`/`persist(index:)` use that path regardless of the `env` passed into the indexer (`CodexSessionAnalyticsIndex.swift:104`-`:122`).

That means after switching between managed Codex homes or local profile/account routing, `UsageStore.bootstrapCodexSessionAnalyticsCache()` can load and display the previous home’s session titles/tool stats before a refresh (`Sources/CodexBar/UsageStore+CodexSessionAnalytics.swift:28`-`:37`). For this feature, the cache needs to be keyed by the resolved Codex home/account scope, or the cached index needs to store and validate the root identity before it is used.

steipete:
Thanks again for the session analytics work. I am closing this PR rather than keeping it in limbo. The feature idea is still interesting, but this branch is now too large and stale to land safely: 19 files, 3.6k additions, merge conflicts on current main, and only GitGuardian has run on the current head. The earlier blocker still stands too: the persisted analytics cache is global while the indexed data is CODEX_HOME scoped, so switching Codex homes/accounts can briefly show another home session titles/tool stats before refresh. For local session analytics that is a privacy/correctness boundary, not just a polish issue. Best next path is a fresh smaller PR on current main, with the cache keyed/validated by resolved Codex home/account scope from the start, then menu UI as a follow-up if the data layer is solid. Appreciate the work here.

PR #627, opened 2026-03-31
Fix duplicated Codex token counting in forked sessions

sszzz830:
## Summary

This fixes double counting for Codex token usage when a session is forked.

Forked Codex sessions get a new `session_id`, but they inherit the parent session's cumulative `token_count` history. The existing scanner treated the forked session as fully independent and counted that inherited prefix again, which inflated token and cost totals.

## What changed

- Read `session_meta.forked_from_id` from Codex session logs
- Track raw `token_count` entries per session during scanning
- When a child session is forked from a parent, compare the child's raw prefix against the parent's existing prefix
- Skip only the inherited prefix that exactly matches the parent history
- Continue counting all new usage generated after the fork in both parent and child sessions
- Bump the Codex cost usage cache artifact version from `v2` to `v3` so previously overcounted cached results are invalidated

## Why this approach

This is intentionally strict and narrow:

- It only applies to explicit Codex forks with `forked_from_id`
- It does not dedupe unrelated sessions that merely happen to have similar totals
- It preserves post-fork usage correctly for both branches
- It avoids over-deduping by only removing the exact inherited prefix

## Validation

Validated against forked-session scenarios to confirm that:

- inherited parent usage is not counted twice
- new child usage after the fork is still counted
- new parent usage after the fork is still counted
- incremental rescans still work correctly
- the Codex cache now rolls to `codex-v3.json`

### DIFF
--- a/Sources/CodexBarCore/Vendored/CostUsage/CostUsageCache.swift
+++ b/Sources/CodexBarCore/Vendored/CostUsage/CostUsageCache.swift
@@ -4,7 +4,7 @@ enum CostUsageCacheIO {
     private static func artifactVersion(for provider: UsageProvider) -> Int {
         switch provider {
         case .codex:
-            2
+            3
         default:
             1
         }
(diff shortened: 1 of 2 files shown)


ratulsarna:
Thank you for working on the Codex forked-session token counting issue. The Codex session/cache code has changed significantly since this PR was opened, and current main appears to have moved past this older implementation. Since the PR is stale and still has unresolved review feedback, I am closing it as obsolete. We appreciate the original fix direction.

PR #647, opened 2026-04-04
fix: deduplicate message hashes across files, not just within each file

enzonaute:
parseClaudeFile() resets its seenKeys set per file. Claude Code logs subagent messages to both the parent session JSONL and the subagent's own JSONL under <session>/subagents/. Since both files share the same messageId:requestId pairs, per-file dedup misses the cross-file duplicates and inflates the totals.

On a real dataset (30 days, 538 files):
  - per-file dedup:  7.98B tokens (current behavior)
  - global dedup:    2.59B tokens (correct, matches ccusage)
  - ratio: 3.08x overcounting

Fix: pass the seenKeys set through ClaudeScanState so it accumulates across all files in a single scan pass. The new existingSeenKeys parameter defaults to [] so existing call sites and tests are unaffected.


ratulsarna:
A fix has been merged for the issue this PR fixes. Thanks for the investigtion @enzonaute ! 

91146a28c308f20633946cc5d1bf57eb89a81e8e

Issue #572, opened 2026-03-19
Support Intel Mac (non-Apple Silicon) builds


barbarhan:
Hey Peter,

CodexBar doesnt work on Intel Macs - requires Apple Silicon.

I have a 2019 MacBook Pro with Intel and would love to use this. Any chance we could get Intel Mac builds working?

Cheers

ngutman:
Thanks for the request. I’m going to close this as not planned for the current product scope. CodexBar is currently focused on Apple Silicon / newer macOS targets, and keeping Intel support open indefinitely here isn’t actionable.

If platform support changes in the future, it can be revisited with a fresh tracking issue.

Issue #563, opened 2026-03-18
After installing the Codexbar app, a prompt appears Error


yuanCodeLab:
brew install --cask steipete/tap/codexbar，After installing the Codexbar app, a prompt appears Error: Cask codexbar depends on hardware architecture being one of [{type: :arm, bits: 64}], but you are running {type: :intel, bits: 64}.

ngutman:
Thanks for the report. I’m closing this as not planned / out of scope for the current supported platform matrix. This error is the Intel architecture restriction in the Homebrew cask rather than a separate CodexBar bug.

If official Intel support returns in the future, it can be tracked separately.

PR #650, opened 2026-04-04
Remove Intel support claims on website

### LINKED ISSUE #231: Mismatch between Homebrew and codexbar.app site
If I run `brew install --cask steipete/tap/codexbar` on my Intel mac, it fails, saying: 

>Error: Cask codexbar depends on hardware architecture being one of [{type: :arm, bits: 64}], but you are running {type: :intel, bits: 64}.

However, on the first page of https://codexbar.app, it says "Requirements: macOS 14+ (Apple Silicon + **Intel**)."

---

One of these is wrong, but I'm not sure which.

abotkin:
This PR fixes #231 by changing the website to remove claims of Intel support. The current Homebrew setup does not support Intel Macs and while there are some PRs trying to add such support (#331, #457, #544), these PRs don't appear to be moving to acceptance and @steipete has [said he would prefer this not be added in favor of Intel users manual compiling themselves](https://github.com/steipete/CodexBar/issues/52#issuecomment-3691444010).

It doesn't make sense to advertise Intel support for CodexBar to users when it's not officially supported and there's ample evidence that this wording has led to wasted time from many Intel users trying to install via Homebrew: #231, #300, #563, #572

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -44,7 +44,7 @@
         <div class="copy" data-animate="2">
           <div class="badges" aria-label="Highlights">
             <span class="badge">macOS 14+</span>
-            <span class="badge">Apple Silicon + Intel</span>
+            <span class="badge">Apple Silicon only</span>
             <span class="badge">Free &amp; open source</span>
             <span class="badge">No passwords stored</span>
           </div>
@@ -57,7 +57,7 @@
               </p>
             </div>
           </div>
-          <p class="requirement">Requirements: macOS 14+ (Apple Silicon + Intel).</p>
+          <p class="requirement">Requirements: macOS 14+ (Apple Silicon only).</p>
           <div class="cta">
             <a class="btn primary" href="https://github.com/steipete/CodexBar/releases/latest">Download latest</a>
             <a class="btn" href="https://github.com/steipete/CodexBar">View on GitHub</a>


admdly:
Intel Macs work fine with the official DMG releases and have for some time; they are Universal macOS binaries. It just appears that the Homebrew releases haven't been updated for this.

abotkin:
> Intel Macs work fine with the official DMG releases and have for some time; they are Universal macOS binaries. It just appears that the Homebrew releases haven't been updated for this.

@admdly Thank you for pointing this out! You are right - the ZIPs from 0.16.1 onwards are Universal binaries. This was not obvious to me from the website (since it mentions Homebrew as an install method so I'd assume parity with Download Latest) or from a search of the GitHub issues/PRs (where it seems there are multiple Intel folks confused and being told to manually compile even after 0.16.1 released).  So it feels like there could be better documentation of this to avoid user confusion.

CodexBar-Website

@ngutman could you help clarify the project's position on Intel support? I saw in [this thread](https://github.com/steipete/CodexBar/issues/572#issuecomment-4231154365) you mentioned that Intel support is not desired to be supported moving forward. Should the project remove the Intel support claim from the homepage then? If not, it seems like I should be proposing changes to the website to make it obvious that only Apple Silicon is supported on Homebrew - perhaps by adding "(Apple Silicon only)" next to the "Homebrew" and adding a newline to better separate the "Download Latest" button. What are your thoughts?

steipete:
Thanks @abotkin. I landed a narrower version on main in cbefd7da: the site now says GitHub Releases downloads are universal, while the Homebrew cask is Apple Silicon-only. That resolves #231 without removing Intel support claims for the official download. Co-author credit preserved.

PR #663, opened 2026-04-08
Fix Cursor fetch crash path

ratulsarna:
## Summary
- replace the Cursor usage-summary/user-info `async let` flow with a cancellation-safe task-group implementation
- add regression coverage for the success/failure cases around the Cursor fetch path

Fixes #661.


anirudhvee:
It looks like the failures are from the new `urlSession: URLSession = .shared` in `CursorStatusProbe.swift`. It probably just needs the usual `#if canImport(FoundationNetworking)` import there for Linux.

hhh2210:
when can we publish a release for that?

ratulsarna:
@hhh2210 Over the weekend

PR #672, opened 2026-04-09
[codex] Clear ad-hoc keychain caches

magnaprog:
## Summary

During ad-hoc rebuilds, `./Scripts/compile_and_run.sh` only cleared the legacy `com.steipete.CodexBar` service.
The current runtime cache under `com.steipete.codexbar.cache` was left behind, so a fresh build could still hit stale ACLs and trigger another macOS keychain approval/password prompt.

This patch clears both CodexBar-owned services before relaunching and suppresses the noisy `security delete-generic-password` stdout output.

## Notes

This only covers CodexBar-owned keychain entries.
Third-party items such as `Claude Code-credentials` still need stable signing if you want macOS to keep the **Always Allow** grant across rebuilds.

I also updated the development setup doc to spell that out.

## Other platforms

I checked the non-macOS paths in this repo and did not find an equivalent issue:

- `Package.swift` targets macOS for the app bundle.
- `KeychainCacheStore` uses Security.framework only on macOS and falls back to non-keychain behavior elsewhere.
- Linux platform tests already gate unsupported web/keychain-adjacent paths.

## Validation

- `./Scripts/lint.sh lint`
- `bash -n Scripts/compile_and_run.sh`
- `./Scripts/compile_and_run.sh --test` reaches the updated ad-hoc keychain-clearing path cleanly, then stops because this environment has Swift 6.0.0 while the package requires Swift tools 6.2.

### DIFF
--- a/Scripts/compile_and_run.sh
+++ b/Scripts/compile_and_run.sh
@@ -20,6 +20,14 @@ SIGNING_MODE="${CODEXBAR_SIGNING:-}"
 log()  { printf '%s\n' "$*"; }
 fail() { printf 'ERROR: %s\n' "$*" >&2; exit 1; }
 
+delete_keychain_service_items() {
+  local service="$1"
+  security delete-generic-password -s "${service}" >/dev/null 2>&1 || true
+  while security delete-generic-password -s "${service}" >/dev/null 2>&1; do
+    :
+  done
+}
+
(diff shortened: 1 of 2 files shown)


magnaprog:
Replacing this with a new PR under a neutral branch/title.

PR #673, opened 2026-04-09
Clear ad-hoc keychain caches

magnaprog:
## Summary

- Clear both CodexBar-owned keychain services during ad-hoc rebuilds: `com.steipete.CodexBar` and `com.steipete.codexbar.cache`
- Silence `security delete-generic-password` output in the rebuild log
- Update the development setup doc to explain what this change covers and what still requires stable signing

## Why

`./Scripts/compile_and_run.sh` was only clearing the legacy keychain service.
The current runtime cache under `com.steipete.codexbar.cache` was left behind, so a fresh ad-hoc-signed build could still run into stale ACLs and prompt again.

This change is limited to CodexBar-owned keychain entries.
Third-party items such as `Claude Code-credentials` still need stable signing if you want macOS to keep the `Always Allow` grant across rebuilds.

## Validation

- `./Scripts/lint.sh lint`
- `bash -n Scripts/compile_and_run.sh`
- `./Scripts/compile_and_run.sh --test` reaches the updated keychain-clearing path, then stops because this environment has Swift 6.0.0 while the package requires Swift tools 6.2

## Platform check

The app target in this repo is macOS-only, and the keychain cache path is guarded behind `#if os(macOS)`.
I did not find the same issue in the non-macOS code paths in this package.

### DIFF
--- a/Scripts/compile_and_run.sh
+++ b/Scripts/compile_and_run.sh
@@ -20,6 +20,14 @@ SIGNING_MODE="${CODEXBAR_SIGNING:-}"
 log()  { printf '%s\n' "$*"; }
 fail() { printf 'ERROR: %s\n' "$*" >&2; exit 1; }
 
+delete_keychain_service_items() {
+  local service="$1"
+  security delete-generic-password -s "${service}" >/dev/null 2>&1 || true
+  while security delete-generic-password -s "${service}" >/dev/null 2>&1; do
+    :
+  done
+}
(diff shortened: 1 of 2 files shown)


ratulsarna:
Opened a follow-up PR that supersedes this one: https://github.com/steipete/CodexBar/pull/688

Thanks again @magnaprog for the original investigation and patch direction here.

Issue #473, opened 2026-03-03
The `Kimi K2` provider is not official and is misleading


ZeroAurora:
Source: https://github.com/steipete/CodexBar/tree/main/Sources/CodexBarCore/Providers/KimiK2

The website kimi-k2.ai is clearly not official. The official Kimi model is offered by Moonshot at [moonshot.ai](https://www.moonshot.ai/) and [moonshot.cn](https://www.moonshot.cn/). Kimi for Coding is offered [here](https://www.kimi.com/code).

The PR (#147) for this shows significant hallucination that suprisingly passes human review.

Another PR #146 that adds Kimi provider is correct and far more better (though outdated in billing calculation).

mturac:
I can take a small pass at this: clarify the existing Kimi K2 provider as an unofficial legacy kimi-k2.ai endpoint and point users toward the Moonshot / Kimi API provider for the official API surface.

PR #705, opened 2026-04-13
Add Moonshot provider (official platform API)

maxitoon:
## Summary
- add a new Moonshot provider that uses the official Moonshot balance endpoint instead of the third-party Kimi K2 service
- support International and China regions via config/env + a settings picker
- document the provider and add focused parser/settings/UI regression tests

## Details
- adds `Sources/CodexBarCore/Providers/Moonshot/` and `Sources/CodexBar/Providers/Moonshot/`
- keeps Kimi K2 untouched; this only addresses the official Moonshot provider path discussed in #473
- stores Moonshot API keys in `~/.codexbar/config.json` and honors `MOONSHOT_API_KEY` / `MOONSHOT_KEY`
- honors `MOONSHOT_REGION` and persists a `region` value in provider config
- shows Moonshot as a balance-only provider (`Balance: $X.XX`, `No limit set`) and surfaces negative cash balance as `in deficit`

## Verification
- `swift build --target CodexBarCore`
- `swift build --target CodexBarCLI`
- `swift build --target CodexBarWidget`
- `swift run CodexBarCLI --help` (confirms `moonshot` is registered as a provider)
- `MOONSHOT_API_KEY=moonshot-invalid-key MOONSHOT_REGION=international swift run --skip-build CodexBarCLI usage --provider moonshot --format json` (confirms a clean Moonshot API error path on invalid auth)

## Environment blockers in this session
- `swift build --target CodexBar` fails in the upstream `KeyboardShortcuts` dependency because `PreviewsMacros` is unavailable in this Command Line Tools-only environment
- `./Scripts/lint.sh lint` crashes inside `swiftlint` because SourceKit/Xcode is unavailable here
- no real Moonshot API key was available in the environment, so I could not run the live success-path smoke test or capture screenshots

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -40,3 +40,4 @@
 - Keep provider data siloed: when rendering usage or account info for a provider (Claude vs Codex), never display identity/plan fields sourced from a different provider.***
 - Claude CLI status line is custom + user-configurable; never rely on it for usage parsing.
 - Cookie imports: default Chrome-only when possible to avoid other browser prompts; override via browser list when needed.
+- Moonshot provider: official Moonshot uses `GET /v1/users/me/balance` on `api.moonshot.ai` or `api.moonshot.cn`, exposes a Region picker, shows balance-only/no-limit semantics, and must stay distinct from the third-party Kimi K2 provider tracked in issue #473.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.21 — Unreleased
 
 ### Highlights
+- Moonshot: add an official Moonshot platform provider with International/China region selection, config/env API-key support, and balance-only display semantics distinct from Kimi K2 (#473).
 - z.ai: preserve weekly and 5-hour token quotas together, surface the 5-hour lane correctly across the menu/menu bar, and add regression coverage (#662). Thanks to @takumi3488 for the original fix and investigation.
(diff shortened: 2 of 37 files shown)


giuseppebisemi:
Nice provider addition. I tested Moonshot locally and two small UI follow-ups made it fit better with the existing provider set:

- Use the shared Kimi branding for Moonshot (`iconStyle: .kimi`, `iconResourceName: "ProviderIcon-kimi"`) instead of a separate Moonshot icon. In the app it reads more consistently with Kimi/Kimi K2, especially because the provider is commonly recognized through the Kimi branding.
- Surface the Moonshot balance directly in the menu bar label when Moonshot is selected. The fetcher already exposes `Balance: $X.XX` in the identity/login method, so the menu bar can strip the `Balance:` prefix and show just `$X.XX` instead of falling back to an empty/percent-style label.

One tiny polish item while doing that: the shared Kimi SVG looked slightly low compared with neighboring icons in the menu bar; moving the path up a bit improved optical alignment for Kimi, Kimi K2, and Moonshot together.

steipete:
Thanks for putting this together. I’m closing this older Moonshot PR because #793 supersedes it: that branch contains the Moonshot provider work from here and adds the Kimi CLI config discovery follow-up in the same thread. Keeping one Moonshot PR open should make review/rebase/landing much cleaner.

PR #707, opened 2026-04-13
fix: recognize Ollama __Secure-session cookies

### LINKED ISSUE #704: Ollama login detection fails because CodexBar does not recognize current `__Secure-session` cookie
SUMMARY

CodexBar reports:

No Ollama session cookie found. Please log in to ollama.com in your browser.

But Safari is already signed into `ollama.com`.

This appears to be caused by CodexBar’s Ollama cookie detector recognizing older cookie names only, while the current Ollama web session is stored under __Secure-session.

Observed behavior

CodexBar fails to detect an existing Ollama web session and keeps prompting the user to log in.

Expected behavior

If the browser has a valid signed-in ollama.com session, CodexBar should detect it and use it.

Evidence

From the installed `CodexBar.app` binary, the Ollama detector appears to recognize these cookie names:

- __Secure-next-auth.session-token
- next-auth.session-token
- __Host-ollama_session

From Safari’s local cookie store for `ollama.com`, the active session cookie present is:
__Secure-session.

Environment

- CodexBar: 0.19.0
- Browser: Safari
- macOS
- Full Disk Access / restart / re-login were already attempted

Thank you!

anirudhvee:
Fixes #704

## Summary

- recognize `__Secure-session` as a valid Ollama session cookie
- add focused Ollama cookie recognition tests

## Validation

- `swift test --filter OllamaUsageFetcherTests`
- verified Ollama web usage loads locally in the CLI and the menu bar

### DIFF
--- a/Sources/CodexBarCore/Providers/Ollama/OllamaUsageFetcher.swift
+++ b/Sources/CodexBarCore/Providers/Ollama/OllamaUsageFetcher.swift
@@ -9,6 +9,7 @@ import SweetCookieKit
 
 private let ollamaSessionCookieNames: Set<String> = [
     "session",
+    "__Secure-session",
     "ollama_session",
     "__Host-ollama_session",
     "__Secure-next-auth.session-token",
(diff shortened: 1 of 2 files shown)


ratulsarna:
Thanks for the fix @anirudhvee !

PR #718, opened 2026-04-14
Recover CLI rate limits from Pro Lite errors

ImLukeF:
## Summary
- recover Codex CLI usage and credits from the JSON body embedded in `account/rateLimits/read` decode errors
- keep the existing TTY `/status` fallback, but avoid depending on it when the RPC error already contains usable Pro Lite usage data
- add regression coverage for `prolite` decode mismatch payload recovery

## Root Cause
`codex app-server` can return `unknown variant "prolite"` for `account/rateLimits/read`, while embedding the full `/wham/usage` response body in the error. In explicit Codex CLI mode this left CodexBar with no usable usage or credits when the text status fallback also failed.

## Validation
- `pnpm check`
- `swift test --filter CodexUsageFetcherFallbackTests`
- `./Scripts/compile_and_run.sh --wait`

## Notes
Full `swift test` still has an unrelated existing failure in `OpenAIDashboardNavigationDelegateTests`: `commit completes navigation successfully after grace period`.

### DIFF
--- a/Sources/CodexBarCore/UsageFetcher.swift
+++ b/Sources/CodexBarCore/UsageFetcher.swift
@@ -324,6 +324,20 @@ private struct RPCCreditsSnapshot: Decodable, Encodable {
     let balance: String?
 }
 
+private struct RPCRateLimitsErrorBody: Decodable {
+    let email: String?
+    let planType: String?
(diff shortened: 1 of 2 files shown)


ratulsarna:
Thanks a lot @ImLukeF ! Rebased onto the base branch.

PR #710, opened 2026-04-13
Handle Codex Pro Lite plan responses and OAuth fallback edge cases

ratulsarna:
Fixes #691 and #709.

Codex started returning new `plan_type` values like `prolite`, which left a few different paths in CodexBar out of sync. This change teaches the Codex-specific surfaces to render those plan values consistently as `Pro Lite` without pushing provider-specific behavior into the shared formatter.

It also makes the OAuth usage decoder more tolerant of optional payload drift, but keeps `auto` mode honest about when that partial decode is no longer good enough. If OAuth loses the 5h/session lane because a window payload is malformed, we now fall back to the CLI. If a valid session lane still survives after reconciliation, including the reversed weekly/session shape, we keep the OAuth result instead of discarding usable data.

The test coverage here is intentionally broad because the blast radius crosses a few boundaries: OAuth decoding, dashboard parsing, menu and CLI rendering, user-facing error sanitization, and the CLI fallback path when Codex RPC chokes on newer plan variants.

PR #719, opened 2026-04-15
Recover Codex CLI usage from prolite decode errors

### LINKED ISSUE #717: Unhandled new OpenAI prolite plan causes Codex usage fetch failure
not trying to flex on you guys with that newly bought 100usd codex plan, trust me i was forced 😭

but just figured out it breaks codexbar

openai has added a new `prolite` plan and i do not think it is supported here yet. when codexbar tries to fetch usage / credits / limits, it throws:

unknown variant 'prolite'

repro:
1. use an openai account on the new `prolite` plan 
2. open codex / credits / usage section
3. let it fetch usage
4. watch it fail

expected:
it should support the new `prolite` plan and load usage normally

actual:
usage fetch fails because `prolite` seems unhandled
Image

waleedsworld:
## Summary
- recover Codex CLI usage and credits from the JSON body embedded in account/rateLimits/read decode errors
- keep the existing TTY /status fallback, but avoid depending on it when the RPC error already contains usable Pro Lite usage data
- add regression coverage for prolite decode mismatch payload recovery

## Validation
- `swift run CodexUsageRecoveryHarness` (temporary local verification harness, removed before commit)
- `./.build/lint-tools/bin/swiftformat Sources Tests --lint`
- `env DYLD_FRAMEWORK_PATH=/Library/Developer/CommandLineTools/usr/lib ./.build/lint-tools/bin/swiftlint --strict`

## Notes
- `swift test --filter CodexUsageFetcherFallbackTests` is blocked in this local CLT-only environment because `XCTest` is unavailable
- `pnpm check` is blocked on this machine because `Scripts/lint.sh` invokes `swiftlint` without the required `DYLD_FRAMEWORK_PATH`

Closes #717

### DIFF
--- a/Sources/CodexBarCore/UsageFetcher.swift
+++ b/Sources/CodexBarCore/UsageFetcher.swift
@@ -316,12 +316,28 @@ private struct RPCRateLimitWindow: Decodable, Encodable {
     let usedPercent: Double
     let windowDurationMins: Int?
     let resetsAt: Int?
+
+    static func from(_ window: CodexUsageResponse.WindowSnapshot?) -> Self? {
+        guard let window else { return nil }
+        return Self(
+            usedPercent: Double(window.usedPercent),
+            windowDurationMins: window.limitWindowSeconds / 60,
+            resetsAt: window.resetAt)
+    }
 }
 
 private struct RPCCreditsSnapshot: Decodable, Encodable {
     let hasCredits: Bool
     let unlimited: Bool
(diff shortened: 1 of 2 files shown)


ratulsarna:
Thanks for the PR @beginnerslvl ! This fix is already covered and merged via #710 .

waleedsworld:
my bad for missing that. sorry for the extra noise here

PR #727, opened 2026-04-16
Fix Antigravity probe: TLS delegate mismatch, extension token, and port ranking

icey-zhang:
## Summary

- **Fix URLSession delegate signature mismatch**: Swift 6.2 SDK changed `URLSessionDelegate` TLS challenge methods to require `@MainActor @Sendable` on the `completionHandler`. The old `@Sendable`-only signature caused the delegate to silently never be called, making all HTTPS connections to Antigravity's self-signed localhost server fail. Switched to the async delegate variant which matches the current protocol.
- **Use correct CSRF token for extension port fallback**: Antigravity uses separate tokens — `--csrf_token` for the HTTPS API port and `--extension_server_csrf_token` for the HTTP extension port. The HTTP fallback path now extracts and uses the correct extension token instead of the API token.
- **Rank port probe results**: Port probing now uses a three-tier priority (200 success > HTTP error > unreachable) so a 403 on the extension port doesn't shadow a working API port.

## Test plan

- [x] `swift build` passes with zero warnings on the modified file
- [x] All 13 `AntigravityStatusProbeTests` pass
- [x] Manual verification: `curl` confirms HTTPS API port returns 200 for `GetUserStatus`
- [ ] Verify Antigravity quota displays correctly in CodexBar menu bar after rebuild

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Providers & Usage
 - Synthetic: parse live five-hour, weekly, and search quota payloads, including continuous reset/regeneration details (#732). Thanks @baanish!
+- Antigravity: restore localhost probing with async TLS challenge handling, extension-token fallback, and best-effort port selection (#727). Thanks @icey-zhang!
 - Gemini: discover OAuth config in fnm/Homebrew/bundled CLI layouts so expired-token refresh keeps working (#723). Thanks @Leechael!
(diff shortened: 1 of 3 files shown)


steipete:

- Land commit: c5556c4aaf0244dcba01ab93c94fc95125ff2f03
- Merge commit: 2be8ae213134f1710eac76a224f0ca68d46d09d1

Thanks @icey-zhang!

PR #728, opened 2026-04-16
Support bundled Gemini CLI layout for OAuth credential extraction

icey-zhang:
## Summary

- Newer versions of Gemini CLI (installed via npm) use a bundled format where all code is packed into `bundle/chunk-*.js` files instead of the traditional `dist/src/code_assist/oauth2.js` path
- This caused token refresh to fail with "Could not find Gemini CLI OAuth configuration" because none of the existing search paths matched
- Add `searchBundleDirectory` fallback that scans `chunk-*.js` files in the bundle directory for embedded `OAUTH_CLIENT_ID` and `OAUTH_CLIENT_SECRET`

## Test plan

- [x] All 39 `GeminiStatusProbe` tests pass
- [x] Manual verification: Gemini quota displays correctly after rebuild with bundled Gemini CLI (`~/.nvm/.../gemini-cli/bundle/chunk-*.js` layout)

### DIFF
--- a/Sources/CodexBarCore/Providers/Gemini/GeminiStatusProbe.swift
+++ b/Sources/CodexBarCore/Providers/Gemini/GeminiStatusProbe.swift
@@ -514,6 +514,15 @@ public struct GeminiStatusProbe: Sendable {
             }
         }
 
+        // Bundled format: realPath points into a bundle/ dir containing chunk-*.js files
+        // with OAuth credentials embedded in one of the chunks.
+        if let result = self.searchBundleDirectory(binDir, fm: fm) {
+            return result
+        }
+        if let result = self.searchBundleDirectory(baseDir + "/bundle", fm: fm) {
+            return result
+        }
+
         return nil
     }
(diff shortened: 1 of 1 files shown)


ratulsarna:
Thanks for catching and fixing the bundled Gemini CLI layout issue, @icey-zhang. This was a real compatibility gap.

Since this PR was opened,  has gained a broader Gemini CLI package-root discovery path that handles the bundled layout by walking from  through its relative imports and parsing the OAuth constants from the imported bundle chunks. I also re-checked the current published  package shape, and that newer path covers the same  layout this PR targets.

I’m going to close this one as superseded so the queue stays tidy, but thank you again for the sharp report and patch.

ratulsarna:
Small correction: the inline code names were stripped from my previous comment by the shell. The intended details were: current `main` walks from `bundle/gemini.js`, parses its relative imports, and covers the current `@google/gemini-cli` `bundle/chunk-*.js` layout.

PR #729, opened 2026-04-16
Add Abacus AI provider

ratulsarna:
Supersedes #479.

Carries forward ChrisGVE's Abacus AI provider work onto the current `main` baseline and includes the follow-up fixes from review.

Thanks again to @ChrisGVE for the original implementation and the review iterations.


ratulsarna:
@codex review

PR #739, opened 2026-04-18
Prefer verification_uri_complete for Copilot login

### LINKED ISSUE #738: Copilot login opens the bare device-flow URL instead of verification_uri_complete
When logging into GitHub Copilot from CodexBar, the browser opens the generic device-flow verification page, but CodexBar does not use `verification_uri_complete` from the device code response.

That makes the login flow fragile on the desktop side, because users may not see an obvious pre-filled device code and can get stuck even though CodexBar already received the `user_code`.

Observed behavior:
- CodexBar requests a device code
- It copies `user_code` to the clipboard
- It opens `verification_uri`
- On the GitHub/Copilot side, the user may land on a page where the code is not visible or not obviously applied

Expected behavior:
- If `verification_uri_complete` is present, CodexBar should open that URL instead
- Fall back to `verification_uri` only when the complete URL is absent

I prepared a small fix for this by decoding `verification_uri_complete` and preferring it when opening the browser.

skhe:
Fixes #738

This updates the Copilot device login flow to prefer `verification_uri_complete` over `verification_uri` when opening the browser.

Why:
- GitHub device flow can return a complete verification URL that already carries the user code
- CodexBar currently ignores that field and always opens the bare verification page
- In practice, this can make the Copilot login flow confusing because the user code is not reliably visible in the browser flow

Changes:
- Decode `verification_uri_complete` in `CopilotDeviceFlow.DeviceCodeResponse`
- Add a small `verificationURLToOpen` helper with fallback to `verification_uri`
- Use that helper in `CopilotLoginFlow`
- Add tests for both the preferred and fallback paths

Verification:
- `swift build`
- `swift test --filter CopilotDeviceFlowTests`

Notes:
- `pnpm check` could not complete in my environment because the lint script attempted to download SwiftFormat and failed before running the checks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Providers & Usage
 - Gemini: discover OAuth config in fnm/Homebrew/bundled CLI layouts so expired-token refresh keeps working (#723). Thanks @Leechael!
+- Copilot: open the complete device-login verification URL when available so the browser flow carries the user code (#739). Thanks @skhe!
 
 ### Fixes
 - Keychain cache: preserve cached credentials when macOS temporarily denies keychain UI after wake, avoiding repeated prompts (#594). Thanks @josepe98!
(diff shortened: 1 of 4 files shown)


steipete:

- Land commit: ef86cdac
- Merge commit: d968dd9696125fad5e2dd4eb016eb520f3bb6262

Thanks @skhe!

Issue #533, opened 2026-03-15
macOS widget is registered but never appears in Widget Gallery


Theoz001:
### Summary

CodexBar installs correctly on macOS, and the widget extension is present and registered, but no CodexBar widgets appear in the Widget Gallery at all.

Searching for `CodexBar` returns nothing, and there is no `CodexBar` section in the widget list.

### Environment

- CodexBar version: `0.18.0-beta.3`
- macOS version: `26.2 (25C56)`
- Install location: `/Applications/CodexBar.app`

### What I checked

- The widget extension exists:
  - `/Applications/CodexBar.app/Contents/PlugIns/CodexBarWidget.appex`
- PlugInKit sees it and marks it as elected:
  - `com.steipete.codexbar.widget`
- Code signing verification passes for both the app and the widget extension
- I restarted `pkd`, `Dock`, and `NotificationCenter`
- I also re-added the widget with `pluginkit` and re-seeded LaunchServices

### Actual behavior

When opening the Widget Gallery, `chronod` attempts to load `com.steipete.codexbar.widget`, but the extension exits immediately and the gallery never shows any CodexBar widgets.

### Relevant logs

```text
chronod ... [com.steipete.codexbar::com.steipete.codexbar.widget] scheduling query
chronod ... [com.steipete.codexbar::com.steipete.codexbar.widget] performing query

containermanagerd ... [com.steipete.codexbar.widget] requesting [group.com.steipete.codexbar]: REJECTED.
Requestor's signature does not allow it to access a TCC-protected group container.
Group containers identifiers should be prefixed by requestor's team ID to allow access on this platform.

chronod ... query failed - will try lazy reload later:
Error Domain=NSCocoaErrorDomain Code=4099
"The connection to service with pid -1 named (null) was invalidated."

steipete:
Fixed on current `main`.

The reported failure was the legacy app group `group.com.steipete.codexbar` being rejected by macOS. Current builds use the team-prefixed app group for both app and widget: `Y5PE65HELJ.com.steipete.codexbar`.

Proof from the current packaged app:
- `CodexBar.app` entitlement `com.apple.security.application-groups`: `Y5PE65HELJ.com.steipete.codexbar`
- `CodexBarWidget.appex` entitlement `com.apple.security.application-groups`: `Y5PE65HELJ.com.steipete.codexbar`
- widget Info.plist includes `CodexBarTeamID = Y5PE65HELJ`

[...]

PR #782, opened 2026-04-21
Fix widget App Intents packaging

vincentyangch:
## Summary
- generate and bundle `Metadata.appintents` for `CodexBarWidget` during packaging so the widget extension ships the App Intents metadata that Apple WidgetKit expects
- make widget configuration intent parameters optional and resolve safe defaults for unset provider/metric selections
- add focused widget tests for the new fallback resolution helpers

## Why
On the official `0.22` release, the widget extension bundle is registered and signed correctly but ships with an empty widget `Resources/` directory. During local debugging on macOS `26.4.1`, the installed widget repeatedly failed descriptor loading in `chronod` with `NSCocoaErrorDomain Code=4099` during `getAllDescriptors`, and the widget never appeared in the gallery.

Apple's own working widget extensions on this macOS build include `Resources/Metadata.appintents/...`, while the CodexBar release bundle does not. This patch teaches the packaging script to generate and embed that metadata using Apple's `appintentsmetadataprocessor` and also aligns the WidgetConfigurationIntent parameter model with what the metadata processor expects.

## Details
- `Scripts/package_app.sh`
  - add a `generate_widget_appintents_metadata` helper that builds the `CodexBarWidget` Xcode scheme into a derived-data staging area
  - collect generated `*.swiftconstvalues`, dependency metadata file lists, and source lists
  - invoke `appintentsmetadataprocessor`
  - bundle the resulting `Metadata.appintents` directory into `CodexBarWidget.appex/Contents/Resources`
- `Sources/CodexBarWidget/CodexBarWidgetProvider.swift`
  - make `ProviderSelectionIntent.provider` optional
  - make `CompactMetricSelectionIntent.provider` and `.metric` optional
  - add fallback helpers so the runtime still defaults to `Codex` / `Credits`
- `Tests/CodexBarTests/CodexBarWidgetProviderTests.swift`
  - add focused tests for the new fallback helper behavior

## Validation
- `bash -n Scripts/package_app.sh`
- `swift test -q --filter CodexBarWidgetProviderTests`
- confirmed the patched packaged widget bundle contains:
  - `Metadata.appintents/version.json`
  - `Metadata.appintents/extract.actionsdata`

## Notes
- I could not fully prove the user-visible gallery fix end-to-end on the test machine because locally built app bundles were signed with an Apple Development identity rather than the official release signing setup, and WidgetKit app-group authorization still diverged under that signing context.
- Even with that limitation, the release-vs-patched bundle comparison was clear: the official `0.22` widget shipped without `Metadata.appintents`, while the patched build includes it.
- This likely relates to #533.

### DIFF
--- a/Scripts/package_app.sh
+++ b/Scripts/package_app.sh
@@ -98,6 +98,86 @@ path.write_text(text)
 PY
 }
 
+generate_widget_appintents_metadata() {
+  local widget_resources_dir="$1"
+  local xcode_conf
+  local host_arch
+  local derived_dir
+  local build_dir
+  local object_dir
+  local source_file_list
+  local const_values_list
+  local dependency_metadata
+  local static_dependency_metadata
+  local appintents_tool
+  local sdk_root
+  local swiftc_path
+  local toolchain_dir
+  local xcode_version
+
+  xcode_conf="Release"
+  if [[ "$LOWER_CONF" == "debug" ]]; then
+    xcode_conf="Debug"
+  fi
+
+  host_arch=$(uname -m)
+  derived_dir="$ROOT/.build/xcode-widget-metadata-${LOWER_CONF}"
+  build_dir="$derived_dir/Build/Intermediates.noindex/CodexBar.build/${xcode_conf}/CodexBarWidget.build"
+  object_dir="$build_dir/Objects-normal/${host_arch}"
+  source_file_list="$object_dir/CodexBarWidget.SwiftFileList"
+  const_values_list="$object_dir/CodexBarWidget.SwiftConstValuesFileList"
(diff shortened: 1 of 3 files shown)


ngutman:
Thanks for the fix here — I opened #783 to carry this forward with the same packaging change plus a small follow-up cleanup: keeping the widget configuration parameters non-optional and using `@Parameter(default:)` instead of optional fallback helpers. The new PR preserves your original commit credit and supersedes this one for merge.

ngutman:
Closing in favor of #783, which preserves the original packaging fix, carries the cleaner App Intent default-parameter follow-up, and has now been merged. Thanks again @vincentyangch for the original fix.

PR #793, opened 2026-04-26
Moonshot: auto-discover API key from Kimi CLI config

maxitoon:
### 1. Moonshot: auto-discover API key from Kimi CLI config

Users of the [Kimi CLI](https://github.com/moonshot-ai/Kimi-CLI) (which uses Moonshot AI models like `kimi-k2.6`) often enable the **Kimi** provider in CodexBar. That provider only supports web cookies from kimi.com, so terminal-only users see "No available fetch strategy for kimi."

Their API key is actually stored in `~/.kimi/config.toml` under `[providers."managed:moonshot-ai"]`, but CodexBar never looked there.

**Solution:** Extend `MoonshotSettingsReader.apiKey()` to fall back to `~/.kimi/config.toml` when `MOONSHOT_API_KEY` / `MOONSHOT_KEY` env vars are absent.

- `parseKimiConfigAPIKey()` — simple TOML-line parser that respects section boundaries, strips quotes, and ignores comments
- `loadKimiConfigContents()` — reads `~/.kimi/config.toml`
- Fallback priority: env vars → Kimi CLI config → nil
- 7 unit tests covering sections, quotes, comments, and boundaries

### 2. Increase Overview tab provider limit from 3 → 5

Power users often have 4–5 active providers (Codex, Claude, Moonshot, OpenRouter, Cursor). The previous limit of 3 forced switching in the provider picker. All UI logic already derives from the constant, so this is a safe single-line change.

## Verification

- `swift build --target CodexBarCore` ✅
- Standalone logic tests (14/14 passed against real `~/.kimi/config.toml`) ✅
- Built and running locally with both changes ✅

## Checklist

- [x] Logic follows existing CodexBar patterns
- [x] No breaking changes to existing env-var resolution
- [x] Unit tests added
- [x] Minimal diff

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -40,3 +40,4 @@
 - Keep provider data siloed: when rendering usage or account info for a provider (Claude vs Codex), never display identity/plan fields sourced from a different provider.***
 - Claude CLI status line is custom + user-configurable; never rely on it for usage parsing.
 - Cookie imports: default Chrome-only when possible to avoid other browser prompts; override via browser list when needed.
+- Moonshot provider: official Moonshot uses `GET /v1/users/me/balance` on `api.moonshot.ai` or `api.moonshot.cn`, exposes a Region picker, shows balance-only/no-limit semantics, and must stay distinct from the third-party Kimi K2 provider tracked in issue #473.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.21 — Unreleased
 
 ### Highlights
+- Moonshot: add an official Moonshot platform provider with International/China region selection, config/env API-key support, and balance-only display semantics distinct from Kimi K2 (#473).
 - z.ai: preserve weekly and 5-hour token quotas together, surface the 5-hour lane correctly across the menu/menu bar, and add regression coverage (#662). Thanks to @takumi3488 for the original fix and investigation.
 - Cursor: fix a crash in the usage fetch path and add regression coverage (#663). Thanks @anirudhvee for the report and validation!
 - Antigravity: accept localhost TLS challenges when probing the local language server so usage/account info loads again instead of reporting `no working API port found` (#693, fixes #692). Thanks @anirudhvee!
@@ -13,6 +14,7 @@
 - Codex: make OpenAI web extras opt-in for fresh installs, preserve working legacy setups on upgrade, add an OpenAI web battery-saver toggle, and keep account-scoped dashboard state aligned during refreshes and account switches (#529). Thanks @cbrane!
 
 ### Providers & Usage
+- Moonshot: add the official balance endpoint provider for `api.moonshot.ai` / `api.moonshot.cn`, store API keys in `~/.codexbar/config.json`, and keep Kimi K2 untouched so official Moonshot credentials are no longer conflated with the third-party Kimi K2 service (#473).
 - z.ai: preserve both weekly and 5-hour token quotas, keep the existing 2-limit behavior unchanged, and render the 5-hour quota as a tertiary row in provider snapshots and CLI/menu cards (#662). Credit to @takumi3488 for the original fix and investigation.
 - Cursor: fix the usage fetch path so failed or cancelled requests no longer crash, and add Linux build and regression test coverage fixes (#663).
 - Antigravity: scope insecure localhost trust handling to `127.0.0.1` / `localhost`, keep localhost requests cancellable, and restore local quota/account probing on builds that previously failed TLS challenge handling (#693, fixes #692). Thanks @anirudhvee!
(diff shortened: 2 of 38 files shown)


maxitoon:
cc @steipete — this PR builds on top of #705 (Moonshot provider) and adds auto-discovery of the Moonshot API key from the Kimi CLI config file (`~/.kimi/config.toml`).

**Why this matters:** Kimi CLI users (like myself) naturally enable the "Kimi" provider in CodexBar, but that only supports web cookies from kimi.com. Their actual API key lives in `~/.kimi/config.toml` under `[providers."managed:moonshot-ai"]`. This change makes the Moonshot provider "just work" for terminal users without requiring manual env vars.

**Tested:** Built and running locally — the Moonshot provider now fetches usage automatically from my Kimi CLI config. ✅

giuseppebisemi:
I tested this locally and these were the concrete UI changes that worked well for me:

For the descriptor, Moonshot can reuse the existing Kimi branding:

```swift
iconStyle: .kimi,
iconResourceName: "ProviderIcon-kimi",
```

For the menu bar label, I special-cased Moonshot to display the balance amount from the identity text:

```swift
if provider == .moonshot,
   let balance = Self.moonshotBalanceDisplayText(snapshot: snapshot)
{
    return balance
}
```

where `moonshotBalanceDisplayText` strips `Balance:` and keeps only the first amount, e.g. `Balance: $49.58 · $0.42 in deficit` -> `$49.58`.

For the icon alignment, I adjusted the shared `ProviderIcon-kimi.svg` by applying a small upward translation to the main path:

```svg
<path transform="translate(0 -0.8)" ... />
```

That made the Kimi glyph sit closer to the optical center of the 16px menu bar icon slot, and improves alignment for Kimi, Kimi K2, and Moonshot since they share the same asset.

steipete:
Thanks for the provider work. I don’t think this branch is reviewable as-is.

The PR has no merge ref right now, so it appears conflicted. It also mixes several unrelated changes: a Moonshot provider, Kimi CLI config fallback behavior, `AGENTS.md`, a new `CLAUDE.md`, README/docs updates, and an unrelated overview provider limit change from 3 to 5.

Please split this into focused PRs. The useful shape would be:

- one PR for official Moonshot provider support
- a separate PR for any Kimi CLI config fallback
- no `CLAUDE.md` addition
- no overview limit change in the provider PR

That will make the provider/API behavior possible to review cleanly.

steipete:
Thanks for putting this together. I’m going to close this PR rather than merge it as-is because the current branch mixes several separate changes and is based on an older provider surface.\n\nMain blockers:\n- The PR title is about Kimi CLI config discovery, but the diff adds a new Moonshot provider, changes the Overview provider limit, and creates .\n- CodexBar’s maintainer notes live in ; we intentionally ignore .\n- Official Moonshot support needs to be reviewed as its own provider PR, separate from Kimi/Kimi K2 behavior.\n- The Kimi CLI config fallback should be a small follow-up only after the target provider/source exists on current .\n\nA mergeable path would be: open a fresh PR against current  that does exactly one thing, e.g. either (1) add official Moonshot provider support, or (2) add config auto-discovery for an already-existing Moonshot/Kimi source. Please keep the unrelated overview-limit change out of that PR. Thanks again for the investigation.

steipete:
Thanks for putting this together. I am going to close this PR rather than merge it as-is because the current branch mixes several separate changes and is based on an older provider surface.

Main blockers:
- The PR title is about Kimi CLI config discovery, but the diff adds a new Moonshot provider, changes the Overview provider limit, and creates CLAUDE.md.
- CodexBar maintainer notes live in AGENTS.md; we intentionally ignore CLAUDE.md.
- Official Moonshot support needs to be reviewed as its own provider PR, separate from Kimi/Kimi K2 behavior.
- The Kimi CLI config fallback should be a small follow-up only after the target provider/source exists on current main.

A mergeable path would be: open a fresh PR against current main that does exactly one thing, e.g. either (1) add official Moonshot provider support, or (2) add config auto-discovery for an already-existing Moonshot/Kimi source. Please keep the unrelated overview-limit change out of that PR. Thanks again for the investigation.

maxitoon:
Follow-up clarification from my side: my goal is not to stay on this older/home-made branch if the problem is solved upstream.

From the current repo state, #911 landed Moonshot / Kimi API provider support on `main`, so the maintained path is to use upstream `main` or wait for the next release that includes #911. The latest published release I see right now is `v0.25.1`, which does not contain #911 yet.

The only behavior from this PR that I still rely on but do not see in `main` is auto-discovery of the Moonshot API key from Kimi CLI config at `~/.kimi/config.toml` (`[providers."managed:moonshot-ai"] api_key = ...`). If the intended supported path is manual entry in CodexBar Settings or `MOONSHOT_API_KEY` / `MOONSHOT_KEY`, I can drop this branch and use that. If auto-discovery is still welcome, it should be a tiny separate follow-up PR against current `main`, touching only `MoonshotSettingsReader` and its tests.

PR #797, opened 2026-04-26
Restart Codex app when switching accounts

leezhuuuuu:
## Summary
- add a Codex provider setting to restart Codex.app around system account switches
- quit running Codex.app instances before swapping live auth, then relaunch from the detected app path
- cover the setting persistence and restart promotion paths with tests

## Tests
- pnpm check
- swift test --filter CodexAccountPromotionServiceTests --filter CodexActiveSourceConfigTests


ratulsarna:
Thank you for working on the account-switching flow. We are going to close this PR because restarting Codex during account switching is a fairly disruptive behavior and the current implementation still has unresolved review feedback around shutting down before the switch is guaranteed. We appreciate the attempt to improve the flow, but this needs a different approach before it can move forward.

PR #799, opened 2026-04-27
Fix token account switcher not refreshing menu with new account data

### LINKED ISSUE #798: Token account switcher does not refresh menu with new account data
## Description

When switching between token accounts (e.g. multiple Claude session keys) in the menu bar dropdown, the usage card does not update to show the newly selected account's data. The menu continues displaying the previous account's usage until manually refreshed or the menu is closed and reopened.

## Root cause

Two issues in `StatusItemController+Menu.swift` in the `TokenAccountSwitcherView` `onSelect` handler:

1. **`store.refresh()` is silently dropped.** The global `refresh()` method has a `guard !self.isRefreshing else { return }` gate. When the menu opens, `scheduleOpenMenuRefresh` already kicks off a refresh. If the user switches token accounts while that refresh is in-flight, the new `store.refresh()` call returns immediately without fetching data for the newly selected account.

2. **No menu repopulation after async refresh.** Even when the refresh does run, `invalidateMenus()` intentionally skips open menus ("Don't refresh menus while they're open"). The Codex account switcher handles this by explicitly calling `refreshOpenMenuIfStillVisible` after its async refresh completes, but the token account switcher did not.

## Expected behavior

Switching token accounts should immediately fetch the new account's usage data and update the open menu in-place.

## Environment

- CodexBar 0.24 (development)
- macOS 15

Zeko369:
## Summary

- Replace `store.refresh()` with targeted `store.refreshProvider(display.provider)` in the token account switcher's `onSelect` handler — avoids being silently dropped by the global `isRefreshing` guard when a menu-open refresh is already in-flight
- Add `refreshOpenMenuIfStillVisible` after the async fetch completes so the open menu updates in-place (matching the Codex account switcher pattern)

Fixes #798

## Test plan

- [ ] Enable a provider with 2+ token accounts (e.g. Claude with multiple session keys)
- [ ] Open the menu and switch between accounts — the usage card should update each time without needing to manually refresh or close/reopen the menu
- [ ] Verify switching while a background refresh is in-flight still picks up the new account

### DIFF
--- a/Sources/CodexBar/StatusItemController+Menu.swift
+++ b/Sources/CodexBar/StatusItemController+Menu.swift
@@ -730,8 +730,9 @@ extension StatusItemController {
                 self.settings.setActiveTokenAccountIndex(index, for: display.provider)
                 Task { @MainActor in
                     await ProviderInteractionContext.$current.withValue(.userInitiated) {
-                        await self.store.refresh()
+                        await self.store.refreshProvider(display.provider)
                     }
+                    self.refreshOpenMenuIfStillVisible(menu, provider: display.provider)
                 }
                 self.populateMenu(menu, provider: display.provider)
                 self.markMenuFresh(menu)


steipete:
Thanks @Zeko369 — I landed this on main in 1164cb89 with changelog credit and a regression test.

The final patch keeps your provider-scoped refresh idea, adapted to current main: token account selections now call `refreshProvider(display.provider)` so a menu-open global refresh cannot drop the account switch, then the open menu is refreshed again after fresh data arrives.

Verified:
- `swift test --filter StatusMenuTokenAccountSwitcherTests`
- `swift test --quiet`
- `pnpm check`
- `swiftlint --strict`

Issue #776, opened 2026-04-21
Add Notification System


danielsimao:
I would like to receive notifications when I'm about to reach usage limits or even when my usage is about to reset.

gianpaj:
We already have some notifications 

<img width="461" height="305" alt="Image" src="https://github.com/user-attachments/assets/12bf768b-efa2-4fce-a242-a5ac70adf32a" />

<img width="397" height="207" alt="Image" src="https://github.com/user-attachments/assets/a2fb1280-ddf3-48d7-b356-b5be9adf7626" />

ratulsarna:
As highlighted by @gianpaj , notifications exist in the app.

Issue #617, opened 2026-03-29
Allow configurable alert thresholds with notification bubbles and sounds.


victorrentea:
I run often into being out of quota. I'd like to be notified, for example, when 5% of my quota is left for the 5 hours. Maybe 10% customizable with the notification bubble in macOS as well as sound. Then maybe allow 2 alarm levels. It's sometimes frustrating when I start a large task and I run out of quota just halfway. Restarting the whole thing after the quota refills can be very painful, being token-intensive, so I might not start the task if I knew that I'm only 5% left.

steipete:
Closing as completed. Opt-in quota warning notifications, warning markers, configurable thresholds, and provider-level session/weekly overrides landed on main (#852).

Issue #456, opened 2026-03-02
Feature Request: Desktop / System Notification When Permission Prompt is Waiting


luoling8192:
#### Pain Point Description
CodexBar already does an amazing job at passive monitoring — showing usage quotas, reset times, and limits right in the menu bar without any login hassle. It's become an essential part of my workflow.

One very frequent frustration with the underlying CLIs (especially **Claude Code** and **Codex CLI**) is the silent hang when they request permission for tool use:
- File write (Write)
- Shell command execution (Bash/Execute)
- Reading sensitive paths, etc.

The CLI just waits indefinitely for `Y/n` or approval in the terminal, with **no visual/audio cue** if you're focused elsewhere (VS Code, browser, another desktop). This leads to:
- "Why is it stuck?" debugging sessions
- Broken flow during long autonomous tasks
- Minutes/hours lost if you step away

Since CodexBar is already deeply integrated with these providers (via RPC, PTY fallback, stdout probing for status/usage), it seems like a natural fit to also detect and surface these "active waiting" states with a system-level alert.

#### Proposed Feature
Add an **optional system notification** (macOS Notification Center) when CodexBar detects a permission prompt is active in a monitored CLI session.

Example triggers:
- Output matches patterns like `(Y/n)`, "Approve tool use?", "permission required", "awaiting approval", etc.
- PTY / probing detects idle/waiting state after a tool invocation

Notification details (customizable via Settings?):
- Title: "Claude Code / Codex CLI Waiting for Approval"
- Body: "Permission prompt active → check terminal to continue. Provider: [Claude/Codex] | Session: [short task or PID if available]"
- Optional: Subtle sound (system Glass/Ping), menu bar icon badge/flash/pulse (e.g. orange until cleared)
- Optional: Debounce (max 1 per 30s) to avoid spam

This would complement the existing quota monitoring perfectly: one tells you "how much you have left", the other tells you "hey, act now so you can use it".

#### Why CodexBar is the Ideal Place for This
- Already runs persistently and observes CLI state (usage parsing, PTY fallback for Claude/Codex)
- No need to patch upstream CLIs — just pattern-match on output or state transitions (similar to how quota/usage is detected)
- Keeps everything local, lightweight, privacy-respecting — aligns with CodexBar's philosophy
- Many users (including me) keep CodexBar running 24/7 → perfect for catching these moments without extra tools

#### Implementation Ideas (Non-Binding – Happy to Contribute Code)
[...]

PR #801, opened 2026-04-27
Add Apple Shortcuts and webhook actions for CodexBar notifications

kibermaks:
## Summary

Adds automation delivery options for CodexBar notification events, plus a dedicated Notifications tab in Preferences.

CodexBar can now route important events to:

- local macOS notifications
- configurable notification sounds
- Apple Shortcuts
- webhook-style URLs
- custom URL schemes

This makes notifications useful when the user steps away from the Mac, and it also enables autonomous workflows. For example, a quota-restored event can run an Apple Shortcut that alerts an iPhone or Apple Watch, triggers HomeKit/Home Assistant automation, calls Pushover/ntfy/Telegram/Pushcut, or invokes a webhook that resumes a paused agent/task once quota is available again.

## What Changed

- Adds a new Preferences → Notifications tab.
- Adds a global notifications toggle.
- Adds global custom sound volume.
- Adds per-event settings for:
  - session quota depleted
  - session quota restored
  - provider login successful
  - Augment session expired
- Adds per-event sound selection and preview.
- Adds per-event webhook URL delivery.
- Adds per-event Apple Shortcut execution by shortcut name.
- Adds starter Shortcut templates/docs under `docs/shortcuts`.
- Migrates existing session quota depleted/restored notifications onto the new configurable delivery path.
- Preserves the legacy session quota notifications setting by mapping it to the split depleted/restored event settings.

## Why

The current local macOS notification is useful while actively sitting at the computer, but it does not cover common “step away from the machine” workflows or longer autonomous runs.

This PR lets CodexBar act as the event source and lets users decide what happens next:

- send an iPhone or Apple Watch notification through Shortcuts
- flash a light or trigger a HomeKit/Home Assistant scene when quota is restored
- call a webhook when a provider becomes usable again
- resume an external automation, queue, or agent workflow after quota recovers
- route provider-login or Augment-session-expired events to external automations
- keep local banners/sounds for desktop use
- disable or customize noisy events independently

The goal is not to make CodexBar own every integration. It exposes clean notification events that can be connected to the automation system users already rely on.

## Scope

This PR focuses on notification delivery and notification settings.

It does not add configurable quota threshold warnings yet. That remains a separate feature request, but this gives that work a cleaner delivery layer to build on.

It also does not implement CLI permission-prompt detection yet. The event model should make that easier to add later without duplicating delivery logic.

## Related

Related to:

- #617
- #456
- #263
- #776
- #433
- #619

## Testing

- `git diff --check`
- `swift test`
- `swift test --filter SubprocessRunnerTests`


steipete:
Thanks for the solid test coverage here. I found one thing that should be fixed before this is merge-ready:

`Sources/CodexBar/AppNotifications.swift` logs the fully rendered hook URL after delivery/failure/custom-scheme open. Webhook URLs often contain bearer-style secrets in the path or query string, and the current log redactor only handles cookie/auth/bearer header text, not generic URL tokens. So a Slack/Discord/Zapier/Raycast-style URL configured in Notifications could end up in `CodexBar.log` with its secret intact.

Please avoid logging the full hook URL. Logging scheme + host, or a redacted URL with query removed, should be enough for diagnostics. Relevant lines on this branch are around `AppNotifications.swift:225`, `:235`, `:239`, and `:248`.

steipete:
Thanks for the detailed work here. I am going to close this PR for queue cleanup, not because the automation idea is bad.

This is no longer a safe merge candidate:

- the branch is dirty against current main
- the diff is large, about 1.6k additions across 29 files
- it combines several concerns: notification delivery, new preferences UI, settings migration, Shortcuts, webhooks, Augment events, dashboard navigation test changes, and pace/widget test updates
- it only has GitGuardian on the current head, so the current branch does not have full CI evidence
- the app already changed quite a bit around quota notifications/settings since this branch was opened

A smaller current-main series would be much easier to land:

1. add an isolated notification event/delivery model with tests
2. add one external action type, probably Apple Shortcuts first
3. add webhook delivery separately, with clear URL validation and privacy notes
4. move existing quota notifications onto the new path only after the delivery seam is proven

That would keep the useful part reviewable and avoid mixing unrelated settings/UI churn into one PR. Appreciate the contribution; closing this branch shape so the active queue stays actionable.

PR #803, opened 2026-04-28
Document disabling Keychain access

davidalee:
## Summary
- document how to disable CodexBar Keychain access before reopening the app
- add a CLI-only Codex + Claude config example that turns cookie sources off
- clarify that the official provider CLIs may still trigger their own credential prompts

## Verification
- README checked first: existing Keychain section only documented allowing access, not disabling it before launch
- ./Scripts/compile_and_run.sh: Swift targets built, then packaging failed because xcodebuild could not load Apple IDESimulatorFoundation/DVTDownloads plugin and suggested xcodebuild -runFirstLaunch

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,27 @@ Wondering if CodexBar scans your disk? It doesn’t crawl your filesystem; it re
     - Find the browser’s “Safe Storage” key (e.g., “Chrome Safe Storage”, “Brave Safe Storage”, “Firefox”, “Microsoft Edge Safe Storage”).
     - Open the item → **Access Control** → add `CodexBar.app` under “Always allow access by these applications”.
     - This removes the prompt when CodexBar decrypts cookies for that browser.
+  - **How do I disable Keychain access instead?**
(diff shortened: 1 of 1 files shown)


steipete:
Thanks for the docs pass here. I covered the current need on `main` in c7950231 with a focused update to `docs/configuration.md`.

I am closing this PR because the branch is now stale against current `main`: the diff includes a very large unrelated rollback/removal set, so it is no longer safe to land as-is. The landed docs now cover the keychain/manual-cookie configuration path without carrying those unrelated changes.

davidalee:
@steipete I still think this is useful information to include. Reason being I was not able to access the app without granting keychain access.

I would deny access, but the same popup would continue re-firing. I hit deny about 10 times before giving up and having to figure out how to manually disable the keychain access prompt. Only then did the app initiate. And everything works (with the exception of "claude code session time remaining" as I mentioned in another issue).

It's not personal, I just don't feel safe granting unfettered keychain access to an open source project given all of the supply chain attacks we've seen proliferate recently.

PR #813, opened 2026-04-29
fix(cursor): parse enterprise overall/pooled usage

fcamus00:
## Summary
- Enterprise/Team Cursor accounts always showed **100% remaining** in the menu because the API response uses `individualUsage.overall` and `teamUsage.pooled` instead of the legacy `individualUsage.plan` block CodexBar's decoder knows about.
- Added `CursorOverallUsage` and `CursorPooledUsage` decoders, extended `CursorIndividualUsage` and `CursorTeamUsage`, and updated `parseUsageSummary` precedence so the personal cap (`overall`) wins, with the shared pool (`pooled`) as a last-resort fallback.
- USD figures now flow through from `overall`/`pooled` into `planUsedUSD` / `planLimitUSD` so the debug log surfaces the right dollars.

## Compatibility
- Existing `plan` precedence (Pro / Hobby / Team with `totalPercentUsed`, lane percents, plan ratio) is **unchanged** — only adds new fallbacks below it.
- Legacy `/api/usage` request-based plan logic untouched.
- Menu UI / labels untouched (no new rows added; default Cursor flow is preserved).

## Tests
3 new tests in `Tests/CodexBarTests/CursorStatusProbeTests.swift`:
- Decodes the live Enterprise JSON shape (`overall` + `pooled`).
- Confirms `individualUsage.overall` (73.84%) wins over `teamUsage.pooled` (45.25%) for the headline.
- Confirms `pooled`-only fallback when no individual data exists.
- Regression guard: legacy `plan` block still wins when present.

\`swift test --filter CursorStatusProbeTests\` → 32/32 green.

## Verification
- Built and ran with \`./Scripts/compile_and_run.sh\`.
- Live Cursor Enterprise account that previously read 100% now correctly reflects ~26% remaining (matches Cursor's own dashboard).
- Verified via **Preferences → Debug → Probe logs → Cursor**.

## Reproducer (sanitized live response)
\`\`\`json
{
  "membershipType": "enterprise",
  "limitType": "team",
  "isUnlimited": false,
  "individualUsage": {
    "overall": { "enabled": true, "used": 7384, "limit": 10000, "remaining": 2616 }
  },
  "teamUsage": {
    "onDemand": { "enabled": true, "used": 0, "limit": null, "remaining": null },
    "pooled":   { "enabled": true, "used": 12725135, "limit": 28122000, "remaining": 15396865 }
  }
}
\`\`\`

Before: menu shows `100% remaining`. After: shows ~`26% remaining` (73.84% used).


ratulsarna:
THanks @fcamus00 for the contribution!

PR #819, opened 2026-04-30
Add Chinese (Simplified) localization

markhome1:
Add Chinese (Simplified) localization support for CodexBar.


ratulsarna:
@codex review

PR #820, opened 2026-04-30
Add feature: Chinese localization and README updates

AidenYang1:
## Hi，Steipete

Ur app has a large users in China, but I noticed CodexBar doesn't have localization (Chinese) support yet — so I went ahead and made some improvements. I spent an afternoon refactoring and decoupling the localization code, and also worked on the translations to make them feel more natural and native — staying true to the original meaning without being machine.

## Here's what I've done:

1. I added an in-app language switcher in the **General** settings.
2. I decoupled all the hardcoded strings from the original code and added `localization.strings` support.
3. I did a manual review of the Chinese translations specifically. (There might be a few minor spots I missed, but overall it should be a solid, trustworthy Chinese localization.)
4. I added a Chinese README and a toggle button to switch between them.

## PR
I did code audit on your latest main branch — as long as there are no major structural changes on your latest, you should be able to apply this PR directly. You're also can review it by Claude to check. But for the Chinese translations themselves, trust me on those.

## Other language support
I refactored the hardcoded strings section, so if other contributors want to add other languages, all they need to do is create a new language folder with a strings file、localize it and add a case to the `AppLanguage` enum.

Feel free to reply if u have any questions — hope this helps CodexBar reach more users

Thanks,
Aiden

## screenshots

image

image

image

image

image


AidenYang1:
Fixed in the latest commit — added value: fallback to all 32 NSLocalizedString calls in CodexBarCore so CLI and non-app contexts always show readable English strings.

steipete:
Thanks for the thoughtful localization work. We just landed Simplified Chinese app localization and the in-app language selector via #819, so this PR is now superseded and conflicts heavily with main. If there are README_zh or translation refinements here that are still missing, please open a smaller follow-up PR on top of current main.

PR #830, opened 2026-05-03
fix(cli): provider selection queries all 28 providers when 3+ are enabled

lhoBas:
## Summary

- `providerSelection()` returned `.all` when `enabled.count >= 3`, which expands to every registered provider (currently 28) instead of just the user's enabled ones. A config with e.g. codex + claude + copilot would query all 28 providers, producing 25 spurious errors and unnecessary overhead.
- Now `.all` only triggers when the enabled set matches the full registry; otherwise uses `.custom(enabled)`.

Bug introduced in a1b94b48 ("refactor: move provider secrets to config", Jan 18 2026) — the `>= 3` threshold was likely safe when there were few providers but broke as the registry grew.

## Test plan

- [x] Added regression test in `TestsLinux/ProviderSelectionTests.swift` that fails without the fix (3 enabled → 28 queried) and passes with it
- [x] All existing `CodexBarLinuxTests` pass (12 total)

### DIFF
--- a/Sources/CodexBarCLI/CLIHelpers.swift
+++ b/Sources/CodexBarCLI/CLIHelpers.swift
@@ -17,9 +17,12 @@ extension CodexBarCLI {
         if let rawOverride, let parsed = ProviderSelection(argument: rawOverride) {
             return parsed
         }
-        if enabled.count >= 3 { return .all }
-        if enabled.count == 2 {
-            let enabledSet = Set(enabled)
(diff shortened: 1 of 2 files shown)


steipete:
Thanks @lhoBas, this is fixed on main in e296a1f6. I folded the behavior into a grouped maintainer patch with regression coverage for the 3+ enabled-provider case and credited you in the changelog.

PR #836, opened 2026-05-04
fix(claude): recognize OAuth subscription type

### LINKED ISSUE #824: Claude Usage Dashboard opens to wrong page
Usage Dashboard (under the Claude tab) takes me here:

https://platform.claude.com/settings/billing

For me, that page shows my API/organization usage.  The correct URL (for personal subscriptions) appears to be:

https://claude.ai/settings/usage

shixy96:
## Summary

- Fixes #824.
- Parse and retain `subscriptionType` from Claude Code OAuth credentials.
- Prefer `subscriptionType` over `rateLimitTier` when deriving the Claude OAuth login method, while keeping `rateLimitTier` as a fallback.
- Preserve `subscriptionType` when refreshed OAuth credentials are written back to CodexBar's cache.
- Add regression coverage for Pro credentials that report `rateLimitTier: "default_claude_ai"`.

## Root Cause

Claude Code can store OAuth credentials for a Pro account with `subscriptionType: "pro"` while `rateLimitTier` is set to `default_claude_ai`, as documented in https://github.com/anthropics/claude-code/issues/38909. CodexBar only used `rateLimitTier` to infer the plan, so it could not derive `Claude Pro`; downstream subscription checks then treated the account as unknown and used the generic Claude dashboard URL.

## Solution

Use the explicit OAuth `subscriptionType` as the primary plan signal and fall back to `rateLimitTier` only when `subscriptionType` is absent. This avoids a broad "all OAuth means subscription" shortcut and keeps non-subscription OAuth accounts from being misclassified.

## Why This Fix

This changes the shared Claude OAuth credential model and plan resolver instead of patching the dashboard action. The dashboard was only the visible symptom; the broken state was that OAuth Pro credentials could not resolve to a subscription plan.

`subscriptionType` is the stable account-level signal in this credential shape. `rateLimitTier` can still identify Max, Team, Enterprise, and other known tiers, so CodexBar keeps it as a fallback. The fix does not map bare `default_claude_ai` to Pro, and it does not treat every OAuth account as a subscription.

The refreshed credential cache now preserves `subscriptionType`, so the fix survives token refresh instead of working only until the next cache write.

## Validation

- `swiftformat Sources Tests`
- `swiftlint --strict`
- `pnpm check`
- `git diff --check`
- `swift test --filter 'ClaudePlanResolverTests|ClaudeOAuthTests'`
- `swift test --filter OpenAIDashboardNavigationDelegateTests`
- `swift test --filter CodexAccountScopedRefreshTests`
- `TZ=UTC swift test --filter MistralUsageParserTests`

### DIFF
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeOAuth/ClaudeOAuthCredentialModels.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeOAuth/ClaudeOAuthCredentialModels.swift
@@ -10,19 +10,22 @@ public struct ClaudeOAuthCredentials: Sendable {
     public let expiresAt: Date?
     public let scopes: [String]
     public let rateLimitTier: String?
+    public let subscriptionType: String?
 
     public init(
         accessToken: String,
         refreshToken: String?,
         expiresAt: Date?,
         scopes: [String],
-        rateLimitTier: String?)
+        rateLimitTier: String?,
+        subscriptionType: String? = nil)
     {
         self.accessToken = accessToken
         self.refreshToken = refreshToken
         self.expiresAt = expiresAt
         self.scopes = scopes
         self.rateLimitTier = rateLimitTier
+        self.subscriptionType = subscriptionType
     }
 
     public var isExpired: Bool {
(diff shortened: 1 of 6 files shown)


steipete:
Thanks @shixy96, this is fixed on main in 275c195b. I landed the same underlying fix with the regression coverage expanded to the actual menu dashboard URL, plus docs/changelog, and preserved co-author credit.

PR #840, opened 2026-05-04
Fix OpenAI dashboard analytics scraping

xiaoqianWX:
This is to amend #825  . It seems to still pick up the wrong usage graph sometimes or not pick up anything. After this change this issue seems to no longer appear with extensive testing. A test has been added.

Note: previously CI didn't pass and there seems to be some other issues, but now that should be fixed


ratulsarna:
@xiaoqianWX Thanks for the fixes!

PR #844, opened 2026-05-05
Fix battery drain from hung Codex RPC + 60 FPS animation loop

### LINKED ISSUE #842: Hung Codex JSON-RPC account/rateLimits/read leaves StatusItem animating at 60 FPS (battery drain path)
## Summary

`UsageFetcher`'s Codex JSON-RPC client has no per-request timeout. When
`codex app-server` accepts `initialize` but never replies to
`account/rateLimits/read`, `loadLatestUsage` hangs forever. Because the
hang produces no snapshot **and** no error, `StatusItemController`'s
animation gate `!hasData && !isStale` stays true and a 60 FPS
`DisplayLinkDriver` redraws the menu bar icons indefinitely on the main
thread — exactly the sustained UI work macOS flags as "Using Significant
Energy."

Related (closed): #269, #139. Related (open): #623, #678. The prior
fixes guarded the *fallback-only* animation path; this report identifies
a separate path into the same 60 FPS animation that none of those fixes
cover.

## Root cause chain (upstream `main`)

1. `CodexRPCClient.request(...)` awaits `readNextMessage()`, which is a
   bare `for await lineData in self.stdoutLineStream` — no per-request
   deadline. (`Sources/CodexBarCore/UsageFetcher.swift:523-571`)
2. `loadRPCUsage` calls `rpc.fetchRateLimits()` with no surrounding
   timeout. (`Sources/CodexBarCore/UsageFetcher.swift:626-659`)
3. `withFallback` only runs the TTY secondary when primary **throws** —
   a hang isn't a throw, so the TTY fallback never executes.
   (`Sources/CodexBarCore/UsageFetcher.swift:718-732`)
4. `UsageStore.isStale(provider:)` returns
   `self.errors[provider] != nil`. A hang records no error →
   `isStale = false`. (`Sources/CodexBar/UsageStore.swift:411-413`)
5. `shouldAnimate` returns `!hasData && !isStale` → `true`.
   (`Sources/CodexBar/StatusItemController+Animation.swift:663-682`)
6. `updateAnimationState` starts `DisplayLinkDriver` at 60 FPS, which
   calls `applyIcon` on every tick — main-thread NSImage redraws at
   60 Hz indefinitely.
   (`Sources/CodexBar/StatusItemController+Animation.swift:684-722`,
   `Sources/CodexBar/DisplayLink.swift:24-42`)

The codebase already documents this animation pattern as a battery
sink (Animation.swift:670-671: *"Animating the fallback causes
unnecessary CPU usage (battery drain). See #269, #139."*) — that guard
just doesn't generalize to the hung-primary case.

## Reproduction

Stub `codex` CLI that responds to `initialize` then hangs on
`account/rateLimits/read`. With no timeout, `loadLatestUsage` never
returns; menu bar enters animation state and stays there. macOS flags
the process for energy use after sustained operation.

(I have a Swift Testing repro along the lines of
`Tests/CodexBarTests/CodexUsageFetcherFallbackTests.swift` that hangs
`account/rateLimits/read` for 30s and asserts the fetch completes via
TTY fallback in <2s when a per-request RPC timeout is wired up. Happy
to PR if useful.)

## Suggested fix shape

1. Per-request RPC deadline in `CodexRPCClient.request(...)` —
   `withThrowingTaskGroup` racing the stdout read against
   `Task.sleep(for: .seconds(timeout))`, throwing on timeout so
   `withFallback` can route to TTY.
2. Treat the resulting timeout as an error in `UsageStore.errors[.codex]`
   so `isStale` flips true and `shouldAnimate` returns false even before
   the TTY fallback completes.
3. (Defense-in-depth) Cap `DisplayLinkDriver` start-up to a lower fps
   (e.g. 12-20) for menu bar status items where 60 Hz isn't perceptible
   anyway, and/or auto-stop the animation after a hard ceiling (e.g.
   30s) regardless of state.

## Environment

- Branch: upstream `main` at HEAD of this repo
- macOS: applies to both `CADisplayLink` (15+) and `CVDisplayLink` (14)
  paths in `DisplayLink.swift`

hyspacex:
Closes #842.

## Summary

Three connected changes that together stop the menu-bar from spending CPU forever when a Codex RPC call hangs:

1. **Per-method RPC timeouts** so `account/rateLimits/read` (and friends) can't await stdout indefinitely.
2. **Process termination on timeout** so the AsyncStream reader actually exits instead of leaking.
3. **Animation hardening** — drop loading FPS 60→30, add a 30s hard ceiling on continuous animation.

## Root cause (verified in #842)

When `codex app-server` accepted `initialize` but never replied to `account/rateLimits/read`:

1. `CodexRPCClient.request` (`Sources/CodexBarCore/UsageFetcher.swift`) awaited `for await lineData in self.stdoutLineStream` with no per-request deadline.
2. `loadRPCUsage` hung, so `withFallback` never saw a throw and the TTY secondary never ran.
3. `UsageStore.isStale(provider:)` returns `errors[provider] != nil`. A hang records no error → `isStale = false`.
4. `shouldAnimate` returned `!hasData && !isStale` → `true`, so `DisplayLinkDriver.start(fps: 60)` ran main-thread NSImage redraws indefinitely. macOS flagged this as "Using Significant Energy."

The existing fallback-only animation guard (`Animation.swift:670` — *"Animating the fallback causes unnecessary CPU usage (battery drain). See #269, #139."*) didn't cover the hung-primary case.

### `Sources/CodexBarCore/UsageFetcher.swift`
- `RPCWireError.timeout(method:)` added.
- `CodexRPCClient` and `UsageFetcher` take internal-only `initializeTimeoutSeconds` (default 8s), `requestTimeoutSeconds` (default 3s), `ttyTimeoutSeconds` (default 10s). Defaults match production behavior; tests inject shorter values.
- `request()` races the stdout read against `Task.sleep` via `withThrowingTaskGroup`. On timeout: `process.terminate()` (closes stdout, unblocks the AsyncStream reader, body task unwinds) → throw `RPCWireError.timeout` → propagates through `withFallback` to TTY.
- `loadTTYUsage` / `loadTTYCredits` get a defensive `runWithTTYTimeout` wrapper. `CodexStatusProbe.fetch()` already has its own internal `timeout`; the outer race is defense-in-depth and bounds injected test fetchers that bypass the probe.
- The terminate-on-timeout log line names the failing method so prod logs distinguish `initialize` vs `account/rateLimits/read` hangs.

### `Sources/CodexBar/StatusItemController+Animation.swift`
- `loadingAnimationFPS = 30.0` (was 60) — halves main-thread NSImage redraw cost. 12 FPS would be choppy; 30 stays smooth.
- `loadingAnimationPhaseIncrement = 2.7 / FPS` derived so visible animation speed is unchanged when FPS is tuned.
- `loadingAnimationMaxContinuousDuration: 30.0` — hard ceiling on continuous animation. Even if a future code path produces `!hasData && !isStale` indefinitely, the driver force-stops after 30s. Defense-in-depth on a bug class that has shipped twice (#269, #139, #842).
- Stop logic factored into `stopLoadingAnimation()` so the ceiling and the natural-stop path use the same teardown.

### `Sources/CodexBar/StatusItemController.swift`
- `var animationStartedAt: Date?` to track ceiling.

### `Tests/CodexBarTests/CodexUsageFetcherFallbackTests.swift`
4 new regression tests, extended in the existing file (no parallel suite):

| Test | What it locks in |
|---|---|
| Hung rate-limits read times out and falls back to TTY within budget | RPC timeout fires fast enough that TTY fallback completes in <2s |
| Repeated hung RPC requests stay bounded across calls | Process termination actually unblocks the reader (no leaked task / wedged pipe) |
| Hanging TTY fallback also times out within budget | Defensive TTY wrapper bounds custom fetchers; specifically asserts `RPCWireError.timeout` not just any throw |
| Recorded error marks provider stale and stops menu-bar animation | Second half of the chain — once an error is recorded, `shouldAnimate` returns false and `animationDriver` stays nil |

The first stub (`makeHungRateLimitsStubCodexCLI`) mirrors the production failure shape: responds to `initialize`, hangs forever on `account/rateLimits/read`, answers `account/read` and TTY normally.

## Decisions

- **Timeouts:** 8s for `initialize` (covers child-process spawn + cold init), 3s for everything else (read calls are sub-second on healthy networks). Single-budget would either be too tight for initialize or too loose for the actual bug path.
- **30 FPS:** The energy cost halves vs 60 FPS but stays perceptually smooth. 12 FPS was on the table but visibly choppy on continuous sweeps (Knight Rider).
- **30s hard ceiling:** Belt-and-suspenders. Max legitimate refresh budget is ~21s (8s init + 3s request + 10s TTY); 30s is well above that. Cheap insurance against future code paths bypassing the error-recording chain.
- **No new user-facing settings.** All timeouts are internal init parameters for tests only.

## Verification

- `swift build` — clean
- `swift test --filter CodexUsageFetcherFallbackTests` — 10/10 (6 existing + 4 new)
- `swift test --filter BatteryDrainDiagnosticTests` — 3/3
- Fast-path test runtimes well under their assertions (Test A: 0.41s vs 2s budget; Test C: 0.82s; Test D: 1.01s)

The macOS "Using Significant Energy" badge is a heuristic over time and not deterministically reproducible on demand. What's verified here is the underlying mechanism: the hung-RPC chain that produced sustained 60 FPS main-thread redraws.

## Out of scope

- WebKit/AuthKit dashboard refresh resilience (already addressed upstream)
- New "battery saver" user-facing toggle (this is a correctness fix, not a configurable behavior)

### DIFF
--- a/Sources/CodexBar/StatusItemController+Animation.swift
+++ b/Sources/CodexBar/StatusItemController+Animation.swift
@@ -6,6 +6,18 @@ extension StatusItemController {
     private static let loadingPercentEpsilon = 0.0001
     private static let blinkActiveTickInterval: Duration = .milliseconds(75)
     private static let blinkIdleFallbackInterval: Duration = .seconds(1)
+    /// Target frame rate for the menu-bar loading animation. Reduced from 60 FPS
+    /// to halve main-thread NSImage redraw cost while still feeling smooth (12 FPS
+    /// looks choppy). The phase increment below is derived from this so visible
+    /// animation speed stays constant if the value changes.
+    static let loadingAnimationFPS: Double = 30.0
+    /// Per-tick phase advance, derived so that animation speed remains 2.7
+    /// phase-units per second regardless of FPS. Originally `0.045` at 60 FPS.
+    static let loadingAnimationPhaseIncrement: Double = 2.7 / StatusItemController.loadingAnimationFPS
+    /// Hard ceiling on continuous animation duration. Guards against future
+    /// regressions of the `!hasData && !isStale` animation loop that pinned the
+    /// menu bar at full FPS until the user quit (#842).
+    private static let loadingAnimationMaxContinuousDuration: TimeInterval = 30.0
 
     func needsMenuBarIconAnimation() -> Bool {
         if self.shouldMergeIcons {
@@ -691,29 +703,44 @@ extension StatusItemController {
                     self.animationPattern = .knightRider
                 }
                 self.animationPhase = 0
+                self.animationStartedAt = Date()
                 let driver = DisplayLinkDriver(onTick: { [weak self] in
                     self?.updateAnimationFrame()
                 })
                 self.animationDriver = driver
-                driver.start(fps: 60)
+                driver.start(fps: Self.loadingAnimationFPS)
             } else if let forced = self.settings.debugLoadingPattern, forced != self.animationPattern {
                 self.animationPattern = forced
                 self.animationPhase = 0
             }
         } else {
-            self.animationDriver?.stop()
-            self.animationDriver = nil
-            self.animationPhase = 0
-            if self.shouldMergeIcons {
-                self.applyIcon(phase: nil)
-            } else {
-                UsageProvider.allCases.forEach { self.applyIcon(for: $0, phase: nil) }
-            }
+            self.stopLoadingAnimation()
+        }
+    }
+
+    private func stopLoadingAnimation() {
+        self.animationDriver?.stop()
+        self.animationDriver = nil
+        self.animationPhase = 0
+        self.animationStartedAt = nil
+        if self.shouldMergeIcons {
+            self.applyIcon(phase: nil)
+        } else {
+            UsageProvider.allCases.forEach { self.applyIcon(for: $0, phase: nil) }
         }
     }
 
     private func updateAnimationFrame() {
-        self.animationPhase += 0.045 // half-speed animation
+        // Hard ceiling guards against future regressions of the !hasData && !isStale
+        // animation loop that caused the menu bar to redraw at full FPS forever
+        // (see #842).
+        if let startedAt = self.animationStartedAt,
+           Date().timeIntervalSince(startedAt) > Self.loadingAnimationMaxContinuousDuration
+        {
+            self.stopLoadingAnimation()
+            return
+        }
+        self.animationPhase += Self.loadingAnimationPhaseIncrement
         if self.shouldMergeIcons {
             self.applyIcon(phase: self.animationPhase)
         } else {
--- a/Sources/CodexBar/StatusItemController.swift
+++ b/Sources/CodexBar/StatusItemController.swift
@@ -112,6 +112,10 @@ final class StatusItemController: NSObject, NSMenuDelegate, StatusItemControllin
     var animationDriver: DisplayLinkDriver?
     var animationPhase: Double = 0
     var animationPattern: LoadingPattern = .knightRider
+    /// Tracks when the loading animation last started. Used by the 30-second hard
+    /// ceiling in `updateAnimationFrame` to guard against future regressions of the
+    /// `!hasData && !isStale` animation loop. See #842.
+    var animationStartedAt: Date?
     private var lastConfigRevision: Int
     private var lastProviderOrder: [UsageProvider]
     private var lastMergeIcons: Bool
--- a/Sources/CodexBarCore/UsageFetcher.swift
+++ b/Sources/CodexBarCore/UsageFetcher.swift
@@ -366,7 +366,7 @@ private struct RPCRateLimitsErrorBody: Decodable {
     }
 }
 
-private enum RPCWireError: Error, LocalizedError {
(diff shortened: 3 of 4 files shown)


steipete:
Thanks for the report and patch, Harry. The root bug was real, but this PR conflicted with the newer shared Codex CLI snapshot path on main, so I landed an adapted fix in 021f7393 instead.

What landed:
- Codex RPC reads now have bounded per-request timeouts and terminate a hung `codex app-server` process on timeout: `Sources/CodexBarCore/UsageFetcher.swift`.
- The loading animation now runs at 30 FPS and has a 30s continuous ceiling: `Sources/CodexBar/StatusItemController+Animation.swift`.
- Regression coverage covers hung `account/rateLimits/read` calls and the “provider errored, do not animate forever” path.

Verification on main: full local `swift test`, `pnpm check`, packaged app relaunch via `./Scripts/compile_and_run.sh`, live bundled `CodexBarCLI usage --provider codex --source cli`, and GitHub CI 25360102746 are green.

I preserved contributor credit via `Co-authored-by` on the landed commit and thanked you in the changelog.

PR #847, opened 2026-05-05
Fix Ollama Safari cookie fallback

pdurlej:
## Summary
- make Ollama auto cookie import fall back from Chrome to Safari/other available browser sessions
- surface Safari/Full Disk Access failures instead of the misleading generic no-session-cookie error
- clarify the no-session-cookie/settings copy and update Ollama docs for Safari-only sessions

## Bug
Ollama provider auto mode claimed to import browser cookies, but the fetch path used the default preferred browser list (`[.chrome]`) without enabling fallback. Users logged in only through Safari saw `No Ollama session cookie found` even though a valid Ollama Cloud browser session existed.

There is a second Safari-specific failure mode: CodexBar can find Safari's cookie store but macOS can deny app access to `~/Library/Containers/com.apple.Safari/Data/Library/Cookies/Cookies.binarycookies`. Previously that also collapsed to `No Ollama session cookie found`, hiding the real fix: grant CodexBar Full Disk Access or use a manual Cookie header.

## Fix
The fetcher now keeps Chrome-first behavior but passes `allowFallbackBrowsers: true` for automatic cookie resolution. Manual cookie mode is unchanged. If fallback reaches a browser cookie store but receives an access-denied error, Ollama now returns a user-facing Full Disk Access message instead of the generic no-session-cookie error.

## Verification
- `git diff --check` passed for changed files
- `CodexBarCLI usage --provider ollama --source web --format json --pretty -v` succeeded locally after the fallback and returned Ollama account/usage JSON
- local app log reproduced the Safari permission failure: Safari cookie file exists, but CodexBar.app cannot read it without Full Disk Access
- `swift test --filter OllamaUsageFetcherTests` is currently blocked locally before reaching project code by `KeyboardShortcuts` failing to build: missing `PreviewsMacros.SwiftUIView` for SwiftUI `#Preview` macros under the Command Line Tools-only toolchain
- `./Scripts/compile_and_run.sh` is also blocked locally by Xcode-only tooling unless local build workarounds are applied; the machine has Xcode installed but its license has not been accepted for CLI use

### DIFF
--- a/Sources/CodexBar/Providers/Ollama/OllamaProviderImplementation.swift
+++ b/Sources/CodexBar/Providers/Ollama/OllamaProviderImplementation.swift
@@ -48,7 +48,7 @@ struct OllamaProviderImplementation: ProviderImplementation {
             ProviderCookieSourceUI.subtitle(
                 source: context.settings.ollamaCookieSource,
                 keychainDisabled: context.settings.debugDisableKeychainAccess,
-                auto: "Automatic imports browser cookies.",
+                auto: "Automatic imports Chrome first, then falls back to Safari and other browser sessions.",
                 manual: "Paste a Cookie header or cURL capture from Ollama settings.",
                 off: "Ollama cookies are disabled.")
         }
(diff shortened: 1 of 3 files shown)


pdurlej:
Closing this approach: requiring Full Disk Access for Safari-only Ollama sessions is not an acceptable UX/trust tradeoff for a menu bar monitor. A better fix should prefer explicit manual cookie import or a lower-permission auth path instead.

PR #855, opened 2026-05-07
Make OpenRouter color-bar preview use the standard resolver path

willytop8:
The color-bar preview was reading `usedPercent` directly instead of going through `IconRemainingResolver.resolvedPercents`. Key-quota state (no limit configured, key data unavailable) had no effect — the preview always showed a non-nil percent even when the live icon renders no color bar.

Routes the preview through the same resolver the live icon uses:
- No key limit → `nil` primary, no color bar
- Key data unavailable → `nil` primary, no color bar
- Key limit set → key-usage percent drives the bar as expected

Adds tests for all three cases, each asserting both the resolved percent and the menu-bar display text.

```
swift test --filter StatusItemControllerMenuTests
```

### DIFF
--- a/Sources/CodexBar/StatusItemController+Animation.swift
+++ b/Sources/CodexBar/StatusItemController+Animation.swift
@@ -223,7 +223,6 @@ extension StatusItemController {
         return false
     }
 
-    // swiftlint:disable function_body_length
     @discardableResult
     func applyIcon(phase: Double?) -> Bool {
         guard let button = self.statusItem.button else { return false }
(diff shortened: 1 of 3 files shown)


steipete:
Thanks for the focused follow-up. This is now superseded by #854, which landed on main and removed the OpenRouter brand-logo fallback from the live menu-bar icon path with pixel-level coverage in . The remaining diff here is based on the pre-#854 fallback helper and now conflicts with current main, so I’m closing this one to keep the queue tidy.

steipete:
Correction to my previous comment: the landed coverage is in StatusItemAnimationTests. The close reason is unchanged: #854 removed the fallback path on main, and this branch now conflicts against that landed fix.

PR #857, opened 2026-05-07
Add command code support

sixhobbits:
Might be slop because I'm not a macos dev and I don't have XCode so my agent had to do a bunch of workarounds to build this locally and test it

But it works for me

CleanShot 2026-05-07 at 12 22 12@2x

image

image


steipete (approved):
Thanks @sixhobbits. I rebased this onto current main, kept the newer Crof/Venice provider entries, fixed the Command Code bare-token cookie default, updated provider-order coverage, and verified local/CI checks.

PR #863, opened 2026-05-08
Feat: Add models.dev pricing metadata pipeline

iam-brain:
# Model pricing metadata

CodexBar has an additive models.dev pricing pipeline for future cost lookup work. Existing hardcoded pricing remains unchanged for now.

## Source and cache

- Source API: `https://models.dev/api.json`
- No API key is required.
- Local cache: `~/Library/Caches/CodexBar/model-pricing/models-dev-v1.json`
- TTL: 24 hours

The pipeline lets future scanner code read the last valid cache synchronously with `ModelsDevPricingPipeline.lookup` and refresh stale metadata separately with `ModelsDevPricingPipeline.refreshIfNeeded`. If a refresh fails, the last valid cache remains usable.

## Lookup rules

Pricing is scoped by provider id and model id. This prevents two providers with the same model id or display name from sharing pricing accidentally.

Planned local source mapping:

- Codex/OpenAI logs: models.dev provider id `openai`
- Claude logs: models.dev provider id `anthropic`
- Vertex AI Claude logs: models.dev provider id `google-vertex-anthropic`

The first integration PR only adds the parser, client, cache, provider-scoped lookup, and tests. It does not route live cost calculations through models.dev yet.

## Units

models.dev publishes costs as USD per 1M tokens. CodexBar converts those to USD per token in the metadata layer:

```text
perToken = modelsDevCost / 1_000_000
```

When models.dev includes `cost.context_over_200k`, CodexBar parses those values as the above-200k-token pricing lane and converts them with the same per-1M-token rule.

## Summary
- add an additive models.dev pricing metadata parser/client/cache pipeline
- support provider-scoped model lookup, cache TTL handling, and per-1M-token to per-token conversion
- document the cache path and future lookup rules
- fix Linux CLI builds by importing FoundationNetworking where Swift exposes URLSession networking types separately

## Reasoning
- reduce reliance on user-curated and logged model pricing, which is brittle as providers add aliases, change prices, and introduce threshold pricing
- transition toward Anomaly's models.dev online catalog because it already does the cataloging work, tracks provider-scoped model metadata, and exposes a free API without requiring an API key
- keep this additive first so CodexBar can validate the catalog/cache pipeline before switching existing cost calculations away from the fallback pricing tables

## Scope
- does not switch existing Codex/Claude cost calculations to models.dev
- does not migrate or change existing fallback pricing tables
- no API key required for models.dev

## References
- models.dev catalog API: `https://models.dev/api.json`

## Validation
- `swift test --filter ModelsDevPricingTests`
- `pnpm check`
- Latest PR CI after rebase: Linux CLI x64 and arm64 checks pass.
- Latest PR CI after rebase: macOS `lint-build-test` currently fails in existing Codex RPC fallback/baseline tests with `timeout(method: "initialize")`; the models.dev pricing suite passes locally.
- `swift test` was also run earlier; it still fails on the pre-existing `MistralUsageParserTests.parses dates from response` expectation (expected month 11, got 10).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.25 — Unreleased
 
 ### Providers & Usage
+- Cost history: add an additive models.dev pricing metadata parser/cache pipeline for future provider-scoped cost lookups (#863). Thanks @iam-brain!
 - Venice: add API-key balance provider support with DIEM/USD balance display and token-account CLI wiring (#865). Thanks @clawSean!
 - Factory/Droid: add token-rate-limit billing windows, Core fallback buckets, and extra usage balance display (#878). Thanks @dantemoon1!
 - Usage pace: compute pace for any explicit reset window instead of a provider allowlist (#875). Thanks @ViperThanks!
--- a/Sources/CodexBarCore/Vendored/CostUsage/ModelsDevPricing.swift
+++ b/Sources/CodexBarCore/Vendored/CostUsage/ModelsDevPricing.swift
@@ -1,4 +1,7 @@
 import Foundation
+#if canImport(FoundationNetworking)
+import FoundationNetworking
+#endif
 
 struct ModelsDevPricingInfo: Codable, Equatable, Sendable {
     var providerID: String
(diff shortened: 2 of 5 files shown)


iam-brain:
Current CI build failure seems to be caused by increased pressure from the new tests introduced for the feature.

steipete (approved):
Reviewed model pricing pipeline changes and local validation; CI is green. Thanks @iam-brain!

PR #881, opened 2026-05-10
Fix models.dev pricing refresh continuity

iam-brain:
## Summary
Follow-up to #863 after the late Codex review comment on the merged models.dev pricing pipeline.

- accept models.dev catalog refreshes when upstream renames model map keys but keeps the same model identity
- keep stale pricing when a fetched matching model is no longer priceable
- compare refresh continuity through the same provider-scoped lookup path used at runtime
- preserve stale pricing when the refreshed catalog only has a different pinned @YYYYMMDD snapshot and no base/default candidate
- ignore unpriceable models already present in the old cache when deciding whether a refresh is safe

## Verification
- swift test --filter ModelsDevPricingTests
- pnpm check
- ./Scripts/compile_and_run.sh

PR #883, opened 2026-05-10
Fix: Align models.dev refresh guard with lookup

iam-brain:
## Summary
Follow-up to #881 / #863 for the pinned snapshot continuity edge case.

- route models.dev refresh continuity through the same provider-scoped `pricing(modelID:)` lookup used at runtime
- keep exact/base/default fallback behavior, but avoid treating different pinned `@YYYYMMDD` snapshots as covering one another when no base/default candidate exists
- add a regression test for `claude-sonnet-4@20250101` vs `claude-sonnet-4@20250201`

## Verification
- swift test --filter ModelsDevPricingTests
- pnpm check


steipete (approved):
Looks good. The continuity guard now uses the same pricing lookup path as runtime, with a pinned-snapshot regression test. Local focused test and make check passed; CI green.

PR #884, opened 2026-05-10
Integrate models.dev pricing flow

iam-brain:
## Summary
- Prefer cached models.dev pricing for Codex and Claude cost calculations before bundled fallback tables.
- Thread provider-scoped models.dev catalog/cache context through Codex, Claude, and Pi session cost scanners.
- Recompute report costs from the current catalog before falling back to cached packed cost rows, with focused regression coverage.

## Reasoning
- PRs #863 and #881 added the models.dev metadata pipeline; #883 aligned refresh continuity. This moves the Codex and Claude pricing layer into that flow while preserving bundled fallback behavior.
- Provider-scoped lookup keeps same-name or overlapping model ids from crossing provider boundaries.

## Scope
- Codex and Claude pricing layer.
- Cost usage scanner report construction.
- Pi session merged cost reports.
- Focused pricing tests for models.dev precedence, threshold pricing, research-preview override, and isolated bundled fallback behavior.

## Screenshots
N/A

## GIFs
N/A

## References
- Builds on #863, #881, and #883.

## Validation
- `swift test --filter CostUsagePricingTests --filter PiSessionCostScannerTests --filter CostUsageScannerTests`
- `make check`
- `./Scripts/compile_and_run.sh`

## Audit
- GPT-5.5 high subagent audit completed. Initial findings were fixed, and the final pass found no material issues.

PR #886, opened 2026-05-10
Preserve tiered pricing boundaries

iam-brain:
## Summary
- Preserve Claude per-request and Pi per-message cost boundaries when rendering daily reports for threshold-priced models.
- Track priced-sample completeness so cached per-sample costs are only trusted when every sample in a model/day bucket was priced.
- Fall back conservatively for legacy or mixed partial caches instead of treating partial cached cost totals as authoritative.

## Reasoning
- Follow-up to #884: automated review correctly caught that re-rating aggregate day/model totals can cross Claude long-context thresholds that individual requests never crossed.
- Threshold/tiered pricing is not linear across aggregate totals, so report building must prefer complete per-sample cost sums and avoid partial cached sums.

## Scope
- Claude cost-usage report construction.
- Pi session cost report construction/cache shape.
- Regression tests for multiple below-threshold samples whose aggregate crosses the threshold.

## Screenshots
N/A

## GIFs
N/A

## References
- Follows #884.
- Addresses Codex bot P1 feedback on preserving threshold pricing boundaries.

## Validation
- `swift test --filter CostUsageScannerTests --filter PiSessionCostScannerTests`
- `make check`

## Audit
- GPT-5.5 high subagent audit completed.
- Initial P2 finding about mixed priced/unpriced caches fixed.
- Follow-up P2 finding about legacy cache completeness fixed.
- Final pass found no remaining P1/P2 findings.

### DIFF
--- a/Sources/CodexBarCore/PiSessionCostCache.swift
+++ b/Sources/CodexBarCore/PiSessionCostCache.swift
@@ -76,6 +76,7 @@ struct PiPackedUsage: Codable, Equatable {
     var totalTokens: Int = 0
     var costNanos: Int64 = 0
     var costSampleCount: Int = 0
+    var usageSampleCount: Int?
 
     var isZero: Bool {
         self.inputTokens == 0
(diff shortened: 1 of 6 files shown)


iam-brain:
@codex review

PR #892, opened 2026-05-10
feat(http): add shared provider client seam

serezha93:
## Summary
Add a shared ProviderHTTPClient seam and route a few network call sites through it:
- Copilot GitHub identity fetch
- Factory status probe
- Google Workspace status fetch

## Why
This is a narrow first step toward the broader proxy work. It keeps the PR small and reviewable while proving the shared HTTP path with tests.

## Testing
- swift test --filter ProviderHTTPClientTests
- swift test --filter CopilotUsageFetcherTests
- swift test --filter GoogleWorkspaceStatusNetworkTests
- swift test --filter FactoryStatusProbeFetchTests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.27.0 — Unreleased
 
 ### Added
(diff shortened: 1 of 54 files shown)


serezha93:
@steipete see old PR #697

PR #946, opened 2026-05-14
Fix account switcher flicker during multi-account menu refresh

kubahasek:
Fixes an issue where the Menubar app flickers when there is more than one account added to a provider


steipete:
Thanks @kubahasek, this was the right bug class. I landed the maintainer-side fix on `main` in https://github.com/steipete/CodexBar/commit/aebe1c91 with your commit credited via `Co-authored-by`.

I changed the shape a bit: instead of only disabling animation around rebuilds, switcher-triggered rebuilds are deferred until after the account-selection action returns, and token-account switcher state is now tracked like the Codex switcher so smart menu refreshes do not drop into unnecessary full rebuilds.

Verified with focused switcher regressions, full `swift test`, and `make check`. Closing this PR since the fix is now on `main`.

PR #947, opened 2026-05-14
[codex] Fix update menu install action

velvet-shark:
## What was happening

When CodexBar had already downloaded an app update, the menu showed the restart/apply-update item as expected. Clicking that item the first time only closed the menu. The app did not restart, the update details did not appear, and reopening the menu showed the same restart/apply-update item again. Clicking the same item a second time was what finally brought up Sparkle’s update information prompt.

That made the update flow feel broken: the first click looked like it had been ignored, and users had to discover that repeating the same action would eventually continue the update.

## Why it happened

CodexBar was marking the update as ready when Sparkle reported that the update download had finished. That was too early for the menu action CodexBar exposed. The app was not retaining Sparkle’s immediate-installation callback, so the menu item’s action fell back to starting another update check with `checkForUpdates(nil)`.

As a result, the first click could dismiss the menu without invoking Sparkle’s prepared install path. A later click worked only because the repeated check flow caused Sparkle to present its update UI.

## What changed

- Capture Sparkle’s immediate installation block when Sparkle says the downloaded update can be installed on quit.
- Treat that delegate callback, rather than the earlier download callback, as the moment the update is actually ready for the restart/apply menu action.
- Route the menu item through the captured install callback so the first click performs the prepared update action.
- Clear the captured callback when the update flow is canceled, fails, aborts, or completes.
- Add regression coverage proving the menu action installs the prepared update instead of starting another update check.

This PR also includes two small local-verification fixes needed to run the full checks on macOS with the current toolchain: exclude the Linux-only test target on macOS, and make lint find SourceKit from the active Xcode/Command Line Tools install.

## Validation

- `swift test --filter StatusItemControllerMenuTests`
- `swift test`
- `make check`

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -60,15 +60,18 @@ let package = Package(
                 swiftSettings: [
                     .enableUpcomingFeature("StrictConcurrency"),
                 ]),
-            .testTarget(
-                name: "CodexBarLinuxTests",
-                dependencies: ["CodexBarCore", "CodexBarCLI"],
-                path: "TestsLinux",
-                swiftSettings: [
-                    .enableUpcomingFeature("StrictConcurrency"),
-                    .enableExperimentalFeature("SwiftTesting"),
-                ]),
         ]
+
+        #if os(Linux)
+        targets.append(.testTarget(
+            name: "CodexBarLinuxTests",
+            dependencies: ["CodexBarCore", "CodexBarCLI"],
+            path: "TestsLinux",
+            swiftSettings: [
+                .enableUpcomingFeature("StrictConcurrency"),
+                .enableExperimentalFeature("SwiftTesting"),
+            ]))
+        #endif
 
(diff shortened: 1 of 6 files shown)


steipete:
Thanks! I landed the Sparkle/update-menu part in https://github.com/steipete/CodexBar/commit/837cdb971ab8e36107c2756b75d6cca55ac333d8, keeping it scoped to the prepared install callback and leaving the unrelated local tooling/style edits out.

Proof:
- `swift test --filter StatusItemControllerMenuTests --filter StatusMenuTokenAccountSwitcherTests --filter FactoryStatusProbeFetchTests`
- `make check`
- `codex-review clean: no accepted/actionable findings reported`

velvet-shark:
Great. This issue annoyed me one too many times so I decided to make a PR. Glad it's going to be fixed in the next release.

PR #969, opened 2026-05-16
Block OpenCode Go cookie leakage on redirects

pavbar:
## Summary
- add a guarded URLSession for OpenCode Go usage requests
- block cross-host and non-HTTPS redirects so cookies are not forwarded away from opencode.ai
- keep same-host HTTPS redirects working and preserve injected test sessions

## Verification
- swift test --filter OpenCodeGoUsageFetcherErrorTests
- fresh review passes against the PR diff found no remaining OpenCode Go issues

Note: branch name intentionally avoids slashes so the fork's current push CI branch pattern can match it.


steipete:
Thanks @pavbar. I landed this via #979 because this token could not push the small regression-test/changelog fixups back to the contributor fork.

The landed version keeps the PR behavior, adds a direct regression test for same-host HTTPS redirect gating, and credits you in the changelog.

Landed: #979 / 53d675f9.

PR #976, opened 2026-05-16
fix(grok): treat expired credentials as missing in fetch

taibaran:
## Summary

Follow-up to the Codex bot's P1 review on #965 (commit `cbd30a4e`). Grok session tokens expire after ~7 days, so this hits real users on a predictable cadence.

## The bug

`GrokStatusProbe.fetch` decides whether to surface the RPC auth error using:

```swift
if billing == nil, credentials == nil {
    throw rpcError ?? GrokRPCError.notAuthenticated
}
```

That check is permissive: an `auth.json` whose `expires_at` is already in the past still loads successfully (`isExpired` is computed but never consulted), so `credentials != nil` and the auth error is silently swallowed. The UI then renders a "successful" snapshot with stale email/team and no `grok login` hint while billing keeps 401-ing in the background.

## The fix

Treat expired records the same as missing credentials when deciding whether to swallow the RPC error, **and** drop them from the rendered snapshot so the UI doesn't surface stale identity for an inactive session:

```swift
let activeCredentials = credentials.flatMap { $0.isExpired ? nil : $0 }
if billing == nil, activeCredentials == nil {
    throw rpcError ?? GrokRPCError.notAuthenticated
}
```

Behaviour matrix after this change:

| `expires_at` | Old behaviour | New behaviour |
|---|---|---|
| missing | rendered as fresh | rendered as fresh (unchanged — pre-`expires_at` shape stays supported) |
| in the future | rendered as fresh | rendered as fresh (unchanged) |
| in the past | rendered with stale data, RPC error swallowed | RPC error raised, snapshot omits stale identity |

## Tests

Adds `isExpired reflects past expires_at` covering past, future, and missing `expires_at`. All 11 Grok tests pass under Xcode 26.5.

## Test plan

- [ ] `make test` passes
- [ ] With a freshly-logged-in `grok login` session, Grok shows identity as before
- [ ] After `rm ~/.grok/auth.json` (or letting it expire), the provider surfaces the `Not authenticated to Grok. Run \`grok login\`.` hint instead of rendering empty

### DIFF
--- a/Sources/CodexBarCore/Providers/Grok/GrokStatusProbe.swift
+++ b/Sources/CodexBarCore/Providers/Grok/GrokStatusProbe.swift
@@ -100,18 +100,24 @@ public struct GrokStatusProbe: Sendable {
         let localSummary = GrokLocalSessionScanner.summarize(env: env)
         let cliVersion = Self.detectVersion(env: env)
 
+        // Grok session tokens expire after ~7 days. An expired record on disk
+        // must be treated like a missing credential when deciding whether to
+        // mask the RPC auth error: otherwise we render a snapshot with stale
+        // identity and no `grok login` hint while billing silently 401s.
+        let activeCredentials = credentials.flatMap { $0.isExpired ? nil : $0 }
+
         // `localSummary` is *not* currently projected into a visible RateWindow or
         // identity field, so a stale `~/.grok/sessions/` directory must not
         // suppress the auth-required hint. Only swallow the RPC error when we
-        // actually have something rendrable for the user — credentials or a
-        // billing response.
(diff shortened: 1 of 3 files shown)


taibaran:
CI was failing on `SettingsStoreTests / "provider order persists and appends new providers"` because the hardcoded expected list in `SettingsStoreTests.swift:1161` was never updated when grok was added in #965 — my Grok-only `swift test --filter Grok` skipped this file so the regression became visible only on the next post-merge CI run.

Pushed `64585525` to append `.grok` to the fixture. Full suite (`swift test`, all 2565 tests) passes locally under Xcode 26.5.

PR #996, opened 2026-05-16
fix(cli): validate serve Host headers

### LINKED ISSUE #995: Harden codexbar serve against non-local Host headers
## Summary

The local `codexbar serve` HTTP server binds to `127.0.0.1`, but request handling did not validate the HTTP `Host` header before serving `/usage` and `/cost`.

A browser or local client using a non-local Host value should not be able to read local usage/cost/account metadata from the server. This is a common hardening control for localhost HTTP services, especially against DNS-rebinding-style access patterns.

## Impact

`/usage` and `/cost` can include provider usage, cost, plan, and account metadata. These routes should only be reachable through expected loopback hosts.

## Proposed fix

- Parse request headers in the local HTTP server.
- Reject duplicate `Host` headers.
- Allow only loopback Host values: `127.0.0.1`, `localhost`, `localhost.`, and `[::1]`, with optional numeric ports.
- Apply the guard before route dispatch, cache lookup, or provider fetch work.
- Add unit tests and smoke coverage for hostile/missing/duplicate Host values.

## Validation

- `swift test --filter CLIServeRouterTests`: passed
- `swift test`: passed
- `make check`: passed
- Live smoke: localhost Host succeeds, hostile Host returns 403, duplicate Host returns 400

rohitjavvadi:
## Summary

Fixes #995.

This hardens `codexbar serve` by validating the HTTP `Host` header before any route dispatch, cache lookup, or provider fetch work runs.

Changes:
- parse request headers in the local HTTP server
- reject duplicate `Host` headers at parse time
- allow only loopback hosts: `127.0.0.1`, `localhost`, `localhost.`, and `[::1]`, with optional numeric ports
- return 403 for missing, malformed, or non-loopback Host values
- add regression coverage for hostile, missing, duplicate, and loopback Host cases

## Validation

- `swift test --filter CLIServeRouterTests`
- `swift test`
- `make check`
- live smoke: localhost Host succeeds, hostile Host returns 403, duplicate Host returns 400

## Review notes

- Claude review identified the duplicate-Host last-wins edge case in the first patch; this PR rejects duplicate Host headers.
- Two independent review agents found no actionable issues after the duplicate-Host fix.
- A final Claude retry hit a temporary Claude API 500 twice, so it could not complete; the earlier actionable Claude finding was addressed and verified.

### DIFF
--- a/Sources/CodexBarCLI/CLILocalHTTPServer.swift
+++ b/Sources/CodexBarCLI/CLILocalHTTPServer.swift
@@ -12,6 +12,7 @@ struct CLILocalHTTPRequest {
     let target: String
     let path: String
     let queryItems: [String: String]
+    let headers: [String: String]
 
     static func parse(_ data: Data) -> CLILocalHTTPRequest? {
         guard let raw = String(data: data, encoding: .utf8),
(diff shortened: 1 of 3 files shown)


steipete:
Thanks @rohitjavvadi — this is fixed on `main` in e5eb944d.

I landed the same hardening path with a couple of final tweaks:
- parse headers before route/cache/provider work
- reject missing/duplicate Host with 400
- reject non-loopback Host with 403
- allow localhost, localhost., 127.0.0.1, and [::1] with numeric ports
- fail closed on incomplete oversized headers

Validation on the landed commit:
- `swift test --filter CLIServeRouterTests`
- `make check`
- live smoke: localhost Host -> 200, hostile Host -> 403, duplicate Host via raw socket -> 400
- Codex review clean

Closing this PR as superseded by the landed fix. Appreciate the report and patch.

Issue #988, opened 2026-05-16
v0.26.1: Menu bar icon does not appear on macOS 15.7.4 Sequoia


yanlin1026:
## Summary
  CodexBar 0.26.1 launches and runs normally, but its menu bar icon never appears in the menu
  bar. The symptom matches #861 (v0.24) and #890 (v0.25), but I'm on macOS Sequoia (not
  Tahoe), so PR #945's allow-list fix doesn't apply.

  ## Environment
  - macOS: 15.7.4 Sequoia (Build 24G517)
  - Hardware: MacBook Pro 16" (Liquid Retina XDR Display, 3456 x 2234)
  - CodexBar: 0.26.1 (build 63)
  - Git commit: f642907c
  - Build timestamp: 2026-05-15T14:22:08Z

  ## Steps to reproduce
  1. Download and install CodexBar 0.26.1.
  2. Launch the app.

  ## Expected
  A CodexBar icon appears in the menu bar.

  ## Actual
  No icon appears. Menu bar is not crowded — there's plenty of room and no notch overflow.

  ## Diagnostics

  **1. Process is running:**
  yanlin   5478   /Applications/CodexBar.app/Contents/MacOS/CodexBar

  **2. `Info.plist` is correct** (`LSUIElement = 1`).

  **3. Preferences show the status item is "visible" with a fixed position:**
  "NSStatusItem Visible Item-0" = 1;
  "NSStatusItem Preferred Position Item-0" = 892;

  **4. Logs show no NSStatusItem-related activity.** Only CoreFoundation / CFPasteboard /
  HIToolbox calls — no `status`, `menu`, `item`, or `visible` keywords appear in `log show
  --predicate 'process == "CodexBar"'`.

  **5. Sequoia has no "Allow in Menu Bar" allow-list** (verified in System Settings → Control
  Center — no third-party app entries), so PR #945's root cause does not apply here.

  ## Things I already tried (none worked)
  - `killall CodexBar` and relaunch
  - `defaults delete com.steipete.codexbar "NSStatusItem Preferred Position Item-0"`
  - `defaults delete com.steipete.codexbar "NSStatusItem Visible Item-0"`
  - `killall SystemUIServer`
  - ⌘-drag across the menu bar to find an invisible item (none found)

  ## Related issues
  - #861 (v0.24, Sequoia) — same symptom
  - #890 (v0.25) — same symptom, Tahoe, fixed by #945
  - #915 — related disappearance

steipete:
[...]

PR #998, opened 2026-05-16
[codex] Recover status item after display changes

### LINKED ISSUE #997: Menu bar icon does not recover after unplugging the external display that hosts it
## Summary

CodexBar can leave its menu bar icon unavailable after the external display that currently hosts the status item is unplugged.

This looks distinct from the existing startup recovery fixes. The app process keeps running, but the `NSStatusItem` appears to remain tied to the removed display's menu bar instead of being rematerialized on the remaining built-in display.

## Steps to Reproduce

1. Connect an external display to a Mac.
2. Launch CodexBar.
3. Have the CodexBar menu bar item visible on the external display's menu bar.
4. Unplug the external display while CodexBar is still running.

## Actual Behavior

The CodexBar process keeps running, but the menu bar icon does not appear on the built-in display. In my local reproduction, relaunching CodexBar also did not immediately make the icon visible on the built-in menu bar.

## Expected Behavior

When macOS screen parameters change and the display hosting the menu bar item disappears, CodexBar should recover the status item on one of the currently active screens, ideally after a short debounce to let AppKit settle.

## Related Work

- #988 / `a75d276f` adds startup recovery for a missing status item.
- #890 / #945 handles the Tahoe "Allow in Menu Bar" setting guidance.
- #805 / #849 restored the icon for the earlier macOS 26.4 visibility issue.

Those fixes are valuable, but I do not see runtime display topology changes being handled today. `StatusItemController` schedules a startup visibility check, but it does not currently observe `NSApplication.didChangeScreenParametersNotification`.

## Root Cause Hypothesis

`NSStatusItem` can report as present while its backing button/window is associated with a screen that is no longer part of `NSScreen.screens`. Because CodexBar only checks visibility at startup, a hot-unplug event can leave the app running with an inaccessible status item.

## Proposed Fix

Observe `NSApplication.didChangeScreenParametersNotification`, debounce the recovery check, and recreate status items when:

- the screen count decreases while any status item is expected to be visible, or
- a visible status item snapshot is blocked/missing/detached from the current `NSScreen.screens`.

I have a local patch for this and will open a PR.

Llldmiao:
Fixes #997.

## Summary

- Detect status items whose backing window is attached to a detached/non-current screen.
- Observe `NSApplication.didChangeScreenParametersNotification` and debounce display-change recovery.
- Recreate status items when a display is removed while a visible item exists, or when a visible item snapshot is blocked.
- Add focused coverage for detached-screen snapshots and display-removal recovery decisions.

## Root Cause

The startup recovery added in `a75d276f` catches status items that fail to materialize shortly after launch, but CodexBar did not observe runtime display topology changes. If the external display hosting the `NSStatusItem` is unplugged, AppKit can leave the app process alive while the status item is still associated with a removed screen/menu bar.

## Validation

- `git diff --check` passes.
- `swift test --filter MenuBarVisibilityWatcherTests` passes with Xcode 26.5 / Apple Swift 6.3.2.

## Notes

This PR intentionally targets the runtime hot-unplug case. It is complementary to the startup recovery in #988, the Tahoe allow-list guidance in #945, and the earlier macOS 26.4 visibility fix in #849.

### DIFF
--- a/Sources/CodexBar/MenuBarVisibilityWatcher.swift
+++ b/Sources/CodexBar/MenuBarVisibilityWatcher.swift
@@ -6,13 +6,31 @@ struct StatusItemVisibilitySnapshot: Equatable {
     let hasButton: Bool
     let hasWindow: Bool
     let hasScreen: Bool
+    let isOnCurrentScreen: Bool
     let buttonWidth: CGFloat
+
+    init(
+        isVisible: Bool,
+        hasButton: Bool,
+        hasWindow: Bool,
+        hasScreen: Bool,
+        isOnCurrentScreen: Bool = true,
+        buttonWidth: CGFloat)
+    {
+        self.isVisible = isVisible
+        self.hasButton = hasButton
+        self.hasWindow = hasWindow
+        self.hasScreen = hasScreen
+        self.isOnCurrentScreen = isOnCurrentScreen
+        self.buttonWidth = buttonWidth
+    }
 }
 
 extension StatusItemVisibilitySnapshot: CustomStringConvertible {
     var description: String {
         "visible=\(self.isVisible),button=\(self.hasButton),window=\(self.hasWindow),"
-            + "screen=\(self.hasScreen),width=\(String(format: "%.1f", Double(self.buttonWidth)))"
+            + "screen=\(self.hasScreen),currentScreen=\(self.isOnCurrentScreen),"
+            + "width=\(String(format: "%.1f", Double(self.buttonWidth)))"
     }
 }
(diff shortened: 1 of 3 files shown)


steipete:
Thanks for the focused fix. I landed this on `main` with a small follow-up for the debounce cancellation path so stale display-change checks cannot run after a newer notification cancels them.

Landed as fcbd46f3.

Verified with:
- `swift test --filter MenuBarVisibilityWatcherTests`
- `make check`
- `codex-review --mode auto`

PR #1000, opened 2026-05-16
[codex] Fix OpenAI costs parsing

### LINKED ISSUE #999: OpenAI Admin API usage fails when costs amount value is returned as a string
## Summary

OpenAI Admin API usage currently fails when parsing the organization costs response if `amount.value` is returned as a JSON string instead of a number.

## Reproduction

1. Configure an OpenAI Admin API key for the OpenAI provider.
2. Refresh OpenAI usage.
3. The costs endpoint returns HTTP 200, but CodexBar reports:

```text
Failed to parse OpenAI API usage costs: The data couldn’t be read because it isn’t in the correct format.
```

## Observed API shape

A live `GET /v1/organization/costs?...&group_by=line_item` response included:

```json
{
  "amount": {
    "value": "12.50",
    "currency": "usd"
  },
  "line_item": "Text tokens"
}
```

The parser currently expects `amount.value` to decode as `Double`, so the whole provider refresh fails.

## Expected behavior

OpenAI usage should accept both numeric and numeric-string cost values and continue rendering Admin API usage.

SergeyLavrentev:
## Summary

- Decode OpenAI organization cost `amount.value` when it is returned as a numeric string.
- Keep existing numeric decoding support for documented numeric responses.
- Add a regression fixture covering the string value shape observed from the live API.

Fixes #999

## Root cause

The OpenAI Costs API can return `amount.value` as a string such as `"12.50"`. `OpenAIAPIUsageFetcher` decoded this field as `Double?`, so a successful HTTP 200 costs response failed JSON decoding and OpenAI Admin API usage rendered an error.

## Validation

- Live probe against `/v1/organization/costs` with an Admin API key decoded 31 cost buckets using the tolerant decoder.
- `swift test --filter OpenAIAPICreditBalanceTests` could not complete in this local environment because only Command Line Tools are installed; the macOS test target pulls `KeyboardShortcuts`, whose SwiftUI `#Preview` usage requires `PreviewsMacros` from full Xcode.

### DIFF
--- a/Sources/CodexBarCore/Providers/OpenAI/OpenAIAPIUsageFetcher.swift
+++ b/Sources/CodexBarCore/Providers/OpenAI/OpenAIAPIUsageFetcher.swift
@@ -353,6 +353,17 @@ private struct CostResult: Decodable {
     struct Amount: Decodable {
         let value: Double?
         let currency: String?
+
+        private enum CodingKeys: String, CodingKey {
+            case value
+            case currency
+        }
+
+        init(from decoder: Decoder) throws {
+            let container = try decoder.container(keyedBy: CodingKeys.self)
+            self.value = try container.decodeFlexibleDoubleIfPresent(forKey: .value)
+            self.currency = try container.decodeIfPresent(String.self, forKey: .currency)
+        }
     }
 
     let amount: Amount?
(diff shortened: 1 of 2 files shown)


steipete:
Thanks for the clean repro and patch. I landed the fix on `main` in f23a478e with the same tolerant decoding behavior, plus a small test tweak so the fixture covers both documented numeric values and the observed numeric-string value.

Validation:
- `swift test --filter OpenAIAPICreditBalanceTests`
- `make check`
- Codex review: no actionable findings

Closing this draft PR since the fix is now shipped on `main`.

PR #1005, opened 2026-05-17
fix: set explicit HTTP timeouts to prevent OOM when network is blocked

### LINKED ISSUE #1004: CodexBarCLI: catastrophic memory leak when network is blocked (28 GB per process)
## Bug

When a firewall (e.g. LuLu on macOS) blocks `CodexBarCLI`'s network access, the process grows to **28 GB of compressed memory** per instance, eventually triggering Jetsam (OOM killer) and crashing the system.

## Reproduction

1. Install CodexBar via `brew install --cask steipete/tap/codexbar`
2. Have a network firewall (LuLu, Little Snitch, etc.) that blocks new apps by default
3. Run `codexbar` or `codexbar cost` from terminal
4. The firewall blocks the connection silently (no allow prompt for CLI tools in LuLu)
5. Process hangs indefinitely, memory grows to 28+ GB

## Root Cause

`ProviderHTTPClient.swift` line 53 uses `URLSession.shared` with no custom timeout configuration:

```swift
public init(session: URLSession = .shared) {
    self.session = session
}
```

`URLSession.shared` defaults to:
- `timeoutIntervalForRequest` = 60s
- `timeoutIntervalForResource` = **604800s (7 days!)**

When the firewall blocks TCP connections, URLSession enters a state where it retries internally for up to 7 days. Combined with the `usage` command spawning concurrent async tasks for all providers, and the `cost` command loading session JSONL files into memory, the process accumulates tens of GB.

## Additional Factor

`CLIEntry.swift` line 163 defaults bare `codexbar` (no args) to the `usage` subcommand, which fetches ALL enabled providers simultaneously. A user exploring the CLI with no arguments triggers the worst case.

## Suggested Fix

Set an explicit, reasonable timeout on the URLSession configuration:

```swift
public final class ProviderHTTPClient: ProviderHTTPTransport, @unchecked Sendable {
    public static let shared = ProviderHTTPClient()

    private let session: URLSession

    public init(session: URLSession? = nil) {
        if let session {
            self.session = session
        } else {
            let config = URLSessionConfiguration.default
            config.timeoutIntervalForRequest = 30
            config.timeoutIntervalForResource = 60
            config.waitsForConnectivity = false
            self.session = URLSession(configuration: config)
        }
    }
    // ...
}
```

This ensures that blocked connections fail within 30-60 seconds instead of hanging for 7 days.

## Environment

- CodexBar 0.26.1 (installed via Homebrew)
- macOS 15 (Sequoia)
- LuLu firewall (blocks unsigned/new CLI binaries by default)
- Heavy Claude Code user (1B+ tokens/month → large session JSONL files)

msmolkin:
## Summary

- Set `timeoutIntervalForRequest = 30s` and `timeoutIntervalForResource = 90s` on `ProviderHTTPClient`'s URLSession
- Set `waitsForConnectivity = false` so blocked connections fail immediately instead of waiting

## Problem

`URLSession.shared` defaults to `timeoutIntervalForResource = 604800` (7 days). When a firewall blocks `CodexBarCLI` (e.g. LuLu on macOS blocks unsigned/new CLI binaries by default), every provider fetch hangs indefinitely. With the `usage` command spawning concurrent tasks for all enabled providers, the process accumulates **28 GB+ of memory per instance**, triggering Jetsam and crashing the system.

Observed on v0.26.1 with LuLu firewall on macOS 15. Two stuck `CodexBarCLI` processes totaling 56 GB crashed the machine.

## Fix

One-line change to `ProviderHTTPClient.init`: use a custom `URLSessionConfiguration` with sane timeouts instead of `URLSession.shared`.

Fixes #1004

## Test plan

- [ ] Run `codexbar usage` with network blocked → should fail within 30s with timeout error instead of hanging
- [ ] Run `codexbar usage` with network allowed → should work as before
- [ ] Existing tests pass (timeout doesn't affect normal response times)

### DIFF
--- a/Sources/CodexBarCore/ProviderHTTPClient.swift
+++ b/Sources/CodexBarCore/ProviderHTTPClient.swift
@@ -57,7 +57,9 @@ public final class ProviderHTTPClient: ProviderHTTPTransport, @unchecked Sendabl
             let configuration = URLSessionConfiguration.default
             configuration.timeoutIntervalForRequest = 30
             configuration.timeoutIntervalForResource = 90
+            #if !canImport(FoundationNetworking)
             configuration.waitsForConnectivity = false
+            #endif
             self.session = URLSession(configuration: configuration)
         }
     }
--- a/Sources/CodexBarCore/ProviderHTTPClient.swift
+++ b/Sources/CodexBarCore/ProviderHTTPClient.swift
@@ -50,8 +50,16 @@ public final class ProviderHTTPClient: ProviderHTTPTransport, @unchecked Sendabl
 
     private let session: URLSession
 
-    public init(session: URLSession = .shared) {
-        self.session = session
+    public init(session: URLSession? = nil) {
+        if let session {
+            self.session = session
+        } else {
+            let configuration = URLSessionConfiguration.default
+            configuration.timeoutIntervalForRequest = 30
+            configuration.timeoutIntervalForResource = 90
+            configuration.waitsForConnectivity = false
+            self.session = URLSession(configuration: configuration)
+        }
     }
 
     public func data(for request: URLRequest) async throws -> (Data, URLResponse) {


steipete:
Thanks for the report and patch. I landed a maintainer-polished version on main as 3ce41010.

Changes before landing:
- Kept the shared ProviderHTTPClient timeout fix.
- Preserved injected URLSession behavior for tests/custom transports.
- Added a regression test for the default client configuration.
- Added changelog coverage.

Proof:
- swift test --filter ProviderHTTPClientTests
- make check
- codex-review clean: no accepted/actionable findings reported

Co-authored-by credit preserved in the landed commit. Thank you.

PR #1008, opened 2026-05-17
feat(proxy): add invisible network proxy config core

serezha93:
Adds the invisible core for network proxy support.

What's included:
- `NetworkProxyConfiguration` in `CodexBarCore`
- proxy validation in `CodexBarConfigValidator`
- config store round-trip coverage

What is intentionally not included:
- Settings UI
- password/keychain storage
- transport wiring

Tests:
- `swift test --filter ConfigValidationTests`


steipete:
What's the use case for this, why is this needed?

steipete:
Closing this draft because it adds a persistent config/API surface without wiring any transport or delivering a user-visible proxy workflow. The maintainer asked for the concrete use case on May 20, and there has been no follow-up.

A future proxy contribution should start from the actual affected provider/transport path, include the authentication and credential-storage model, and prove requests use the configured proxy end to end. That will let us judge the smallest durable config contract instead of landing unused compatibility surface first. Thanks @serezha93.

PR #1038, opened 2026-05-18
Avoid auto-launching blocked Codex CLI candidates

m-rokai:
## Summary
- add a macOS Codex launch preflight for non-override CLI candidates before CodexBar auto-launches them
- detect npm `codex.js` shims and assess their vendored native Codex binary without launching it
- skip candidates with quarantine/malware xattrs or explicit policy failures such as revoked signing certificates, then fall back to the signed Codex.app CLI when present

## Why
A local repro machine kept seeing macOS "malware" warnings even though `CodexBar.app` itself was notarized and accepted by Gatekeeper. The suspicious launch target was the Codex CLI candidate selected from PATH:

```text
/usr/local/bin/codex -> ../lib/node_modules/@openai/codex/bin/codex.js
/usr/local/lib/node_modules/@openai/codex/.../vendor/aarch64-apple-darwin/codex/codex: CSSMERR_TP_CERT_REVOKED
```

The same machine also had a signed Codex.app CLI available:

```text
/Applications/Codex.app/Contents/Resources/codex --version
codex-cli 0.131.0-alpha.9
```

Before this change, CodexBar would pick the PATH shim first and could keep touching the blocked native binary through background/version probes. This change makes Codex resolution treat blocked candidates as unavailable for automatic selection, while preserving `CODEX_CLI_PATH` as an explicit user override.

## Validation
- `swift test --filter PathBuilderTests`
- `swift test --filter CodexCLILaunchGateTests`
- `git diff --check`

Not run: `swiftformat` / `swiftlint` because they are not installed in this local environment.

### DIFF
--- a/Sources/CodexBarCore/PathEnvironment.swift
+++ b/Sources/CodexBarCore/PathEnvironment.swift
@@ -83,6 +83,7 @@ public enum BinaryLocator {
         commandV: (String, String?, TimeInterval, FileManager) -> String? = ShellCommandLocator.commandV,
         aliasResolver: (String, String?, TimeInterval, FileManager, String) -> String? = ShellCommandLocator
             .resolveAlias,
+        launchCandidateFilter: (String, FileManager) -> Bool = CodexLaunchPreflight.isLaunchCandidateAllowed,
         fileManager: FileManager = .default,
         home: String = NSHomeDirectory()) -> String?
     {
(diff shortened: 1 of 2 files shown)


steipete:
Thanks @m-rokai, this is landed on `main` in 8e8b7405 with your contribution preserved as co-author.

I kept the core fix and added a maintainer follow-up before landing:
- automatic Codex binary resolution now skips macOS-blocked candidates and can fall back to the signed Codex.app CLI
- quarantine is treated as a Gatekeeper assessment signal, not an unconditional block, so notarized apps still work
- generic `spctl rejected` / revoked / malware results are treated as blocked
- CLI auto mode now tries web, then OAuth, then `codex app-server`, and the CLI strategy is unavailable when no allowed CLI can be resolved

Proof run:
- `swift test --filter PathBuilderTests --filter CodexBaselineCharacterizationTests --filter CodexCLILaunchGateTests`
- `make check`
- Codex review clean after the Gatekeeper/quarantine fixes

I could not push the rewritten fixup back to the PR branch without force-pushing over the original head, so I landed it directly on `main` and am closing this PR. Thanks again for the starting point.

Issue #1016, opened 2026-05-17
Explore faster local JSONL scanning for cost history


hhh2210:
Related to #1013 and #1014, but intentionally separate from that correctness fix.

While debugging the long `turn_context` attribution bug, one broader question came up: should CodexBar's local cost-history scanner eventually have a faster path for large JSONL histories?

The current bug is not primarily a raw throughput issue. In #1013, the failure was that the scanner dropped model state from a long Codex `turn_context` row and later fell back to `gpt-5` for model-less `token_count` rows. #1014 keeps that fix small and in Swift: retain truncated prefixes, recover only the Codex turn-context model from that prefix, keep normal JSON parsing away from partial rows, and invalidate the Codex cost cache.

This issue is only for the follow-up performance/design question:

- A Rust/SIMD helper could plausibly make sense if local histories grow large enough. Libraries such as `simdjson`, Rust `simd-json`, or byte-search primitives like `memchr`/`memmem` are a good fit for NDJSON-style scanning and selective parsing.
- A pure Swift optimization may be enough first: mmap-style reads, fewer allocations, SIMD-friendly byte searches, and selective JSON parsing only for rows containing relevant markers.
- Metal/GPU is probably a much weaker fit for this specific workload unless profiling proves otherwise. JSONL scanning has irregular control flow, string escaping, row boundaries, and I/O/memory bandwidth constraints. GPU JSON parsers usually need a structural-index pipeline, which would add a lot of complexity to a menu-bar app.

Open questions for maintainers:

- Is scanner throughput currently a real user-facing issue, or should this remain speculative until there is profiling data?
- If it becomes a problem, would you prefer keeping the scanner pure Swift, or would an optional/native helper in Rust be acceptable for this project?
- Should we add a small benchmark fixture around Codex/Claude JSONL scanning before considering any non-Swift implementation?

My bias would be: land the correctness fix in #1014 first, keep this issue as a design/profiling placeholder, and only consider Rust/SIMD after a benchmark shows the Swift scanner is actually the bottleneck.

hhh2210:
I think a benchmark would make this issue much easier to evaluate without turning it into a premature rewrite discussion.

One useful shape would be a two-layer benchmark:

[...]

PR #1042, opened 2026-05-19
refactor: derive Codex cost cache producer key from parser source hash

### LINKED ISSUE #1020: Replace manual Codex cost cache version bumps with auto-invalidation
The local Codex cost cache (`~/Library/Caches/CodexBar/cost-usage/codex-vN.json`) is invalidated by hand: every time the scanner gets a correctness fix that changes how rows are interpreted, someone has to bump `N` in `CostUsageCacheIO.artifactVersion(for: .codex)` so existing user caches get rebuilt.

Visible bump chain on this file:

| version | commit | reason |
| --- | --- | --- |
| 2 → 3 | d058431f | Fix Codex cost usage session replay |
| 3 → 4 | fd194450 (#698) | Codex cost scanner overcounting / cross-day undercounting |
| 4 → 6 | 2f6a31c2 | Codex priority cost pricing |
| 6 → 7 | 60f4a4e8 (#968) | Prefer Codex last token usage |
| 7 → 8 | 0487693b / 036b4975 (#1014) | Codex cost model attribution / long turn contexts |

That's six manual bumps over the lifetime of the scanner, each tied to a correctness fix. Two failure modes the current pattern does not catch:

1. **Forgotten bump.** A fix that silently expects rebuilt cache looks correct on a clean install and stays wrong on every existing user machine until they nuke `~/Library/Caches/CodexBar`. There is no review-time signal that says "this scanner diff changes semantics, you must also bump the cache version."
2. **Upstream-driven drift.** When OpenAI Codex CLI ships a new CLI version that changes JSONL shape, every existing user has a stale cache and no one in this repo knows to bump. The manual bump only catches *our* changes, not theirs.

## What might be smarter

Make the cache key carry enough about its own inputs that any meaningful change auto-invalidates it. Two sources of change matter, and they can stack:

```text
cacheKey = "codex-p<parserContentHash>-c<observedCodexVersionsHash>.json"
```

- `parserContentHash`: derived from the scanner source itself at build time (or kept as an explicit constant with a CI check that it changes when relevant files change). Catches forgotten bumps.
- `observedCodexVersionsHash`: hash of the set of `session_meta.codex_version` values seen during the scan. Catches upstream-driven drift without anyone in this repo noticing.

## Why this is separate from #1016

#1016 is about scanner throughput. This is about the cache-invalidation treadmill — independently useful even if the Swift scanner stayed exactly as fast as it is today.

## Open questions for direction

- Is the manual bump actually painful enough to warrant the design churn, or is the current pattern acceptable because it's caught by review?
- Would you accept a `session_meta.codex_version`-driven cache key, or does that add too much coupling to upstream Codex CLI internals?
- Any preference on parser-version source: explicit constant (status quo, just self-checked in CI) vs. file content hash vs. something else?

Happy to prototype if there's a direction you'd like to see. Not blocking on anything — filing because the pattern keeps showing up in scanner-fix PRs.

hhh2210:
## Summary

- stamp Codex cost caches with `producerKey = "codex:cu:p<hash>"`, where `<hash>` is a generated SHA256 prefix over the included non-Claude `Vendored/CostUsage/*.swift` inputs
- keep the existing `codex-v8.json` artifact path stable while ignoring legacy or mismatched Codex caches so they rebuild once
- add a tracked generated `CodexParserHash` file plus a non-mutating `Scripts/lint.sh` check that fails when the hash is stale
- leave non-Codex cost caches on their existing artifact-version behavior, so a Codex scanner/cache-source change does not cold-invalidate Claude caches

## Why

This implements the auto-invalidation design proposed in #1020 along the **forgotten-bump axis**. The earlier correctness fix in #1014 made oversized `turn_context` rows recover the model from the retained prefix, but a fixed parser can still be hidden by an already-written stale `codex-v8.json` cache. Continuing to solve that by renaming the file to `codex-v9.json`, then `codex-v10.json`, is fragile because every future Codex parser semantic fix has to remember to bump the filename.

The parser content hash is a conservative build-time producer identity for Codex cost-cache output. It changes when the included non-Claude `Vendored/CostUsage/*.swift` inputs change, so Codex caches rebuild after scanner/cache-source changes without requiring a manual `codex-vN.json` path bump. This may over-invalidate on refactors or comments in the included files, but it avoids under-invalidation after parser fixes.

The existing artifact version remains the coarse file/schema boundary. This PR removes the need to bump the Codex cache path for ordinary Codex scanner semantic changes; it does not remove the artifact-version mechanism for incompatible cache schema/path changes.

## Implementation

- `Scripts/regenerate-codex-parser-hash.sh` reads every `*.swift` under `Sources/CodexBarCore/Vendored/CostUsage/` excluding Claude-specific files, concatenates them in a stable order with a separator, takes SHA256, and writes the first 16 hex characters into `Sources/CodexBarCore/Generated/CodexParserHash.generated.swift`.
- The generated file is tracked in git so contributors and CI can compile without requiring a SwiftPM plugin or pre-build step.
- `Scripts/lint.sh` runs the generator in `--check` mode, comparing against a temporary expected file without mutating the worktree.

## Why this does not close #1020 entirely

#1020 also proposes invalidating on observed `cli_version` drift (an upstream-shape signal). I prototyped that on top of this PR and pulled back. The expected producer key needs to be known before `CostUsageCacheIO.load` returns, but the observed `session_meta.payload.cli_version` / `originator` set is only knowable after reading session files. Making that real would require either a preflight metadata scan or a separate producer-version index.

More importantly, the cache stores scanner output, not raw JSONL bytes. If upstream Codex CLI changes JSONL shape and CodexBar has not shipped a scanner fix, invalidating the cache only reruns the same unchanged scanner and produces the same wrong output. The useful follow-up for that axis is likely telemetry/logging for newly observed producer versions, not cache invalidation itself.

## Local benchmark context

From the merged #1017 synthetic shape benchmark on this machine after this patch:

```text
Codex JSONL shape benchmark: divisor=20 bytes=19621792 lines=7290 truncated=129 current=127.1MB/s baseline=47.3MB/s speedup=2.7x
```

From the real local 30-day Codex sessions tree used in #1016 (~1.24 GB / 545 files / 160k lines), a cold CLI cost refresh took about 14.18s wall-clock, while a cached read took about 1.15s. So this trades a one-time Codex cache rebuild when the included Codex scanner/cache-source inputs actually change for a stable invalidation mechanism that no longer needs manual path-version patches.

## Tests

- `Scripts/regenerate-codex-parser-hash.sh --check`
- `swift test --filter CostUsageCacheTests`
- `swift test --filter CostUsage`
- `Scripts/lint.sh lint`
- `swift test`

Refs #1020 (forgotten-bump axis only)


steipete:
Thanks @hhh2210. I landed this on `main` as 5aded294 with your co-author credit and changelog entry.

I kept the parser source-hash approach, wired the freshness check into lint, and verified it with focused cache/cost tests plus a synthetic CLI cost run. Closing this PR because the fix is now on `main`.

Issue #678, opened 2026-04-09
Codex provider OpenAI web refresh can spike WebKit.WebContent CPU and slow the whole machine


Iamadig:
**Summary**
When CodexBar is running with Codex enabled, the app can make the whole machine feel slow, not just the menu bar app itself.

This looks tied to the Codex OpenAI-web/dashboard refresh path. On launch, CodexBar triggers offscreen WKWebView work, and fresh com.apple.WebKit.WebContent helpers spike CPU hard.

**What I observed**
On April 9, 2026, I reproduced this locally and saw:

CodexBar briefly hit around 100% CPU on launch
fresh com.apple.WebKit.WebContent processes hit roughly 160–185% CPU
another WebKit content process stayed hot afterward around 50–110%
the slowdown was noticeable system-wide, not only inside CodexBar
Relevant logs
From ~/Library/Logs/CodexBar/CodexBar.log:

OpenAI web refresh start
OpenAI webview show
browser cookie import begins
second OpenAI webview show
Probe timed out (href=https://chatgpt.com/)
No matching browser session found
import mismatch: no signed-in session detected ...
There were also Safari cookie permission-denied messages, but those looked secondary. The main issue appears to be the expensive WebKit scrape / probe path.

**Suspected cause**
The expensive part seems to be the Codex OpenAI dashboard scraping flow using an offscreen WKWebView, especially when:

the browser session does not match the Codex account
the dashboard probe times out
startup still awaits this path
retries continue after mismatch / timeout
Expected behavior
launching CodexBar should stay lightweight
a failed Codex web/dashboard probe should not cause prolonged high CPU
a menu bar app should not make the whole machine feel slow
Suggested fixes
do not await OpenAI web refresh during startup
move it to a detached / delayed background refresh
use a much shorter timeout for the browser-session probe
add aggressive backoff after timeout or account/session mismatch
stop retrying once the app knows there is no matching signed-in browser session
**Environment**
macOS 14.7.7
CodexBar 0.20 (build 55)

amatiash:
I have constant 100% CPU usage for a single core.
macOS 15.7.4
CodexBar 0.20 (55)

hhh2210:
Still reproducible on **CodexBar 0.22 (build 57)**, macOS 26.4.1 (arm64).

Trigger: clicking the refresh (↻) button on the Codex row in the Providers pane (OAuth source, Pro account).

[...]

PR #1073, opened 2026-05-20
perf: fix status item icon observer CPU loop, background web scrapes, fix switcher first-click

ptstory:
## Summary

Fixes a CPU hot loop where `StatusItemController.observeStoreIconChanges()` was
pegging a full core at ~99% continuously, even when the rendered menu bar icon
was unchanged. Also backgrounds blocking web scrape work during regular refresh
cycles and fixes a first-click acceptance bug in the Codex account switcher.

Related: #678 — this PR addresses a separate root cause (icon observer loop, not
WebKit scrape) for the same user-facing symptom (CodexBar making the machine slow).

## Problem

A 5-second spindump of a stuck CodexBar process (v0.25, build 60) showed 38 of
42 main-thread dispatch samples landing in `observeStoreIconChanges()` →
`updateIcons()`. The root issue is `UsageStore.iconObservationToken`, which
registers Swift Observation dependencies on 10+ store fields. `isRefreshing`
toggles on every refresh cycle regardless of icon state, causing the observer to
re-fire immediately after re-registration.

The existing render-signature caching inside `applyIcon()` mitigated redundant
image/title mutation but did not suppress the observer callback itself —
`updateIcons()` still ran on every fire, computing render inputs, checking menu
attachment, and updating animation/blink state.

Separately, the regular refresh cycle was `await`-ing
`refreshOpenAIDashboardIfNeeded()` and `refreshCreditsIfNeeded()` inline, which
could block the main thread for seconds during web scrapes.

## Changes

**Fix 1 — Icon observer loop** (`StatusItemController.swift`, `UsageStore.swift`)

- Remove `isRefreshing` from `iconObservationToken` (noisy, not icon-relevant).
- Add `storeIconObservationSignature()` that derives a string signature from
  actual icon-render inputs (snapshot, status indicator, animation state, display
  text, provider visibility).
- Guard the observer callback: skip `updateIcons()` when the signature is
  unchanged.

**Fix 2 — Background web scrapes** (`UsageStore.swift`)

- For non-forced refreshes, run `refreshCreditsIfNeeded()` and
  `refreshOpenAIDashboardIfNeeded()` as unstructured background tasks so the UI
  never blocks on slow network operations.
- Forced refreshes (manual) still await inline.
- Tradeoff: unstructured tasks can't be cancelled if the parent refresh is
  cancelled. Acceptable for a menu bar app on a 2-minute poll cycle.

**Fix 3 — Switcher first-click** (`StatusItemController+SwitcherViews.swift`)

- Override `acceptsFirstMouse(for:)` → `true`.
- Swallow child button hit testing so the parent view handles mouse events.
- Implement `mouseDown`/`mouseUp` with press tracking to prevent drag-off
  misfires.
- Extract shared `applySelection(_:)` to deduplicate the action path.

## Testing

- `./Scripts/lint.sh lint`
- `swift test --skip-build --filter StatusItemIconObservationSignatureTests`
- `swift test --skip-build --filter CodexManagedOpenAIWebRefreshTests`
- `swift test --skip-build --filter StatusMenuCodexSwitcherTests`
- `swift test` currently hits a pre-existing unrelated failure in
  `OpenAIDashboardWebViewCacheTests` (`Preserved page expiry is scheduled without
  future cache activity`).

## Reproduction

1. Enable Codex provider with `openAIWebAccessEnabled = true` and
   `historicalTrackingEnabled = true`.
2. Observe main CodexBar process in Activity Monitor — CPU pegged at ~99%.
3. `spindump <PID> 5` confirms hot path in `observeStoreIconChanges()`.
4. After this patch, CPU drops to idle between refresh intervals.

### DIFF
--- a/Sources/CodexBar/StatusItemController+SwitcherViews.swift
+++ b/Sources/CodexBar/StatusItemController+SwitcherViews.swift
@@ -1293,20 +1293,24 @@ final class CodexAccountSwitcherView: NSView {
 
     override func mouseDown(with event: NSEvent) {
         let location = self.convert(event.locationInWindow, from: nil)
-        self.pressedAccountID = self.buttons.first(where: { $0.frame.contains(location) })?.identifier?.rawValue
+        self.pressedAccountID = self.accountID(at: location)
     }
 
     override func mouseUp(with event: NSEvent) {
         defer { self.pressedAccountID = nil }
         guard let pressedAccountID = self.pressedAccountID else { return }
         let location = self.convert(event.locationInWindow, from: nil)
-        guard let releasedAccountID = self.buttons.first(where: { $0.frame.contains(location) })?.identifier?.rawValue,
+        guard let releasedAccountID = self.accountID(at: location),
               releasedAccountID == pressedAccountID,
               let account = self.accounts.first(where: { $0.id == pressedAccountID })
         else {
             return
         }
         self.applySelection(account)
+    }
+
+    private func accountID(at pointInSelf: NSPoint) -> String? {
+        self.buttons.first(where: { self.convert($0.bounds, from: $0).contains(pointInSelf) })?.identifier?.rawValue
     }
 
     @objc private func handleSelect(_ sender: NSButton) {
@@ -1336,7 +1340,17 @@ final class CodexAccountSwitcherView: NSView {
     }
 
     func _test_simulateRuntimeClick(id: String) -> Bool {
-        guard let account = self.accounts.first(where: { $0.id == id }) else { return false }
+        guard let button = self.buttons.first(where: { $0.identifier?.rawValue == id }) else { return false }
+        let point = self.convert(NSPoint(x: button.bounds.midX, y: button.bounds.midY), from: button)
+        self.pressedAccountID = self.accountID(at: point)
+        defer { self.pressedAccountID = nil }
+        guard let pressedAccountID = self.pressedAccountID,
+              let releasedAccountID = self.accountID(at: point),
+              releasedAccountID == pressedAccountID,
+              let account = self.accounts.first(where: { $0.id == pressedAccountID })
+        else {
+            return false
+        }
         self.applySelection(account)
         return true
     }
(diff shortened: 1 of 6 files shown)


steipete:
Thanks for this. The direction makes sense: normal refresh should not feel stuck behind slow Codex credits/OpenAI dashboard work, and reducing menu-bar icon churn is worth looking at.

I am not comfortable merging this as-is yet:

- The switcher part is currently red locally. `swift test --filter StatusMenuCodexSwitcherTests` fails the new runtime-click tests: the simulated click still selects `live@example.com` instead of the requested managed/second-row account. Since that is exactly the first-click fix surface, this needs to be fixed before merge.
- The icon-loop claim needs measurement. `iconObservationToken` currently observes `isRefreshing`, so refresh start/finish can call `updateIcons()`, but `applyIcon()` already has render signatures and skips unchanged icon rendering. That may be extra observer noise, not necessarily a CPU hot loop.
- Backgrounding credits/OpenAI web is plausible, but please add proof that repeated normal refreshes do not stack unbounded credits/dashboard tasks. OpenAI dashboard has coalescing; credits currently does not look coalesced in this PR.

What would make this mergeable:

1. Fix the switcher tests and implementation.
2. Add before/after measurement for normal refresh wall time and icon work count/CPU. Suggested points: `UsageStore.refresh`, provider refresh group, `refreshCreditsIfNeeded`, `refreshOpenAIDashboardIfNeeded`, `StatusItemController.updateIcons`, and actual `IconRenderer.makeIcon`/render calls.
3. Consider splitting this into smaller PRs: icon observation, background Codex extras, and switcher first-click handling are separate risk surfaces.

I ran:

- `swift test --filter StatusItemIconObservationSignatureTests` passes.
- `swift test --filter CodexManagedOpenAIWebRefreshTests` passes isolated.
- `swift test --filter StatusMenuCodexSwitcherTests` fails 2 tests in the new click path.

ptstory:
Thanks for the detailed review. Agreed on all points.

Plan:

* Split into three smaller PRs as suggested: icon observation, background Codex extras, and switcher first-click handling.
* Investigate and fix the switcher click test failures. 
* Add before/after measurement for the icon observation change (normal refresh wall time, icon work count, and actual render calls).
* Add explicit coalescing/guarding (and coverage) for credits refresh to prevent repeated normal refreshes from stacking unbounded background work.

I'll convert this PR to draft and link the split follow up PRs here.

ptstory:
Split into three focused PRs per review feedback:
- #1078 — background codex refresh
- #1079 — switcher first-click
- #1081 — icon observer signature (this PR)

Closing this draft in favor of the focused PRs.

PR #1081, opened 2026-05-21
perf: suppress redundant icon observer callbacks

ptstory:
## Summary

Narrow the icon observation path so `updateIcons()` only fires when icon-relevant state changes.

The existing render signature inside `applyIcon()` already avoided redundant icon rendering, but it did not prevent redundant observer work from reaching `updateIcons()`. This change reduces observer callback and `updateIcons()` churn before render-signature checking is reached.

Split from #1073 per review feedback. Related: #678.

## Not included

- No background refresh changes (see #1078)
- No switcher first-click changes (see #1079)
- No unrelated cleanup from #1073

## Problem

`StatusItemController.observeStoreIconChanges()` watches `UsageStore.iconObservationToken`, which registers Swift Observation dependencies on multiple store fields.

`isRefreshing` toggles on every refresh cycle regardless of icon state. That causes the observer to re-fire and call `updateIcons()` even when no icon-relevant state changed.

That does not necessarily mean every observer fire performs an expensive render: `applyIcon()` already has render-signature checks and often skips unchanged icon rendering. The issue fixed here is narrower: unnecessary observer/updateIcons work was still happening before those render-signature checks were reached.

## Changes

- Remove `isRefreshing` from `iconObservationToken` because refresh start/finish state is not itself icon-relevant.
- Add `storeIconObservationSignature()` to derive a signature from actual icon-relevant state.
- Guard the icon observer callback so it skips `updateIcons()` when the derived store icon signature is unchanged.
- Add lightweight `[perf]` instrumentation gated behind `debugLogLevel=verbose` for future diagnosis.

## Measurement

Measured with `[perf]` instrumentation gated by `debugLogLevel=verbose`.

Scope: startup refresh cycles only.

| Metric | Before (`main`) | After (this branch) |
|--------|------------------|---------------------|
| `updateIcons()` calls per refresh cycle | 9–11 (avg 10) | 5–6 (avg 6) |
| Actual renders per refresh cycle | 1 | 1–2 |
| `applyIcon()` render-signature skips per refresh cycle | 10–12 (avg 11) | 6–8 (avg 7) |

## Scope note

This PR does not claim to fully explain or fix the original 99% CPU spike from #1073. After local account pruning from 7 managed Codex accounts to 3 kept/paid accounts, I can no longer reproduce the sustained CPU peg on demand.

The scoped claim here is narrower: refresh-state churn was causing `observeStoreIconChanges()` to call `updateIcons()` even when the icon-relevant store signature had not changed. This PR removes that redundant observer work before render-signature checks are reached.

## Testing

- `StatusItemIconObservationSignatureTests`
  - verifies signature stability across `isRefreshing` churn
  - verifies signature stability across status description/timestamp churn when the status indicator is unchanged
  - verifies signature changes when the status indicator changes
- `swiftformat`
- `swiftlint --strict`
- `make check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - OpenAI: parse Wednesday and Saturday dashboard reset lines so rate-limit reset times are not dropped on those days (#1080). Thanks @m1qaweb!
 - Localization: translate provider-detail labels and empty states when Simplified Chinese is selected (#1051). Thanks @wang93wei!
 - Antigravity: discover OAuth credentials from the bundled extension language server in newer IDE builds so Add Account works again (#1076). Thanks @xARSENICx!
+- Menu bar: suppress redundant icon observer work during refresh cycles, reducing icon update passes without changing rendered state (#1081). Thanks @ptstory!
 - Menu bar: wait for display changes to settle before recovering status items and retry if macOS still leaves the icon detached (#1074). Thanks @yipjunkai!
 - Menu: keep lower action rows stable when Refresh is highlighted or pressed (#1071). Thanks @MadanChaollaPark!
 - Linux CLI: avoid linking JetBrains provider parsing against `libxml2.so.2`, improving compatibility with newer distros that ship libxml2 2.15+ (#1046). Thanks @semsemyonoff!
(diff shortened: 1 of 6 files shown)


steipete (approved):
Reviewed the icon observer suppression path, ran focused tests, lint/check, Codex autoreview, and a release bundle relaunch smoke. Added the maintainer changelog entry and rebased onto current main.

PR #1094, opened 2026-05-21
Restore full-width provider switcher quota bars

bcharleson:
## Summary

- Reverts the v0.27.0 corner-badge quota indicators back to the pre-0.27 full-width bottom progress bar used through 0.26.1
- Reserves bottom space in stacked/inline switcher buttons so the bar no longer overlaps tab labels in multi-row layouts
- Refreshes switcher quota indicator fill when the open menu preserves the provider switcher during usage updates (⌘R / background refresh)

## Context

In `f9344673` (shipped in v0.27.0), the switcher meter was changed from a full-width bottom bar to a 16px top-trailing badge to keep buttons centered. That badge reads poorly when many providers are enabled and the switcher uses the stacked two-row layout: the meter floats near labels, misaligns, and looks like a broken underline instead of a quota bar.

This restores the readable bar while keeping the centering fix by reserving dedicated space below icon/title content instead of overlaying the badge on button chrome.

## Test plan

- [x] `swift test --filter StatusMenuSwitcherClickTests`
- [ ] Open merged menu with 8+ enabled providers; confirm each unselected tab shows a full-width bottom bar with visible track + fill
- [ ] Press ⌘R while menu is open; confirm switcher bars update without rebuilding the whole switcher
- [ ] Select a provider tab; confirm its quota bar hides while others remain visible
- [ ] Compare against CodexBar 0.26.1 on a second machine for visual parity

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Localizations: add Spanish and Catalan language packs and fill missing localization keys (#1041). Thanks @seifreed!
 
 ### Fixed
+- Menu: restore full-width provider switcher quota bars and refresh them while the menu stays open (#1094). Thanks @bcharleson!
 - Codex: accept the first click in the account switcher inside menu popovers (#1079). Thanks @ptstory!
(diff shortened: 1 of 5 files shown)


bcharleson:
@steipete 🚀🦾

PR #1105, opened 2026-05-22
Fix Linux build of AlibabaTokenPlanCookieHeader

m1qaweb:
AlibabaTokenPlanCookieHeader.swift uses HTTPCookie (lines 79, 94, 95, 97-99, 102, 113, 118, 134, 146) but only imports Foundation. On macOS HTTPCookie lives in Foundation, but on Linux it was moved to FoundationNetworking and is "explicitly marked unavailable" in Foundation -- so the new file added in #1098 (Alibaba Token Plan provider) breaks the Linux build of CodexBarCore outright. Sibling new files in the same PR (AlibabaTokenPlanUsageFetcher, T3ChatUsageFetcher) already use the right conditional import; this one was overlooked.

VERIFICATION

Ubuntu 24.04 x86_64 with Swift 6.2.4:

Before:
    $ swift build
    error: 'HTTPCookie' is unavailable: This type has moved to the
    FoundationNetworking module. Import that module to use it.
    (10+ occurrences across the file, build aborts)

After:
    $ swift build
    Build complete! (72.86s)
    $ swift test
    Test run with 14 tests in 3 suites passed

No source-of-truth change beyond restoring buildability -- adds the same
"#if canImport(FoundationNetworking) / import FoundationNetworking / #endif"
block that already guards GeminiStatusProbe, ManusCookieHeader,
AlibabaTokenPlanUsageFetcher, T3ChatUsageFetcher, and others.

### DIFF
--- a/Sources/CodexBarCore/Providers/Alibaba/AlibabaTokenPlanCookieHeader.swift
+++ b/Sources/CodexBarCore/Providers/Alibaba/AlibabaTokenPlanCookieHeader.swift
@@ -1,4 +1,7 @@
 import Foundation
+#if canImport(FoundationNetworking)
+import FoundationNetworking
+#endif
 
 struct AlibabaTokenPlanCookieHeaders: Sendable {
     private static let cachedAPIHeaderName = "__codexbar_alibaba_token_plan_api"


steipete:
Thanks for catching this. The exact Linux build fix is already on `main` via #1104 / fd4878ba, including the conditional `FoundationNetworking` import in `AlibabaTokenPlanCookieHeader.swift`.

Proof: the post-merge #1104 checks were green for Linux x64, Linux arm64, and `lint-build-test`, and current `main` contains the same import block this PR adds.

Closing this one as superseded so the queue stays tidy. Appreciate the tight repro and verification.

PR #1106, opened 2026-05-22
Fix length-1 quote trap in shared settings-reader cleaned() helper

m1qaweb:
PROBLEM

The provider settings readers in CodexBarCore (29 of them, plus
CodexBarConfig, CLIConfigCommand, and ProviderTokenResolver -- 32 files
total) share a copy-pasted cleaned() helper that strips one layer of
wrapping quotes:

    if (value.hasPrefix("\"") && value.hasSuffix("\"")) ||
        (value.hasPrefix("'") && value.hasSuffix("'"))
    {
        value.removeFirst()
        value.removeLast()
    }

When value is exactly one character (a lone " or '), both hasPrefix and
hasSuffix return true, removeFirst() empties the string, and removeLast()
then traps the process with:

    Swift/RangeReplaceableCollection.swift:867: Fatal error:
    Can't remove last element from an empty collection

The trap fires deep inside libswiftCore.so's
RangeReplaceableCollection.removeLast() and is not catchable. The whole
process dies (menu bar app or CLI command). Restarting with the same bad
env var or config value crashes again on next probe.

REACHABILITY

Triggered by these user-controllable inputs:

  - env var set to a lone quote (e.g. OLLAMA_API_KEY='"' from a
    misquoted shell script, launchd plist, systemd unit, or paste error)
  - apiKey / cookieHeader values in ~/.codexbar/config.json
  - codexbar config set-api-key --api-key '"' on the CLI

VERIFIED REPRODUCTION

Ubuntu 24.04 x86_64 with Swift 6.2.4, against upstream HEAD 123fe84e
(no patches applied). The PR 1098 follow-up fix for FoundationNetworking
was applied first so CodexBarCore could compile, but the cleaned()
helpers were left unpatched. Then a single test:

    @Test
    func alibabaTokenPlanCookieHeader_lonelyDoubleQuote_shouldNotCrash() {
        let env = ["ALIBABA_TOKEN_PLAN_COOKIE": "\""]
        let result = AlibabaTokenPlanSettingsReader.cookieHeader(
            environment: env)
        #expect(result == nil)
    }

Output:

    Swift/RangeReplaceableCollection.swift:867: Fatal error:
    Can't remove last element from an empty collection
    *** Program crashed: Illegal instruction at 0x00007eab486e2d76 ***
    Thread 2 crashed:
      1 [ra] 0x00007eab485ee6ac
        RangeReplaceableCollection<>.removeLast() + 443
        in libswiftCore.so

After applying this patch, the same test passes in 0.001s.

FIX

Replace the two-line removeFirst() + removeLast() pair in all 32
locations with:

    value = String(value.dropFirst().dropLast())

dropFirst() and dropLast() are empty-safe (they return an empty
Substring on empty input). For value.count >= 2 the result is
byte-identical to the old code. For the single trap case
(value.count == 1 with matching quote prefix/suffix), the new code
returns "" which the surrounding .isEmpty guards convert to nil --
strictly better than a process crash.

REGRESSION COVERAGE

New TestsLinux/SettingsReaderQuoteUnwrapTrapTests.swift exercises two
representative public readers (Alibaba Token Plan, Ollama API key) with
lone-quote and properly-quoted inputs across 7 test cases. 7/7 pass on
the fixed code; the lone-quote tests would have killed the test process
before this patch.

  swift build  ->  Build complete!
  swift test   ->  21 tests in 4 suites passed (was 14 before this PR)
  make check   ->  no new lint violations introduced by this PR

NOTES

  1. The trigger (lone " or ' as a config value) is unusual but well
     within "user provides unexpected input" territory. The
     disproportion of the consequence (full process crash, repeated on
     restart while the bad value persists) justifies the fix even
     though the trigger is rare.

  2. I chose dropFirst/dropLast over a length guard or a shared helper
     because the diff per file is the smallest possible (2 lines
     removed, 1 line added) and the idiom is natively empty-safe.
     Happy to rework to a different shape (value.count >= 2 guard, or
     extracted CodexBarCore helper) if you prefer either approach.

  3. Scope is 32 files because the pattern is copy-pasted that many
     times. Patching only one copy would leave the other 31 still
     crashable. A future refactor to a shared CodexBarCore helper
     would prevent future copy-paste of the pattern.

### DIFF
--- a/Sources/CodexBarCLI/CLIConfigCommand.swift
+++ b/Sources/CodexBarCLI/CLIConfigCommand.swift
@@ -221,8 +221,7 @@ extension CodexBarCLI {
         if (value.hasPrefix("\"") && value.hasSuffix("\"")) ||
             (value.hasPrefix("'") && value.hasSuffix("'"))
         {
-            value.removeFirst()
-            value.removeLast()
+            value = String(value.dropFirst().dropLast())
         }
         value = value.trimmingCharacters(in: .whitespacesAndNewlines)
         return value.isEmpty ? nil : value
--- a/Sources/CodexBarCore/Config/CodexBarConfig.swift
+++ b/Sources/CodexBarCore/Config/CodexBarConfig.swift
@@ -157,8 +157,7 @@ public struct ProviderConfig: Codable, Sendable, Identifiable {
         if (value.hasPrefix("\"") && value.hasSuffix("\"")) ||
             (value.hasPrefix("'") && value.hasSuffix("'"))
         {
-            value.removeFirst()
-            value.removeLast()
+            value = String(value.dropFirst().dropLast())
         }
         value = value.trimmingCharacters(in: .whitespacesAndNewlines)
         return value.isEmpty ? nil : value
--- a/Sources/CodexBarCore/Providers/Alibaba/AlibabaCodingPlanSettingsReader.swift
+++ b/Sources/CodexBarCore/Providers/Alibaba/AlibabaCodingPlanSettingsReader.swift
@@ -52,8 +52,7 @@ public struct AlibabaCodingPlanSettingsReader: Sendable {
         if (value.hasPrefix("\"") && value.hasSuffix("\"")) ||
             (value.hasPrefix("'") && value.hasSuffix("'"))
         {
-            value.removeFirst()
-            value.removeLast()
+            value = String(value.dropFirst().dropLast())
         }
 
         value = value.trimmingCharacters(in: .whitespacesAndNewlines)
--- a/Sources/CodexBarCore/Providers/Alibaba/AlibabaTokenPlanSettingsReader.swift
+++ b/Sources/CodexBarCore/Providers/Alibaba/AlibabaTokenPlanSettingsReader.swift
@@ -34,8 +34,7 @@ public struct AlibabaTokenPlanSettingsReader: Sendable {
         if (value.hasPrefix("\"") && value.hasSuffix("\"")) ||
             (value.hasPrefix("'") && value.hasSuffix("'"))
         {
-            value.removeFirst()
-            value.removeLast()
+            value = String(value.dropFirst().dropLast())
         }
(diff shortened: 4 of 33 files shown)


m1qaweb:
The two Linux CI failures here (build-linux-cli on linux-x64 and
linux-arm64) are not caused by this PR.

They are the pre-existing Linux build break introduced by #1098 (Alibaba
Token Plan provider), which added
Sources/CodexBarCore/Providers/Alibaba/AlibabaTokenPlanCookieHeader.swift
but omitted the conditional FoundationNetworking import. HTTPCookie is
unavailable on Linux without that import, so swift build on Linux has
been failing on every commit to main since #1098 landed.

I have a separate one-line PR open at #1105 that fixes exactly
this. Once that PR merges, this PR's CI will turn green on a re-run with
no changes here.

Verified by reproducing on upstream main 123fe84e with no patches
applied -- same error, same files, same exit code 1.

steipete:
@clawsweeper re-review

Maintainer proof for the real CLI path on the patched branch:

```text
$ tmp_home=$(mktemp -d)
$ HOME="$tmp_home" .build/debug/CodexBarCLI config set-api-key --provider ollama --api-key '"' --json --pretty
[
  {
    "error" : {
      "code" : 1,
      "kind" : "args",
      "message" : "Missing API key. Pass --api-key <key> or pipe it with --stdin."
    },
    "provider" : "cli",
    "source" : "cli"
  }
]
exit=1
```

That exercises the user-facing `codexbar config set-api-key` path with a lone double quote. The fixed branch exits cleanly with a normal argument error instead of trapping in `removeLast()`.

Additional local proof:

```text
$ swift test --filter SettingsReaderQuoteUnwrapTrapTests
Test run with 7 tests in 1 suite passed
```

The earlier Linux failures are superseded by #1104 / fd4878ba. The current rerun for this PR has Linux x64 and Linux arm64 green; macOS `lint-build-test` is still running.

Issue #227, opened 2026-01-21
Extra usage values displayed 100x too high


SCPAbee:
## Bug Description

The Extra usage section displays values that are 100x higher than the actual values shown in Claude.ai settings.

## Expected vs Actual

| Field | Claude.ai (correct) | CodexBar (wrong) |
|-------|---------------------|------------------|
| Spent | €7.63 | $763.00 |
| Limit | €20 | $2000.00 |
| Percentage | 38% | 38% ✓ |

## Details

The percentage calculation is correct (38%), suggesting the raw values are being multiplied by 100 somewhere, or cents are being displayed as dollars/euros.

## Environment
- macOS
- Claude Max plan
- CodexBar (latest version)

## Reproduction
1. Open CodexBar menu bar
2. Compare "Extra usage" values with Claude.ai Settings → Nutzung → Zusätzliche Nutzung
3. Notice the 100x discrepancy in absolute values while percentage matches

PR #1114, opened 2026-05-23
Claude: normalize OAuth extra usage spend limit from minor units

### LINKED ISSUE #1111: Extra usage displayed 100x too high in v0.29 — possible re-occurrence of #227
## Bug Description

Seeing extra usage values displayed 100x higher than Claude.ai, similar to #227.

| Field | Claude.ai (correct) | CodexBar (wrong) |
|-------|---------------------|------------------|
| Spent | \$32 | \$3200 |
| Limit | \$100 | \$10000 |
| Percentage | 32% ✓ | 32% ✓ |

The percentage is correct while the absolute values are 100x too high, suggesting cents are being treated as dollars somewhere.

Not sure if this is a re-occurrence of #227, a regression introduced by its fix, or a case where the Anthropic API returns values in different units depending on the plan/endpoint — Claude Enterprise plan may return values in cents while another path returns dollars, causing the fix to over- or under-correct depending on the plan.

## Environment
- macOS
- Claude Enterprise plan
- CodexBar v0.29

Yuxin-Qiao:
Fixes #1111.

## Summary
- Normalize Claude OAuth `extra_usage.used_credits` and `monthly_limit` from minor units to major currency units before creating `ProviderCostSnapshot`.
- Prevents recurrence of the #227-style 100x extra usage display bug.
- Example fixed behavior: `used_credits=763`, `monthly_limit=2000`, `currency=EUR` now displays as `€7.63 / €20.00`, while utilization remains ~`38.15%`.
- Scope is intentionally limited to Claude OAuth extra usage normalization.

## Scope
- Claude OAuth only
- No Cursor changes
- No formatter/UI changes
- No `ProviderCostSnapshot` semantic changes
- No Claude Web changes
- No provider fetcher or network behavior changes beyond the normalization boundary

## Validation
- `swift test --filter ClaudeOAuth`
- `swift test --filter Cursor`
- `swift test --filter MenuCard`
- `make check`
- `git diff --check`

Note: `swift test --filter Claude` previously hit an unrelated flaky `ClaudeOAuthCredentialsStoreSecurityCLITests` failure outside the #1111 files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Display: add optional workday markers for weekly progress bars (#1102). Thanks @Yuxin-Qiao!
 
 ### Fixed
+- Claude: normalize OAuth extra-usage spend limits from minor units so Enterprise spend displays as currency instead of 100x too high (#1114, fixes #1111). Thanks @Yuxin-Qiao!
 - OpenCode Go: read local usage history before falling back to browser-cookie dashboard fetches (#1021). Thanks @sopenlaz0!
 - Menu bar: show extra-usage spend as currency text for Claude and Cursor when that metric is selected (#1107). Thanks @Yuxin-Qiao!
 - Codex: run regular credits and OpenAI dashboard refreshes in the background while coalescing overlapping refresh work (#1078). Thanks @ptstory!
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeUsageFetcher.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeUsageFetcher.swift
@@ -954,7 +954,7 @@ extension ClaudeUsageFetcher {
         let normalized = Self.normalizeClaudeExtraUsageAmounts(
             used: used,
             limit: limit,
-            treatAsMajorUnits: isSpendLimit)
+            treatAsMajorUnits: false)
         return ProviderCostSnapshot(
             used: normalized.used,
             limit: normalized.limit,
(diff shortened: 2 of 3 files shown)


Yuxin-Qiao:
Thanks for the review.

I agree the remaining question is the Claude OAuth unit contract for spend-limit-only payloads.

What I can prove from this PR:

* current main has a concrete 100x path for OAuth `extra_usage` spend-limit payloads
* `used_credits=763`, `monthly_limit=2000` currently displays as `€763.00 / €2,000.00`
* after this patch, the same fixture displays as `€7.63 / €20.00`
* utilization remains ~`38.15%`, matching the reported symptom where percentage is correct but absolute currency is 100x high
* Claude Web and Cursor paths already normalize minor units before display

What I cannot currently provide:

* a live affected Claude OAuth account payload from the reporter
* confirmation from Anthropic that all OAuth spend-limit `extra_usage` payloads are always minor units

Given that, I’m intentionally leaving this as a small normalization-boundary patch with focused regression tests rather than adding heuristics or UI-level scaling.

Maintainer decision point:

* If maintainers accept that Claude OAuth `extra_usage.used_credits/monthly_limit` follows the same minor-unit contract as Claude Web and the reported #1111/#227 symptom, this patch fixes the reproduced 100x path in current main.
* If maintainers suspect mixed major/minor-unit OAuth payloads by plan/endpoint, I’m happy to pause and investigate a safer discriminator before merge.

No further code changes from me unless you prefer the mixed-unit investigation path.

PR #1118, opened 2026-05-23
Fix the status item missing issue.

HsiangHo:
I am the developer of **Barbee**, a menu bar management app like Bartender. 
I previously found that icons would often disappear when using Codexbar (due to off-screen issues), and this problem has been fixed.

To reproduce this problem, try installing **Barbee**(https://testflight.apple.com/join/HPKNaKKv), then hiding the CodexBar menu bar icon, and restarting CodexBar. 
****
This commit should resolve the issue of the menu bar icon disappearing.


steipete:
Thanks for digging into the menu bar manager/status item recovery path. I landed #1122 instead because it keeps the existing `NSStatusItem` identity for displaced live items, adds stable non-persistent AX identity, keeps destructive recreation for real materialization failures, and includes the same-count display-remap guard.

Landed in #1122: https://github.com/steipete/CodexBar/commit/22cf7d4d2b7631022901a6e62fe39227b21610b8

Closing this PR as superseded.

PR #1122, opened 2026-05-23
Fix menu bar item identity

lederniermagicien:
## Problem

CodexBar can be reclassified by menu bar managers as a new item after the user pins it visible. Thaw issue https://github.com/stonerl/Thaw/issues/605 reports CodexBar repeatedly moving back into the hidden section.

The risky CodexBar behavior was in status item visibility recovery: a live status item whose backing window had moved off the current screen could be treated as failed materialization, causing CodexBar to remove and recreate the `NSStatusItem`. For managers such as Thaw/Ice, that churn can make the pinned item look like a new menu bar item.

## Solution

Give CodexBar status items stable manager-facing identity and split recovery into destructive vs non-destructive paths:

- Real materialization failures still recreate the status item: missing button, missing window, or zero width.
- Live-but-displaced status items are no longer treated as broken. When a display removal leaves a visible status item attached to a stale/detached screen, CodexBar toggles visibility on the existing `NSStatusItem` and re-renders it instead of replacing it.
- Merged and per-provider status items now publish stable `autosaveName`, accessibility identifier/title, and tooltip values.

## Safety details

- Keeps destructive recreation for the cases that actually indicate a broken AppKit status item.
- Avoids replacing `NSStatusItem` instances when a menu bar manager may have intentionally parked or moved the item.
- Preserves existing split/merged item behavior; per-provider status items still get removed when switching back to merged mode.
- Tests lock both replacement recovery and non-destructive refresh behavior so the two recovery paths do not collapse into one another again.

## Changes

**Status item identity**
- Add stable identity metadata for merged and provider status items.
- Reapply that identity when recovery creates replacement items.

**Visibility recovery**
- Distinguish blocked status items from displaced live status items.
- Use non-destructive visibility refresh for stale/detached-screen status item windows after display removal.
- Continue destructive recreation for missing button/window or zero-width items.

**Tests**
- Cover detached and stale-screen live status items.
- Cover missing-window recovery.
- Cover stable status item identity.
- Cover non-destructive refresh preserving exact `NSStatusItem`/button instances.

## Verification

- [x] `swift test --filter 'MenuBarVisibilityWatcherTests|StatusItemControllerSplitLifecycleTests'`
- [x] `make check`
- [x] `./Scripts/compile_and_run.sh`
- [x] SOAK/AppKit probe: title/image/visibility changes preserved the same status item button/window.
- [x] AX probe against the running packaged app reported `AXMenuBarItem` with title `CodexBar` and identifier `CodexBar.StatusItem.codex`.

Full `swift test` was also run. It still fails in unrelated existing tests outside this status-item path:

- `ProviderStorageFootprintTests/forced scheduled storage refresh does not restart identical in flight scan`
- `CodexManagedOpenAIWebRefreshTests/regular refresh does not await OpenAI web scrape`
- `CodexBackgroundRefreshCoalescingTests/rapid regular refreshes coalesce concurrent Codex credits fetches`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Fixed
 - Groq: show a distinct Groq provider icon instead of reusing the Grok glyph (#1112). Thanks @kiankyars!
 - Claude: normalize OAuth extra-usage spend limits from minor units so Enterprise spend displays as currency instead of 100x too high (#1114, fixes #1111). Thanks @Yuxin-Qiao!
+- Menu bar: preserve status item identity during display-change recovery so menu bar managers do not treat CodexBar as a new hidden item (#1122, fixes #1109). Thanks @lederniermagicien!
 - OpenAI: retry transient Admin API usage failures once before surfacing an access error (#1117).
 - OpenCode Go: read local usage history before falling back to browser-cookie dashboard fetches (#1021). Thanks @sopenlaz0!
 - Menu bar: show extra-usage spend as currency text for Claude and Cursor when that metric is selected (#1107). Thanks @Yuxin-Qiao!
--- a/Sources/CodexBar/MenuBarVisibilityWatcher.swift
+++ b/Sources/CodexBar/MenuBarVisibilityWatcher.swift
@@ -135,13 +135,12 @@ enum MenuBarVisibilityWatcher {
     }
 
     static func shouldRefreshScreenChangePlacement(
-        previousScreenCount: Int,
-        currentScreenCount: Int,
+        previousScreenCount _: Int,
+        currentScreenCount _: Int,
         snapshots: [StatusItemVisibilitySnapshot])
         -> Bool
     {
-        guard currentScreenCount < previousScreenCount else { return false }
-        return self.hasAnyDisplacedVisibleSnapshot(snapshots)
+        self.hasAnyDisplacedVisibleSnapshot(snapshots)
     }
 
     static func shouldAttemptScreenChangeRecovery(snapshots: [StatusItemVisibilitySnapshot]) -> Bool {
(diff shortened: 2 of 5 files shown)


lederniermagicien:
Addressed the ClawSweeper blocker in `c41f6184`.

- Removed the new `NSStatusItem.autosaveName` assignments.
- Kept the non-destructive recovery path for live displaced status items.
- Kept stable non-persistent identity via accessibility identifier/title/tooltip.
- Updated tests to assert we do **not** introduce CodexBar autosave keys while still preserving the AX identity and same `NSStatusItem`/button instances through non-destructive refresh.

Verification after the update:

- `swift test --filter 'MenuBarVisibilityWatcherTests|StatusItemControllerSplitLifecycleTests'` passed.
- `make check` passed.
- `./Scripts/compile_and_run.sh` passed.
- AX probe against the running packaged app still reports `AXMenuBarItem` title `CodexBar` with identifier `CodexBar.StatusItem.codex`.

@clawsweeper re-review

diazdesandi:
Hi, from Thaw here. This fix looks related to an issue we’ve seen in internal testing with CodexBar 0.29.0 on macOS 26, where CodexBar’s menu bar item keeps getting recreated and ends up “stuck” until users reset Control Center’s preferences, which also corrupts its plist files.

We’re working on mitigating this from the Thaw side, but a CodexBar release with this change would solve it properly for users. Is there a rough timeline for when a build including this fix might ship?

cc. @stonerl @nightah

nk-tedo-001:
@steipete @lederniermagicien 
I'm not sure, but it looks like the unmerged icons could still be affected by this issue. 
I don't have 100% reproducibility, so I'll keep an eye on it. Could you please confirm or reject this on your end?

PR #1123, opened 2026-05-23
Add toggles to hide extra rate windows from menu bar

elijahfriedman:
Adds toggles in the provider preferences to hide individual extra rate windows from the menu bar. Hidden state is persisted per provider.

Screenshots show:
- Toggle visible in preferences
- Extra rate window hidden from menu bar
- Window restored when toggled back on
Test coverage also added: HiddenExtraRateWindowsTests.swift (7 test cases)

Screenshot 2026-05-23 at 2 07 36 PM
Screenshot 2026-05-23 at 2 07 50 PM
Screenshot 2026-05-23 at 2 08 04 PM
Screenshot 2026-05-23 at 2 08 39 PM
Screenshot 2026-05-23 at 2 10 38 PM

### Testing
1. Open preferences for a provider with extra rate windows
2. Toggle "Show in menu bar" to hide/show them
3. Verify state persists across app restarts

### DIFF
--- a/Sources/CodexBar/MenuCardView.swift
+++ b/Sources/CodexBar/MenuCardView.swift
@@ -692,6 +692,7 @@ extension UsageMenuCardView.Model {
         let resetTimeDisplayStyle: ResetTimeDisplayStyle
         let tokenCostUsageEnabled: Bool
         let showOptionalCreditsAndExtraUsage: Bool
+        let hiddenExtraRateWindowIDs: Set<String>
         let sourceLabel: String?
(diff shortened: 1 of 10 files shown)


steipete:
Thanks for the clean implementation and screenshots. I am going to close this one as a product call: per-extra-rate-window switches add too much settings/menu-management surface for the value they provide. The app should keep the provider detail UI focused, and if we need less menu noise we should solve it with a broader display model rather than per-row toggles.

Issue #1119, opened 2026-05-23
Show paid-plan expiration / subscription renewal date separately from usage reset timing


Yuxin-Qiao:
Users with monthly paid AI coding subscriptions need to know when their current paid plan expires or renews. This is useful for understanding how long current paid access remains valid and when the next billing cycle begins.

This should be displayed separately from quota or usage reset timing.

This feature is **not** about:

* quota reset date, such as \`resetsAt\`
* weekly/monthly usage reset
* billing-window reset
* OAuth access token expiry, such as \`expiresAt\`
* OpenAI API credit-grant expiry, such as \`nextGrantExpiry\`
* billing history timestamps, such as \`createdAt\` or \`ymd\`

The field needed here is specifically the paid subscription's own validity / expiration / renewal date, for example:

* plan expiration date
* subscription renewal date
* valid until date
* current paid plan validity end date

Current investigation findings:

* Claude: current code exposes quota reset windows and OAuth token expiry, but no confirmed paid-plan renewal or expiration field.
* Codex / OpenAI / ChatGPT: current code exposes plan tier, quota reset windows, and OpenAI API credit-grant expiry, but no confirmed ChatGPT Plus/Pro/Team renewal field.
* MiniMax: current code exposes quota reset and billing-history records, but no confirmed paid-plan expiration or renewal field.

Requested data:

Please provide a redacted response sample from any relevant endpoint or page payload that contains a real paid-plan validity / renewal / expiration date.

Likely places to check include:

* ChatGPT / Codex OAuth usage endpoint, such as \`/backend-api/wham/usage\`
* Claude account / subscription / billing payloads, if available from claude.ai
* MiniMax coding-plan / billing / subscription payloads, such as \`/v1/api/openplatform/coding_plan/remains\` or \`/user-center/payment/coding-plan\`

Please redact:

* cookies
* authorization headers
* access tokens / refresh tokens
* user IDs / organization IDs / group IDs
* email addresses / phone numbers / names
* invoice or payment identifiers
* private billing identifiers

Please keep:

* JSON field names
* date fields
* response structure
* plan/status fields if non-sensitive
* quota/reset fields only if needed for context

Once the real field names and types are confirmed, the implementation should be straightforward:

* add an optional provider-neutral field such as \`subscriptionExpiresAt\`, \`subscriptionRenewsAt\`, or \`planValidUntil\`
* parse provider-specific fields safely
* return nil if the field is missing or unparseable
[...]

PR #1127, opened 2026-05-23
Add redacted MiniMax diagnostic report CLI

Yuxin-Qiao:
This adds a small, MiniMax-only diagnostic report command for safe issue proof collection.

Scope:

* Adds `codexbar diagnose --provider minimax --format json`
* Output is always redacted and contains boolean auth-source presence only
* Does not perform a live fetch yet
* Emits `responseShape: null` when no live/fixture payload shape is available
* Leaves `suspectedSubscriptionFields` empty because no real subscription/renewal/valid-until payload field has been observed yet
* Does not implement paid-plan expiration parsing for #1119
* Does not change menu bar UI or normal usage output

Example output includes:

* `schemaVersion`
* `generatedAt` as ISO8601
* `provider`
* `liveFetch`
* `authSourcesPresent`
* `endpointsAttempted`
* `responseShape`
* `suspectedPlanFields`
* `suspectedDateFields`
* `suspectedSubscriptionFields`
* `redaction`

Validation:

* `swift build --target CodexBarCLI`
* `swift test --filter MiniMaxDiagnostic`
* `swift test --filter CLI`

Related issue:

* #1119

### DIFF
--- a/Sources/CodexBarCLI/CLIEntry.swift
+++ b/Sources/CodexBarCLI/CLIEntry.swift
@@ -52,6 +52,8 @@ enum CodexBarCLI {
                 self.runConfigSetAPIKey(invocation.parsedValues)
             case ["cache", "clear"]:
                 self.runCacheClear(invocation.parsedValues)
+            case ["diagnose"]:
+                await self.runDiagnose(invocation.parsedValues)
             default:
                 Self.exit(
                     code: .failure,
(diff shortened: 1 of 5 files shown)


steipete:
Closing this in favor of #1128 as the better MiniMax diagnostics direction. This branch also still has concrete diagnostic correctness findings around supported cookie env sources, config-backed auth presence, and redaction-summary output, so keeping both diagnostic PRs open would add noise.

PR #1132, opened 2026-05-24
Fix CommandCode subscription fallback

### LINKED ISSUE #1131: CommandCode usage fails when subscription enrichment is unavailable
## Problem

CommandCode usage can fail completely when the subscription enrichment endpoint is unavailable, even if the required credits endpoint succeeds. That leaves users with no CommandCode menu usage instead of a usable remaining-credit snapshot.

## Source evidence

On current `upstream/main` (`3c0dbe96`), `Sources/CodexBarCore/Providers/CommandCode/CommandCodeUsageFetcher.swift:24-28` starts both requests as sibling `async let` tasks and awaits both with `try`:

```swift
async let creditsResult = self.fetchCredits(cookieHeader: cookieHeader, transport: transport)
async let subscriptionResult = self.fetchSubscription(cookieHeader: cookieHeader, transport: transport)

let credits = try await creditsResult
let subscription = try await subscriptionResult
```

The subscription payload is already modeled as optional (`SubscriptionPayload?`) because free-tier responses can return `data: null`, and the analogous Codebuff provider treats subscription data as optional enrichment with a short grace period (`Sources/CodexBarCore/Providers/Codebuff/CodebuffUsageFetcher.swift:11-14`).

## Reproduction idea

Use a test transport where:

- `GET /internal/billing/credits` returns the existing successful fixture with `monthlyCredits`
- `GET /internal/billing/subscriptions` returns `503` or times out

Expected: CommandCode still shows the credits snapshot without plan/reset enrichment.

Actual: `fetchUsage` throws the subscription endpoint error and no CommandCode usage snapshot is returned.

## Expected behavior

Keep the credits request strict because it is the required usage payload. Treat subscription as best-effort enrichment, bounded by a short grace period after credits returns. If subscription succeeds with an active but unknown plan, keep failing explicitly so the catalog gap is still visible.

## Duplicate check

I checked existing issues and PRs for `CommandCode subscription usage`, `CommandCode credits subscription endpoint`, and `commandcode timeout`; I did not find an overlapping report or open PR.

## Regression coverage

Add transport-level tests for successful credits plus failing subscription, and for the existing unknown active-plan failure path.

rohitjavvadi:
Fixes #1131

## Summary
- keep `/internal/billing/credits` as the required CommandCode usage payload
- treat `/internal/billing/subscriptions` as best-effort enrichment with a short grace window after credits returns
- keep active unknown subscription plans as hard failures so the plan catalog does not silently drift

## Proof
Before the fix, `swift test --filter CommandCodeUsageFetcherTests` failed when credits returned 200 but subscriptions returned 503 because the subscription error bubbled as `apiError(503)`.

The new coverage proves:
- credits 200 + subscriptions 503 still returns a credits-only snapshot
- slow subscriptions do not block the required credits snapshot past the grace window
- subscriptions that finish first are still included
- credits failures still fail closed
- active unknown plans still throw
- non-finite test grace values cannot trap the timer conversion

## Checks
- `swift test --filter CommandCodeUsageFetcherTests`
- `swift test --filter CommandCode`
- focused web-refresh regression test
- `swift build`
- `make check`
- `git diff --check`

Full `swift test` was attempted twice but did not complete because `swiftpm-testing-helper` stopped producing output. The affected provider tests, build, and lint checks above passed.

### DIFF
--- a/Sources/CodexBarCore/Providers/CommandCode/CommandCodeUsageFetcher.swift
+++ b/Sources/CodexBarCore/Providers/CommandCode/CommandCodeUsageFetcher.swift
@@ -8,6 +8,7 @@ import FoundationNetworking
 public enum CommandCodeUsageFetcher {
     private static let log = CodexBarLog.logger(LogCategories.commandcodeUsage)
     private static let requestTimeoutSeconds: TimeInterval = 15
+    private static let subscriptionGraceSeconds: TimeInterval = 2
     private static let apiBase = URL(string: "https://api.commandcode.ai")!
     private static let creditsPath = "/internal/billing/credits"
     private static let subscriptionsPath = "/internal/billing/subscriptions"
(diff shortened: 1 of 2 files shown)


rohitjavvadi:
Closing this for now because I do not have a signed-in Command Code session available to capture the redacted live behavior proof requested by ClawSweeper. I’ll revisit this with proper live proof and send it again.

PR #1139, opened 2026-05-25
Antigravity: expose full per-model quotas via extraRateWindows

guhyun9454:
## What

`AntigravityStatusSnapshot.toUsageSnapshot()` currently collapses antigravity's 8+ per-model quotas into a 3-family representative summary (`primary`/`secondary`/`tertiary` = Claude / Gemini Pro / Gemini Flash). This PR additionally fills `extraRateWindows` with **every** model quota, so consumers can render the complete per-model breakdown the Antigravity IDE itself shows.

## Why

The 3-slot summary loses information that's available from the API:

- **Models with no family slot are dropped entirely.** GPT-OSS 120B classifies as `unknown` in `family(from:)`, so it never appears anywhere.
- **Per-effort variants are folded together.** Gemini 3.5 Flash (Medium/High/Low) and Gemini 3.1 Pro (Low/High) collapse to one representative each; Claude Sonnet vs Opus collapse to one "Claude".

There was no way for a consumer to recover the full list, even though `GetUserStatus` already returns all of it in `modelQuotas`.

## Change

10 lines in `toUsageSnapshot()`: map every `modelQuota` into a `NamedRateWindow` (`id`=modelId, `title`=label, window via the existing `rateWindow(for:)` helper) and pass it as `extraRateWindows`.

- **Purely additive.** `primary`/`secondary`/`tertiary` are unchanged, so the menubar's existing summary view and any current consumer keep working.
- Reuses the same `extraRateWindows` channel Claude already uses for its "Designs"/"Daily Routines" buckets — no schema or payload changes.

## Effect

Any surface that renders `extraRateWindows` (the CLI `serve`/`usage` JSON, and the menubar's expanded view) will now show antigravity's full per-model quota list. Example `serve --port … /usage?provider=antigravity` now emits 8 windows including `GPT-OSS 120B (Medium)`.

## Testing

- Built `CodexBarCLI` (release) and ran `serve` against a live, logged-in Antigravity instance: `/usage?provider=antigravity` returns all 8 model windows (Gemini 3.5 Flash Med/High/Low, Gemini 3.1 Pro Low/High, Claude Sonnet 4.6, Claude Opus 4.6, GPT-OSS 120B) with `usedPercent` matching the IDE's quota panel.
- `primary`/`secondary`/`tertiary` summary values unchanged.

### DIFF
--- a/Sources/CodexBar/MenuCardView.swift
+++ b/Sources/CodexBar/MenuCardView.swift
@@ -1095,27 +1095,10 @@ extension UsageMenuCardView.Model {
                     thresholds: input.quotaWarningThresholds[.weekly],
                     showUsed: input.usageBarsShowUsed)))
         }
-        if let extraRateWindows = snapshot.extraRateWindows {
-            metrics.append(contentsOf: extraRateWindows.map { namedWindow in
-                Metric(
-                    id: namedWindow.id,
-                    title: namedWindow.title,
-                    percent: Self.clamped(
-                        input.usageBarsShowUsed
-                            ? namedWindow.window.usedPercent
-                            : namedWindow.window.remainingPercent),
-                    percentStyle: percentStyle,
-                    resetText: Self.resetText(
-                        for: namedWindow.window,
-                        style: input.resetTimeDisplayStyle,
-                        now: input.now),
-                    detailText: nil,
-                    detailLeftText: nil,
-                    detailRightText: nil,
-                    pacePercent: nil,
-                    paceOnTop: true)
-            })
-        }
+        metrics.append(contentsOf: Self.extraRateWindowMetrics(
+            snapshot: snapshot,
+            input: input,
+            percentStyle: percentStyle))
         if input.provider == .kilo,
            metrics.contains(where: { $0.id == "primary" }),
            metrics.contains(where: { $0.id == "secondary" })
@@ -1441,7 +1424,7 @@ extension UsageMenuCardView.Model {
 
     private static func antigravityMetrics(input: Input, snapshot: UsageSnapshot) -> [Metric] {
         let percentStyle: PercentStyle = input.usageBarsShowUsed ? .used : .left
-        return [
+        var metrics = [
             Self.antigravityMetric(
                 id: "primary",
                 title: input.metadata.sessionLabel,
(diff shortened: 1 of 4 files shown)


Ksld154:
does this PR cause the model list expanded to 15+ items?   
seems not really cool

image

guhyun9454:
Thanks for flagging this @Ksld154 — you're right that the expanded list can blow up to 15+ rows, and I dug into the root cause.

**What's happening**

Antigravity usage comes from two sources:
- **Local probe** (Antigravity IDE language server, `source: local`) — returns a small curated model list (8 on my account, no internal models).
- **Remote OAuth fallback** (`fetchAvailableModels` on cloudcode-pa, used when the IDE isn't running) — returns **every** model the account can see, including internal ones like `tab_flash_lite_vertex` / `tab_jump_flash_lite_vertex` and legacy variants. Models without a `displayName` fall back to their raw model ID, which is why the screenshot shows those odd entries.

The PR's `extraRateWindows` mapping maps **all** `modelQuotas` with no filtering and re-sorts them alphabetically by label. The existing representative-slot logic already classifies `tab_*` / `lite` / `autocomplete` / unknown-family models (via `selectionPriority`), but `extraRateWindows` bypasses that classification, so everything leaks into the menu.

I also confirmed empirically (CLI, `--source cli`) that the language-server array order is **not** the IDE dropdown order, so simply preserving source order won't reproduce the nice grouping.

**Proposed fix**
1. Order the extra windows by a stable, parsed key from the label (family → version → tier) so it's grouped sensibly and new models (e.g. a future 3.6) slot in automatically — no hardcoded model list.
2. Reuse the existing `isAutocomplete` / `isLite` / family classification: show curated/known models, and only surface other models (`tab_*`, unknown) when their quota has actually been consumed.

I'm verifying the remote/OAuth path directly before opening the follow-up PR. Will link it here.

guhyun9454:
Follow-up PR opened: #1209.

It makes the `extraRateWindows` projection source-aware — local (curated IDE list) shows all models, remote OAuth fallback hides internal `tab_*`/lite/image/unknown models unless their quota was consumed — and replaces the alphabetical sort with a stable family-grouped order parsed from the label (so new model versions slot in automatically, no hardcoded list).

Verified against a real account: remote went 13 → 7 (the `tab_*_vertex` / image / lite noise from the screenshot is gone), local stays at its curated 8, both family-grouped. Before/after CLI output is in the PR description.

Ksld154:
Great!

thanks homie

7 is closer to antigravity cli displayed

PR #1146, opened 2026-05-25
Keep legacy credentials when config save fails

### LINKED ISSUE #1140: Config migration can clear legacy credentials after failed config save
`CodexBarConfigMigrator.loadOrMigrate` can clear legacy token/cookie stores even if persisting the migrated `~/.codexbar/config.json` fails.

Repro I used locally:

1. Point `CodexBarConfigStore` at a path whose parent is a regular file, so saving `config.json` fails with `NSPOSIXErrorDomain Code=20 "Not a directory"`.
2. Seed a fake legacy token store.
3. Call `CodexBarConfigMigrator.loadOrMigrate`.

Observed: the legacy secret is cleared, `codexbar.legacySecretsMigrationCompleted` is set, and `configStore.load()` still returns nil.

The root cause is that save failures are logged but migration cleanup still runs:

```swift
if state.didUpdate {
    do {
        try configStore.save(config)
    } catch {
        log.error("Failed to persist config: \(error)")
    }
}

if state.sawLegacySecrets || state.sawLegacyAccounts {
    let cleared = self.clearLegacyStores(...)
    if cleared {
        userDefaults.set(true, forKey: Self.legacyMigrationCompletedKey)
    }
}
```

Expected: legacy stores should only be cleared after the replacement config has been successfully persisted. If saving fails, keep the old credentials and leave the migration-completed flag unset so a later launch can retry safely.

RajvardhanPatil07:
Fixes #1140.

## Summary
- Stop legacy credential cleanup when migrated settings cannot be persisted.
- Keep the migration completion flag unset after a failed save so a later launch can retry.
- Add regression coverage for a blocked settings path and the retry path after persistence works again.

## Validation
- Parser hash check passed.
- SwiftFormat lint passed.
- SwiftLint strict passed with the local SourceKit framework path.
- Core target build passed.

## Maintainer validation
- `make check`
- `swift test --filter CodexBarConfigMigratorTests`

Focused test output covered the failed-save path preserving the legacy token and leaving the migration completion flag unset, then the retry path clearing the legacy token after persistence recovered.


steipete:
@clawsweeper re-review

Maintainer validation added to the PR body. I ran `make check` and `swift test --filter CodexBarConfigMigratorTests`; the focused test covers the failed-save preservation path and the retry cleanup path.

PR #1152, opened 2026-05-26
Refine Simplified Chinese token tagline

Jerry-CodeHub:
## Summary

- Update the Simplified Chinese tagline to use "Token" instead of "词元"
- Keep the existing informal "你" tone for consistency with the rest of zh-Hans localization
- Apply the same wording to the About tagline

## Why

"Token" is more natural in Chinese developer-facing AI product copy than the more literal "词元", while still matching the English source text:

> May your tokens never run out—keep agent limits in view.

This is intended as a tagline-specific wording choice. Other usage labels can continue using the existing glossary terms where they read better as formal UI labels.

## Proof

Verified the changed Simplified Chinese localization entries are present and parseable:

```bash
plutil -p Sources/CodexBar/Resources/zh-Hans.lproj/Localizable.strings | rg 'May your tokens never run out|about_tagline'
```

Output:

```text
"about_tagline" => "愿你的 Token 永不耗尽——随时关注智能体限制。"
"May your tokens never run out—keep agent limits in view." => "愿你的 Token 永不耗尽——随时关注智能体限制。"
```

## Testing

- Verified zh-Hans localization resource parsing with `plutil -p`

### DIFF
--- a/Sources/CodexBar/Resources/zh-Hans.lproj/Localizable.strings
+++ b/Sources/CodexBar/Resources/zh-Hans.lproj/Localizable.strings
@@ -171,7 +171,7 @@
 "Managed Codex accounts unavailable" = "托管 Codex 账户不可用";
 "Managed account storage is unreadable. Live account access is still available, " = "托管账户存储不可读。实时账户访问仍可用，";
 "Manual" = "手动";
-"May your tokens never run out—keep agent limits in view." = "愿你的词元永不耗尽——随时关注智能体限制。";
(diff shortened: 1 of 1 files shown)


Jerry-CodeHub:
@clawsweeper re-review

steipete:
Thanks for the careful wording pass. This is now covered by the newer Simplified Chinese localization polish that landed in #1165 (`62535877`): current `main` already keeps the tagline on `token`, preserves consistency with the rest of zh-Hans, and also updates the agent-limits wording plus the missing notification strings.

Closing this as superseded so we do not merge a conflicting tagline-only branch.

PR #1157, opened 2026-05-26
Handle provider switcher shortcuts in the open menu

### LINKED ISSUE #1144: Add Cmd-number shortcuts for provider switcher
It would be nice to add Cmd-number shortcuts for the provider switcher while the menu is open.

Expected behavior:

- Cmd+1 selects Overview
- Cmd+2 selects the first provider
- Cmd+3 selects the second provider
- and so on in the visible switcher order
- numbers past the visible items do nothing

This makes it faster to jump between Overview and providers without clicking.

### LINKED ISSUE #1156: Left/right provider navigation doesn't work
## Description

In CodexBar `0.29.1`, the merged provider switcher has left/right arrow navigation in the code, but it does not work for me in the actual open menu.

This is separate from #1144. Cmd-number shortcuts would be new shortcuts, while this is existing left/right navigation code that does not seem to work through the real menu.

## Environment

- CodexBar version: `0.29.1`
- Base/source context: current `main` after `v0.29.1` still has the same left/right navigation path
- macOS version: `26.5 (25F71)`
- Hardware: Apple Silicon (`MacBookPro17,1`)
- Menu mode: merged provider switcher menu

## Steps to Reproduce

1. Enable merged icon mode with multiple providers
2. Open the CodexBar menu from the menu bar
3. Press the right arrow key
4. Press the left arrow key

## Actual Behavior

Nothing changes in the provider switcher.

## Expected Behavior

- right arrow moves to the next switcher item
- left arrow moves to the previous switcher item

## Reproducibility

This is reproducible on my machine.

I verified:

- the provider switcher is visible at the top of the open menu
- clicking the provider switcher works
- pressing left/right while the menu is open does not switch providers

## Evidence

The code already appears to have left/right provider switcher navigation, but the existing test path calls `performKeyEquivalent` directly. That does not prove the real open `NSMenu` receives left/right keypresses.

anirudhvee:
Fixes #1156
Fixes #1144

## Summary

- make provider switcher keypresses work in the real open menu
- add Cmd-number shortcuts for the visible switcher order
- keep the menu open after switching providers from the keyboard

## Validation

- `make check`
- `swift test --filter StatusMenuSwitcherClickTests`
- verified the freshly built app with the menu open: Cmd+1..4 and left/right switch tabs, and the menu stays open

## Proof

- Video proof: shows Cmd+1..4 and left/right switching providers while the menu stays open.

https://github.com/user-attachments/assets/57008c51-f849-4a69-b533-c9188ef1e433

### DIFF
--- a/Sources/CodexBar/StatusItemController+Menu.swift
+++ b/Sources/CodexBar/StatusItemController+Menu.swift
@@ -127,9 +127,6 @@ extension StatusItemController {
 
     func menuDidClose(_ menu: NSMenu) {
         let wasHostedSubviewMenu = self.isHostedSubviewMenu(menu)
-        if ObjectIdentifier(menu) == self.providerSwitcherShortcutMenuID {
-            self.removeProviderSwitcherShortcutMonitor()
-        }
         self.forgetClosedMenu(menu)
         if wasHostedSubviewMenu {
             self.refreshOpenMenusIfNeeded()
@@ -138,6 +135,10 @@ extension StatusItemController {
 
     func forgetClosedMenu(_ menu: NSMenu) {
         let key = ObjectIdentifier(menu)
+
+        if key == self.providerSwitcherShortcutMenuID {
+            self.removeProviderSwitcherShortcutMonitor()
+        }
 
         self.openMenus.removeValue(forKey: key)
         self.menuRefreshTasks.removeValue(forKey: key)?.cancel()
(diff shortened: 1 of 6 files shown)


anirudhvee:
@clawsweeper re-review

anirudhvee:
Looks like the failed job is from `SubprocessRunnerTests.reads large stdout without deadlock`, not the provider switcher changes here.

I ran `swift test --filter SubprocessRunnerTests` locally and it passed. Could someone rerun the failed job?

PR #1162, opened 2026-05-26
Refresh expired StepFun tokens

LeoLin990405:
## Summary
- Add StepFun RefreshToken support and retry usage once after auth failures
- Persist refreshed tokens back to token accounts in the app and CLI paths
- Keep username/password login fallback for non-manual auth and improve StepFun auth error guidance

## Testing
- swift build --target CodexBarCore
- swift build --product CodexBarCLI
- swift run CodexBarCLI usage --provider stepfun --format json --pretty --no-color
- swift test --filter StepFun *(blocked locally: test build fails before StepFun tests because Command Line Tools cannot build the existing Swift Testing/preview-macro test targets)*


steipete:
Maintainer follow-up: I fixed the stale-cache recovery noted by Codex, narrowed auth-error detection, and wired StepFun manual-token persistence through both app settings and CLI config/token accounts.

Proof:
- `swift test --filter 'StepFunTokenRefreshTests|TokenAccountEnvironmentPrecedenceTests'` (37 tests)
- `make check`
- `swift test` (2,951 tests)
- Live CLI credential check: StepFun credentials are not present locally, so a real expired-token endpoint run cannot be performed here; `.build/debug/CodexBarCLI usage --provider stepfun --format json --pretty --no-color` correctly fails with missing StepFun username/password.

PR #1168, opened 2026-05-26
Add OpenAI project usage scoping

mstallone:
## Summary
- add OpenAI Admin API project scoping from `OPENAI_PROJECT_ID` or `providers[].workspaceID`
- pass `project_ids` to costs and completions usage requests, and label project-scoped snapshots in the menu
- keep selected OpenAI token accounts from inheriting a configured project filter, and fail closed instead of showing unscoped billing data

A single OpenAI org can have production, development, and Codex experiment projects. Showing one project keeps those costs and tokens from being mixed when the project is the useful view.

## Testing
- `git diff --check origin/main`
- `swift test`
- `make check`


mstallone:
I was debating the `OPENAI_PROJECT_ID` tradeoff that `@clawsweeper` mentioned.

The goal of this PR is to add OpenAI project scoping, especially for setups like ours where production and Codex usage live in different OpenAI projects. I do not want org-level or production-project usage showing up when I just want to look at my personal Codex usage.

I chose `OPENAI_PROJECT_ID` because it is the clearest environment-level signal for project scoping. For the configured/env OpenAI credential path, if that variable is set, I felt CodexBar should treat it as an intentional request for project-scoped OpenAI usage and fail closed rather than falling back to unscoped org billing.

That matches the config behavior and seems like the more consistent long-term policy, even if it creates a one-time surprise for a small set of users who happen to have `OPENAI_PROJECT_ID` in their shell environment.

That said, I am happy to adjusting this if maintainers prefer the compatibility-preserving behavior.

steipete:
Maintainer follow-up: I pushed a small policy fixup and merged the branch onto current `main`.

What changed:
- kept configured/Admin-key project scoping fail-closed so project-filtered usage never falls back to unscoped org billing
- preserved compatibility for ambient `OPENAI_API_KEY` + `OPENAI_PROJECT_ID`: it can attempt scoped Admin usage, but still keeps the legacy billing fallback when the key is not an Admin key
- kept selected OpenAI token accounts scrubbing `OPENAI_PROJECT_ID`

Proof:
- official OpenAI API docs confirm `project_ids` is supported on organization usage/costs endpoints
- `swift test --filter 'OpenAIAPIProjectScopeTests|OpenAIAPICreditBalanceTests|OpenAIAPIUsageFetcherTests|ProviderConfigEnvironmentTests|ConfigValidationTests|ProviderSettingsDescriptorTests'` (68 tests)
- `make check`
- autoreview clean: no accepted/actionable findings reported
- redacted CLI smoke with fake OpenAI credentials exercised the project-scoped API error path (`HTTP 400`) without leaking secrets

PR #1180, opened 2026-05-27
Avoid repeated display-change status item recreation

### LINKED ISSUE #1175: [Bug]: Display-change recovery retry loop corrupts Control Center and confuses menu bar managers
Hi, coming from https://github.com/stonerl/Thaw

### Summary

When a display configuration change occurs (monitor connected/disconnected,
resolution changed) while CodexBar's status item is hidden by a menu bar
manager (Bartender, Ice, Hidden Bar, etc.), CodexBar can destroy and recreate
its `NSStatusItem` up to four times in rapid succession. Each recreation is
visible to the system as a brand-new status item, which corrupts Control
Center's internal layout and causes menu bar managers to lose track of or
re-expose the previously hidden item.

We discovered this issue while testing for the release of Thaw 2.0.0 beta 13. CodexBar corrupted the Control Center files and broke the app functionality itself. By looking at the logs, it was discovered that there were issues happening with CodexBar. We are working on a way to handle them from our side. If the build with issues is required, we can provide it. We also have multiple logs files and we can run a live test if required.

### Environment

- CodexBar: /`0.29.x`
- macOS: 26.5
- Menu bar manager: Thaw

### Expected behavior

On a display configuration change, CodexBar refreshes or recreates its status
item once if recovery is needed, then stops — leaving menu bar managers free
to manage its position without interference.

### Actual behavior

The `verifyScreenChangeRecoveryIfNeeded` method introduced a retry loop in
commit c305858c ("fix: retry menu bar display recovery"). After the initial
`recreateStatusItemsForVisibilityRecovery()` call, the loop reschedules itself
up to three more times (2 s apart). When a menu bar manager re-parks or
re-hides the newly recreated item, the item can temporarily appear blocked
(`hasWindow: false` or `buttonWidth: 0`) before the manager completes its own
bookkeeping. Each such appearance triggers another recreation cycle, for up to
four total NSStatusItem destructions per display event.

### Reproduction

1. Install Thaw  and hide CodexBar's icon in its
   overflow section.
2. Run CodexBar.
3. Trigger a menu bar change
4. Observe Control Center icons shift or glitch, and Thaw re-exposing CodexBar
   as if it were a new unknown item.

image

Yuxin-Qiao:
## Summary

Fixes #1175 by making display-change follow-up verification observe-only.

The initial display-change recovery path may still perform the one allowed destructive `NSStatusItem` recreation when a visible item is blocked. Follow-up verification no longer recreates status items or reschedules another retry; it only logs and shows guidance if the item still appears blocked.

This avoids repeatedly destroying/recreating menu bar status items while menu bar managers such as Thaw, Bartender, Ice, or Hidden Bar are re-parking hidden items.

## Tests

- `swift test --filter MenuBarVisibilityWatcherTests`
- `swift build`

## Manual validation

Manual validation with Thaw/Bartender/Ice/Hidden Bar is still recommended because the original report depends on third-party menu bar manager behavior.

### DIFF
--- a/Sources/CodexBar/MenuBarVisibilityWatcher.swift
+++ b/Sources/CodexBar/MenuBarVisibilityWatcher.swift
@@ -147,7 +147,11 @@ enum MenuBarVisibilityWatcher {
         self.hasAnyBlockedVisibleSnapshot(snapshots)
     }
 
-    static func shouldRetryScreenChangeRecovery(
+    /// Returns true when a follow-up verification should report "still blocked" for logging/guidance.
+    /// The actual verifyScreenChangeRecoveryIfNeeded is observe-only and never calls
+    /// recreateStatusItemsForVisibilityRecovery — the single allowed recreate happens in
+    /// checkScreenChangeStatusItemVisibility before any follow-up is scheduled.
+    static func shouldObserveBlockedScreenChangeFollowUp(
         attempt: Int,
         snapshots: [StatusItemVisibilitySnapshot])
         -> Bool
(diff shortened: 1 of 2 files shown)


steipete:
Thanks for the cleanup here. We landed the reporter-validated fix in #1176 as 63d9373b, which removes the repeated display-change recovery retry path and keeps the macOS guidance check covered by tests.

Closing this as superseded by #1176 so we do not carry two competing fixes for #1175.

PR #1206, opened 2026-05-29
fix: preserve Codex web credits-only refresh

### LINKED ISSUE #1204: Codex web dashboard rejects authorized credits-only snapshots
### Summary
Codex web dashboard refresh can fail the whole provider when the authenticated dashboard snapshot has credits/account data but no usage-limit windows yet. CLI/OAuth paths already preserve credits-only responses; the web path throws `.missingUsage`.

### Evidence
- CodexBar 0.31.0 on macOS.
- OpenAI web auth was valid: `/api/auth/session` returned 200 and the dashboard signed-in email matched the Codex account.
- The menu could show Codex cost/credits data, but the Codex usage header stayed at "Not fetched yet" during the partial web state.
- Local log showed dashboard parsing reached authenticated dashboard content but failed on missing/non-usage chart data: `usage breakdown error: Only non-usage chart candidates found: Daily threads by client, Daily skill invocations`.
- A regression test against `CodexWebDashboardStrategy.makeAuthorizedDashboardResultForTesting` reproduces the failure: authorized dashboard + `creditsRemaining` + no primary/secondary limits currently throws `.missingUsage`.

### Expected
When ownership is authorized and credits are present, Codex web should return a partial usage result with nil limit windows plus credits/identity, matching the existing CLI/OAuth credits-only fallback behavior.

### Actual
The authorized web dashboard result throws `.missingUsage`, which can surface as a failed Codex fetch / "Not fetched yet" instead of preserving useful dashboard data.

### Proposed fix
If `dashboard.toUsageSnapshot(...)` returns nil but `dashboard.toCreditsSnapshot()` exists, create a credits-only `UsageSnapshot` with the authorized identity and attach the dashboard/credits.

### Validation plan
- `swift test --filter CodexWebDashboardStrategyAuthorityTests`
- `make check`
- `TZ=UTC swift test`

soumikbhatta:
## Summary
- preserve authorized Codex web dashboard snapshots when credits are present but usage-limit windows are absent
- synthesize a credits-only Codex `UsageSnapshot` with the attached dashboard identity, matching existing CLI/OAuth partial-result behavior
- add regression coverage for the authorized credits-only web dashboard path

## Why
Fixes #1204.

The Codex web strategy already verifies dashboard ownership before attachment, but it still threw `.missingUsage` when `dashboard.toUsageSnapshot(...)` returned nil. That made a useful authenticated dashboard response with credits/account data look like a failed provider fetch. CLI and OAuth already preserve credits-only responses, so this brings the web path in line with those sources.

## Validation
- `swift test --filter CodexWebDashboardStrategyAuthorityTests`
- `make check`
- `TZ=UTC swift test`


soumikbhatta:
Behavior proof / validation update for ClawSweeper:

This PR is scoped to the Codex web dashboard attach path only. The local failure that led to #1204 was an authenticated OpenAI dashboard refresh where account/credits data was available but usage-limit windows were absent, causing CodexBar to treat the provider refresh as failed instead of preserving the partial result.

Redacted local evidence from the failing setup before the patch:

```text
CodexBar 0.31.0, provider=codex, source=openaiDashboard
/api/auth/session status=200
matched signed-in dashboard account: [redacted]
usage API status=401
usage breakdown parse rejected dashboard charts: Only non-usage chart candidates found: Daily threads by client, Daily skill invocations
openaiDashboard credits/account data present
usage primary=nil, secondary=nil -> provider surfaced as not fetched / missing usage
```

Source-level failure on `main` matched that behavior: after authority evaluation returned `.attach`, `CodexWebDashboardStrategy` threw `.missingUsage` when `dashboard.toUsageSnapshot(...)` returned nil, before returning the dashboard credits.

Patch behavior on this branch:

```text
commit: 79ae80972de2b44fe7c1115adad52af74f4a7c1f
swift test --filter CodexWebDashboardStrategyAuthorityTests
Result: 9 tests passed, 0 failures
```

The new regression test constructs the authorized credits-only dashboard state and verifies the returned result preserves:

```text
usage.primary == nil
usage.secondary == nil
usage.identity.accountEmail == owner@example.com
usage.identity.loginMethod == pro
credits.remaining == 42
dashboard == original dashboard
```

Additional local validation on the patched branch:

```text
make check
SwiftFormat: 0/984 files require formatting
SwiftLint: 0 violations in 983 files
Codex parser hash is current
```

Full-suite validation was also run before the timestamp-only polish commit:

```text
TZ=UTC swift test
Result: 3131 tests passed, 0 failures
```

Honest boundary: my current live Codex account no longer naturally stays in the credits-only state on demand after refresh, so I cannot attach a fresh after-patch live screenshot that still has credits with no usage windows. The reproduced failure state above is from the real authenticated local setup; the branch adds the narrow regression that proves the same authorized credits-only state now returns credits and identity instead of `.missingUsage`.

@clawsweeper re-review

PR #1208, opened 2026-05-29
Bound serve requests and coalesce concurrent /usage cache misses

enieuwy:
## Summary

`codexbar serve` can leave `/usage` hanging indefinitely and lets concurrent cache-miss requests each launch a full provider fetch. This surfaced downstream as a SketchyBar widget stuck grey/stale: the server's `/health` kept returning `200` while `/usage` never responded, and queued callers stampeded the providers.

Two root causes in `CLIServeCommand`:

- **No total request deadline.** `serveUsage` loops providers with `webTimeout: 60` each, and the socket `requestReadTimeoutMilliseconds = 5000` only bounds the *request read*, not the response. A slow upstream wedges `/usage` with no overall cap.
- **No single-flight on cache miss.** `cachedServeResponse` was check-then-fetch; the cache actor only guarded the dict, so N concurrent callers on an expired key each ran the provider fetch.

## Changes

- Add a total per-request deadline that races route work against a timeout; on expiry return **`504 Gateway Timeout`** and do **not** cache the timeout.
- **Make the deadline configurable: `--request-timeout <seconds>` (default 30; `0` disables it / unbounded)**, mirroring the existing `--port` / `--refresh-interval` options. This addresses the compatibility concern: setups that prefer slow-but-eventual success over fail-fast can set `--request-timeout 0` (or a high value) to keep the old behavior.
- Add per-key single-flight in `CLIServeResponseCache`: concurrent callers for the same in-flight key await the one producing fetch. In-flight state always clears and waiters are always resumed, including on timeout.
- Add `CLIHTTPStatus.gatewayTimeout` (504); make `CLIHTTPStatus` and `CLILocalHTTPResponse` `Sendable`.
- `/health` stays cheap and static (process liveness), unchanged.

No public API churn beyond the new status case; the cache/deadline helpers are internal/file-scoped.

### A note on cancellation
On timeout the producing `Task` is cancelled, but the underlying provider fetch is only best-effort cancellable (still independently bounded by `webTimeout: 60`). This cannot corrupt state: `completeFetch` runs exactly once with the 504 (clearing in-flight, resuming waiters), and a late-completing fetch result is discarded and never written to the cache.

## Tests

`Tests/CodexBarTests/CLIServeRouterTests.swift`:

- 5 concurrent cache misses invoke the producer **exactly once** and all receive the same response.
- A timeout returns `504`, is **not** cached, a later success caches, and a third call is served from cache (producer called exactly twice total).
- `--request-timeout 0` disables the deadline: a slow producer finishes and returns its real response instead of a 504.
- `decodeServeRequestTimeout` validation (non-numeric / negative -> rejected; `0`, explicit value, and default 30 accepted).

These prove the deadline / single-flight behavior deterministically with an injected producer and **no live provider, Keychain, or browser access** (per `AGENTS.md`, which warns against validation that can trigger Keychain/account prompts).

## Real `codexbar serve` proof

Built from this branch (`CodexBarCLI`), provider-free paths so no account data is touched:

```
$ CodexBarCLI serve --request-timeout abc
[{"error":{"message":"--request-timeout must be zero or greater.","code":1,"kind":"args"},"provider":"cli","source":"cli"}]   # exit 1

$ CodexBarCLI serve --port 63907 --request-timeout 30
CodexBar server listening on http://127.0.0.1:63907

$ curl /health   -> {"status":"ok"}                http=200 time=0.0007s
$ curl -X POST /usage -> {"error":"method not allowed"} http=405
$ curl /nope     -> {"error":"not found"}          http=404
```

A live `/usage` 504 demonstration is intentionally omitted: it requires a real provider fetch, which an unsigned debug build can trigger a Keychain prompt for — exactly the validation `AGENTS.md` says to avoid. The timeout/coalescing/unbounded behaviors are instead covered deterministically by the unit tests above.

## Commands run

- `swift test --filter CLIServeRouterTests` -> 10 passed
- `swift test` -> full suite passed (re-verified focused suite + `make check` after rebasing onto latest `main`)
- `make check` -> SwiftFormat + SwiftLint clean (0 violations)

No UI changes.


enieuwy:
Thanks @clawsweeper — addressed in `ab39b023` (PR body updated with proof + the cancellation note):

**30s cap (P1) — now configurable.** Added `--request-timeout <seconds>` (default 30), matching `--port` / `--refresh-interval`. `0` disables the deadline entirely (unbounded), so existing slow-but-eventual setups can opt out of fail-fast and keep prior behavior. Covered by decode-validation tests and a behavior test asserting `0` lets a slow producer finish instead of 504.

**Real serve proof (P1).** Added redacted output from a branch-built `CodexBarCLI serve` covering the provider-free paths: server boot/listen, `/health` 200, `POST /usage` 405, `/nope` 404, and the new flag's validation error. A live `/usage` 504 demo is intentionally omitted because it requires a real provider fetch that an unsigned debug build can raise a Keychain prompt for — the exact validation `AGENTS.md` warns against. The deadline, single-flight coalescing, timeout-not-cached, and `0`-unbounded behaviors are instead proven deterministically by 10 unit tests with an injected producer (no providers/Keychain/browser).

**Cooperative cancellation (P2).** Confirmed and documented: on timeout the work `Task` is cancelled, but the provider fetch is only best-effort cancellable (still bounded by `webTimeout: 60`). It can't corrupt state — `completeFetch` runs once with the 504 (clears in-flight, resumes waiters) and any late fetch result is discarded and never cached.

PR #1209, opened 2026-05-29
fix: filter and order Antigravity per-model quota windows

guhyun9454:
## Summary

Follow-up to #1139. That PR exposed **every** Antigravity model quota in `extraRateWindows`. On the **remote OAuth fallback** path this dumps the full account catalog — 13+ entries including internal models (`tab_flash_lite_vertex`, `tab_jump_flash_lite_vertex`), image models, and lite variants — alphabetically sorted. The menu expanded to 15+ rows of noise (reported in #1139).

## Root cause

Both data sources flow through `AntigravityStatusSnapshot.toUsageSnapshot()`:
- **Local probe** (running Antigravity IDE language server): a small **curated** list.
- **Remote OAuth fallback** (`fetchAvailableModels`): the **entire** account catalog, including internal/legacy models; entries without a `displayName` show their raw model id.

PR #1139's mapping took **all** `modelQuotas`, unfiltered, sorted alphabetically. The existing `normalizeModel` classification (`isLite`, `isAutocomplete`/`tab_`, family) was bypassed.

## Change

Make the projection **source-aware**:
- **Local** (curated, trusted): show all models, ordering only.
- **Remote** (noisy fallback): show a model only if it is family-recognized and not lite/autocomplete/image, **or** its quota was actually consumed (`remainingFraction < 0.999`).

Replace the alphabetical sort with a stable **family-grouped order parsed from the label** (family rank → version desc → tier → label) — new model versions slot in automatically, no hardcoded model list. The three summary slots (Claude / Gemini Pro / Gemini Flash) are unchanged.

## Real behavior proof

Verified with `codexbar usage --provider antigravity` against a real account (redacted), built from this branch.

**Remote (OAuth) — BEFORE (current `main`):** 13 entries, alphabetical, with noise
```
Gemini 2.5 Flash Image, tab_flash_lite_vertex, Gemini 2.5 Pro, Gemini 3 Pro (High),
Gemini 2.5 Flash Lite, Gemini 3 Pro Image, Gemini 3 Flash, Gemini 3.1 Flash Lite,
Gemini 3.1 Pro (Low), Gemini 3.1 Pro (High), tab_jump_flash_lite_vertex,
Gemini 3 Pro (Low), Gemini 2.5 Flash
```

**Remote (OAuth) — AFTER (this branch):** 7 entries, junk removed, family-grouped
```
Gemini 3.1 Pro (High), Gemini 3.1 Pro (Low), Gemini 3 Pro (High), Gemini 3 Pro (Low),
Gemini 2.5 Pro, Gemini 3 Flash, Gemini 2.5 Flash
```

**Local — AFTER (this branch):** 8 curated models, family-grouped, unchanged set
```
Claude Opus 4.6 (Thinking), Claude Sonnet 4.6 (Thinking),
Gemini 3.1 Pro (High), Gemini 3.1 Pro (Low),
Gemini 3.5 Flash (High), Gemini 3.5 Flash (Medium), Gemini 3.5 Flash (Low),
GPT-OSS 120B (Medium)
```

## Tests

`Tests/CodexBarTests/AntigravityStatusProbeTests.swift` adds local (show-all) and remote (filtered) fixtures built from the exact verified catalogs above, plus conditional-display (consumed junk shown) and ordering edge cases. `MenuCardAntigravityTests.swift` updated for the source-aware path.

Acceptance: `swift test --filter AntigravityStatusProbeTests`, `make build`.

### DIFF
--- a/Sources/CodexBarCore/Providers/Antigravity/AntigravityRemoteUsageFetcher.swift
+++ b/Sources/CodexBarCore/Providers/Antigravity/AntigravityRemoteUsageFetcher.swift
@@ -146,7 +146,8 @@ public struct AntigravityRemoteUsageFetcher: Sendable {
         return AntigravityStatusSnapshot(
             modelQuotas: models,
             accountEmail: claims.email,
-            accountPlan: Self.resolvePlan(response: codeAssist, claims: claims))
+            accountPlan: Self.resolvePlan(response: codeAssist, claims: claims),
+            source: .remote)
     }
 
     private static func shouldRefresh(expiryDate: Date?, now: Date) -> Bool {
--- a/Sources/CodexBarCore/Providers/Antigravity/AntigravityStatusProbe.swift
+++ b/Sources/CodexBarCore/Providers/Antigravity/AntigravityStatusProbe.swift
@@ -106,12 +106,11 @@ public struct AntigravityStatusSnapshot: Sendable {
         // extraRateWindows carries a source-aware set: the full curated list for
         // .local (verified junk-free), and a filtered list for .remote (catalog noise
         // hidden, consumed quota always kept). Sorted by family→version→tier.
-        let shownModels: [AntigravityNormalizedModel]
-        switch self.source {
+        let shownModels: [AntigravityNormalizedModel] = switch self.source {
         case .local:
-            shownModels = normalized
+            normalized
         case .remote:
-            shownModels = normalized.filter { m in
+            normalized.filter { m in
                 (m.family != .unknown && !m.isLite && !m.isAutocomplete && !m.isImage)
                     || (m.quota.remainingFraction ?? 1.0) < 0.999
             }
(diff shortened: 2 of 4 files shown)


guhyun9454:
Addressed the [P2] raw-id version parsing finding in 464472c9.

`parseVersion` now accepts `-` as well as `.` between major and minor, so a raw model id used as the label when `displayName`/`label` is absent (e.g. `gemini-3-1-pro-low`) parses as 3.1 and keeps newest-first ordering. Added a regression test (`hyphenated raw model ids without display name parse minor version`).

For context: in the live catalogs I captured, the actual Antigravity model ids already use dotted minors (`gemini-3.1-pro-low`, `gemini-2.5-pro`) and every family-recognized model carries a `displayName`, while the only raw-id-as-label entries (`tab_*_vertex`) are filtered out — so this didn't misorder observed data. The fix is defensive against the hyphenated-id shape and verified to leave dotted-version parsing unchanged.

On the remote `extraRateWindows` shape change (13 → 7): that's the intended behavior of this PR and is a maintainer call — happy to gate it behind an option or keep a separate full-catalog projection if preferred over changing the existing output.

PR #1215, opened 2026-05-29
fix: add Opus 4.8 to built-in Claude pricing table

### LINKED ISSUE #1210: Opus 4.8 cost is not calculated
In v0.31.0 cost for Opus 4.8 tokens is not calculated. It is always empty:
Image

chhoumann:
Fixes #1210.

## Problem
In v0.31.0, cost for Opus 4.8 tokens is never calculated — it always shows empty.

`claudeCostUSD` first tries the models.dev cache, then falls back to the built-in Claude pricing map. That map had entries for Opus 4.5/4.6/4.7 but **no `claude-opus-4-8` entry**, so when the models.dev cache is missing or stale (e.g. right after a new model ships) the lookup returns `nil` and the cost renders empty.

## Fix
- Add a `claude-opus-4-8` entry to the built-in Claude pricing table, using the same rates as Opus 4.5/4.6/4.7 ($5/M input, $25/M output, $6.25/M cache write, $0.50/M cache read — matching current models.dev pricing for Opus 4.8).
- Add a regression test (`claude cost supports opus48`) mirroring the existing opus47 test.

## Testing
```
swift test --filter CostUsagePricingTests
```
All 24 tests pass, including the new `opus48` case.

### DIFF
--- a/Sources/CodexBarCore/Vendored/CostUsage/CostUsagePricing.swift
+++ b/Sources/CodexBarCore/Vendored/CostUsage/CostUsagePricing.swift
@@ -275,6 +275,16 @@ enum CostUsagePricing {
             outputCostPerTokenAboveThreshold: nil,
             cacheCreationInputCostPerTokenAboveThreshold: nil,
             cacheReadInputCostPerTokenAboveThreshold: nil),
+        "claude-opus-4-8": ClaudePricing(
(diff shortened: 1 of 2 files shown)


chhoumann:
Closing — opened by mistake. #1214 already covers this. Sorry for the noise!

PR #1220, opened 2026-05-30
Preserve Claude web usage on transient refresh auth failures

LeoLin990405:
## Summary
- Preserve the last good Claude Web usage snapshot when a manual/background refresh hits `ClaudeWebAPIFetcher.FetchError.unauthorized`.
- Clear the transient refresh error instead of replacing an otherwise usable Claude card with `Unauthorized`.
- Keep first-run behavior unchanged: if there is no prior Claude snapshot, the Unauthorized error still surfaces.

## Why
Claude Web can return intermittent 401/403 responses during refresh even after a successful usage fetch. Before this change, a later refresh could replace the visible Claude session/weekly usage with `Unauthorized. Your Claude session may have expired.` despite having a fresh prior snapshot.

## Tests
- Added `ClaudeWebRefreshResilienceTests` covering:
  - prior Claude Web snapshot + unauthorized refresh keeps the prior snapshot and hides the transient error
  - no prior Claude Web snapshot + unauthorized refresh still surfaces the error

## Validation
- `swift build --target CodexBarCore` passes locally.
- `.build/lint-tools/bin/swiftformat Sources Tests --lint` passes locally with `0/985 files require formatting`.
- Local `swift test --filter ClaudeWebRefreshResilienceTests` is blocked by the local CommandLineTools environment (`Testing` module / SwiftUI `@Entry` macro plugins are unavailable without full Xcode); CI should be the test signal.
- Verified manually with a local patched app: after a successful Claude Web snapshot, clicking Refresh while Claude Web API returned 403 kept the Claude session/weekly usage visible instead of showing Unauthorized.

## Scope
No provider ordering, settings UI, localization, auth storage, credential reads, token/cookie logging, or package/test target changes.

### DIFF
--- a/Sources/CodexBar/UsageStore+Refresh.swift
+++ b/Sources/CodexBar/UsageStore+Refresh.swift
@@ -300,9 +300,12 @@ extension UsageStore {
             let shouldSurface =
                 self.failureGates[provider]?
                     .shouldSurfaceError(onFailureWithPriorData: hadPriorData) ?? true
-            if provider == .claude,
-               hadPriorData,
-               Self.isClaudeWebSessionRefreshFailure(error)
+            let preservesClaudeWebSessionFailure =
+                provider == .claude &&
+                hadPriorData &&
+                Self.isClaudeWebSessionRefreshFailure(error)
+            if preservesClaudeWebSessionFailure,
+               !shouldSurface
             {
                 self.errors[provider] = nil
                 return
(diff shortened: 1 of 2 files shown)


LeoLin990405:
Addressed the Codex review feedback in `5f967485`: Claude Web unauthorized refresh failures now respect the existing failure gate. With prior data, the first failure is still treated as a transient refresh flake and hidden; once the gate decides to surface repeated failures, the Unauthorized error is shown while preserving the last good snapshot. The no-prior-data path still surfaces Unauthorized immediately.\n\nLocal validation:\n- `.build/lint-tools/bin/swiftformat Sources Tests --lint` ✅\n- `swift build --target CodexBarCore` ✅\n- `swift test --filter ClaudeWebRefreshResilienceTests` remains blocked locally by the CommandLineTools Swift Testing/SwiftUI macro environment, same as before; CI is the test signal.

PR #1224, opened 2026-05-30
fix(augment): parse updated auggie account status output

bcharleson:
## Summary

Fixes Augment usage failing with `Failed to parse auggie output: Could not extract credits from output` after the `auggie` CLI changed how it prints account status.

**Why this changed:** `auggie account status` no longer prints the legacy used/total breakdown. It now renders credits in a boxed layout like:

```
319,054 credits remaining                     Max Plan
450,000 credits / month
9 days remaining in this billing cycle (ends 6/9/2026)
```

CodexBar still expected the old format:

```
11,657 remaining · 953,170 / 964,827 credits used
```

Because parsing failed with `AuggieCLIError.parseError`, the CLI strategy did not fall back to browser cookies, so Augment appeared broken even when `auggie` and the browser session were healthy.

This PR:
- Parses the new `auggie account status` output (remaining + credits/month), while keeping legacy output support
- Falls back to browser cookie fetch when CLI parsing fails
- Restores `session` and `web_rpc_proxy_session` to Augment session cookie detection
- Persists imported cookies after session refresh and clears stale cookie cache on force refresh

## Test plan

- [x] `swift test --filter 'AuggieCLIProbeParseTests|AugmentCLIFetchStrategyFallbackTests'`
- [x] `swift run CodexBarCLI usage augment` returns Augment credits via CLI on a machine with authenticated `auggie`
- [x] Local CodexBar.app build shows Augment usage (71% left, 450,000 credits/month) instead of the parse error

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@
 - Settings: add search to the Providers pane so large provider lists can be filtered by name or id (#1184). Thanks @046081-dotcom!
 
 ### Fixed
-- Augment: parse the updated `auggie account status` output format, fall back to browser cookies when CLI parsing fails, and restore session cookie detection (#1224). Thanks @bcharleson!
 - Codex: cancel OpenAI WebKit dashboard refreshes promptly and avoid an immediate second background WebView retry after timeouts, reducing launch-time Web Content CPU spikes (#1217).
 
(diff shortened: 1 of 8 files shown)


bcharleson:
Merged to `main` on fork `bcharleson/codexbar` (commit 7d40fb47).

Issue #1211, opened 2026-05-29
tries to fetch status before connection was made


ghost:
after mac startup, it tries it fetch the status before the mac connected to the internet, it only load correctly after clicking on the codexapp icon

PR #1229, opened 2026-05-30
Retry failed startup usage refresh once

Yuxin-Qiao:
Summary:

- Retry the initial usage refresh once after a startup network/status failure.
- Keep the retry scoped to providers that actually run background refresh work.
- Cancel the pending startup retry during UsageStore cancellation.
- Keep the retry bounded to one delayed attempt.

Validation:

- ./Scripts/lint.sh lint
- swift test --filter UsageStoreStartupRetryTests
- swift test --filter UsageStoreManualTokenRefreshTests
- CI / lint-build-test
- CI / build-linux-cli (linux-x64, ubuntu-24.04)
- CI / build-linux-cli (linux-arm64, ubuntu-24.04-arm)
- GitGuardian Security Checks

Proof:

- Local lint and focused tests passed before push.
- CI is green on d6af342ffd0c8545ab3d50b102e553a4a6656a8c.
- Runtime app proof still to be added separately if maintainers require it.

### DIFF
--- a/Sources/CodexBar/UsageStore.swift
+++ b/Sources/CodexBar/UsageStore.swift
@@ -234,6 +234,9 @@ final class UsageStore {
     @ObservationIgnored var managedCodexAccountsForStorageOverride: [ManagedCodexAccount]?
     @ObservationIgnored private var pathDebugRefreshTask: Task<Void, Never>?
     @ObservationIgnored var codexPlanHistoryBackfillTask: Task<Void, Never>?
+    @ObservationIgnored private(set) var startupRetryTask: Task<Void, Never>?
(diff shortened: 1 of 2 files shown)


Yuxin-Qiao:
@clawsweeper re-review

steipete:
Thanks for the PR. Current `main` already has the narrower startup connectivity retry path in `UsageStore+StartupConnectivityRetry.swift` from #1211, with bounded retry classification and focused coverage.

This branch adds a parallel startup retry task that can retry non-connectivity provider errors, so I’m closing it as superseded. If there’s still a reproducible gap in the current retry helper, the best follow-up is a narrow patch against that existing path.

PR #1251, opened 2026-05-31
Avoid redundant storage refresh scheduling during provider-switcher m…

Yuxin-Qiao:
## Summary
Add a small performance guard to avoid redundant storage-refresh scheduling during frequent provider-switcher menu rebuilds.

## Change
In `refreshStorageFootprintsForOverview()`, return early when:
1. a storage refresh task is already in flight
2. the previous refresh is still within the automatic refresh interval

## Why
Switching providers can trigger repeated menu rebuilds. This keeps the overview-triggered storage refresh path cheap and reduces potential UI jank risk.

## Scope
- 1 file changed
- 8 lines added
- no intended behavior change
- If helpful, I can follow up with a focused regression test for this guard path.

### DIFF
--- a/Sources/CodexBar/UsageStore+ProviderStorage.swift
+++ b/Sources/CodexBar/UsageStore+ProviderStorage.swift
@@ -29,6 +29,14 @@ extension UsageStore {
     }
 
     func refreshStorageFootprintsForOverview() {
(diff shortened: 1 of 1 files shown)


Yuxin-Qiao:
Closing — no longer observing the jank in practice, so this guard is no longer needed. Thanks for the review on the diff.

PR #1254, opened 2026-05-31
Fix System language fallback for locale variants

Yuxin-Qiao:
## Summary
Fixes a first-launch localization issue where CodexBar could show English even when macOS is Chinese and app language is set to **System**.

## Root cause
System locale variants like `zh-Hans-CN` could miss existing resources such as `zh-Hans.lproj`, causing fallback to English.

## Changes
- Use `Bundle.preferredLocalizations(from:forPreferences:)` with `Locale.preferredLanguages` in System mode.
- Expand lproj candidates with a fallback chain:
  - `zh-Hans-CN -> zh-Hans -> zh`
  - normalize `_` to `-`
- Add regression coverage in `LocalizationBundleTests`.

## Validation
- `swift test --filter LocalizationBundleTests`
- 7/7 passed

### DIFF
--- a/Sources/CodexBar/Localization.swift
+++ b/Sources/CodexBar/Localization.swift
@@ -73,8 +73,10 @@ private func localizedBundle() -> Bundle {
         }
     } else {
         // System mode: follow macOS language preferences
-        if let preferred = resourceBundle.preferredLocalizations.first,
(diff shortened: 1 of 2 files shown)


Yuxin-Qiao:
Closing — the underlying issue has already been resolved, so this fix is no longer needed. Thanks for the review work on the diff.

PR #1257, opened 2026-06-01
Fix Alibaba token plan SEC token fallback

YanxinXue:
## Summary

- Add a user info JSON fallback for resolving Alibaba Token Plan `sec_token`
- Treat `PostonlyOrTokenError` / expired token payloads as `loginRequired`
- Cover the new SEC token fallback and token-expired error mapping in tests

## Why

Alibaba Token Plan requests can fail when the dashboard HTML preflight does not expose `sec_token`, even though the authenticated user info endpoint can still return it.

The API can also return expired-token payloads as `successResponse: false`. Those should prompt login/token refresh instead of surfacing as a generic API error.

## Proof

Tested the packaged app from PR head against a real Alibaba Token Plan session.

- Commit: `44bcfc07`
- App version: `0.32.2 (77)`
- Result: Alibaba Token Plan usage loaded successfully
- Evidence: 
截屏2026-06-01 12 35 07

The screenshot shows the Token Plan provider loading usage successfully from a real authenticated session after this change. Sensitive values are redacted:

- Cookies
- `secToken`
- Account identifiers
- Private request details

Additional runtime proof:

Captured from a real Alibaba Token Plan session using the packaged app from commit `44bcfc07`:

```text
Resolved Alibaba Token Plan sec_token from user info
Fetching Alibaba Token Plan usage
Alibaba Token Plan HTTP response status=200
```
Sensitive values were redacted, including cookies, secToken, account identifiers, and private request details.

## Validation

- `swift test --filter AlibabaTokenPlan`
- `swiftlint --strict --no-cache`

### DIFF
--- a/Sources/CodexBarCore/Providers/Alibaba/AlibabaTokenPlanUsageFetcher.swift
+++ b/Sources/CodexBarCore/Providers/Alibaba/AlibabaTokenPlanUsageFetcher.swift
@@ -26,6 +26,7 @@ public enum AlibabaTokenPlanUsageError: LocalizedError, Sendable, Equatable {
     }
 }
 
+// swiftlint:disable:next type_body_length
 public struct AlibabaTokenPlanUsageFetcher: Sendable {
     private static let log = CodexBarLog.logger("alibaba-token-plan")
     private static let gatewayBaseURLString = "https://bailian.console.aliyun.com"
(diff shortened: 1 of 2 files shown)


YanxinXue:
@clawsweeper re-review

YanxinXue:
@clawsweeper re-review

PR #1260, opened 2026-06-01
Fix provider switcher shortcuts

ProspectOre:
## Summary
Route provider switcher command-number shortcuts through `StatusItemMenu` instead of a custom event-tracking monitor.

## Context
Part of the UI hang/lag cleanup set: this removes a custom event-tracking shortcut monitor and uses the normal menu key-equivalent path instead.

## Validation
- `swift test --filter 'StatusMenuSwitcherClickTests|StatusMenuPersistentRefreshTests|StatusItemControllerShutdownTests'`
- `make check`
- `git diff --check`

## Runtime Proof
Built and launched this PR branch from `7d3e97f3990a3a60e8cafacaf6a4b77f2dd0b800` with the PR-built app isolated as the only running CodexBar status-item app.

Redacted live state monitor while the visible menu was operated manually:

```text
23:25:55 PDT selected=codex overview=0
23:26:31 PDT selected=claude overview=0   # after Cmd-3
23:26:35 PDT selected=claude overview=1   # after Cmd-1
23:26:42 PDT selected=cursor overview=0   # extra command-number check
```

No screenshots or raw AX menu dumps were posted because the menu can contain private account/cost data.

### DIFF
--- a/Sources/CodexBar/StatusItemController+Actions.swift
+++ b/Sources/CodexBar/StatusItemController+Actions.swift
@@ -77,6 +77,12 @@ extension StatusItemController: StatusItemMenuPersistentActionDelegate {
         }
     }
 
+    nonisolated func performProviderSelection(at index: Int) {
+        Task { @MainActor [weak self] in
+            self?.selectOpenProviderSwitcherSegment(at: index)
+        }
+    }
+
     @objc func refreshAugmentSession() {
         Task {
             await self.store.forceRefreshAugmentSession()
--- a/Sources/CodexBar/StatusItemController+Menu.swift
+++ b/Sources/CodexBar/StatusItemController+Menu.swift
@@ -131,7 +131,6 @@ extension StatusItemController {
             // Intentionally skip open-menu tracking when refresh is disabled (tests).
             // If refresh is re-enabled while this menu stays open, it will not be backfilled until next open.
             self.openMenus[ObjectIdentifier(menu)] = menu
-            self.installProviderSwitcherShortcutMonitorIfNeeded(for: menu)
             // Only schedule refresh after menu is registered as open - refreshNow is called async
             self.scheduleOpenMenuRefresh(for: menu)
         }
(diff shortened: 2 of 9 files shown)


ProspectOre:
@clawsweeper re-review

ProspectOre:
Closing this PR: live validation on macOS 26.5.1 (25F80) invalidated its premise.

## Retest protocol

Packaged release build of this change rebased onto current main (`cde92cfb`), merged icons, codex + claude + Overview, Apple Silicon. A polling state monitor recorded `selectedMenuProvider` / `mergedMenuLastSelectedWasOverview` at 0.5s while the menu was operated manually: open → ⌘3 → ⌘1 → ⌘2 → ← / → → Esc, with the dropdown visibly tracking and the system otherwise idle (`sample` confirmed no background scan activity and an idle main thread before the test).

Result: zero selection changes recorded. None of the shortcuts worked.

```text
12:49:50 selected=codex overview=0   # armed; no further transitions during the test
```

## What this means

1. `StatusItemMenu.performKeyEquivalent` does not receive key events while AppKit is tracking the menu on macOS 26.5. The persistent ⌘R/⌘,/⌘Q shortcuts keep working because they are real `NSMenuItem.keyEquivalent`s matched natively by the tracking session — not because of the override.
2. The `ProviderSwitcherShortcutEventMonitor` run-loop observer is therefore load-bearing for Command-number *and* arrow shortcut delivery during tracking, and since 463ec914 / 9778128b it is also the switcher's pointer-interaction fast path. Removing it (this PR) or narrowing it to pointer events only (retested separately on the same hardware) silently breaks the keyboard shortcuts.
3. Hidden `NSMenuItem`s do not participate in key-equivalent matching, so attaching ⌘1–⌘9 to invisible items is not a viable replacement either.

I could not reproduce this PR's original June 1 live proof on 26.5.1; on today's hardware retest the key-equivalent route never fired during tracking. Treat the monitor as the supported mechanism until AppKit offers a real key-equivalent hook for tracking sessions.

PR #1266, opened 2026-06-01
Fix MiniMax token plan usage display

XWind18:
## Summary
- Handle MiniMax token plan quota responses with `general` and `video` buckets.
- Preserve bucket-style display labels (`General`, `Video`) in the MiniMax menu card.
- Support boosted quota limits from `interval_boost_permill` / `weekly_boost_permill`.
- Treat Plus `general` weekly status-3 quota as unlimited, while hiding unavailable Plus `video` placeholders.
- Keep MiniMax subscription metadata merged into the displayed plan name.

## Validation
- `swift test --filter MiniMaxTokenPlanChangeTests`
- `swift test --filter MiniMaxMenuCardModelPlanTests`
- `make check`
- `CODEXBAR_SIGNING=adhoc ./Scripts/package_app.sh`

## Notes
- Full `swift test` currently fails in unrelated `ProviderStorageFootprintTests.swift:401` with `storageRefreshGeneration` expected `41`, actual `42`.
- The GitHub Actions failure shown earlier occurs while installing Swift 6.2.1 via swiftly, before project build/test starts.

image
image

### DIFF
--- a/Sources/CodexBar/MenuBarMetricWindowResolver.swift
+++ b/Sources/CodexBar/MenuBarMetricWindowResolver.swift
@@ -104,7 +104,7 @@ enum MenuBarMetricWindowResolver {
         {
             return primary.usedPercent >= secondary.usedPercent ? primary : secondary
         }
-        if provider == .cursor {
+        if provider == .cursor || provider == .minimax {
             return Self.mostConstrainedWindow(
                 primary: snapshot.primary,
                 secondary: snapshot.secondary,
--- a/Sources/CodexBar/MenuCardView+Costs.swift
+++ b/Sources/CodexBar/MenuCardView+Costs.swift
@@ -166,6 +166,15 @@ extension UsageMenuCardView.Model {
                 percentLine: nil)
         }
 
+        if provider == .minimax, cost.period == "MiniMax points balance" {
+            let balance = String(format: "%.0f", cost.used)
+            return ProviderCostSection(
+                title: L("Credits"),
+                percentUsed: nil,
+                spendLine: "\(L("Balance")): \(balance)",
+                percentLine: nil)
+        }
+
         if provider == .openai || provider == .claude, cost.limit <= 0 {
             let spend = UsageFormatter.currencyString(cost.used, currencyCode: cost.currencyCode)
             let periodLabel = Self.localizedPeriodLabel(cost.period ?? "Last 30 days")
--- a/Sources/CodexBar/MenuCardView+MiniMax.swift
+++ b/Sources/CodexBar/MenuCardView+MiniMax.swift
@@ -13,9 +13,6 @@ extension UsageMenuCardView.Model {
                 format: L("minimax_usage_amount_format"),
                 used.formatted(),
                 service.limit.formatted())
-            let usedLabel = String(
-                format: L("minimax_used_percent_format"),
-                String(format: "%.0f%%", displayPercent))
             let localizedName = Self.localizedMiniMaxServiceName(service.displayName)
             let title = if localizedName == L("minimax_service_text_generation"), textGenerationCount > 1 {
                 "\(L("minimax_service_text_generation")) · \(Self.displayWindowBadge(for: service.windowType))"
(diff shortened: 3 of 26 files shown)


Yuxin-Qiao:
I took a quick pass at the MiniMax compatibility blocker.

The red arm64 Linux check looks unrelated to this patch: it failed while Swiftly was downloading Swift 6.2.1 (`downloadSwiftToolchain` transport error), before source build/test.

The source-side blocker is the endpoint compatibility issue ClawSweeper called out: API-token fetches now try only `/v1/token_plan/remains`, while existing API-token setups may still need the old `/v1/api/openplatform/coding_plan/remains` endpoint.

I prepared a small helper commit here:

- Branch: https://github.com/Yuxin-Qiao/CodexBar/tree/help/pr1266-minimax-fallback
- Commit: https://github.com/Yuxin-Qiao/CodexBar/commit/abb52f9d

What it does:

- Keeps `/v1/token_plan/remains` as the primary API-token endpoint.
- Falls back to `region.apiRemainsURL` (`/v1/api/openplatform/coding_plan/remains`) when the new endpoint fails with compatible fallback cases.
- Adds focused tests for:
  - new token-plan endpoint success
  - new endpoint auth failure followed by legacy endpoint success
  - both endpoints failing auth

Validation:

```text
swift test --filter MiniMaxTokenPlanChangeTests
✔ Suite MiniMaxTokenPlanChangeTests passed after 0.007 seconds.
✔ Test run with 14 tests in 1 suite passed after 0.007 seconds.
```

I also checked the current merge conflict locally; it appears limited to release/version files (`CHANGELOG.md`, `appcast.xml`, `version.env`), not the MiniMax source files.

XWind18:
Thanks @Yuxin-Qiao for the MiniMax compatibility patch and test direction. I incorporated the API-token fallback approach here.

What changed:
- Keep `/v1/token_plan/remains` as the primary API-token endpoint.
- Fall back to the legacy `/v1/api/openplatform/coding_plan/remains` endpoint on compatible failures.
- Preserve the existing global-to-China retry behavior.
- Add/update regression coverage for the new endpoint, legacy fallback, both-endpoints-failing auth, and the updated global/China request order.

Validation:
- `swift test --filter MiniMaxTokenPlanChangeTests`
- `swift test --filter MiniMaxAPITokenFetchTests`
- `make check`
- `CODEXBAR_SIGNING=adhoc ./Scripts/package_app.sh`

Full `swift test` was previously still blocked only by the known unrelated `ProviderStorageFootprintTests.swift:401` generation mismatch.

coygeek:
I confirmed one adjacent acceptance case from a live MiniMax console session and from a focused local test against current `main`:

- Live MiniMax `/console/usage` had `5h limit Used 0%` and `Weekly limit Used 97%`.
- Current `MenuBarMetricWindowResolver` only applies `mostConstrainedWindow(...)` automatically for `.cursor`, so if MiniMax parses into primary=5h and secondary=weekly, the menu-bar automatic metric can still show the 0% 5h lane while the weekly lane is nearly exhausted.

The parser/token-plan direction in this PR looks aligned with the live console shape. The one small follow-up I would consider before merge is extending the automatic metric resolver to MiniMax too, with a regression case like:

```swift
let snapshot = UsageSnapshot(
    primary: RateWindow(usedPercent: 0, windowMinutes: 300, resetsAt: nil, resetDescription: nil),
    secondary: RateWindow(usedPercent: 97, windowMinutes: 7 * 24 * 60, resetsAt: nil, resetDescription: nil),
    updatedAt: Date())

#expect(MenuBarMetricWindowResolver.rateWindow(
    preference: .automatic,
    provider: .minimax,
    snapshot: snapshot,
    supportsAverage: false)?.usedPercent == 97)
```

I verified that focused case locally on current main with `swift test --filter MenuBarMetricWindowResolverTests` after applying the resolver change.

XWind18:
Addressed the MiniMax automatic menu-bar metric follow-up. MiniMax now uses the most constrained quota window for automatic metrics, so a near-exhausted weekly lane is surfaced over a low 5h lane.

Added coverage for the 5h 0% + weekly 97% case in `MenuBarMetricWindowResolverTests`.

Validation:
- `swift test --filter MenuBarMetricWindowResolverTests`
- `swift test --filter MiniMaxTokenPlanChangeTests`
- `swift test --filter MiniMaxAPITokenFetchTests`
- `make check`

XWind18:
I synced this branch with the latest main. The failing CI case was from the new #1286 StatusMenuOpenRefreshTests coverage, not from the MiniMax provider changes.

The test used short Task.sleep windows and could miss the pre-rebuild observation window on CI. I added a test-only stabilization that keeps the same coalescing coverage but uses a deterministic gate instead of real-time sleeps. MiniMax production behavior is unchanged.

LeoLin990405:
Just closed my own #1285 (narrow `modelRemains` filter to drop auxiliary services) in favor of this PR — your `shouldRenderWeeklyWindow` + most-constrained-window approach is the right level of fix and covers the same user-visible symptom (menu surfacing `video 0/3` on a coding-plan account that's idle in the current window) plus more.

If maintainers want a second pair of eyes on the boosted-quota-percent commit (1b7dfaa) once CI is green, I can do a focused review against a local checkout of this branch. Real-behavior tested on a `sk-cp-…` account that returns `general(0/0)` + `video(3/3)` in the API response — same scenario from screenshot in the PR description.

Thanks for the thorough work here.

Yuxin-Qiao:
I checked this against the current MiniMax `/console/usage` page shape from the latest screenshots. This PR already covers the main token-plan parsing/display issue and CI is green; one small compatibility follow-up may still help now that the visible console has moved away from the legacy Coding Plan page.

I prepared a helper commit on top of this PR's current head (`1b7dfaa`):

- Branch: https://github.com/Yuxin-Qiao/CodexBar/tree/help/pr1266-minimax-web-remains-first
- Commit: https://github.com/Yuxin-Qiao/CodexBar/commit/02ed64342d40c67149b86beb7cc5bff1b5bd8a92

What it changes:

- Web/cookie MiniMax usage now tries `/v1/api/openplatform/coding_plan/remains` first.
- The legacy `/user-center/payment/coding-plan` HTML path remains as fallback.
- Auth failures and cancellation still preserve the existing behavior.
- `docs/minimax.md` documents the request order.

Validation on the helper worktree:

```text
swift test --filter MiniMax
✔ Test run with 89 tests in 14 suites passed after 0.053 seconds.
```

This is not a blocker if maintainers are comfortable with the current HTML-first fallback, but it makes the web-session path better aligned with the new MiniMax console/usage surface while keeping the old page fallback.

Yuxin-Qiao:
Follow-up update on the helper branch: I pushed `6ea9a16` to https://github.com/Yuxin-Qiao/CodexBar/tree/help/pr1266-minimax-web-remains-first.

This adds support for the current MiniMax console usage page data path:

- Fetches `/backend/account/token_plan/usage_summary` from the MiniMax `www` backend host before falling back to the older `/account/amount` billing-history endpoint.
- Parses `daily_token_usage`, `total_days`, and `most_active_day` into the existing MiniMax inline usage dashboard model.
- Preserves the existing quota fetch flow from `/v1/api/openplatform/coding_plan/remains` and the legacy Coding Plan HTML fallback.
- Sends `x-group-id` for web-session billing/activity requests when CodexBar has it from browser storage.

Local validation in this managed Codex environment: `git diff --check` passes. I could not complete `swift test --filter MiniMax` here because SwiftPM is blocked by the local sandbox (`sandbox-exec: sandbox_apply: Operation not permitted`), so this should still get a normal CI/focused test pass before merging.

XWind18:
Thanks @Yuxin-Qiao for the additional MiniMax console compatibility investigation.

My main goal with this PR is to restore MiniMax usage data first, since the recent MiniMax token-plan changes already broke the old parsing/display path. The current branch keeps the fix focused on token-plan parsing/display, API-token fallback, and the automatic metric behavior, and CI is green now.

The newer `/console/usage` web-session path support looks useful, but it expands the auth/web-session surface beyond the original fix. Since I’m not a maintainer, I’d prefer not to broaden this PR further unless maintainers want that included before merge.

If maintainers prefer to include the web-session compatibility follow-up in #1266, I’m happy to take a look and incorporate the minimal part needed. Otherwise, I think it may be better as a follow-up PR after this fix lands.

Issue #1169, opened 2026-05-26
macOS 26.5 / CodexBar 0.29.1: ControlCenter still turns CodexBar into an ephemeral status item even though “Allow in the Menu Bar” is enabled


kieranmcshane:
## Summary

CodexBar 0.29.1 still fails to show a visible menu bar item on macOS 26.5 on my machine, even though:

- the app launches normally
- the process stays alive
- `System Settings > Menu Bar > Allow in the Menu Bar > CodexBar` is enabled
- toggling that setting off/on does not restore the icon

The strongest signal is that `ControlCenter` creates a menu-bar displayable for CodexBar and then immediately converts it into an `ephemaral` item with `.ephemeral` positioning.

This looks different from the simple "Allow in the Menu Bar is off" case.

## Environment

- CodexBar: `0.29.1` (`69`)
- macOS: `26.5` (`25F71`)
- Hardware: `MacBook Air` (`Mac16,12`), Apple `M4`, `16 GB` RAM
- Install path: `/Applications/CodexBar.app`

## Expected behavior

CodexBar should create a visible menu bar item after launch and keep it visible while the app is running.

## Actual behavior

CodexBar launches and stays running, but no visible menu bar item appears.

## Confirmed state

`System Settings > Menu Bar > Allow in the Menu Bar > CodexBar` is ON.

I rechecked it directly in System Settings and also toggled it OFF -> ON again. The checkbox ends up enabled, but the icon still does not appear.

## Reproduction

On this machine, the problem is reproducible with a normal launch:

1. Launch CodexBar from `/Applications/CodexBar.app`
2. Verify that `Allow in the Menu Bar > CodexBar` is enabled
3. Observe that no visible CodexBar item appears in the menu bar
4. Quit and relaunch CodexBar
5. Restart `ControlCenter` / `SystemUIServer`
6. Observe that the problem persists

## What I tried

- Quit and relaunch CodexBar
- Force-kill and relaunch CodexBar
- Toggle `System Settings > Menu Bar > Allow in the Menu Bar > CodexBar` OFF -> ON
- Restart `ControlCenter`
- Restart `SystemUIServer`
- Relaunch after Control Center rebuild
- Reduce CodexBar menu-bar text width / compact display options
- Temporarily disable other third-party “Allow in the Menu Bar” apps to rule out simple crowding
- Verify that CodexBar is still running after launch

None of these restored a visible status item.

## Evidence: process is alive

The main process remains running from:

```text
[...]

Issue #1109, opened 2026-05-22
Menu bar icon does not appear on macOS 26.5, CLI works (v0.28.0)


lexleizou:
### Description

CodexBar starts successfully and the CLI works, but no CodexBar item is visible in the macOS menu bar. This still reproduces after a clean reinstall and after toggling the macOS menu bar permission for CodexBar.

This looks related to earlier menu-bar/status-item issues such as #890, #988, and #1077, but in this case `System Settings > Menu Bar > Allow in the Menu Bar` is already enabled and has been toggled off/on without restoring the icon.

### Environment

- CodexBar: 0.28.0 (66)
- macOS: 26.5 (25F71)
- Install method: official ZIP from the appcast / GitHub release, copied to `/Applications/CodexBar.app`
- CLI path: `/opt/homebrew/bin/codexbar`

### What I tried

1. Quit CodexBar and relaunched it.
2. Toggled `System Settings > Menu Bar > Allow in the Menu Bar > CodexBar` off and back on.
3. Quit/disabled Ice to rule out menu-bar hiding interference.
4. Restarted `ControlCenter` and `SystemUIServer`.
5. Fully removed the app/config/cache-related files I could access, then reinstalled CodexBar from the official ZIP.

The app process runs, but the menu bar item is still not visible.

### CLI diagnostics

The CLI appears to be functional:

```console
$ codexbar --version
CodexBar

$ codexbar config validate --format json --pretty
[

]
```

Provider usage commands also return data successfully, so this does not appear to be a general backend/config failure.

### Additional observations

Before the clean reinstall, Accessibility inspection briefly showed a CodexBar status item present but effectively positioned at the far right edge of the menu bar (`x=1437`, `w=63` on an approximately `1470`-wide logical screen). After the clean reinstall, CodexBar still launches, but the visible status item does not appear to materialize.

### Expected behavior

CodexBar should show a visible status item in the macOS menu bar after launch.

### Actual behavior

CodexBar launches and CLI commands work, but no menu bar icon/item is visible.

### Related issues / PRs

- #890: similar symptom, but there the cause was the macOS menu bar allow-list setting.
- #988: status item visible in prefs but not actually materialized.
- #1077: macOS 26.5 display/status-item materialization issue.
- #849: previous macOS 26.4 menu bar icon restore fix.

andreineculau:
Same on 0.29.0

steipete:
Folks, I’d love help - I can’t reproduce this.

[...]

Issue #861, opened 2026-05-07
v0.24: Menu bar icon does not appear (process runs but icon is invisible)


h3x89:
## Summary

After upgrading to 0.24 (build 59), the CodexBar menu bar icon does not appear at all. The process starts successfully but the icon is never visible in the menu bar. Downgrading to 0.23 (build 58) immediately fixes the issue.

## Environment

- macOS 15.4 (Darwin 25.4.0)
- MacBook Pro M1/M2
- CodexBar 0.24 (build 59) — broken
- CodexBar 0.23 (build 58) — works correctly

## Steps to Reproduce

1. Install CodexBar 0.24
2. Launch the app (or let it launch at login)
3. Process appears in `pgrep -x CodexBar` — it IS running
4. No icon appears in the menu bar

## Observed Behavior

- Process running: ✅ (verified via `ps`, PID confirmed)
- CPU usage: normal (~1% after startup spike of ~20%)
- Menu bar icon: ❌ not visible anywhere
- `NSStatusItem Preferred Position Item-0 = 585` is set in defaults, but icon never renders

## Expected Behavior

Menu bar icon should appear as it does in 0.23.

## Diagnosis Notes

- stdout/stderr are redirected to `/dev/null` — no console output available
- `log show --process CodexBar` returns nothing (no unified log entries)
- `NSStatusItem VisibleCC Item-1` through `Item-26` all set to `0` in preferences
- No crash reports generated
- This is **different from issue #858** (no 100% CPU, app is not freezing — icon simply never appears)

## Workaround

Downgrade to 0.23: https://github.com/steipete/CodexBar/releases/download/v0.23/CodexBar-0.23.zip

marijnbent:
Same here.

jrdinatl:
same here

rafachavantes:
same here

steipete:
Fixed on main in 35cd513 by removing the explicit codexbar-* NSStatusItem autosave names so macOS cannot relaunch CodexBar into a persisted hidden menu-extra state. Also covered by StatusMenuTests.

PR #1306, opened 2026-06-05
Repair hidden status item defaults repeatedly

Thinkscape:
## Why?

CodexBar randomly disappearing after updates, external screen disconnections, or when interacting with Bartender, is a major pain. Restarting the app would be a logical recovery method, but it currently doesn't work b/c the repair doesn't happen on repair. This PR fixes it.

## Summary
- remove the one-shot gate from hidden macOS status-item visibility defaults repair
- keep the existing repair marker as historical state, but continue scanning current defaults on each startup
- preserve the CodexBar-owned key filter so unrelated apps' `NSStatusItem VisibleCC ...` keys and true values are left intact

## Related
- #1109
- #1169
- #861

## Tests

### DIFF
--- a/Sources/CodexBar/MenuBarStatusItemDefaultsRepair.swift
+++ b/Sources/CodexBar/MenuBarStatusItemDefaultsRepair.swift
@@ -6,8 +6,6 @@ enum MenuBarStatusItemDefaultsRepair {
     private static let legacyAutosavePrefix = "codexbar-"
 
     static func repairHiddenVisibilityDefaultsIfNeeded(defaults: UserDefaults) -> [String] {
-        guard !defaults.bool(forKey: self.didRepairKey) else { return [] }
-
         let repairedKeys = defaults.dictionaryRepresentation().keys
             .filter { key in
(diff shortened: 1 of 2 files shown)


hhh2210:
Opened a small stacked follow-up for the ClawSweeper P1 blocker here: https://github.com/Thinkscape/CodexBar/pull/1

It keeps repeated cleanup for stale `NSStatusItem VisibleCC Item-* = false` keys, but preserves user-created `codexbar-* = false` visibility defaults after the repair marker already exists.

Validation run locally:
- `make check`
- `swift test`
- independent reviewer pass: no findings

steipete:
Thanks for the focused patch and tests. I am closing this because repeated startup cleanup is not safe: it would clear intentional hidden status-item choices made by users or menu-bar managers on every launch.

The later evidence on #1169 also shows the one-time cleanup already ran successfully while ControlCenter still classified the item as ephemeral; a minimal AppKit status-item probe under the same bundle identifier reproduced the same behavior. That points to a ControlCenter/app-identity state problem rather than stale visibility defaults, so repeating this cleanup would add a regression without fixing the remaining root cause.

PR #1320, opened 2026-06-05
Fix Copilot unlimited chat quota display

soumikbhatta:
## Summary
- Decode Copilot quota snapshots that explicitly report `unlimited`.
- Keep unlimited zero-entitlement Copilot chat quota snapshots instead of treating them as placeholders.
- Add regression coverage for an unlimited chat-like quota without `percent_remaining`.

## Why
GitHub Copilot can report chat as an unlimited quota snapshot with `entitlement=0` and `remaining=0`. CodexBar's placeholder filter treated that shape as unusable, so the Copilot menu only showed Premium and dropped the unlimited Chat lane.

This keeps the existing placeholder behavior for ordinary zero-entitlement snapshots. The exception only applies when the payload explicitly says `unlimited: true`.

## Validation
- `swift test --filter CopilotUsageModelsTests`
- `swift test --filter CopilotUsageFetcherTests`
- `./Scripts/lint.sh lint`
- Live CLI proof against the same Copilot account: installed CodexBar `0.32.4` returns `secondary: null`; this branch returns `secondary.usedPercent: 0`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.32.6 — Unreleased
 
 ### Fixed
+- Copilot: keep explicitly unlimited chat quotas visible instead of dropping their zero-entitlement payload as unavailable (#1320). Thanks @soumikbhatta!
 - Doubao: confirm zero-remaining HTTP 200 request limits before falling back, preserving genuine exhaustion and avoiding false 100% usage (#1383). Thanks @LeoLin990405 and @foobra!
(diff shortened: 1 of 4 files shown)


steipete:
Landed as `a6bb67720277aade0f336a738e52ebe599c117b1`.

Verification:
- `swift test --filter CopilotUsage` - 29 tests passed.
- `make check` - SwiftFormat and strict SwiftLint passed.
- Full branch autoreview against `origin/main` - clean, confidence 0.96.
- GitHub CI - macOS `lint-build-test`, Linux x64, Linux arm64, and GitGuardian all passed.

The final implementation normalizes explicitly unlimited Copilot chat quota even when the response also supplies `percent_remaining: 0`, with model and fetcher regressions covering that precedence.

Thanks @soumikbhatta!

Issue #1311, opened 2026-06-05
High CPU usage


RoversX:
Hello, thanks for making this. After many recent updates to the codex bar, the CPU usage has become pretty high. I have disabled the extra web and other similar functions, but it is still pretty high. When I scroll to the menu bar items, they also become very laggy after the recent updates. I'm not sure if this is caused by the graph. Are there any other way to decrease the cpu usage? Thank you.

Stable Channel, 0.32.4(79)

<img width="566" height="150" alt="Image" src="https://github.com/user-attachments/assets/7860c071-1812-479f-b476-1de658f276fa" />

RoversX:
It's also very laggy when clicking any buttons. Codexbar used to be very good.

dengshu2:
Adding a process sample tying the sustained CPU to a specific hot path, since the thread asked for one.

**Environment**
- CodexBar 0.32.4 (79), Stable channel
- macOS 26.5.1 (25F80), Apple Silicon
- No popover/menu open — idle in the menu bar
- Relevant config: `selectedMenuProvider = codex`, `tokenCostUsageEnabled = 1`, `launchAtLogin = 1`

**Observed**
Single `CodexBar` main process sitting at a sustained **~8–14% CPU while idle** (no UI open). Cumulative CPU time ≈ 60s over ~12 min uptime → ~8.5% average, with `sample` snapshots catching 13.5% spikes. Two `sample` runs taken ~5h apart show the same dominant stack.

**Hot path — main thread**
The CPU is overwhelmingly the closed-menu rebuild loop tearing down and re-evaluating the SwiftUI graph. Top of the tree (446 samples in a 5s run):

```
StatusItemController.rebuildClosedMenuIfNeeded(_:)
  → closure #1 in rebuildClosedMenuIfNeeded(_:)
    → SwiftUI graph re-evaluation:  AGGraphGetInputValue / AGGraphGetValue  (AttributeGraph)
```

and in parallel, repeated full teardown of the hosting view:

```
NSHostingView.__deallocating_deinit
  → GraphHost.invalidate()
    → AGGraphInvalidate
      → AG::Graph::invalidate_subgraphs()
```

So even with the menu closed, the status item's SwiftUI view tree is being rebuilt/invalidated and the AttributeGraph re-evaluated on a recurring tick — consistent with the "stale closed-menu rebuild loop" root-caused in #1347.

[...]

PR #1322, opened 2026-06-05
Memoize models.dev catalog to fix cost-scan CPU spike

turbothad:
## Summary

Fixes a CPU spike / menu freeze during token-cost usage refreshes (reported in #1311).

When a cost lookup runs **without** a pre-resolved catalog, `CostUsagePricing.modelsDevLookup` falls back to `ModelsDevPricingPipeline.lookup` → `ModelsDevCache.load`, which reads and JSON-decodes the full **~800 KB** `models.dev` catalog from disk. That fallback is hit **once per usage row** (`codexModeSplitMaps` / Claude scan loops call `codexCostUSD` / `claudeCostUSD` per row). Whenever the top-level catalog load returns nil for a window — cold start, the 24 h TTL refresh writing the cache mid-scan, or a decode hiccup — the entire scan of a large `~/.codex` history re-decodes the catalog hundreds of thousands of times.

A macOS CPU resource report captured this precisely: **90 s of CPU at 95%** over a 94 s window, heaviest stack:

```
UsageStore.refreshTokenUsage
 └ CostUsageFetcher.loadTokenSnapshot
    └ CostUsageScanner.loadCodexDaily
       └ scanCodexFile → rescanCodexFile → codexModeSplitMaps
          └ CostUsagePricing.codexCostUSD
             └ ModelsDevCache.load
                └ JSONDecoder.decode   ← full ~800 KB catalog, per row
```

Because `refreshTokenUsage` runs when the menu is opened, this also presents as the menu **freezing on click**.

## Fix

Memoize the **load outcome** inside `ModelsDevCache`, keyed by the cache file's **path + size + modification date**:

- Repeated `load`s of an unchanged file reuse the in-memory result instead of re-reading/decoding, so the per-row fallback collapses to a stat + dictionary lookup.
- The whole outcome is cached, **including failures** — a corrupt or wrong-version cache is read and decode-attempted exactly as expensively as a valid one (a wrong-version file is fully decoded and only then rejected), so caching only successes would leave the per-row storm in place whenever the cache is unreadable.
- `isStale` still depends on `now`, so it is recomputed on every call; only the read+decode outcome is memoized.
- `save(...)` drops the memo entry; the size/mtime key also self-invalidates when the file changes underneath.
- Thread-safe via `NSLock` (cross-platform; the vendored CostUsage code also builds for Linux), mirroring the existing `PathEnvironment` / `ProviderStorageFootprint` cache helpers.

This does not change the threading of `modelsDevCatalog` through the scanner (which already exists); it just makes the nil-catalog fallback cheap so a transient nil can no longer turn a full history scan into a CPU storm.

The codex parser hash (which covers files under `Vendored/CostUsage`) is regenerated; this also invalidates stale cost-usage caches so they rebuild on the fixed path.

## Behavior proof (before/after)

Micro-benchmark of the hot path — repeated `ModelsDevCache.load` against a synthetic **2.1 MB** catalog (larger than the real ~788 KB, so conservative), 1000 iterations, debug build, same machine, `main` vs this branch:

| Cache state | `main` (unfixed) | this branch | speedup |
|---|---|---|---|
| valid | **42.54 ms/load** | **0.108 ms/load** | ~394× |
| invalid-version | **42.69 ms/load** | **0.107 ms/load** | ~399× |

The `invalid-version` row demonstrates the failed-load path: on `main` it costs the same ~43 ms/load as a valid decode (full decode, then version rejection); with the fix it is memoized to ~0.1 ms/load. At hundreds of thousands of rows per refresh this is the difference between a multi-second main-thread stall and a single decode. (Benchmark was run with a temporary, uncommitted test; happy to share the harness.)

## Tests

Added to `ModelsDevPricingTests`:

- `serves decoded catalog from memo while the file is unchanged` — corrupts the cache bytes while preserving size + mtime; asserts the load still returns the originally decoded artifact (no re-decode).
- `serves a failed load from memo while the file is unchanged` — primes a failure, then restores valid bytes of identical size + mtime; asserts the cached failure is still returned (the now-valid file is not read/decoded again).
- `saving a new catalog invalidates the memo` — asserts a subsequent load reflects freshly written data.

## Commands executed

```
swift build                                       # Build complete
swift test --filter 'ModelsDevPricingTests|CostUsage'  # 185 passed
make check                                        # swiftformat --lint + swiftlint --strict → 0 violations
./Scripts/regenerate-codex-parser-hash.sh
```

No UI changes, so no screenshots.

Refs #1311


turbothad:
Thanks for the review — both P1s addressed in 1f9f6e7:

**P1 (failed-load fallback):** The memo now caches the full load *outcome* (decoded artifact **or** error), not just successes. An unchanged corrupt / wrong-version cache is now read at most once instead of per row. Added a regression test (`serves a failed load from memo while the file is unchanged`) that primes a failure then swaps in valid bytes of identical size+mtime and asserts the cached failure is still served (proving no re-read).

**P1 (real behavior proof):** Added before/after numbers from a hot-path micro-benchmark (repeated `ModelsDevCache.load`, 2.1 MB catalog, 1000 iterations, debug, `main` vs branch):

| Cache state | `main` | branch | speedup |
|---|---|---|---|
| valid | 42.54 ms/load | 0.108 ms/load | ~394× |
| invalid-version | 42.69 ms/load | 0.107 ms/load | ~399× |

The `invalid-version` row is the failed-load path you flagged — same ~43 ms/load as a valid decode on `main`, memoized to ~0.1 ms/load with the fix. Full details in the updated PR description.

`swift test --filter 'ModelsDevPricingTests|CostUsage'` → 185 passed; `make check` → 0 violations.

@clawsweeper re-review

steipete:
## Verification

Local:
- Replayed the two contributor commits onto current `main` because the PR head was 21 commits behind.
- Added the maintainer changelog entry after replay.
- `Scripts/regenerate-codex-parser-hash.sh` — updated/generated hash on current tree.
- `Scripts/regenerate-codex-parser-hash.sh check` — passed; hash current at `7049d0743a7cd98f`.
- `swift test --filter 'ModelsDevPricingTests|CostUsage'` — passed (185 tests).
- `make check` — passed; parser hash current, SwiftFormat lint mode clean, SwiftLint found 0 violations.
- `git diff --check origin/main..HEAD` — passed.

Review:
- Codex Auto Review on the replayed branch — clean, no accepted/actionable findings.

CI:
- Approved the fork PR CI run after GitHub marked it `action_required`.
- GitGuardian Security Checks — passed.
- CI / build-linux-cli (linux-arm64, ubuntu-24.04-arm) — passed.
- CI / build-linux-cli (linux-x64, ubuntu-24.04) — passed.
- CI / lint-build-test — passed.

Landed:
- PR head: `f9d54d1f5732fe97b7b222c2feedb69a2b675ea6`
- Squash merge commit: `88b00e557bd2b19aeefe5a8fcd0961b3b84beb85`
- #1311 remains open intentionally; this PR references it but does not claim to close the broader high-CPU report.

Verification images:
- Not applicable: this was a deterministic cost-cache performance/memoization fix covered by regression and cost-scan tests, not a visual UI rendering change needing screenshot proof.

Issue #1321, opened 2026-06-05
can someone fix the performance?


guocity:
can someone fix the performance? popup menu its really lagging, please fix it!!!

mattdaniell:
+1 to this, when I open codexbar it freezes and I am unable to do anything on mac until it resolves.
I suspect its the multiple codex accounts, or windsurf / opencode.

AmirAjaj:
I downgraded to 0.31.0, which runs pretty stable for me with no performance regressions. I advice you to do the same if you are also experiencing the laggy behaviour.  The downgrade fixed it for me and afaik there were no meaningful feature additions in 0.32.0

kcahrens:
+1 — Very laggy hover + slow provider switching / combined page
Hover interactions feel extremely laggy, especially when:

Hovering over providers
Switching between different providers
Using the combined page

The UI response to hover is noticeably delayed and makes the app feel sluggish.
Video showing the issue:
https://github.com/user-attachments/assets/708d6cc2-c176-435a-a3f1-4355b32d331e
Environment

CodexBar: 0.32.4 (79) — built Jun 2, 2026
macOS: 26.3.1 (25D2128)

This is on top of the general performance complaints already in this thread. Happy to test any fixes or provide more details.

markwylde:
> I downgraded to 0.31.0, which runs pretty stable for me with no performance regressions. I advice you to do the same if you are also experiencing the laggy behaviour. The downgrade fixed it for me and afaik there were no meaningful feature additions in 0.32.0

Thanks, can confirm downgrading to 0.31.0 fixes the performance issues.

Yuxin-Qiao:
I dug into this (the "0.31.0 is fine, 0.32.x lags, downgrading fixes it" reports were the key signal) and traced two concrete main-thread costs that were added/grew in the 0.32 series and still exist on `main`. Opened two focused PRs:

**#1351 — menu readiness signature cost**
`observeStoreChanges` recomputes a "readiness signature" on *every* store mutation. It serializes every enabled provider's token snapshot + 30-day daily breakdown + the OpenAI dashboard per-service breakdown into a string (with `String(format: "%.8f", …)` per value), so the cost scales with provider/account count and fires on every background refresh tick.
- Skip computing it entirely when no menu is open (the result is only consulted while a menu is open).
[...]

PR #1351, opened 2026-06-07
perf: cut menu readiness signature cost on store changes

Yuxin-Qiao:
## Summary

Targets the popup-menu lag reported in #1321. Multiple users confirmed the lag appeared in the 0.32 series and that downgrading to 0.31.0 restores responsiveness, so this focuses on a 0.32-era main-thread regression that still exists on `main`.

The menu *readiness signature* (`menuAdjunctReadinessSignature()`) serializes every enabled provider's token snapshot plus its 30-day daily breakdown (and the OpenAI dashboard per-service breakdown) into a string on **every** store mutation, via `observeStoreChanges`. This runs on the main actor on each background refresh tick and scales with provider/account count.

Three changes:

- **Skip the signature entirely when no menu is open.** `observeStoreChanges` always computed `didMenuAdjunctReadinessChange()` to pass as `refreshOpenMenus`, but `invalidateMenus` only consults that argument while a menu is open. When the menu is closed (the common case during background refresh) the result was computed and discarded.
- **Re-anchor the baseline when a root menu opens and actually shows current data (correctness for the skip).** Because the baseline is no longer recomputed on every closed-menu store change, it can drift from live store data. Without re-anchoring, a menu reopened from new data followed by an open-menu change that *reverts* to the previous baseline value would compare equal and skip the rebuild, leaving stale content visible. The baseline is now resynced when a **root** menu opens **and** `!menuNeedsRefresh(menu)` after `refreshMenuForOpenIfNeeded` — i.e. the menu was rebuilt or was already fresh for the current `menuContentVersion`. If an in-flight provider refresh causes `refreshMenuForOpenIfNeeded` to preserve stale content, resync is skipped so the refresh-completion update is not masked. Nested submenu opens intentionally do **not** resync, so a pending parent refresh can't be masked. Thanks to @clawsweeper for catching both the original drift case and the in-flight stale-preservation case.
- **Make the signature cheaper.** Replace `String(format: "%.8f", …)` (a hot per-value cost run for every daily/service value of every provider) with the raw `Double` bit pattern. The signature is only ever compared for equality against the previous signature, so the bit pattern is both exact (no rounding collisions) and far cheaper. This also helps while a menu *is* open, where the signature is recomputed on each tick.

Net effect: less main-thread work both at idle (closed) and during interaction (open), which is the responsiveness path that regressed in 0.32.

## Proof

**Real app — freshly built ad-hoc bundle of this branch, 7 providers enabled** (codex, claude, cursor, gemini, antigravity, minimax, deepseek), `refreshFrequency = oneMinute`, menu **closed**. A 65 s `sample` of the running `CodexBar` process — spanning at least one background refresh tick — shows the menu signature/rebuild path never lands on the main thread:

```
$ sample CodexBar 65
# grep across ALL threads:
menuAdjunctReadinessSignature | formatDoubleForSignature | populateMenu
  | updateMenuContentPreservingSwitcher | sizeThatFits  -> 0 hits
observeStoreChanges | invalidateMenus                   -> 0 hits

main thread: 54126/54126 samples parked in App.main -> __CFRunLoopRun -> mach_msg
idle CPU: ~0.3%
```

So with a multi-provider setup and the menu closed, the readiness/rebuild work is no longer a sustained main-thread cost. (Honest caveat: `sample` is 1 ms-coarse and a single per-tick signature call is ~tens of µs, so this demonstrates the path is not a sustained closed-path cost rather than a precise per-call delta.)

**Real app — menu *open* with 7 providers enabled** (same ad-hoc bundle: codex, claude, cursor, gemini, antigravity, minimax, deepseek). Opened via `osascript` + System Events (`click menu bar item "CodexBar"`):

```
click-to-open (11 visible items): ~190 ms real
20 s sample while menu held open:
  menuAdjunctReadinessSignature | populateMenu | observeStoreChanges
    | invalidateMenus | updateMenuContentPreservingSwitcher | sizeThatFits  -> 0 hits
  main thread: 17133/17138 samples parked in mach_msg
```

So the open-menu interaction path is responsive on first open, and there is no sustained signature/rebuild churn while the menu stays open across background refresh ticks.

**Runtime cost of the work skipped when closed**, measured on the real `menuAdjunctReadinessSignature()` with a realistic multi-provider store (4 providers, each with a 30-day token daily breakdown, 5000 calls, warm):

```
menuAdjunctReadinessSignature(): 0.391s / 5000 = ~78 us/call
```

That full ~78 µs call ran on the main actor on **every** store mutation pre-fix, even with all menus closed (i.e. on each background refresh tick). It now runs only when a menu is actually open, and the cost grows with enabled providers/accounts.

**Why the per-call cost itself also dropped — signature double-formatting micro-benchmark** (`swiftc -O`, 60 doubles/iter × 200k iters):

```
old(%.8f):       10.799s
new(bitPattern):  1.166s
speedup: 9.3x
```

**Regression tests are negative-validated.**

`reopening root menu resyncs readiness baseline so reverted store data still refreshes` — fails when resync is disabled, passes with it:

```
# resync disabled:
✘ Expectation failed: controller.didMenuAdjunctReadinessChange()
# resync enabled:
✔ Test "reopening root menu resyncs readiness baseline …" passed
```

`root open during in flight refresh preserves stale content and does not resync baseline` — fails when baseline resync is unconditional on root open, passes when gated on `!menuNeedsRefresh(menu)`:

```
# unconditional resync on root open:
✘ Expectation failed: controller.didMenuAdjunctReadinessChange()
# gated resync (only when menu is fresh):
✔ Test "root open during in flight refresh …" passed
```

## Test plan

- [x] `swift build`
- [x] `make check` (SwiftFormat + SwiftLint, 0 violations, 1018 files)
- [x] `swift test --filter StatusMenuTests` — includes existing open-menu refresh coverage plus both baseline-resync regression tests (closed reopen revert + in-flight stale preservation)
- [x] Negative validation of both regression tests (each fails without its respective fix)
- [x] Runtime measurement of the skipped signature call + the formatting speedup (see Proof)
- [x] Real-app 65 s `sample`, 7 providers, menu closed — no main-thread signature/rebuild churn (see Proof)
- [x] Menu-*open* responsiveness with 7 enabled providers (real app): ~190 ms click-to-open, 11 items; 20 s open-menu `sample` shows 0 signature/rebuild hot-path hits, main thread parked (see Proof). Maintainer/reporter multi-account spot-check still welcome on merge but is not author-blocking (clawsweeper: ready for maintainer look).

## Notes / follow-ups (not in this PR)

- Chart submenus still `removeAllItems` + rebuild their `NSHostingView` when underlying data changes; gating that on a per-provider content fingerprint is a possible follow-up (overlaps active menu-card-height work).
- The multi-account "freeze on open" reports appear largely mitigated on current `main` by the recent Codex token-scan budget commits.

### DIFF
--- a/Sources/CodexBar/StatusItemController+Menu.swift
+++ b/Sources/CodexBar/StatusItemController+Menu.swift
@@ -85,6 +85,11 @@ extension StatusItemController {
         self.cancelDeferredMenuInteractionRefreshTask()
         self.cancelClosedMenuRebuild(menu)
 
+        // Track whether this is the root menu opening (no menus were open). Only the root open rebuilds
+        // all content from current data, so the readiness baseline is re-anchored only here — re-anchoring
+        // on a nested submenu open could mask a pending refresh for the already-open parent menu.
+        let menuTrackingWasIdle = self.openMenus.isEmpty
+
         if self.isHostedSubviewMenu(menu) {
             self.hydrateHostedSubviewMenuIfNeeded(menu)
             self.refreshHostedSubviewHeights(in: menu)
@@ -95,6 +100,9 @@ extension StatusItemController {
                 // Intentionally skip open-menu tracking when refresh is disabled (tests).
                 // If refresh is re-enabled while this menu stays open, it will not be backfilled until next open.
                 self.openMenus[ObjectIdentifier(menu)] = menu
+                if menuTrackingWasIdle {
+                    self.resyncMenuAdjunctReadinessBaseline()
+                }
             }
             // Removed redundant async refresh - single pass is sufficient after initial layout
             return
@@ -128,6 +136,9 @@ extension StatusItemController {
             // Intentionally skip open-menu tracking when refresh is disabled (tests).
             // If refresh is re-enabled while this menu stays open, it will not be backfilled until next open.
             self.openMenus[ObjectIdentifier(menu)] = menu
+            if menuTrackingWasIdle {
+                self.resyncMenuAdjunctReadinessBaseline()
+            }
             self.installProviderSwitcherShortcutMonitorIfNeeded(for: menu)
             // Only schedule refresh after menu is registered as open - refreshNow is called async
             self.scheduleOpenMenuRefresh(for: menu)
--- a/Sources/CodexBar/StatusItemController+Menu.swift
+++ b/Sources/CodexBar/StatusItemController+Menu.swift
@@ -136,7 +136,10 @@ extension StatusItemController {
             // Intentionally skip open-menu tracking when refresh is disabled (tests).
             // If refresh is re-enabled while this menu stays open, it will not be backfilled until next open.
             self.openMenus[ObjectIdentifier(menu)] = menu
-            if menuTrackingWasIdle {
+            // Only re-anchor when the opened menu actually shows current data. During an in-flight provider
+            // refresh `refreshMenuForOpenIfNeeded` can preserve stale content; resyncing the baseline to
+            // live store data in that case would mask the refresh-completion update (#1351).
+            if menuTrackingWasIdle, !self.menuNeedsRefresh(menu) {
                 self.resyncMenuAdjunctReadinessBaseline()
             }
             self.installProviderSwitcherShortcutMonitorIfNeeded(for: menu)
--- a/Sources/CodexBar/StatusItemController+Menu.swift
+++ b/Sources/CodexBar/StatusItemController+Menu.swift
@@ -131,6 +131,7 @@ extension StatusItemController {
             self.deferOpenAIDashboardRefreshUntilMenuCloses(reason: "parent menu open")
         }
 
+        let menuWasFreshBeforeOpen = !self.menuNeedsRefresh(menu)
         self.refreshMenuForOpenIfNeeded(menu, provider: provider)
         if self.isMenuRefreshEnabled {
             // Intentionally skip open-menu tracking when refresh is disabled (tests).
@@ -140,7 +141,10 @@ extension StatusItemController {
             // refresh `refreshMenuForOpenIfNeeded` can preserve stale content; resyncing the baseline to
             // live store data in that case would mask the refresh-completion update (#1351).
             if menuTrackingWasIdle, !self.menuNeedsRefresh(menu) {
-                self.resyncMenuAdjunctReadinessBaseline()
+                self.resyncMenuAdjunctReadinessBaselineForRootOpen(
+                    menu,
+                    provider: provider,
+                    menuWasFreshBeforeOpen: menuWasFreshBeforeOpen)
             }
             self.installProviderSwitcherShortcutMonitorIfNeeded(for: menu)
             // Only schedule refresh after menu is registered as open - refreshNow is called async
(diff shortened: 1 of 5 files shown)


Yuxin-Qiao:
Pushed a follow-up commit addressing the review:

- **Correctness fix (the P1 baseline-staleness risk):** the readiness baseline is now re-anchored when a **root** menu opens, so a closed→reopen→revert sequence can no longer compare equal against a stale baseline and skip a needed rebuild. Nested submenu opens deliberately do not resync, to avoid masking a pending parent refresh. Good catch.
- **Focused regression test** added (`reopening root menu resyncs readiness baseline so reverted store data still refreshes`), and negative-validated: it fails with the resync removed and passes with it.
- **Proof** added to the PR body: a 9.3x micro-benchmark for the per-tick double serialization this PR removes, plus the negative-validation output for the new test.

PR body updated with the above. @clawsweeper re-review

Yuxin-Qiao:
Added a direct runtime measurement to the PR body (mirroring what #1352 used): the real `menuAdjunctReadinessSignature()` costs ~78 µs/call with 4 providers × 30-day daily, and that full call ran on the main actor on **every** store mutation even with all menus closed — now skipped while closed. The per-call cost itself also drops via the 9.3x formatting change, and the baseline-resync correctness fix is negative-validated. @clawsweeper re-review

Yuxin-Qiao:
Added **real-app proof** to the PR body: a freshly built ad-hoc bundle of this branch with **7 providers enabled** and the menu closed, sampled for 65 s across a background refresh tick (`refreshFrequency = oneMinute`). The readiness/rebuild path (`menuAdjunctReadinessSignature`, `populateMenu`, `observeStoreChanges`, …) gets **0 hits on any thread**, main thread stays parked in `mach_msg`, idle CPU ~0.3%. Plus the ~78 µs/call cost of the now-skipped signature and the 9.3x formatting speedup. Honest caveat noted: couldn't automate the menu-*open* capture locally (no Accessibility/Peekaboo). @clawsweeper re-review

Yuxin-Qiao:
Addressed the @clawsweeper P1 stale-preservation blocker:

- **Gate baseline resync on fresh visible content:** root-menu resync now runs only when `!menuNeedsRefresh(menu)` after `refreshMenuForOpenIfNeeded`. If an in-flight provider refresh preserves stale content on open, the baseline is **not** re-anchored to live store data, so the refresh-completion update still registers as a readiness change and triggers a rebuild.
- **Focused regression test:** `root open during in flight refresh preserves stale content and does not resync baseline` — negative-validated (fails with unconditional resync, passes with the gate).
- Existing `reopening root menu resyncs readiness baseline …` test still passes.

`make check` clean; `swift test --filter StatusMenuReadinessBaselineTests` 2/2 pass.

@clawsweeper re-review

Yuxin-Qiao:
Checked off the last Test plan item — menu-*open* proof (7 providers, same ad-hoc bundle):

- **Open:** \`osascript\` click on the CodexBar menu-bar item → **11 items** visible, **~190 ms** real time.
- **While open:** held the menu open for **20 s** and ran \`sample CodexBar 20\` — **0 hits** on \`menuAdjunctReadinessSignature\`, \`populateMenu\`, \`observeStoreChanges\`, \`invalidateMenus\`, \`updateMenuContentPreservingSwitcher\`, \`sizeThatFits\`; main thread **17133/17138** samples parked in \`mach_msg\`.

PR body updated; all Test plan checkboxes are now complete. Clawsweeper already has this at **diamond lobster / ready for maintainer look** — the remaining step is maintainer merge.

PR #1352, opened 2026-06-07
perf: size hosted menu charts without a throwaway hosting controller

Yuxin-Qiao:
## Summary

Follow-up to the popup-menu lag reported in #1321, focused on the hosted chart submenus (cost/usage/credits history, storage breakdown, z.ai hourly, usage breakdown).

Every hosted chart item built **two** SwiftUI hierarchies on each (re)build:
1. the `MenuHostingView` that is actually displayed, and
2. a separate `NSHostingController(rootView:)` created *solely* to measure height via `sizeThatFits`.

The controller-measured height was always immediately overwritten by the subsequent `refreshHostedSubviewHeights()` `fittingSize` pass — which runs from **both** call paths (`menuWillOpen` after hydrate, and `refreshHostedSubviewMenu`). So the second SwiftUI tree was pure overhead on a popup-menu hot path that scales with provider/account count and re-runs whenever provider data changes while a chart submenu is open.

This change measures the **live displayed view** via `fittingSize` instead (the exact mechanism `refreshHostedSubviewHeights` already uses), extracted into a shared `hostedSubviewFittingHeight(for:width:)` helper. Final heights are identical; only the redundant hosting-controller hierarchy is removed.

Why it's safe:
- The displayed view is always the source of truth for the final height (`refreshHostedSubviewHeights` re-measures it on every open/refresh today).
- Chart **content** is still rebuilt from current data on every refresh — no caching, no fingerprints, so no risk of stale charts or clipped heights.
- No change to *when* or *whether* menus rebuild, so the existing AppKit menu-tracking / retry behavior is untouched.

## Proof

**Runtime measurement of the removed work**, on the real `CostHistoryChartMenuView` (30 daily entries, menu-card width 310, 400 builds, warm), comparing the two paths directly:

```
old (2 hierarchies: throwaway NSHostingController.sizeThatFits + live fittingSize): 2.080s total, 5.201 ms/build
new (1 hierarchy: live fittingSize only):                                           1.866s total, 4.664 ms/build
saved per submenu build: ~0.54 ms (10.3% less)
```

That ~0.5 ms is per chart-submenu build, on the main thread, and recurs on every open/refresh while a chart submenu is visible — multiplied across enabled providers/accounts, which is exactly the multi-provider setup the #1321/#1325 reporters describe.

**Correctness:** the new regression test asserts the height the append path assigns equals the authoritative `refreshHostedSubviewHeights` re-measure (and is non-trivial) across cost-history / usage-history / storage-breakdown — i.e. dropping the throwaway controller did **not** change sizing behavior.

(What I can't provide locally: an end-to-end popup recording on the reporters' exact multi-account machine. The change is a strict reduction of main-thread SwiftUI work on the documented hot path, with identical output.)

## Test plan

- [x] `swift build`
- [x] `make check` (SwiftFormat + SwiftLint, 0 violations)
- [x] New test `hosted chart items size to the displayed view without a throwaway controller` — asserts the append-path height matches the authoritative `refreshHostedSubviewHeights` re-measure (and is non-trivial) across cost-history / usage-history / storage-breakdown for a seeded provider
- [x] `StatusMenuHostedSubmenuRefreshTests`, `StatusMenuOpenRefreshTests`, `StatusMenuHeightCacheTests` (41 tests) pass
- [x] Ad-hoc local bundle build + launch, app healthy
- [x] Runtime measurement of removed hierarchy (see Proof)

## Notes

Complements #1351 (menu readiness signature cost). Touches the hosted-submenu sizing path that overlaps @hhh2210's recent menu-card height-cache work, so review from that area owner is welcome.

### DIFF
--- a/Sources/CodexBar/StatusItemController+HostedSubmenus.swift
+++ b/Sources/CodexBar/StatusItemController+HostedSubmenus.swift
@@ -186,9 +186,9 @@ extension StatusItemController {
 
         let chartView = UsageBreakdownChartMenuView(breakdown: breakdown, width: width)
         let hosting = MenuHostingView(rootView: chartView)
-        let controller = NSHostingController(rootView: chartView)
-        let size = controller.sizeThatFits(in: CGSize(width: width, height: .greatestFiniteMagnitude))
-        hosting.frame = NSRect(origin: .zero, size: NSSize(width: width, height: size.height))
+        hosting.frame = NSRect(
+            origin: .zero,
+            size: NSSize(width: width, height: self.hostedSubviewFittingHeight(for: hosting, width: width)))
 
         let chartItem = NSMenuItem()
         chartItem.view = hosting
@@ -213,9 +213,9 @@ extension StatusItemController {
 
         let chartView = CreditsHistoryChartMenuView(breakdown: breakdown, width: width)
         let hosting = MenuHostingView(rootView: chartView)
-        let controller = NSHostingController(rootView: chartView)
-        let size = controller.sizeThatFits(in: CGSize(width: width, height: .greatestFiniteMagnitude))
-        hosting.frame = NSRect(origin: .zero, size: NSSize(width: width, height: size.height))
+        hosting.frame = NSRect(
+            origin: .zero,
+            size: NSSize(width: width, height: self.hostedSubviewFittingHeight(for: hosting, width: width)))
 
         let chartItem = NSMenuItem()
         chartItem.view = hosting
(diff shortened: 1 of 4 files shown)


Yuxin-Qiao:
Added runtime proof to the PR body: a direct measurement on the real `CostHistoryChartMenuView` shows the removed throwaway `NSHostingController` cost ~0.54 ms/submenu build (10.3% less), which recurs per open/refresh and scales with enabled providers/accounts. Correctness is covered by the new height-parity regression test. @clawsweeper re-review

PR #1359, opened 2026-06-08
feat: iOS Companion App & iCloud KVS Sync

tiagodocouto:
Adds an iOS Companion App and switches the macOS app's device-to-device sync to **iCloud Key-Value Store (KVS)**, removing the previous Tailscale/HTTP tunnel entirely.

## iOS Companion App
- New `CodexBarCompanion`, a lightweight SwiftUI iOS app.
- Mirrors the macOS menubar usage stats, charts, and progress bars.
- Reads its data exclusively from iCloud KVS (no local network dependency).
- Adjustable refresh interval (1–10 min) via the clock control in the header.

## iCloud KVS Sync
- macOS serializes the `CompanionCardModel` payload into `NSUbiquitousKeyValueStore` on every refresh (`StatusItemController+Companion`).
- iOS observes `didChangeExternallyNotification` and reads the same key, so updates propagate automatically across the user's Apple devices.
- The KVS identifier is `$(TeamIdentifierPrefix)com.steipete.codexbar`, shared by both the macOS app entitlement and the iOS companion so they address the same container.

## UI & Architecture
- **Empty states**: `StatusItemController`/`MenuCardView` hide providers with no data or that are logged out (e.g. Gemini "not logged in").
- **iCloud-only**: removed the local HTTP server, the Tailscale IP entry field, and the cleartext-HTTP fallback (and its ATS exception). The companion entitlement is now KVS-only — no CloudKit/APS containers.

## Build script
- `package_app.sh` injects `com.apple.application-identifier`, `com.apple.developer.team-identifier`, and `com.apple.developer.ubiquity-kvstore-identifier` from the resolved `$APP_TEAM_ID`/`$BUNDLE_ID`.
- Optionally embeds a local `codexbar-macos.provisionprofile` (gitignored) for AMFI validation of iCloud entitlements during local debug builds.

## Fork owners
To build with iCloud KVS support:
1. Register your App ID with the **iCloud (Key-Value Store)** capability.
2. Generate a **macOS App Development** provisioning profile for it.
3. Save it as `codexbar-macos.provisionprofile` in the repo root (gitignored; auto-embedded by the build script).
4. Keep only ONE matching development certificate in your Keychain to avoid `codesign` ambiguity.
5. Build with `APP_TEAM_ID` set to your own team so the macOS app and the companion share the same KVS container.

## Verification
- macOS (`swift build`): ✅ build complete.
- iOS companion (`xcodebuild`, iOS 26 simulator): ✅ build succeeded.

## Screenshots — iOS Companion

| Overview | Codex | Claude | Antigravity |
|:---:|:---:|:---:|:---:|
|  |  |  |  |


tiagodocouto:
Thanks for the thorough review — all four findings addressed in `37830505`.

## Review findings

| Finding | Status | Change |
|---|---|---|
| [P1] Gate iCloud sync behind explicit opt-in (`StatusItemController.swift:379`) | ✅ Fixed | New `companionSyncEnabled` setting, **OFF by default** (Preferences → Advanced). `syncCompanion()` early-returns when disabled and **removes** any previously-synced `latestUsageSync` value from KVS, so upgraded users never get data uploaded without consent. Email is also dropped from the payload when `hidePersonalInfo` is on (data minimization). |
| [P1] Keep default packaging independent of iCloud (`package_app.sh:161-166`) | ✅ Fixed | The iCloud entitlements (`application-identifier`, `team-identifier`, `ubiquity-kvstore-identifier`) and the provisioning-profile embed are now gated behind `CODEXBAR_ICLOUD_COMPANION=1`. Default / local / adhoc / fork / release builds are unchanged and need no Apple iCloud App ID or profile. |
| [P2] Do not drop non-quota companion cards (`ContentView.swift:193`) | ✅ Fixed | Filter is now `hasDisplayableContent` (metrics **or** providerCost / tokenUsage / credits / usageNotes / placeholder), so cost-only / note-only / error-only providers are kept. |
| [P2] Sync companion after settings changes (`StatusItemController.swift:445`) | ✅ Fixed | `syncCompanion()` is now also invoked from `handleSettingsChange()`, so toggling `hidePersonalInfo` / hiding / reordering providers refreshes the KVS payload instead of leaving stale email/plan on iOS. |

## Proof

**Unit proof** — the KVS value is plain JSON, so this verifies the macOS-writer → iOS-reader round-trip and the content filter (`Tests/CodexBarTests/CompanionSyncProofTests.swift`):

```
$ swift test --filter CompanionSync
✔ Test "Payload survives the JSON round-trip that crosses iCloud KVS" passed
✔ Test "Content filter keeps content-bearing cards that have no quota metrics" passed
✔ Suite "Companion KVS sync" passed — 2 tests
```

**Live-sync observability** — added `os_log` markers on both ends so the real cross-device path is verifiable in Console.app:
- macOS writer: `Companion sync → iCloud KVS: <N> providers, <M> bytes`
- iOS reader: `Companion read ← iCloud KVS: <N> cards`

To reproduce end-to-end (opt-in on, same iCloud account on both devices): enable Preferences → Advanced → “Sync to iOS app (iCloud)”, trigger a refresh on macOS, and both log lines fire in order. I'll attach a redacted screen recording of that path once I have both devices on the same signed build.

Builds: macOS `swift build` ✅ · iOS companion `xcodebuild` (iOS 26 sim) ✅.

@clawsweeper re-review

tiagodocouto:
Rebased onto latest `main` and addressed the open **[P2]** rendering finding.

## Merge conflict resolved

`main` refactored the store-change observation callback into `handleObservedStoreMenuChange()` (extracting `observeStoreChanges()` + `invalidateMenus(...)`), which collided with the `syncCompanion()` call this PR had inserted in that block. Resolved by keeping upstream's refactor and re-applying `self.syncCompanion()` **inside** `handleObservedStoreMenuChange()` (right after `observeStoreChanges()`), so the iOS companion KVS payload still refreshes on every usage refresh. The startup and `handleSettingsChange()` call sites auto-merged cleanly. PR is now mergeable.

## Review finding

| Finding | Status | Change |
|---|---|---|
| [P2] Render content-bearing cards without quota metrics (`CompanionCardView.swift:38-46`) | ✅ Fixed | The view's `body` branched on `metrics.isEmpty` and only rendered usage notes or a placeholder in that case, so cost-only / credits-only / token-usage-only cards (which `hasDisplayableContent` keeps) showed a blank header + divider. Hoisted `hasUsage` / `hasCredits` / `hasCost` out of the metrics-empty branch and now render the full content block whenever **any** of them is present; the placeholder is shown only when there is no content at all. Also added `creditsText` to `hasDetails` so credits-only cards get the header divider/padding for consistency. |

## Verification

- macOS `swift build` ✅
- iOS companion `xcodebuild` (iOS Simulator) ✅ — `CompanionCardView.swift` compiles
- `swift test --filter CompanionSync` ✅ — 3/3 (JSON round-trip, opt-in defaults OFF, content filter keeps content-bearing cards)
- Note: `StatusMenuSwitcherRefreshTests."merged provider switch restores cached tab content"` fails locally, but it fails **identically on a clean `main` checkout** as well — pre-existing/environment-sensitive, not introduced by this merge.

@clawsweeper re-review

steipete:
Thanks @tiagodocouto for the companion implementation and follow-up fixes. Closing after the current-main audit. The branch still lacks signed end-to-end proof that the macOS writer and iOS reader use the same iCloud KVS container; the iOS target hardcodes a development team while macOS derives its container from `APP_TEAM_ID`; and the iOS selection is not reset when the selected provider disappears. The opt-in iCloud path also changes CodexBar's data-storage, privacy, signing, and release boundary, which needs explicit product approval before implementation lands.

If revisited, the path is a fresh current-main proposal/PR with an approved narrow scope, one configurable signing/container contract for both targets, the disappearing-provider regression fixed, and redacted signed macOS-to-iOS live proof. Given the 41-file platform and release surface, rewriting from current main is safer than continuing this branch. Thank you for the considerable work here.

PR #1313, opened 2026-06-05
Add Antigravity agy CLI fallback

enieuwy:
## Summary
- add an Antigravity `agy` CLI HTTPS fallback when the desktop app is closed
- harden warm-session ownership so live CodexBar owners keep their persisted CLI records
- add focused coverage for CLI session lifecycle, fetch strategy fallback, and menu metric selection
- make Antigravity usage source mode authoritative even when Google OAuth accounts are configured
- leave `CHANGELOG.md` untouched for the release flow

## Rationale
Antigravity desktop exposes a local `GetUserStatus` API while the desktop app is running. `agy` exposes the same local HTTPS status API without launching the full desktop app, so CodexBar can use it as the closed-desktop local fallback.

In observed live Antigravity OAuth responses, CodexBar received Gemini-oriented account/quota data and did not receive the Claude model quota lanes returned by the local Antigravity/`agy` `GetUserStatus` endpoint. The remote OAuth parser can still parse Claude quotas if Google returns them, so this PR does not assume OAuth can never provide Claude models. The practical reason for the fallback is that the `agy` local path returned the full tested quota shape: Claude + Gemini per-model lanes, used percentages, and reset timestamps while the desktop app was closed.

## Verification
- `swift test --filter AntigravityCLISessionTests`
- `swift test --filter AntigravityCLIHTTPSFetchStrategyTests`
- `make check`
- `swift test`
- bundled local app was updated/relaunched; bundled `CodexBarCLI usage --provider antigravity --source auto` now returns `source: "cli"` with the normal Google account config present

## Runtime proof: closed-desktop `agy` fallback

Local bundle built and launched from this branch with `./Scripts/compile_and_run.sh`.

Desktop Antigravity/language server was not running:

```text
$ pgrep -fl "/Applications/Antigravity|Google Antigravity|language_server"
# no output
```

Auto mode with a normal config that includes an Antigravity Google OAuth account now still follows the source-mode pipeline and falls through to `agy` before OAuth:

```text
$ ./CodexBar.app/Contents/Helpers/CodexBarCLI usage --provider antigravity --source auto --format json --pretty --log-level debug
2026-06-05T23:38:55+0800 debug ... [CodexBarCore] Antigravity CLI session started binary=agy pid=<redacted>
2026-06-05T23:38:58+0800 debug ... [CodexBarCore] Antigravity CLI session stopping pid=<redacted> reason=one-shot CLI fetch
[
  {
    "account" : "<redacted>",
    "provider" : "antigravity",
[...]

PR #1365, opened 2026-06-09
feat(antigravity): fetch usage from agy CLI via PTY /usage capture

bcharleson:
## Summary
- Add Antigravity usage via the `agy` CLI interactive `/usage` Model Quota panel using PTY capture (Claude + Gemini buckets without Antigravity.app)
- Layer fallbacks: Gemini `retrieveUserQuota` API when PTY capture is partial; fix remote quota 403 from `User-Agent: antigravity`
- Drop the Antigravity IDE local port-scan probe from default auto/cli pipelines; prefer `agy` when installed and auto-disable Gemini when `agy` is present

## Test plan
- [x] `swift test --filter AntigravityAgy`
- [x] `swift test --filter AntigravityStatusProbe`
- [x] `swift run CodexBarCLI diagnose --provider agy --format json` on agy-only machine (no Antigravity.app): source `agy`, Claude primary, Gemini Pro/Flash secondary/tertiary


steipete:
Thanks for the substantial Antigravity investigation and PTY parser work. Closing this competing implementation in favor of #1313 as the canonical `agy` fallback track.

#1313 preserves explicit OAuth account routing, keeps the existing IDE path, has redacted live closed-desktop proof, and now covers warm-session and transient-error behavior. This branch still replaces established source semantics, can cross the selected-account boundary in OAuth mode, and lacks the requested runtime proof. Any distinct PTY-only evidence can be carried into the #1313 discussion if the local HTTPS path proves insufficient.

bcharleson:
Copy that, thanks @steipete

Issue #1334, opened 2026-06-07
[Bug] Antigravity summary bar ignores Gemini family quota (shows 100%)


dhruv-anand-aintech:
## Summary
Antigravity summary bar appears to report `100%` remaining while Gemini family quota is exhausted.

## Environment
- macOS: 26.4.1 (`25E253`)
- CodexBar process observed: `/Applications/CodexBar.app/Contents/MacOS/CodexBar`
- PID observed: `27497`
- Date/time observed: `Sun Jun 7 01:00:50 2026`

## Steps / behavior
- In Antigravity summary mode, overall quota remains at `100%`.
- Gemini family quota is actually exhausted.
- Claude family is also at `100%`.

## Expected
Summary percentage should reflect exhausted Gemini family quota rather than showing `100%`.

## Actual
Summary bar remains at `100%`, which is misleading when one family is exhausted.

## Impact
Users can miss that Gemini is unavailable / depleted and assume usage headroom remains.

PR #1369, opened 2026-06-09
Fix Antigravity summaries for untracked quotas

Martin-Hausleitner:
## Summary
- Exclude Antigravity models without a reported `remainingFraction` from the three family summary lanes.
- Preserve those models in per-model quota windows, including reset metadata.
- Add regression coverage proving automatic mode selects tracked constrained usage instead of an untracked family.

## Verification
- `swift test --filter AntigravityStatusProbeTests`
- `swift test --filter MenuCardAntigravityTests`
- `swift test --filter MenuBarMetricWindowResolverTests`
- `make check`
- Structured autoreview: clean, no accepted/actionable findings

Refs #1334.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 0.32.6 — Unreleased
 
+### Fixed
+- Antigravity: exclude model quotas without a remaining fraction from family summaries so they no longer mask tracked usage in the automatic menu-bar metric (#1369). Thanks @Martin-Hausleitner!
+
 ## 0.32.5 — 2026-06-09
 
 ### Added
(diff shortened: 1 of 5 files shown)


steipete:
Landed as `6f6cb097dc58a5290bc61cf1f44d1dbe98eee780`.

Proof:
- full GitHub CI green: macOS lint/build/test plus Linux x64 and arm64
- focused Antigravity probe, menu-card, and metric-resolver tests passed locally
- `make check` passed
- final autoreview clean with no accepted/actionable findings

The landed version excludes untracked nil-fraction quotas from family summaries while preserving their per-model metadata. Thanks @Martin-Hausleitner!

PR #1373, opened 2026-06-10
Fix Antigravity menu card: drop redundant family summary lanes and placeholder quota

akunzai:
## Summary
The Antigravity menu card duplicated quota rows and drew placeholders for
families that have no real data:

- When per-model quota windows are available, the card prepended the three
  generic family summary lanes ("Claude", "Gemini Pro", "Gemini Flash") on top
  of the detailed per-model rows, showing the same quota twice.
- When a Claude/Gemini family lane was missing, it was rendered as a fake
  `0% left` / `100% used` placeholder instead of being omitted.

This PR renders the real per-model `extraRateWindows` when present, and
otherwise falls back to the summary lanes while skipping any missing (nil)
lane — so there are no duplicated or placeholder rows.

## Changes
- `MenuCardView+ModelHelpers.swift`: prefer per-model extra windows and render
  only those when present; otherwise render the summary lanes, skipping nil
  ones. Removes the placeholder branch that drew `0% left` / `100% used` for
  absent windows.
- `MenuCardAntigravityTests.swift`: missing families are omitted (not shown at
  zero percent), an exhausted real model still shows its reset text, and
  per-model rows render without duplicated summary lanes.

## Before / After
| Before | After |
| --- | --- |
|  Antigravity_Usage_Before | Antigravity_Usage_After |

## Testing
- `swift test --filter MenuCardAntigravityTests --filter AntigravityStatusProbeTests` — 56 tests / 2 suites passed
- `make check` — 0 violations / 0 serious in 1022 files

## Scope notes
- Presentation-only change. The remote display-filter rework is intentionally
  left out (keeps the #1209 behaviour), and the `nil`-fraction `remainingPercent`
  fix is deferred to #1369.

### DIFF
--- a/Sources/CodexBar/MenuCardView+ModelHelpers.swift
+++ b/Sources/CodexBar/MenuCardView+ModelHelpers.swift
@@ -154,30 +154,39 @@ extension UsageMenuCardView.Model {
 
     static func antigravityMetrics(input: Input, snapshot: UsageSnapshot) -> [Metric] {
         let percentStyle: PercentStyle = input.usageBarsShowUsed ? .used : .left
-        var metrics = [
-            Self.antigravityMetric(
-                id: "primary",
-                title: L(input.metadata.sessionLabel),
-                window: snapshot.primary,
-                input: input,
-                percentStyle: percentStyle),
-            Self.antigravityMetric(
-                id: "secondary",
-                title: L(input.metadata.weeklyLabel),
-                window: snapshot.secondary,
-                input: input,
-                percentStyle: percentStyle),
-            Self.antigravityMetric(
-                id: "tertiary",
-                title: input.metadata.opusLabel.map(L) ?? L("Gemini Flash"),
-                window: snapshot.tertiary,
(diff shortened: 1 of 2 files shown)


akunzai:
Closing this in favor of #1369, which landed first and reworks Antigravity untracked-quota handling.

After rebasing onto #1369 I found the two changes are design-incompatible:
- #1369 deliberately keeps untracked (nil-fraction) models in `extraRateWindows` with their reset metadata (see `AntigravityStatusProbeTests.model without remaining fraction stays out of family summary and preserves reset metadata`), excluding them only from the family summary lanes.
- This PR's menu-card de-dup (render per-model rows, drop the redundant summary lanes) therefore surfaces those untracked models as `0% left` per-model rows, and to ship it I'd have to rewrite #1369's just-merged summary-lane test.

That tradeoff is a maintainer call, so I'm closing rather than forcing it. Happy to reopen / revisit the summary-vs-per-model de-duplication separately if it's wanted.

Issue #1374, opened 2026-06-10
Stale merged dropdown rebuilds synchronously inside menuWillOpen on the first click after a background refresh tick


hhh2210:
## Summary

After #1286 / #1297 / #1314, the remaining Merge Icons click latency has one specific source left: the merged dropdown is deferred-until-next-open while closed (the #1314 fix for the background-freeze path), so every background data tick leaves it stale — and the next click then runs the full `populateMenu` (MenuDescriptor build + `NSHostingView` creation + SwiftUI layout) **synchronously inside `menuWillOpen`, before the dropdown can appear**.

This is exactly the leftover #1314 scoped out: *"This still does not remove the first `populateMenu` or reuse hosting views; it eliminates the redundant **closed** rebuild only."*

## Where it comes from (current main)

1. `observeStoreChanges` invalidates with `allowStaleContentDuringDataRefresh: true` on every store tick, bumping `menuContentVersion` without bumping `latestRequiredMenuRebuildVersion` (`StatusItemController.swift`).
2. While the merged menu is closed, `prepareAttachedClosedMenusIfNeeded` defers it into `closedMenusDeferredUntilNextOpen` instead of pre-warming (`StatusItemController+MenuTracking.swift`, #1314).
3. On the next click, `menuWillOpen → refreshMenuForOpenIfNeeded` finds the menu stale. The stale-preserve path (`canPreserveStaleMenuContentDuringRefresh`) only covers refreshes still **in flight**; a tick that already **completed** falls through to the synchronous `populateMenu(...)` before the menu is displayed.

With `refreshFrequency = oneMinute` and Merge Icons on, effectively every open after ≥1 idle minute pays the full rebuild on the click path.

## Evidence already on file

- #1325's sample caught `populateMenu` / `updateMenuContentPreservingSwitcher` / `addMenuCards` / `addMenuCardSections` as the main-thread hot frames while the menu felt stuck on 0.32.4/0.32.5.
- @giuseppebisemi's post-#1297 retest in #1274 measured the open path at `~306 menuWillOpen → ~153 refreshMenuForOpenIfNeeded` inclusive main-thread samples — that subtree is this synchronous open rebuild.

## Expected

Clicking the status item should show the dropdown immediately. A data-only refresh of already-rendered cards should not block the open; it can land right after the menu is on screen.

## Proposed fix

[...]

Issue #1325, opened 2026-06-06
Menu popover stays open or delays switching between Claude and Codex tabs


mattleonard16:
## Summary

CodexBar's menu popover can stay open and become hard to dismiss. Switching between provider tabs, especially from Claude to Codex, is sometimes delayed or appears to hang.

## What I observed

- Opening the CodexBar menu shows the merged provider dropdown.
- The menu sometimes remains open and does not close promptly when clicking away.
- Switching tabs from Claude to Codex can be delayed or feel stuck.
- The issue is most noticeable with the merged menu containing Claude and Codex providers enabled.

## Environment

- macOS: 26.3.1 (25D2128)
- Installed via Homebrew cask: `brew install --cask codexbar`
- Homebrew cask version tested: `CodexBar 0.32.4`
- Local source build also tested: `CodexBar 0.32.5` / commit `65e39f4`
- Enabled providers at the time included Claude, Codex, and Antigravity.

## Troubleshooting already tried

1. Uninstalled/reinstalled via Homebrew:
   - `brew uninstall --cask --force codexbar`
   - `brew install --cask codexbar`
2. Verified the app bundle and CLI link:
   - `/Applications/CodexBar.app` exists
   - `/opt/homebrew/bin/codexbar --version` reported `CodexBar 0.32.4`
   - Gatekeeper accepted the app as `Notarized Developer ID`
3. Confirmed the app was running as `/Applications/CodexBar.app/Contents/MacOS/CodexBar`.
4. Tried reducing background refresh work via defaults:
   - `tokenCostUsageEnabled = false`
   - `openAIWebAccessEnabled = false`
   - `showOptionalCreditsAndExtraUsage = false`
   - `refreshFrequency = manual`
5. Captured a sample while the menu was stuck. The main thread was spending time in menu rebuild paths including:
   - `StatusItemController.populateMenu`
   - `StatusItemController.updateMenuContentPreservingSwitcher`
   - `StatusItemController.addMenuCards`
   - `StatusItemController.addMenuCardSections`
6. System logs from the 0.32.4 run showed repeated runtime warnings like:
   - `This method should not be called on the main thread as it may lead to UI unresponsiveness.`
7. Built and temporarily installed local source `0.32.5` (`65e39f4`), whose changelog mentions merged-menu dismiss latency fixes. Startup logs looked cleaner, but I am filing this because the user-visible symptom was still observed around menu close/tab switching.

## Expected behavior

[...]

PR #1314, opened 2026-06-05
Cache menu card heights and skip closed merged-menu rebuild

hhh2210:
Refs #1274.

## Summary
- keep menu-card height cache entries reusable across menu-content version invalidations when callers provide a width + content fingerprint
- add a `UsageMenuCardView.Model` height fingerprint and pass it through the merged, stacked, history, and chart menu-card call sites
- keep no-fingerprint callers version-scoped, with focused tests for stable content, changed content, legacy fallback, eviction, and provider scoping
- after the latest #1274 retest, trim the fingerprint/account-info hot paths so the cache key no longer gives back most of the saved measurement work

## Why
The latest #1274 samples after #1297 and #1314 split the remaining Merge Icons lag into two pieces:

1. `populateMenu -> addMenuCardSections -> MenuCardItemHostingView.measuredHeight`, where each cache miss allocates a fresh `NSHostingView` just to measure card height.
2. The follow-up cost introduced by this PR's first pass: `heightFingerprint` string formatting/hash work plus `menuCardModel -> accountInfo -> loadAuthBackedCodexAccount -> CodexOAuthCredentialsStore.parse` on repeated rebuilds.

#1286 moved close rebuild work off the synchronous dismiss path, and #1297 targets icon-observation churn. This PR remains the next separate measured-height slice, but now also addresses the obvious cost regressions called out in the #1314 retest. It still does not claim to close #1274; affected macOS 26.x / Merge Icons machines should retest the packaged build.

## Claw/Copilot Follow-up
- replaced raw text cache-key fields, including account email, with per-process salted text shapes (`chars`, `utf8`, `lines`, hash), so cache keys no longer retain raw PII or user-visible strings
- encoded optional string presence explicitly, including `storageText`, so `nil` and `""` no longer collapse into the same height identity
- pruned stale `version:` fallback keys on `invalidateMenus()` while retaining content-fingerprinted keys, avoiding old version-scoped entries accumulating until the global 256-entry flush
- made the height-cache tests use deterministic `AccountInfo(email: nil, plan: nil)` fixtures instead of `UsageFetcher().loadAccountInfo()`

## Latest #1274 Retest Follow-up
@giuseppebisemi's #1314 retest showed the measured-height leaf dropped, but part of the work moved into the new fingerprint and the broader model construction still reparsed Codex OAuth credentials. Current head `254412f0` responds to that specific critique:

[...]

PR #1297, opened 2026-06-04
Reduce merged icon observation churn

hhh2210:
Refs #1274.

## Summary
- Move the menu bar icon observation signature out of the broad status item controller file into `StatusItemController+IconObservation.swift`.
- Remove full `UsageSnapshot` stringification (`String(describing: snapshot)`) from the icon observation signature.
- In Merge Icons mode, sign only the provider actually rendered by the merged icon plus aggregate provider status, instead of every provider snapshot.
- Replace the Codex menu-bar credit projection in the icon render/observation path with a cheap credits/rate-window helper.
- Make `IconStyle` `String`-raw-represented and sign it via `.rawValue`, removing the last `String(describing:)` reflection (`_adHocPrint_unlocked`) from the icon-observation leaf.

## Why
`@giuseppebisemi` sampled the residual lag after #1286 and the hot path was `observeStoreIconChanges -> storeIconObservationSignature -> providerStoreIconObservationSignature`. The old signature subscribed to every provider snapshot and used `String(describing: snapshot)`, so account/email/timestamp and non-rendered provider churn could retrigger icon work even when the visible merged icon would not change.

## Behavior proof
Live runtime proof on the exact reported platform — **macOS 26.5 (build 25F71)**, Apple Swift 6.3.2 — against this branch head.

**1) Signature contract tests (both directions: ignore churn / react to visible change)**

```text
$ sw_vers -productVersion && sw_vers -buildVersion
26.5
25F71
$ xcrun xctest -XCTest CodexBarTests.StatusItemIconObservationSignatureTests ...
✔ store icon observation signature ignores refresh and status metadata churn
✔ store icon observation signature ignores non visual snapshot churn
✔ merged store icon observation signature ignores non primary snapshot churn
✔ store icon observation signature ignores unused credit balance
✔ store icon observation signature changes when icon percentages change
✔ store icon observation signature changes when credit fallback changes
✔ merged store icon observation signature changes when non primary status changes
✔ store icon observation signature changes when status indicator changes
✔ Test run with 8 tests in 1 suite passed after 0.41 seconds.
```

The four `ignores …` tests cover the stale-content risk (non-rendered snapshot / account / timestamp / unused-credit churn no longer moves the signature); the four `changes when …` tests prove visible percent, Codex credits fallback, primary-vs-aggregate status, and status-indicator changes still move it. So the narrowed signature does not suppress required icon refreshes.

**2) Packaged-app main-thread `sample` (Merge Icons on), before vs after the reflection-removal commit**

[...]

PR #1375, opened 2026-06-10
Open data-stale dropdowns in two phases (populateMenu off the click path)

hhh2210:
Refs #1374, #1274, #1325.

## Summary
- open data-stale dropdowns in two phases: when the only pending invalidations are data-refresh ticks, `menuWillOpen` attaches the menu's existing content immediately and schedules the rebuild through the existing `scheduleOpenMenuRebuildIfStillVisible` machinery, so it runs right after the menu is on screen instead of synchronously on the click path
- structural / privacy / localization invalidations (anything that bumps `latestRequiredMenuRebuildVersion`) still rebuild synchronously before display, so stale structure is never shown
- the in-flight-refresh stale-preserve path keeps precedence (#1351 semantics); refresh completion still drives the rebuild in that case
- closing before the deferred rebuild runs cancels it via the existing `forgetClosedMenu` cleanup, and the merged menu re-defers until next open
- tests-only mode (`isMenuRefreshEnabled == false`) keeps the synchronous rebuild, since the post-display rebuild would never run without open-menu tracking

## Why
#1314 deferred the closed merged menu until next open and explicitly scoped out the consequence: *"This still does not remove the first `populateMenu` or reuse hosting views; it eliminates the redundant **closed** rebuild only."*

Since every background data tick while the menu is closed leaves it deferred-stale, the next click pays the full `populateMenu` (MenuDescriptor build + `NSHostingView` creation + SwiftUI layout) synchronously inside `menuWillOpen` before the dropdown can appear — with `refreshFrequency = oneMinute` and Merge Icons on, effectively every open after ≥1 idle minute. Root-cause walkthrough in #1374; #1325's sample caught `populateMenu` / `addMenuCards` as the hot frames on that path, and the post-#1297 retest in #1274 measured it at `~306 menuWillOpen → ~153 refreshMenuForOpenIfNeeded` inclusive main-thread samples.

## Scope / honesty
- This does not reduce total rebuild work; it moves the rebuild off the click-blocking path, the same way #1286 moved it off the dismiss path. The post-display rebuild runs once while the menu is visible, through the smart-update path that preserves the provider switcher (the path open-menu invalidations already use).
- The content shown during the brief post-display window is the same content the menu showed when it last closed; only data-refresh invalidations can take this path, so it is data-staleness by definition, never stale structure.
- First-ever open (menu has no items yet) and required invalidations are unchanged: synchronous rebuild before display.

## Validation
- three focused regression tests in `StatusMenuTwoPhaseOpenTests.swift`: data-stale open keeps existing items and rebuilds after display; required invalidation still rebuilds synchronously before display; closing before the deferred rebuild cancels it and re-defers the merged menu
- one updated expectation in `StatusMenuOpenRefreshTests.swift`: the non-merged stale open now asserts the two-phase deferral instead of the synchronous rebuild (the mechanism is shared, and the deferral is the same win there)
- `swift test` (full suite) + SwiftFormat/SwiftLint were green on the fork mirror of this branch at the pre-rebase head (`abe46ca3`); the rebase onto `6f6cb097` only resolved a CHANGELOG collision with #1369, no code changes

## Behavior proof status
Packaged before/after proof on macOS 26.5 (peekaboo open-click timing + `sample` stack identity: `populateMenu` under `menuWillOpen` vs under the deferred-task thunk, #1314-style) is being collected and will be posted as a follow-up comment on this PR. A retest from the #1274/#1314 measurement setup would be very welcome — asked in #1374.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 - Antigravity: exclude model quotas without a remaining fraction from family summaries so they no longer mask tracked usage in the automatic menu-bar metric (#1369). Thanks @Martin-Hausleitner!
+- Menu bar: open data-stale dropdowns in two phases — attach the existing content immediately on click and rebuild from current data right after display — so merged-menu opens no longer pay the full SwiftUI card rebuild before the menu appears (#1274, #1325).
 
 ## 0.32.5 — 2026-06-09
 
--- a/Sources/CodexBar/StatusItemController+MenuTracking.swift
+++ b/Sources/CodexBar/StatusItemController+MenuTracking.swift
@@ -125,12 +125,39 @@ extension StatusItemController {
             self.deferMenuInteractionRefreshIfNeeded()
             return
         }
+        if self.canDeferStaleMenuRebuildUntilAfterDisplay(menu) {
+            #if DEBUG
+            self.menuLogger.debug(
+                "menu open deferred stale rebuild until after display",
+                metadata: [
+                    "items": "\(menu.items.count)",
+                    "provider": provider?.rawValue ?? "nil",
+                ])
+            #endif
+            self.scheduleOpenMenuRebuildIfStillVisible(menu, provider: provider)
+            return
+        }
         self.populateMenu(menu, provider: provider)
         self.markMenuFresh(menu)
     }
 
     private func canPreserveStaleMenuContentDuringRefresh(_ menu: NSMenu) -> Bool {
         guard self.isMenuDataRefreshInFlight, !menu.items.isEmpty else { return false }
+        let key = ObjectIdentifier(menu)
+        guard let menuVersion = self.menuVersions[key] else { return false }
+        return menuVersion >= self.latestRequiredMenuRebuildVersion
+    }
+
+    /// Two-phase open: when the only pending invalidations are data-refresh ticks, the dropdown can
+    /// attach its existing content immediately and rebuild from current store data right after the
+    /// menu is on screen, instead of paying the full `populateMenu` (incl. SwiftUI hosting-view
+    /// layout) synchronously inside `menuWillOpen` on the click path (#1274, #1325). Structural,
+    /// privacy, or localization invalidations bump `latestRequiredMenuRebuildVersion` and still
+    /// rebuild synchronously before display.
+    private func canDeferStaleMenuRebuildUntilAfterDisplay(_ menu: NSMenu) -> Bool {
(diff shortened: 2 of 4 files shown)


steipete:
Closing this approach after landing #1376 in `e97bfb0db9eb0883ad92d173d8accc547879bcd2`.

#1376 fixes the observed WindowServer stall by deferring parent-menu recomposition caused by data refreshes until AppKit menu tracking ends. This PR instead schedules the full `populateMenu` rebuild immediately after display while the menu is being tracked, which reintroduces the same expensive parent-menu reconstruction during the unsafe window.

#1374 remains open. Its first-open stale-content latency still needs a design that does not rebuild the full parent menu during tracking.

Thanks @hhh2210 for the investigation, tests, and clearly documented tradeoffs.

hhh2210:
Pushed d85192ae addressing the ClawSweeper P2 (readiness-baseline re-anchor) — the analysis was correct, thanks.

**What was wrong:** the two-phase open returns from `refreshMenuForOpenIfNeeded` while the menu is still marked stale, so the `!menuNeedsRefresh` gate in `menuWillOpen` skipped `resyncMenuAdjunctReadinessBaselineForRootOpen`, and the deferred rebuild (`rebuildOpenMenuIfStillVisible`) repopulated + `markMenuFresh`ed without recording a new baseline. An open-menu store change reverting to the pre-rebuild signature would then compare equal in `didMenuAdjunctReadinessChange()` and skip the refresh, leaving the visible menu on the deferred-rebuild data.

**Fix:** `rebuildOpenMenuIfStillVisible` now calls `resyncMenuAdjunctReadinessBaseline()` right after `markMenuFresh`, gated to non-hosted-subview menus. This is the moment the visible root menu and live store data are in sync, matching the contract documented on `resyncMenuAdjunctReadinessBaseline`. For the existing mid-open invalidation flow this is a no-op, because the observation handler's `didMenuAdjunctReadinessChange()` already advanced the baseline before scheduling the rebuild; the in-flight-refresh stale-content path (#1351) is unaffected since it never reaches the rebuild completion.

**Regression coverage:** added `deferred open rebuild re-anchors readiness baseline so reverted store data still refreshes` to `StatusMenuTwoPhaseOpenTests`. It drives the deferred rebuild completion deterministically (no interleaved observation task can mask the regression) and asserts the reverted store value still registers as a readiness change. Verified red on the previous commit, green with the fix; `swift test` + `make check` pass.

Packaged behavior proof (peekaboo open-click timing + `sample` stack identity, before/after) is still being collected and will follow as a separate comment.

PR #1378, opened 2026-06-10
Add Claude web session recovery action

### LINKED ISSUE #1377: Claude web "Unauthorized" error has no actionable recovery path
# Claude web "Unauthorized" error has no actionable recovery path

## Summary
When CodexBar's Claude provider is configured for **web** mode (`claudeWebExtrasEnabled = 1`) and the cached browser session is rejected by claude.ai with 401/403, the menu card prints a dead string — `Unauthorized. Your Claude session may have expired.` — and the user is stuck. There is no hint that the fix is "open claude.ai in your browser and re-login"; the menu offers no clickable recovery item.

## Steps to reproduce
1. Enable Claude **web** mode in CodexBar settings (`claudeUsageDataSource = .web`).
2. Sign in to claude.ai in Chrome (or any supported browser), wait for CodexBar to pick up the `sessionKey` cookie.
3. Wait long enough for Anthropic to invalidate the session server-side (e.g. signing out in another browser, password rotation, prolonged inactivity / Cloudflare-side re-auth). The `sessionKey` cookie itself can still be present and within its own expiry window.
4. Open the CodexBar menu → Claude tab.

## Expected behaviour
The error card should expose a clear, one-click recovery: a menu item like `Re-login at claude.ai →` that opens `https://claude.ai/` in the user's default browser, where claude.ai will mint a fresh `sessionKey` cookie that CodexBar can pick up on the next refresh.

## Actual behaviour
A static error string is shown. The menu offers no actionable hint. As a result the user often assumes the bug is in CodexBar (the menu also feels "frozen" while the main thread is waiting on the 401 response, around 1–3 s per open), and may attempt unrelated fixes (clearing OAuth keychain entries, reinstalling the app, etc.) before realising the only working recovery is to re-login in the browser.

## Root cause
- Error produced at [`Sources/CodexBarCore/Providers/Claude/ClaudeWeb/ClaudeWebAPIFetcher.swift:72`](../blob/main/Sources/CodexBarCore/Providers/Claude/ClaudeWeb/ClaudeWebAPIFetcher.swift#L72) — `FetchError.unauthorized.errorDescription`.
- Reached when `fetchOrganizationInfo` ([line 442–443](../blob/main/Sources/CodexBarCore/Providers/Claude/ClaudeWeb/ClaudeWebAPIFetcher.swift#L442)) gets HTTP 401/403 from `claude.ai/api/organizations`.
- The error is then stored as a plain `String` in `UsageStore.errors[.claude]` (`Sources/CodexBar/UsageStore.swift:135`), and the menu renderer in [`Sources/CodexBar/StatusItemController+Menu.swift`](../blob/main/Sources/CodexBar/StatusItemController+Menu.swift) only paints it as a label.
- There is no plumbing for a per-provider recovery URL/action, so the UI has no way to render a recoverable surface.

## Proposed fix
A focused, additive change:

1. Add `recoveryURL: URL?` and `recoveryActionLabel: String?` computed properties on `ClaudeWebAPIFetcher.FetchError` for the three session-shaped cases (`.unauthorized`, `.noSessionKeyFound`, `.invalidSessionKey`), returning `https://claude.ai/` + `"Re-login at claude.ai"`.
2. Introduce a small `ProviderRecoveryAction { url, label }` value type in `CodexBarCore` with a `init?(inspecting: Error)` factory that recognises `ClaudeWebAPIFetcher.FetchError`.
3. Store recovery actions alongside the existing `errors` dictionary in `UsageStore` (new `errorRecoveryActions: [UsageProvider: ProviderRecoveryAction]`, observed in `menuObservationToken`, written / cleared inside `handleProviderFetchFailure`).
4. Inject a clickable `NSMenuItem` below the usage card in `StatusItemController+Menu.swift` (covering both the single-card and stacked-cards paths) when an actionable recovery action exists; click routes through `NSWorkspace.shared.open`.

A PR implementing this is in flight.

## Environment
- CodexBar 0.32.5 (universal, macOS arm64)
- macOS 14+
- Chrome v149 (browser used as the cookie source)

LeoLin990405:
## Summary
- show a localized `Re-login at claude.ai` contextual action for expired or missing automatically imported Claude web sessions
- reuse the existing provider login-action/menu infrastructure instead of introducing parallel recovery state
- respect effective credential routing so manual cookies, selected session accounts, OAuth, CLI, and API modes keep their existing actions
- cover direct session errors and unavailable `claude.web` planner attempts

Closes #1377.

## Validation
- `swift test --filter ClaudeWebRecoveryMenuTests`
- `swift test --filter UserFacingLocalizationCoverageTests`
- `make check`
- `swift test` — 3,427 tests in 393 suites
- structured Codex autoreview — clean after addressing auto-mode, manual-cookie, selected-account, and unavailable-web findings

No live provider auth, browser-cookie import, or Keychain probe was run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 - Antigravity: exclude model quotas without a remaining fraction from family summaries so they no longer mask tracked usage in the automatic menu-bar metric (#1369). Thanks @Martin-Hausleitner!
 - Claude: add bundled Fable 5 pricing, account for native 1-hour cache-write usage, and refresh Sonnet 4.6 full-context rates (#1368). Thanks @MoollaMore!
+- Claude: show a direct claude.ai re-login action when a configured web session expires or becomes invalid (#1377). Thanks @LeoLin990405!
 - Menu bar: defer data-refresh rebuilds until the tracked menu closes, avoiding multi-second WindowServer stalls with slower providers such as Grok (#1376). Thanks @jangisaac-dev!
 - Xiaomi MiMo: import automatic session cookies from Safari, Chrome variants, Firefox, and Edge instead of limiting discovery to Chrome (#1304). Thanks @Yuxin-Qiao!
 
--- a/Sources/CodexBar/Providers/Claude/ClaudeProviderImplementation.swift
+++ b/Sources/CodexBar/Providers/Claude/ClaudeProviderImplementation.swift
@@ -228,8 +228,35 @@ struct ClaudeProviderImplementation: ProviderImplementation {
     func loginMenuAction(context: ProviderMenuLoginContext)
         -> (label: String, action: MenuDescriptor.MenuAction)?
     {
+        if self.shouldOpenBrowserForWebSessionError(context: context) {
+            return ("Re-login at claude.ai", .loginToProvider(url: "https://claude.ai/"))
+        }
         guard self.shouldOpenTerminalForOAuthError(store: context.store) else { return nil }
         return ("Open Terminal", .openTerminal(command: "claude"))
+    }
+
+    @MainActor
+    private func shouldOpenBrowserForWebSessionError(context: ProviderMenuLoginContext) -> Bool {
+        let settings = context.settings.claudeSettingsSnapshot(tokenOverride: nil)
+        let source = settings.usageDataSource
+        guard source == .auto || source == .web,
+              settings.cookieSource == .auto,
+              let error = context.store.error(for: .claude)
+        else { return false }
+
+        let sessionErrors = [
+            ClaudeWebAPIFetcher.FetchError.unauthorized.localizedDescription,
+            ClaudeWebAPIFetcher.FetchError.noSessionKeyFound.localizedDescription,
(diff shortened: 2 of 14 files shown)


LeoLin990405:
Thanks for the catch — pushed `4c10de6b` to address both:

1. **Stale recovery actions** — the dual-dict approach was indeed fragile. Introduced `UsageStore.setProviderError(_:message:recovery:)` as the single write funnel and routed all ~20 existing `errors[provider]` mutation sites through it (`BackgroundRefresh`, `OpenAIWeb`, `Refresh`, `TokenAccounts`, log-dump paths in `UsageStore`). Clearing or replacing an error now always clears/replaces the paired recovery. `handleProviderFetchFailure` drops its manual two-line sync block in favour of a single helper call. Added a `ClaudeWebRefreshResilienceTests` case that walks exactly your scenario: web 401 sets recovery → unrelated overwrite drops it → clearing the error drops it again.
2. **Lint** — `StatusItemController+Menu` now uses `self.openProviderRecoveryURL` (`.swiftformat` enforces `--self insert`; the previous bare reference is what broke `lint-build-test` in run 27261317646).

`swift test --filter "ClaudeWeb"` 13/13, full suite 341/341 locally. CI should be green on the new run.

PR #1381, opened 2026-06-10
Fix Antigravity serve flakiness

enieuwy:
## Summary

Fix three bugs in the Antigravity provider pipeline and `codexbar serve` that cause the `/usage?provider=antigravity` endpoint to return intermittent errors.

### Bug 1: Local probe attaches to half-warm `agy` processes

`AntigravityStatusProbe` matched all running Antigravity processes including the `agy` CLI. The probe would find a stale or still-initializing `agy`, wait the full 8s timeout, and return an error — while `AntigravityCLIHTTPSFetchStrategy` also tried to manage the same process.

**Fix:** `ProcessScope` enum (`.ideOnly` / `.ideAndCLI`) on the probe. The local fetch strategy uses `.ideOnly` so it only matches the IDE language server. `isRunning()` keeps `.ideAndCLI` for accurate status reporting.

### Bug 2: `codexbar serve` kills warm `agy` sessions every refresh

`shouldResetSessionAfterFetch` returned `true` for `.cli` runtime, which includes `serve`. Every refresh cycle tore down a warm `agy` and the next poll had to cold-start a new one (~10s). If the cold start didn't finish before the request timeout, the serve endpoint returned an error.

**Fix:** New `persistsCLISessions` flag on `ProviderFetchContext`, defaulting `true` for app/serve and `false` for one-shot CLI. Warm sessions stay alive across fetches in persistent runtimes. Serve also installs SIGINT/SIGTERM handlers to tear down managed processes cleanly (no orphaned `agy`).

### Bug 3: `codexbar serve` returns fresh errors instead of last-good data

`shouldCacheServeResponse` returned `false` for error responses. A single Antigravity timeout or 502 replaced the last good payload with a fresh error, so any client polling the serve endpoint would see an error response instead of stale-but-valid data.

**Fix:** `CLIServeResponseCache` records the last good response per key and serves it in place of a failed refresh, bounded by `CachePolicy.staleTTL` (10× refresh interval, 5-minute floor). Disabled when `--refresh-interval 0`.

### Additional improvements

- `AntigravityStatusProbe.fetch()` falls back to `GetCommandModelConfigs` when `GetUserStatus` fails, matching the CLI strategy's error resilience.
- Cold-start warmup deadline raised from 5s to 15s for persistent runtimes.

### Verification

- New tests: probe `ideOnly` scope (2), persistent-session reset (1), serve stale fallback (2), stale TTL bounds (1). 142 focused tests pass.
- `make check`: 0 violations.
- Live e2e against rebuilt serve: cold-start 6.7s, cached ~30ms, transient-failure poll returned last-good payload, SIGTERM left no orphaned `agy`.

Closes #1313 (replaces with a simpler approach — no process manager needed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Localization: add Vietnamese as a selectable app language (#1247). Thanks @Yuxin-Qiao!
 
 ### Fixed
+- Antigravity: stop the local desktop probe from attaching to `agy` CLI processes, keep the warm managed `agy` session alive across `codexbar serve` refreshes (with teardown on SIGINT/SIGTERM), and serve the last good usage payload for transient refresh failures so external bar integrations (SketchyBar/Zellij) no longer flicker between data and errors.
 - Claude: remove transient ClaudeProbe session artifacts after CLI usage polls so background refreshes no longer fill Claude Code project history with CodexBar `/usage` sessions (#1301). Thanks @LPFchan and @matthewod11-stack!
 - Menu bar: keep z.ai overview rows with detail submenus in Overview so hovering quota details no longer recurses into a nested provider menu (#1279, fixes #1246). Thanks @RajvardhanPatil07!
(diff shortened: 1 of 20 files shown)


enieuwy:
Closing — this PR was built on top of #1313's branch, not main. The diff and description were misleading.

PR #498, opened 2026-03-09
Add Qwen and Doubao providers

LeoLin990405:
## Summary
- Add **Qwen (通义灵码)** provider — Alibaba's DashScope coding plan
- Add **Doubao (豆包)** provider — ByteDance's Volcengine Ark coding plan

Both are popular Chinese AI coding platforms with growing user bases.

## Implementation
- Follows the existing descriptor-driven provider architecture
- Uses minimal API probe (POST with `max_tokens: 1`) to read rate-limit headers
- Qwen: auto-detects `sk-sp-*` coding plan keys vs regular API keys, routes to appropriate endpoint
- Doubao: uses `/api/coding/v3/chat/completions` endpoint with `doubao-seed-2.0-code` model
- Graceful fallback when no rate-limit headers are returned (shows "Active" status)

## Test plan
- [ ] Build succeeds on macOS
- [ ] Enable Qwen in settings, enter DashScope API key → usage data displays
- [ ] Enable Doubao in settings, enter ARK API key → usage data displays
- [ ] CLI: `codexbar --provider qwen` and `codexbar --provider doubao` show usage
- [ ] Providers appear correctly in widget picker

PR #1383, opened 2026-06-10
Fix Doubao false 100% request usage

### LINKED ISSUE #1382: [Bug] Doubao usage always shows 100% used when Volcano Ark returns limit > 0 + remaining = 0
## Summary

Doubao usage card always shows **100% used** even when the API key is healthy and well within quota. Reported by @foobra on PR #498 (the original Doubao provider PR thread): _"Doubao rate show error, it always show 100%"_.

## Steps to reproduce

1. Configure a Doubao API key (`ARK_API_KEY` or via Preferences → Providers → Doubao).
2. Open the CodexBar menu bar → Doubao card.
3. Observe usage shown as `100%` (or "1000/1000 requests" style with no available remaining).
4. Confirm the key is actually healthy by issuing a normal Ark chat-completion call externally — it succeeds without 429.

## Expected behavior

When the key is valid and the request returns HTTP 200, the card should reflect actual usage. When Volcano Ark does not return a trustworthy `x-ratelimit-*-requests` pair, the card should fall back to the existing "Active — check dashboard for details" hint rather than misreporting 100% used.

## Actual behavior

The card displays 100% used despite the key being healthy and no real rate-limiting being applied.

## Root cause

[`DoubaoUsageFetcher.swift:33-36`](https://github.com/steipete/CodexBar/blob/main/Sources/CodexBarCore/Providers/Doubao/DoubaoUsageFetcher.swift#L33-L36):

```swift
if self.limitRequests > 0 {
    let used = max(0, self.limitRequests - self.remainingRequests)
    usedPercent = min(100, max(0, Double(used) / Double(self.limitRequests) * 100))
```

The math only guards `limitRequests > 0`. Volcano Ark on certain account tiers (notably unverified / personal keys) returns HTTP 200 with `x-ratelimit-limit-requests > 0` and `x-ratelimit-remaining-requests = 0` **without** actually rate-limiting the request — a real throttle would return 429. The current implementation reads this combination at face value, computes `used = limit`, and clamps to 100%.

The bug is provoked by Volcano's header reporting policy, not by genuine quota exhaustion. The "Active — check dashboard for details" fallback in [`DoubaoUsageFetcher.swift:37-43`](https://github.com/steipete/CodexBar/blob/main/Sources/CodexBarCore/Providers/Doubao/DoubaoUsageFetcher.swift#L37-L43) already exists for the `limitRequests == 0` case but is never reached when `limitRequests > 0 && remainingRequests == 0`.

## Proposed fix

Tighten the guard so the normal-math branch only runs when both `limitRequests > 0` **and** `remainingRequests > 0`. Otherwise fall through to the existing "Active — check dashboard for details" hint, which is the correct user-facing signal when headers are unreliable.

Additionally emit a `log.warning` when the unreliable-headers pattern is observed (200 + `limit > 0` + `remaining = 0`) so users hitting this can attach evidence from `~/Library/Logs/CodexBar/CodexBar.log` to bug reports.

Both behaviors are additive — they cannot cause real quota exhaustion to be hidden, because real exhaustion arrives as HTTP 429, which is handled separately in the existing 429 branch.

A PR implementing this is in flight.

## Environment

- CodexBar 0.32.5
- macOS 14+
- Volcano Ark endpoint: `https://ark.cn-beijing.volces.com/api/coding/v3/chat/completions`

LeoLin990405:
## Problem

@foobra reported on PR #498 that the Doubao usage card always shows 100% used. Volcano Ark can return HTTP 200 with a complete request-limit pair where `limit > 0` and `remaining = 0`, while later requests still succeed. Treating that single response as exhaustion produces a false 100% card.

Fixes #1382.

## Solution

- Confirm an ambiguous successful zero-remaining response with one immediate probe using the same model.
- If the confirmation is another HTTP 200 with zero remaining, mark the request-limit headers unreliable and show the existing `Active - check dashboard for details` fallback.
- If confirmation returns HTTP 429, preserve genuine exhaustion. A 429 carrying request-limit headers is also treated as request-bucket exhaustion.
- Require trustworthy request-limit evidence before rendering a percentage. Bare standalone 429 responses remain on the active fallback because they can describe another throttle.
- Propagate task and URL cancellation. If ordinary confirmation fails, conservatively preserve the initial exhausted snapshot.
- Log the repeated HTTP 200 unreliable-header pattern for diagnostics.

## Testing

- `swift test --filter Doubao`: 17 tests across 3 suites passed.
- Added transport-level coverage for repeated HTTP 200, HTTP 200 followed by explicit or headerless 429, standalone bare/partial 429, confirmation failure, and both cancellation forms.
- `make check`: SwiftFormat clean; SwiftLint 0 violations across 1026 files.
- Branch autoreview: clean, no accepted/actionable findings, correctness confidence 0.88.

GitHub CI is the final full-suite merge gate after this rewritten head.

## Scope

The exact Volcano account-tier matrix remains provider-controlled. The repeated-success warning keeps the condition observable without exposing credentials or account data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 0.32.6 — Unreleased
 
 ### Fixed
-- Doubao: stop misreading unreliable HTTP 200 request-limit headers as 100% used while preserving explicit HTTP 429 exhaustion (#1383). Thanks @LeoLin990405 and @foobra!
+- Doubao: confirm zero-remaining HTTP 200 request limits before falling back, preserving genuine exhaustion and avoiding false 100% usage (#1383). Thanks @LeoLin990405 and @foobra!
 - Antigravity: exclude model quotas without a remaining fraction from family summaries so they no longer mask tracked usage in the automatic menu-bar metric (#1369). Thanks @Martin-Hausleitner!
 - Claude: add bundled Fable 5 pricing, account for native 1-hour cache-write usage, and refresh Sonnet 4.6 full-context rates (#1368). Thanks @MoollaMore!
 - Claude: show a direct claude.ai re-login action when a configured web session expires or becomes invalid (#1377). Thanks @LeoLin990405!
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.32.6 — Unreleased
 
 ### Fixed
+- Doubao: stop misreading unreliable Volcano Ark rate-limit headers as 100% used; the card now falls back to the existing "Active — check dashboard for details" hint when HTTP 200 is returned with `x-ratelimit-limit-requests > 0` and `x-ratelimit-remaining-requests = 0`, which is the unreliable-headers pattern Volcano returns on some account tiers rather than a real throttle (#1382). Thanks @foobra!
 - Antigravity: exclude model quotas without a remaining fraction from family summaries so they no longer mask tracked usage in the automatic menu-bar metric (#1369). Thanks @Martin-Hausleitner!
 - Claude: add bundled Fable 5 pricing, account for native 1-hour cache-write usage, and refresh Sonnet 4.6 full-context rates (#1368). Thanks @MoollaMore!
 - Claude: show a direct claude.ai re-login action when a configured web session expires or becomes invalid (#1377). Thanks @LeoLin990405!
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 0.32.6 — Unreleased
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


steipete:
Maintainer rewrite pushed at `945b8d45`.

Proof:
- 17 focused Doubao tests pass, including repeated HTTP 200 zero headers, explicit/headerless confirmation 429, standalone bare/partial 429, cancellation, and confirmation failure.
- `make check` passes with SwiftFormat clean and SwiftLint 0 violations across 1026 files.
- Branch autoreview is clean with no accepted/actionable findings (correctness confidence 0.88).

GitHub CI is now the final merge gate.

@clawsweeper re-review

PR #1385, opened 2026-06-10
Add Japanese (ja) localization

naoterumaker:
## Summary

- Adds a complete Japanese (`ja`) localization: `ja.lproj/Localizable.strings` with all 1033 keys translated, following Apple's macOS Japanese UI conventions (キーチェーン for Keychain, メニューバー for menu bar, です・ます style for descriptions, concise noun phrases for buttons/labels).
- Adds `AppLanguage.japanese` so 日本語 appears in the Settings → General language picker.
- Adds the `language_japanese` label key to every existing language catalog (native-language renderings: Japonés, Японська, 日语, …).
- Extends `LocalizationLanguageCatalogTests` so the catalog completeness check covers the new key.

## Testing

- `swift test --filter Localization` — all 21 tests across 6 localization suites pass.
- Full `swift test` (3427 tests): the only failure is `MiniMaxMenuCardModelTests` "minimax token plan model shows weekly quota and points balance", which also fails on `main` on any machine whose system locale is Japanese — `subscriptionDateString` (MenuCardView.swift) formats with `Locale.current`, so the expected English date renders as 2027年5月18日. Pre-existing and unrelated to this change; happy to fix it in a follow-up if useful.
- Verified `ja.lproj` is included in the built `CodexBar_CodexBar.bundle` and the app runs with Japanese UI when the language is selected.


zenyr:
Note: I saw a regression that caused UI to ignore my system locale (Korean / English) and chose Japanese anyways.

PR #1402, opened 2026-06-10
Run cost-usage corpus scans off the Swift cooperative thread pool

ProspectOre:
## Summary
Run cost-usage corpus scans and persisted-cache decoding on a dedicated serial utility queue (`CostUsageScanExecutor`) instead of inline on Swift's cooperative thread pool, with task cancellation bridged into the scanner-level cancellation checks.

## Context
This targets the systemic freeze mechanism behind the #1387 reports (and the environment of #1392). `CostUsageScanner` / `PiSessionCostScanner` scans are synchronous and can run for minutes on large session archives. `CostUsageFetcher.loadTokenSnapshot` executed them inline inside cooperative-pool tasks, which violates the pool's forward-progress contract: while a scan runs, it owns a pool thread for minutes, and overlapping provider scans (startup force-refresh racing the hourly token timer) own several. Every other async continuation in the process — menu rebuild tasks, refresh completions — queues behind them, so the UI freezes while the main thread itself samples idle.

Two amplifiers make this a release-day event for heavy users:
- Cache schema bumps (e.g. the claude artifact moving v2 → v4 in 0.32.6-dev) invalidate the whole per-file scan cache, so the first launch after such a release rescans the full corpus.
- The reporter machine for this PR holds a 2.5 GB corpus (903 MB `~/.claude/projects`, 1.6 GB / 316 files `~/.codex/sessions`); the post-bump rescan ran 7+ minutes at full tilt.

Field sample of that storm (v0.32.6-dev pre-fix, 1 ms interval, 5 s window, redacted to frame counts): 2,437 samples in the claude scan and 1,807 in the codex scan **concurrently on two cooperative-pool threads**, sustained across every capture window for ~3 minutes, while menu opens visibly froze — with the main thread parked in `mach_msg` the whole time.

## Change
- New `CostUsageScanExecutor`: one serial `DispatchQueue` (`com.steipete.codexbar.cost-usage-scan`, utility QoS). `run` bridges Swift task cancellation into a `@Sendable` check handed to the scanners, and work cancelled while still queued resumes immediately with `CancellationError` instead of waiting behind an in-flight scan.
- `CostUsageFetcher.loadTokenSnapshot` moves the scan + vertex-fallback + pi-merge block onto the executor; `loadCachedCodexTokenSnapshot` moves its multi-megabyte cache JSON decode there too (it previously ran on the pool via `Task.detached`).
- Serialization also removes concurrent provider scans hammering the same disk.

## Validation
[...]

Issue #1392, opened 2026-06-10
Optimize Codex cost refresh policy for large history windows


TheAngryPit:
## Context

While investigating local CodexBar menu/CPU lag, I found a narrower cost-history refresh problem that may be contributing to the broader menu performance reports in #1321, #1360, and #1371.

Local setup/signal:

- CodexBar 0.32.5 (80)
- Cost summary enabled
- Cost history window set to 365 days
- `~/.codex/sessions` aggregate shape: ~13 GB across ~403 JSONL files
- A `sample` captured during a high-CPU spike pointed at the background cost path, not the main menu drawing path:
  - `UsageStore.refreshTokenUsage`
  - `CostUsageFetcher.loadTokenSnapshot`
  - `CostUsageScanner.loadCodexDaily`
  - `CostUsageScanner.scanCodexFile`
  - `CostUsageJsonl.scan`

This is related to #1016, but I think the shape is different enough to ask separately. #1016 was about raw JSONL scanner throughput / possible Swift vs Rust/SIMD optimization, and was closed as not worth pursuing without a real bottleneck. This issue is not proposing a scanner rewrite. It is about refresh policy for large history windows.

## Proposal

For large windows such as 365 days, make the steady-state refresh cheaper by separating:

- the selected display/history window, e.g. "show me 365 days"
- from the normal refresh cost, e.g. "only process new or modified local activity"

Possible behavior:

- Run a full scan when the cost history window is first enabled or expanded.
- Run a full scan when the Codex sessions root changes.
- Run a full scan / rebuild when the cache schema, parser source hash, pricing key, or other correctness-critical invalidator changes.
- During normal background refreshes, use the existing cache and scan only new or modified session files.
- Detect modified sessions by file metadata/cache state, not just by date, so resumed/forked/older sessions are still handled.
- Optionally run a periodic full reconciliation, for example weekly, to catch any drift.

This would keep 365-day history useful without making every normal refresh scale like a large history rebuild.

## Questions

- Does this refresh-policy direction fit how you want cost history to behave?
- Should manual "Refresh Now" force a full reconciliation, or should it also stay delta-based unless a setting/root/cache invalidator changed?
- If the direction sounds reasonable, can I claim this and open a small focused PR?

## Intended PR scope if accepted

- Keep this to Codex cost-history refresh/cache policy.
- No Rust/SIMD helper.
- No broad menu-rendering refactor.
- No provider/networking changes.
- Add focused tests for:
  - cached steady-state refresh
[...]

PR #1404, opened 2026-06-10
Resolve codex priority turns incrementally per refresh

ProspectOre:
## Summary
Resolve codex priority-turn metadata incrementally per refresh: accumulate the trace-database query result in process memory and only examine rows appended since the last call, instead of re-running the full-table double-`LIKE` scan on every refresh past the scan interval.

## Context
This is the dominant residual warm-refresh cost identified in #1392's follow-up measurements. `makeCodexRefreshPlan → codexPriorityTurns` runs

```sql
select ts, feedback_log_body from logs
where ts >= ? and ts < ?
  and (feedback_log_body like '%websocket request:%' or feedback_log_body like '%response.completed%')
```

against the codex CLI's `logs_2.sqlite` on every refresh once `refreshMinIntervalSeconds` has elapsed. The leading-`%` `LIKE`s force examining every row body in the window, so the cost grows with the database forever. On this machine the database is 762 MB / 322k rows and the query costs tens of seconds per tick; #1392's reporter measured 26–51 s per CLI refresh on a larger archive — `sample` puts essentially all of that time inside `codexPriorityTurns → sqlite3_step`.

## Change
- The `logs` table is append-only with `INTEGER PRIMARY KEY AUTOINCREMENT`, so rowids are monotonic and never reused. A per-database in-process memo keeps the accumulated turns plus a rowid cursor; each refresh runs the same filters with `rowid > ?` (satisfied by the integer PK, so it touches only new rows).
- Full rescan triggers: database file identity change (inode), `max(rowid)` shrinking (prune/replace), or the requested window expanding earlier than the accumulated coverage.
- Windows that end before today keep the original bounded one-shot query, so historical lookups never pay an open-ended scan and never populate the memo.
- The memo is process-local only — no cache schema change. The scanner source hash regeneration is included (one-time cache rebuild on update, which runs serialized off the cooperative pool after #1402).

## Validation
- `swift test --filter CostUsageScannerCodexPriorityTests` — 11 tests (3 new: incremental append pickup incl. late completed-model upgrade, window-expansion rescan, database-replacement rescan).
- `swift test --filter CostUsage` — 184 tests in 17 suites pass, suite wall time unchanged (the bounded historical path exists precisely because an open-ended first scan regressed a real-database test path 150× during development; that regression is fixed and the suite runs in ~4 s).
- `make check` — 0 violations; `git diff --check` clean.

## Runtime Proof
Timing harness (3 consecutive calls for the live 30-day window against the real 762 MB / 322k-row `logs_2.sqlite`, debug build, warm page cache; harness below, not committed). Result counts were identical in every run on both branches:

```text
# current main (cde92cfb): every call pays the full scan
TIMING_RESULT runs=[34.83 s, 22.38 s, 24.60 s] counts=[5, 5, 5]

# this branch: one full scan per process, then the rowid cursor
TIMING_RESULT runs=[8.51 s, 0.0015 s, 0.0006 s] counts=[5, 5, 5]
```

Steady-state per-tick cost drops from tens of seconds to ~1 ms, with byte-identical query filters and equal results. (Absolute first-call times vary with page cache state; a cold-cache first call measured 79 s, equally one-time.)

<details>
<summary>Timing harness (drop into Tests/CodexBarTests to reproduce)</summary>

```swift
import Foundation
import Testing
@testable import CodexBarCore

struct PriorityTurnsTimingHarness {
    @Test
    func `priority turns timing harness real db`() throws {
        let db = FileManager.default.homeDirectoryForCurrentUser
            .appendingPathComponent(".codex/logs_2.sqlite")
        try #require(FileManager.default.fileExists(atPath: db.path))
        let now = Date()
        let today = CostUsageScanner.CostUsageDayRange.dayKey(from: now)
        let since = CostUsageScanner.CostUsageDayRange.dayKey(
            from: Calendar.current.date(byAdding: .day, value: -29, to: now)!)
        let clock = ContinuousClock()
        var counts: [Int] = []
        var durations: [Duration] = []
        for _ in 0..<3 {
            let duration = clock.measure {
                counts.append(CostUsageScanner.codexPriorityTurns(
                    databaseURL: db,
                    sinceDayKey: since,
                    untilDayKey: today).count)
            }
            durations.append(duration)
        }
        print("TIMING_RESULT runs=\(durations) counts=\(counts)")
        #expect(counts[0] == counts[1])
        #expect(counts[1] == counts[2])
    }
}
```

</details>

## Honesty / scope notes
- The memo accumulates rows from the coverage start without an upper `ts` bound, so completed-model upgrades arriving after a turn's request row are applied even across ticks — marginally more complete attribution than the per-call query, identical for the standard advancing window.
- The first scan per process (app launch / each CLI invocation) still pays the full query; the CLI therefore does not benefit. Persisting the memo would need a cache schema change and is left out deliberately.
- The memo grows with the number of priority turns in the coverage window (small in practice; entries are turn-metadata only, no log bodies).


ProspectOre:
@clawsweeper re-review

ProspectOre:
@clawsweeper re-run

ProspectOre:
Addressed the review findings in `edbca5e0`:

- **[P2] Overlapping memo writeback**: writeback now goes through `storeCodexPriorityTurnsMemoIfNewer`, which discards a snapshot when the stored state already dominates it (same file identity, coverage starting at least as early, cursor at least as far). A slower writer with an older cursor can no longer replace newer accumulated state; non-dominated writers overwrite and converge through the existing rescan checks on the next refresh.
- **Regression test**: `overlapping refresh writeback cannot replace newer memo state` deterministically replays the race — a stale snapshot (older cursor, missing turn) is rejected, while a newer-cursor snapshot and an expanded-coverage snapshot still replace. 12/12 tests in `CostUsageScannerCodexPriorityTests` pass.
- **CHANGELOG**: removed the release-owned edit.
- Also fixed the Linux CLI build from the previous push (`c8597c2b`): the unconditional `import os` (for `OSAllocatedUnfairLock`, which only exists inside the `canImport(SQLite3)` blocks) is now guarded with `canImport(os)`. Parser hash regenerated.

@clawsweeper re-review

ProspectOre:
@clawsweeper re-review

ProspectOre:
Addressed the memory-growth finding from the 04:25 review in `a11eab3b`:

- **[P2] Unbounded completion metadata**: confirmed real — whether a turn is priority is unknowable when its `response.completed` row is parsed, so non-priority completions accumulated in `completedModelsByTurnID` for the process lifetime. Retention is now bounded (`codexPriorityCompletedModelRetentionLimit` = 4096): completions are tracked in insertion order and the oldest are evicted once the map exceeds the limit. Memory stays constant on long-running processes regardless of trace volume.
- **Ordering constraints preserved** (per the review's eviction-repair caveat): completion-before-request ordering within a batch and late model upgrades after a priority request are untouched — direct `turns` upgrades don't go through the bounded map, and eviction only drops the oldest entries.
- **Regression test**: `memo bounds retained completion metadata for non-priority turns` seeds limit+8 non-priority completed turns plus two priority requests, asserting the map holds exactly the limit, an evicted completion leaves the request's alias intact, and a retained completion still upgrades its request. 13/13 tests in `CostUsageScannerCodexPriorityTests` pass.

(The 04:27 review run appears to have failed on infrastructure and replaced the verdict with an off-meta placeholder — requesting a fresh review of the current head.)

@clawsweeper re-review

ProspectOre:
Apologies for the extra re-review trigger — I requested it after the earlier review run failed, not realizing the branch was being actively pushed to at the same time. I'll hold off on further review requests here while the branch is moving.

steipete:
Maintainer rebase/fixup pushed at `4de852ad` on top of current `main` (`dd8cf8b0`). The only source conflict was the generated parser hash after #1430; it was regenerated from the combined source.

Verification on the rebased head:
- `swift test --filter CostUsageScannerCodexPriorityTests` — 19 tests passed.
- `swift test --filter CostUsage` — 196 tests in 18 suites passed.
- `make check` — parser hash current; SwiftFormat clean; SwiftLint 0 violations across 1,049 files.
- `git diff --check` — clean.
- Local `$autoreview` against `origin/main` — no actionable findings; patch correct at 0.78 confidence.

Fresh CI is required before merge.

steipete:

PR #1406, opened 2026-06-10
Decode provider status feeds off the main actor

### LINKED ISSUE #1399: Merge Icons mode causes system-wide input freezes / beachballs on macOS 26 — WindowServer event-buffer overflow evidence (not an in-process hang)
## Summary

With **Merge Icons ON**, interacting with the merged dropdown intermittently freezes **the entire desktop** — beachballs, dropped clicks, keyboard input lockout in *other* apps, and Settings taking multiple attempts to open. With **Merge Icons OFF** (separate per-provider status items), the same configuration is completely smooth, even with provider status checks enabled.

I profiled this with Instruments attached to the running pristine 0.32.5 build while reproducing, and the result is unusual and probably why this class of bug keeps surviving fixes: **CodexBar's own process looks healthy the whole time** — no main-thread hang, near-idle CPU — while WindowServer logs show it **starving every other application's event delivery** during long merged-menu tracking sessions. The evidence below includes the WindowServer ground truth, which I haven't seen in any of the related issues.

I also tested the current 0.32.6-dev fixes (#1376, #1384, #1389, #1391, plus #1271) cherry-picked onto 0.32.5: **they do not fix it**.

## Environment

- CodexBar **0.32.5** (pristine notarized release build; also reproduced on 0.32.5 + the 0.32.6-dev backports listed below)
- macOS 26 (Darwin 26.5.1), Apple Silicon MacBook Pro
- Providers enabled: Codex (auto), Claude (auto), Gemini (OAuth), Antigravity — Gemini since disabled (deprecated CLI, expired token; see "Contributing churn" below)
- Refresh cadence 5 min; reproduced with provider status checks both ON and OFF

## Reproduction

1. Enable 3–4 providers, turn **Merge Icons ON**.
2. Open the merged dropdown; switch between providers and Overview; click around.
3. Within a couple of minutes: system-wide lag, beachballs, clicks in the menu and in *other apps* dropped, keyboard intermittently unresponsive. Opening Settings from the menu can take several attempts.
4. Turn **Merge Icons OFF** → identical configuration is fast and stable (status checks ON or OFF makes no difference to this).

## Evidence

Method: 150-second coordinated captures while reproducing — `xctrace record --template 'Time Profiler' --attach <pid>`, back-to-back 10s `/usr/bin/sample` captures, the app's own verbose unified log (`subsystem == "com.steipete.codexbar"`), and a 1 Hz CPU log.

### 1. It is NOT an in-process hang

During a 150s window in which I reproduced several "long beachballs":

- Instruments' **potential-hangs table is empty** (its only flagged event across all sessions was a single 324 ms microhang, attributed below).
- The main thread's largest single work item was **~40 ms**; CPU peaked at 44%.
- But the main thread sat **inside one `NSMenuTrackingSession.startRunningMenuEventLoop` for ~50 continuous seconds** (stack-sampled across five consecutive 10s windows) while the merged menu was open and being interacted with.

### 2. WindowServer shows the system-wide starvation (the key finding)

During the exact beachball windows, WindowServer logged **overflowed event datagram buffers being cleared for nine distinct client connections**, all marked `clientMayIgnoreEvents: 1` — the state every *other* app is placed in while a menu-tracking session is active:

```
14:27:56.285 E WindowServer [SkyLight] Clearing datagram buffer for cid 0x12449b. Client state: Data count: 16392 connectionIsUnresponsive: 0 clientMayIgnoreEvents: 1 clientIsSuspended: 0
14:28:22.370 E WindowServer [SkyLight] Clearing datagram buffer for cid 0x14d293. Client state: Data count: 16404 ... clientMayIgnoreEvents: 1 ...
14:28:22.370 E WindowServer [SkyLight] Clearing datagram buffer for cid 0x16405f. ...
14:28:23.547 E WindowServer [SkyLight] Clearing datagram buffer for cid 0xd80bb.  ...
14:28:23.548 E WindowServer [SkyLight] Clearing datagram buffer for cid 0x98283.  ...
14:28:26.455 E WindowServer [SkyLight] Clearing datagram buffer for cid 0x68647.  ...
14:28:32.401 E WindowServer [SkyLight] Clearing datagram buffer for cid 0x13c92f. ...
14:30:51.253 E WindowServer [SkyLight] Clearing datagram buffer for cid 0xa731b.  ...
14:30:52.831 E WindowServer [SkyLight] Clearing datagram buffer for cid 0xedd3b.  ...
```

These timestamps align exactly with the sampled menu-tracking residency and with the perceived freezes. Mechanism (hypothesis, strongly correlated): while the merged menu's tracking session is active, input routes exclusively to CodexBar and other clients' event queues back up; the long sessions plus in-tracking rebuild/refresh churn push those queues to the 16 KB buffer limit, WindowServer dumps them, and the whole desktop freezes — while CodexBar itself profiles as responsive. This would explain why per-process hang detectors keep coming up empty for this class of report.

Query to see it: `/usr/bin/log show --last 1h --predicate 'process == "WindowServer" AND eventMessage CONTAINS[c] "spinner" OR eventMessage CONTAINS[c] "unresponsive" OR eventMessage CONTAINS[c] "datagram"'`

### 3. Full refresh cycles run inside the open menu

The app's own debug log, at the exact second of the worst buffer-overflow burst (14:28:22), shows a complete provider refresh executing while the menu was open and tracking:

```
14:28:22.008 [gemini-probe] Token check (expiry=2026-05-21 ... ) → Token expired; attempting refresh
14:28:22.008 [subprocess]   Subprocess start (binary=ps label=antigravity-ps timeout=8.0)
14:28:22.120 [tty-runner]   PTY start (binary=claude cols=160 rows=50 timeout=5.0)
14:28:22.955 [gemini-probe] ERROR Could not extract OAuth credentials from Gemini CLI
```

So menu interaction triggers refreshes, refreshes spawn subprocesses (a Claude CLI PTY, `ps` probes, login-shell PATH captures) and drive `@Observable` updates → menu content invalidation/rebuild inside the tracking transaction.

### 4. Secondary finding: Google Workspace status decode runs on the main thread

With status checks ON, `UsageStore.parseGoogleWorkspaceStatus` decodes the full `https://www.google.com/appsstatus/dashboard/incidents.json` (**418 KB** when measured) **on the MainActor**, with a custom date strategy that allocates a fresh `ISO8601DateFormatter` per date field. Measured cost: **150–340 ms of main-thread stall per decode**, and it runs once per Google-status provider (Gemini *and* Antigravity) per refresh — refreshes that fire during menu interaction. The single 324 ms Instruments microhang across all sessions was exactly this stack (`refreshStatus → fetchWorkspaceStatus → JSONDecoder.decode([GoogleWorkspaceIncident])`). Note the feed is live data, so this cost grows whenever Google has a busy incident week — on all app versions simultaneously. This is real and worth fixing (decode off-main, shared formatter), but it is **not** the merged-icons freeze: that reproduces with status checks off.

### 5. Contributing churn observed (for completeness)

- The new login-shell PATH-capture retry (#1316 / `fa9b7570`) produced 3+ concurrent `$SHELL -l` spawns blocked in `wait4`/`read` during provider switching (login shell here is a fast 0.42 s, so this is amplification, not the cause).
- An expired Gemini CLI token failed its refresh on every cycle for weeks at error level — a doomed-retry loop with no user-visible surfacing. Worth a backoff/notification.

## What I tested that does NOT fix it

Cherry-picked onto 0.32.5 (all five applied cleanly; full test suite green) and re-tested with Merge Icons ON:

- #1376 `e97bfb0d` — defer tracked menu refresh rebuilds
- #1384 `7c0ed036` — reuse hosted chart submenus
- #1389 `4c092ce3` — defer copy-icon click work off the NSMenu tracking loop (macOS 26 beachball)
- #1391 `2ed15bf7` — defer overview provider rebuild
- #1271 `f5618341` — keychain prompt coordinator (unrelated; included for dev-build QoL)

**The system-wide freeze persisted unchanged.** Whatever the remaining driver is, it isn't covered by these.

## Related issues

This appears to be the same underlying defect reported (without the WindowServer evidence) in:

- #1374 — Stale merged dropdown rebuilds synchronously inside menuWillOpen on the first click after a background refresh tick
- #1360 — Opening CodexBar on the menu bar causes a temporary freeze
- #1387 — Severe UI lag and freezing
- #1329 — Menu bar icon hangs and remains memory-heavy
- #1308 / #1371 — Lag/stutter when 2+ providers are configured (i.e., when the merged switcher is in play)
- #1321 / #1347 — General performance complaints consistent with the above

The common thread: all reports involve the merged dropdown, and per-process diagnostics look clean — consistent with the WindowServer-side starvation mechanism above.

## Workaround

Merge Icons **OFF**. Separate per-provider status items are completely smooth with the identical provider set and status checks enabled.

## Available on request

Full Instruments `.trace` files (two 150s sessions: status checks on/off), the rolling `sample` captures, app debug logs, and the CPU timelines for all windows quoted above.

ProspectOre:
## Summary
Decode provider status feeds off the main actor and reuse shared ISO8601 formatters, removing a 150–340ms main-thread stall per Google-status provider per refresh.

## Context
Credit to @kcharlan's instrumentation in #1399 (finding 4): with status checks on, `UsageStore.parseGoogleWorkspaceStatus` decodes the full `https://www.google.com/appsstatus/dashboard/incidents.json` payload (hundreds of kilobytes live) **on the MainActor**, allocating a fresh `ISO8601DateFormatter` per decoded date field. Their Instruments capture measured 150–340ms of main-thread stall per decode, once per Google-status provider (Gemini *and* Antigravity) per refresh — and refreshes fire during menu interaction. The single flagged microhang across their sessions was exactly this stack. The payload is live data, so the cost grows whenever Google has a busy incident week, on all app versions simultaneously.

The same shape applies to the statuspage path (`fetchStatus`), which is also MainActor-isolated with per-field formatter allocation; it is included for consistency.

## Change
- `fetchStatus` and `fetchWorkspaceStatus` are explicitly `@concurrent nonisolated`, so feed fetching and decoding stay on the concurrent executor even if the package later adopts caller-inherited nonisolated async behavior. The pure parse helpers remain `nonisolated`; the caller (`refreshStatus`) already hops back to the main actor to publish results.
- Per-date-field `ISO8601DateFormatter()` allocations are replaced with shared lock-guarded formatters (`StatusISO8601FormatterBox`), mirroring the existing `CostUsageISO8601FormatterBox` pattern in CodexBarCore.

## Validation
- `swift test --filter GoogleWorkspaceStatus` — 4 tests pass, including a regression guard that records the executor immediately before JSON decoding begins.
- `swift test` — 3,532 tests in 402 suites pass.
- `make check` — 0 violations; `git diff --check` clean.
- autoreview — no accepted/actionable findings.

## Runtime Proof

**Live-feed artifact** (terminal capture; decode of the real `https://www.google.com/appsstatus/dashboard/incidents.json` payload fetched June 10, 2026 — 420,521 bytes, matching the ~418 KB measured in #1399; 5 iterations per strategy, same process, this branch's packaged toolchain):

```text
LIVE_FEED_PROOF payload_bytes=420521
LIVE_FEED_PROOF before(per-field formatters) ms=[69.9, 69.1, 69.4, 68.0, 71.1]
LIVE_FEED_PROOF after(shared formatters, full parse) ms=[25.8, 22.3, 22.1, 21.7, 21.6]
LIVE_FEED_PROOF gemini_status_indicator=Optional(CodexBar.ProviderStatusIndicator.major)
```

~3.1× faster on the live payload (which happened to carry a real active Gemini incident that parsed correctly through the new path), and the entire cost now lands on a background executor instead of the main thread. The committed regression observes the exact pre-decode boundary rather than the actor-backed transport stub:

```text
Test "fetchWorkspaceStatus decodes off the main thread when called from the main actor" passed
Test run with 4 tests in 2 suites passed
```

Synthetic stress check (699 KB fixture, 300 incidents x 4 dated updates, 10 iterations): 332.6–356.4 ms → 101.2–104.3 ms per decode, same ratio. The #1399 Instruments measurement (150–340 ms live) sits between these two payload sizes, consistent with both.

## Honesty / scope notes
- This fixes #1399's *secondary* finding only. The primary mechanism reported there — WindowServer event-buffer starvation during long merged-menu tracking sessions — is untouched by this PR and still needs a design that keeps refresh churn out of active tracking sessions.
- The decode still costs ~100ms of CPU somewhere; it is now off the UI's critical path rather than free.

<details>
<summary>Timing harness (drop into Tests/CodexBarTests to reproduce)</summary>

```swift
import Foundation
import Testing
@testable import CodexBar

struct StatusDecodeTimingHarness {
    @Test
    func `workspace status decode timing harness`() throws {
        // Synthetic feed shaped like incidents.json: ~300 incidents, each with several
        // updates and date fields, sized to match the measured live payload (~420 KB).
        var incidents: [[String: Any]] = []
        for index in 0..<300 {
            var updates: [[String: Any]] = []
            for u in 0..<4 {
                updates.append([
                    "when": "2026-05-\(String(format: "%02d", (index % 27) + 1))T0\(u):15:00+00:00",
                    "status": "SERVICE_INFORMATION",
                    "text": String(repeating: "status update body text ", count: 12),
                ])
            }
            incidents.append([
                "begin": "2026-05-\(String(format: "%02d", (index % 27) + 1))T00:00:00+00:00",
                "end": index % 7 == 0 ? NSNull() : "2026-05-\(String(format: "%02d", (index % 27) + 2))T00:00:00+00:00",
                "modified": "2026-05-\(String(format: "%02d", (index % 27) + 1))T12:00:00.123+00:00",
                "external_desc": String(repeating: "incident description ", count: 10),
                "severity": "medium",
                "status_impact": "SERVICE_INFORMATION",
                "affected_products": [["title": "Gemini", "id": "npdyhgECDJ6tB66MxXyo"]],
                "most_recent_update": updates[0],
                "updates": updates,
            ])
        }
        let data = try JSONSerialization.data(withJSONObject: incidents)
        struct Incident: Decodable { let begin: Date?; let end: Date?; let modified: Date?
            let updates: [Update]?
            struct Update: Decodable { let when: Date? } }
        func measure(_ strategy: JSONDecoder.DateDecodingStrategy) -> [String] {
            let clock = ContinuousClock()
            var ms: [Double] = []
            for _ in 0..<10 {
                let d = clock.measure {
                    let decoder = JSONDecoder()
                    decoder.keyDecodingStrategy = .convertFromSnakeCase
                    decoder.dateDecodingStrategy = strategy
                    _ = try? decoder.decode([Incident].self, from: data)
                }
                ms.append(Double(d.components.attoseconds) / 1e15 + Double(d.components.seconds) * 1000)
            }
            return ms.map { String(format: "%.1f", $0) }
        }
        let previousStrategy = JSONDecoder.DateDecodingStrategy.custom { decoder in
            let container = try decoder.singleValueContainer()
            let raw = try container.decode(String.self)
            let formatter = ISO8601DateFormatter()
            formatter.formatOptions = [.withInternetDateTime, .withFractionalSeconds]
            if let date = formatter.date(from: raw) { return date }
            formatter.formatOptions = [.withInternetDateTime]
            if let date = formatter.date(from: raw) { return date }
            throw DecodingError.dataCorruptedError(in: container, debugDescription: "bad date")
        }
        print("TIMING_RESULT payload_bytes=\(data.count)")
        print("BEFORE per-field formatters: \(measure(previousStrategy))")
        let clock = ContinuousClock()
        var ms2: [Double] = []
        for _ in 0..<10 {
            let d = clock.measure {
                _ = try? UsageStore.parseGoogleWorkspaceStatus(data: data, productID: "npdyhgECDJ6tB66MxXyo")
            }
            ms2.append(Double(d.components.attoseconds) / 1e15 + Double(d.components.seconds) * 1000)
        }
        print("AFTER shared formatters (full parse incl. filtering): \(ms2.map { String(format: "%.1f", $0) })")
    }
}
```

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Doubao: confirm zero-remaining HTTP 200 request limits before falling back, preserving genuine exhaustion and avoiding false 100% usage (#1383). Thanks @LeoLin990405 and @foobra!
 - Menu bar: defer pasteboard writes and copy feedback outside the `NSMenu` tracking callback so in-menu copy buttons no longer beachball on macOS 26 (#1388). Thanks @LeoLin990405!
 - Menu bar: defer merged status-icon redraws until the tracked menu closes while preserving animation lifecycle and quota-warning timing, reducing WindowServer churn during long menu sessions (#1409, fixes #1399). Thanks @kiranmagic7!
+- Provider status: decode status feeds on the concurrent executor and reuse ISO8601 formatters, removing a measured main-thread stall during refreshes (#1406). Thanks @ProspectOre!
 - Menu bar: keep one stable width across merged provider tabs and resize every hosted card row to AppKit's final menu width so provider switching no longer leaves a widened menu with inset submenu arrows (#1410).
 - Menu bar: defer Overview-row provider transitions out of AppKit's click callback so opening provider detail no longer performs a full synchronous menu rebuild (#1325).
 - Menu bar: open cached menus immediately after data-only invalidations, then refresh missing or stale provider data asynchronously without queuing redundant work on close (#1398). Thanks @joshuavial!
--- a/Sources/CodexBar/UsageStore+Status.swift
+++ b/Sources/CodexBar/UsageStore+Status.swift
@@ -44,6 +44,7 @@ extension UsageStore {
     /// Status feeds decode off the main actor: the Google Workspace incidents payload alone
     /// can be hundreds of kilobytes and cost 150-340ms to decode (#1399), and these helpers
     /// touch no store state.
+    @concurrent
     nonisolated static func fetchStatus(
         from baseURL: URL,
         transport: any ProviderHTTPTransport = ProviderHTTPClient.shared)
@@ -84,9 +85,11 @@ extension UsageStore {
             updatedAt: response.page?.updatedAt)
     }
 
+    @concurrent
     nonisolated static func fetchWorkspaceStatus(
         productID: String,
-        transport: any ProviderHTTPTransport = ProviderHTTPClient.shared)
+        transport: any ProviderHTTPTransport = ProviderHTTPClient.shared,
+        beforeDecoding: (@Sendable () -> Void)? = nil)
         async throws -> ProviderStatus
     {
         guard let url = URL(string: "https://www.google.com/appsstatus/dashboard/incidents.json") else {
@@ -95,6 +98,7 @@ extension UsageStore {
         var request = URLRequest(url: url)
         request.timeoutInterval = 10
         let (data, _) = try await transport.data(for: request)
+        beforeDecoding?()
         return try Self.parseGoogleWorkspaceStatus(data: data, productID: productID)
     }
 
--- a/Sources/CodexBar/UsageStore+Status.swift
+++ b/Sources/CodexBar/UsageStore+Status.swift
@@ -1,8 +1,50 @@
 import CodexBarCore
 import Foundation
 
+/// Shared, lock-guarded ISO8601 formatters for status feeds. Allocating a fresh
+/// `ISO8601DateFormatter` per decoded date field is a measurable share of decoding the
+/// Google Workspace incidents feed, which can run to hundreds of kilobytes (#1399).
+private final class StatusISO8601FormatterBox: @unchecked Sendable {
+    let lock = NSLock()
+    let withFractional: ISO8601DateFormatter = {
+        let fmt = ISO8601DateFormatter()
+        fmt.formatOptions = [.withInternetDateTime, .withFractionalSeconds]
+        return fmt
+    }()
+
+    let plain: ISO8601DateFormatter = {
+        let fmt = ISO8601DateFormatter()
+        fmt.formatOptions = [.withInternetDateTime]
+        return fmt
+    }()
+}
+
+private enum StatusFeedDateParser {
+    static let box = StatusISO8601FormatterBox()
+
+    static func parse(_ text: String) -> Date? {
+        self.box.lock.lock()
+        defer { self.box.lock.unlock() }
+        return self.box.withFractional.date(from: text) ?? self.box.plain.date(from: text)
+    }
+
+    static func decodingStrategy() -> JSONDecoder.DateDecodingStrategy {
+        .custom { decoder in
+            let container = try decoder.singleValueContainer()
+            let raw = try container.decode(String.self)
+            guard let date = StatusFeedDateParser.parse(raw) else {
+                throw DecodingError.dataCorruptedError(in: container, debugDescription: "Invalid ISO8601 date")
+            }
+            return date
+        }
+    }
+}
+
 extension UsageStore {
-    static func fetchStatus(
+    /// Status feeds decode off the main actor: the Google Workspace incidents payload alone
+    /// can be hundreds of kilobytes and cost 150-340ms to decode (#1399), and these helpers
+    /// touch no store state.
+    nonisolated static func fetchStatus(
         from baseURL: URL,
         transport: any ProviderHTTPTransport = ProviderHTTPClient.shared)
         async throws -> ProviderStatus
@@ -32,16 +74,7 @@ extension UsageStore {
         }
 
         let decoder = JSONDecoder()
-        decoder.dateDecodingStrategy = .custom { decoder in
-            let container = try decoder.singleValueContainer()
-            let raw = try container.decode(String.self)
-            let formatter = ISO8601DateFormatter()
-            formatter.formatOptions = [.withInternetDateTime, .withFractionalSeconds]
-            if let date = formatter.date(from: raw) { return date }
-            formatter.formatOptions = [.withInternetDateTime]
-            if let date = formatter.date(from: raw) { return date }
-            throw DecodingError.dataCorruptedError(in: container, debugDescription: "Invalid ISO8601 date")
-        }
+        decoder.dateDecodingStrategy = StatusFeedDateParser.decodingStrategy()
 
         let response = try decoder.decode(Response.self, from: data)
         let indicator = ProviderStatusIndicator(rawValue: response.status.indicator) ?? .unknown
@@ -51,7 +84,7 @@ extension UsageStore {
             updatedAt: response.page?.updatedAt)
     }
 
-    static func fetchWorkspaceStatus(
+    nonisolated static func fetchWorkspaceStatus(
         productID: String,
         transport: any ProviderHTTPTransport = ProviderHTTPClient.shared)
         async throws -> ProviderStatus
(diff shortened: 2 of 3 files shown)


ProspectOre:
@clawsweeper re-review

ProspectOre:
@clawsweeper re-review

steipete:
Landed as `02c94032c79a23b95658b37c2f85ea59d8de713a`.

Proof:
- `swift test --filter GoogleWorkspaceStatus` — 4 tests pass
- `swift test` — 3,532 tests in 402 suites pass
- `make check` and `git diff --check` — clean
- autoreview — no accepted/actionable findings
- exact-head CI — all required checks green
- live 420,521-byte feed decode improved from ~68–71ms to ~22–26ms and now runs off the main actor

Thanks @ProspectOre.

PR #1384, opened 2026-06-10
Reduce hosted chart submenu stalls

steipete:
## Summary
- skip the duplicate hosted-chart fitting pass on submenu open
- retain unchanged hosted chart views across refreshes using content, account, time-range, and width signatures
- precompute plan utilization chart models once per hosted view and replace quadratic duplicate merging
- preserve z.ai dynamic height updates while avoiding unnecessary chart reconstruction

Closes #1379.

## Validation
- `swift test --filter StatusMenuHostedSubmenuRefreshTests`
- `swift test --filter PlanUtilizationHistoryChartMenuViewTests`
- `swift test --filter UsageStorePlanUtilization`
- `swift test --filter UsageStorePlanUtilizationClaudeIdentityTests`
- `make check`
- `swift test` — 3,430 tests in 394 suites passed
- structured Codex autoreview — clean after fixing z.ai height/time invalidation and account-scoped utilization invalidation

No live provider, browser-cookie, or Keychain probe was run.

PR #1416, opened 2026-06-11
perf: skip storage footprint republish when scan is unchanged

soohanpark:
## What

`UsageStore.applyStorageFootprints` reassigned the observable `providerStorageFootprints` on **every** scan, even when the scanned bytes were identical. Storage scans run on every menu open (`refreshStorageFootprintsForOverview`) and roughly every 5 minutes, so each no-op scan woke the `menuObservationToken` observation → `invalidateMenus`, which clears the merged-menu content caches (`clearMergedSwitcherContentCaches`) and schedules avoidable rebuild work — pure churn when nothing the user sees changed.

The reason a guard wasn't already effective: `ProviderStorageFootprint` is `Equatable`, but it carries `updatedAt: Date`, stamped fresh on each scan, so two scans of identical on-disk data never compare equal.

## Change

- Add `ProviderStorageFootprint.hasSameContents(as:)` — value equality over everything **except** `updatedAt` (which is never surfaced to the UI).
- In `applyStorageFootprints`, reuse the existing footprint when only the timestamp would change, then guard the observable write with `!=`. An unchanged scan is now a true no-op and fires no observation.

## Test

Adds `repeated identical storage refresh does not republish observable footprints` — asserts (via `withObservationTracking`) that a second scan over identical on-disk data does not republish `providerStorageFootprints`. Fails before the change, passes after.

Full `swift test`: green (the only failure on my machine is the pre-existing locale-dependent MiniMax date assertion at `MenuCardModelTests.swift:735`, unrelated to this change — it hardcodes an en_US date string).

## Manual verification (release build)

Built and ran a release build (`Scripts/compile_and_run.sh`) and exercised the merged Overview menu — repeated open/close, Claude↔Codex tab switching, and chart-row hovers — while streaming the app's own os_log (`subsystem == "com.steipete.codexbar"`). Both `logMenuOperationDurationIfSlow` (≥150 ms) and `logChartRenderDurationIfSlow` (≥50 ms) log via the same `.warning` path, which persists to the log store, so a slow op would have been captured.

Over the session:
- `slow menu operation` warnings (open/close/tab-switch blocking the main thread ≥150 ms): **0**
- crashes / hangs: **0**
- `slow chart render`: **1** — `55.1 ms` on the first hydrate of `usageHistoryChart` (first hover only; subsequent hovers are skipped by the existing render-signature guard).

So the open/close freeze does not reproduce on this branch, and the one remaining sub-threshold hitch is the inherent first-build cost of a SwiftUI chart (not addressed here, and not cacheable on first render).

## Scope / notes

This is a small, low-risk churn-reduction. It is **not** the main menu open/close freeze fix — recent `main` already addresses that via hosted-chart render signatures (#1384), card hosting-view harvest/recycling, height caching, and deferred rebuilds (#1376). This just removes one remaining source of needless menu invalidation that those caches then have to recover from.

## Maintainer proof

Rebased head `6412dec8` passes 16 focused storage tests, `make check`, `git diff --check`, and clean autoreview. The focused Observation regression performs two identical scans over a real temporary Codex directory and verifies that the second scan does not republish observable storage state.

A freshly packaged release bundle rendered the storage row in Overview and completed three Peekaboo menu close/open cycles without crash, hang, `slow menu operation`, or `slow chart render` diagnostics. The temporary storage-display preference was restored afterward.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Menu bar: move the highlighted Overview provider with trackpad or mouse-wheel scrolling while preserving native submenu and keyboard behavior (#1436). Thanks @joshuavial!
 
 ### Fixed
+- Menu bar: avoid republishing unchanged provider storage footprints so background scans no longer trigger unnecessary menu observation work (#1416). Thanks @soohanpark!
 - Claude: explain that an unauthorized Web session requires signing in at claude.ai or refreshing imported cookies (#1287). Thanks @LeoLin990405!
 - CLI server: reload provider config for every usage and cost request, invalidate config-dependent cache entries, and prune expired config variants without restarting `codexbar serve`. Thanks @enieuwy!
 - Menu bar: reserve quota-bar space consistently across Overview and provider switcher segments so selection no longer changes segment height (#1445). Thanks @Zihao-Qi!
@@ -17,7 +18,6 @@
 - Cost usage: memoize Codex priority-turn trace metadata incrementally so warm refreshes scan only appended rows instead of rescanning large trace databases (#1404). Thanks @ProspectOre!
 - Security: reject insecure or malformed MiniMax and Alibaba endpoint overrides while preserving valid custom HTTPS deployments (#1269). Thanks @Hinotoi-agent!
 - Security: reject insecure or malformed OpenRouter, Codebuff, Groq, and ElevenLabs endpoint overrides before sending provider credentials (#1256). Thanks @Hinotoi-agent!
-- Menu bar: avoid invalidating menu content when repeated provider storage scans return unchanged values (#1416). Thanks @soohanpark!
 
 ## 0.33.0 — 2026-06-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Cost usage: memoize Codex priority-turn trace metadata incrementally so warm refreshes scan only appended rows instead of rescanning large trace databases (#1404). Thanks @ProspectOre!
 - Security: reject insecure or malformed MiniMax and Alibaba endpoint overrides while preserving valid custom HTTPS deployments (#1269). Thanks @Hinotoi-agent!
 - Security: reject insecure or malformed OpenRouter, Codebuff, Groq, and ElevenLabs endpoint overrides before sending provider credentials (#1256). Thanks @Hinotoi-agent!
+- Menu bar: avoid invalidating menu content when repeated provider storage scans return unchanged values (#1416). Thanks @soohanpark!
 
 ## 0.33.0 — 2026-06-11
 
--- a/Sources/CodexBar/UsageStore+ProviderStorage.swift
+++ b/Sources/CodexBar/UsageStore+ProviderStorage.swift
@@ -177,9 +177,25 @@ extension UsageStore {
         updatedAt: Date)
     {
         let providerSet = Set(providers)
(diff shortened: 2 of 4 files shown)


steipete:
Rebased onto current `main` (`3c2d23d1`) and pushed reviewed head `6412dec8`.

Validation:
- `swift test --filter ProviderStorageFootprintTests`: 16 tests passed, including two identical real temporary-directory scans with Observation tracking
- `make check`: passed, 0 violations
- autoreview: clean, confidence 0.86
- `./Scripts/compile_and_run.sh`: release bundle packaged, signed, launched, and remained running
- `git diff --check`: passed

Peekaboo/runtime proof with provider storage display temporarily enabled:
- Overview rendered the scanned storage row (`Storage: 44 GB`)
- repeated three close/open cycles completed without crash or hang
- redacted unified-log query found 0 `slow menu operation` or `slow chart render` diagnostics during the exercise
- screenshot: `/tmp/codexbar-pr1416-storage.png`
- the user setting was restored to disabled and the app restarted afterward

The focused regression is the direct proof for the optimization boundary: an unchanged second scan does not republish `providerStorageFootprints`, while refresh bookkeeping still advances.

@clawsweeper re-review

steipete:
Rebased onto current `main` and force-pushed exact reviewed head `a262b299e4d870015791dbb20243c8a5753c09eb`.

Proof:
- `swift test --filter ProviderStorageFootprintTests` — 16/16 passed
- `make check` — SwiftFormat/SwiftLint clean
- branch autoreview against current `origin/main` — no actionable findings

Waiting for exact-head CI before merge.

steipete:
Exact-head maintainer proof for `a262b299e4d870015791dbb20243c8a5753c09eb`:

- Rebased onto current `origin/main` (`b5edc46c0872cd0ecd431cbafc11ccb6062198ed`).
- `swift test --filter ProviderStorageFootprintTests`: 16/16 passed.
- `make check`: passed; SwiftFormat and SwiftLint report zero violations.
- Branch autoreview against current `origin/main`: no accepted/actionable findings.
- Contributor authorship remains in the branch history; changelog credits `@soohanpark`.
- Exact-head CI: https://github.com/steipete/CodexBar/actions/runs/27370769398

steipete:
Additional exact-head proof for `a262b299e4d870015791dbb20243c8a5753c09eb`:

- Full `swift test`: 3,634 tests across 412 suites passed; zero failures.

steipete:
Landed in [`ad2a035f`](https://github.com/steipete/CodexBar/commit/ad2a035f66014985a11dd6b360cbb9f6d4e11d15).

Proof on the rebased exact head `ce5b5951`:
- `swift test --filter ProviderStorageFootprintTests` — 16 passed
- `make check` — clean
- autoreview — clean (0.88)
- GitHub CI — macOS lint/full tests, Linux x64/arm64, and GitGuardian all green

Thanks @soohanpark!

steipete:
Validated rebased exact head `ce5b5951996f2007ae0a16decfd5c2d2ce6ea3c6`.
Landed as `ad2a035f66014985a11dd6b360cbb9f6d4e11d15`.

- `swift test --filter ProviderStorageFootprintTests` (16 tests)
- regression test runs two real Codex storage scans over a temporary session tree
- second unchanged scan emits no `providerStorageFootprints` Observation update, so `menuObservationToken` is not awakened
- `swift test` (3,634 tests across 412 suites) on the source-identical rebased tree
- `make check`
- branch autoreview clean
- released/fictitious model-name gate clean
- exact-head GitHub CI is the authoritative full-suite gate

PR #1421, opened 2026-06-11
Persist the codex priority-turns memo across launches

ProspectOre:
> **Stacked on #1404** — this branch contains #1404's commits plus one new commit (`e7f3e20e`). Once #1404 lands I'll rebase so only the persistence change remains in the diff.

## Problem

The priority-turns memo introduced in #1404 makes live refreshes incremental, but it is process-local. Every app launch and every CLI invocation pays one full `logs_2.sqlite` scan (tens of seconds on large trace databases) before incremental refreshes resume — the disclosed P1 limitation from #1404's review.

## Change

Persist the memo to disk following the existing cost-usage cache conventions:

- New artifact `cost-usage/codex-priority-turns-v1.json` with a `producerKey` embedding the parser hash (`codex:pt:p<hash>`), so a parser change falls back to one full rescan — same invalidation contract as the codex cost cache.
- Atomic tmp-file + `replaceItemAt` writes; corrupted or version-mismatched artifacts are ignored.
- Seeding happens once per process **through the monotonic store** from #1404, so persisted state can never regress a scan that already ran; the cursor/coverage/file-identity rescan checks all still apply to the seeded state (a replaced or truncated database invalidates exactly as before).
- A dirty flag gates writes: the artifact is rewritten only after a scan actually advances the memo.
- Load and persist are wired at the single `codexPriorityTurns` call site in the codex scan, which runs on the `CostUsageScanExecutor` serial queue (#1402), so the synchronous file IO stays off the cooperative pool and the main actor.
- Memo contents are metadata only (thread/turn IDs, model names, timestamps) — no prompt or response bodies, matching what the parsers already strip.

## Runtime Proof (real database, redacted)

Two separate `swift test` processes against the live trace database — a true relaunch boundary. Cold first, then warm from the persisted artifact:

```text
=== RUN 1: cold (no artifact) ===
PROOF[cold] db=logs_2.sqlite dbBytes=798629888
PROOF[cold] diskLoad=3.3ms scan=12313.6ms persist=2.7ms
PROOF[cold] priorityTurns=5 memoCursor=387421516 retainedCompletions=288 artifactBytes=28698

=== RUN 2: warm (fresh process, seeded from artifact) ===
PROOF[warm] db=logs_2.sqlite dbBytes=798629888
PROOF[warm] diskLoad=1.8ms scan=2.8ms persist=10.5ms
PROOF[warm] priorityTurns=5 memoCursor=387421705 retainedCompletions=288 artifactBytes=28698
```

- Relaunch cost: **12,313.6 ms → 4.6 ms** (1.8 ms artifact load + 2.8 ms incremental scan) on a 798 MB database.
- The warm run's cursor advanced past the cold run's (387421705 > 387421516): rows appended between the two runs were picked up incrementally from the persisted cursor, with identical turn results.
- Artifact size: 28 KB for a 798 MB database.

## Validation

- `swift test --filter CostUsageScannerCodexPriority` — 16 tests pass (13 from #1404 + 3 new: relaunch roundtrip with cursor continuity, producer-key/version mismatch discard, corrupted artifact ignored).
- `make check` — 0 violations in 1043 files.
- `Scripts/regenerate-codex-parser-hash.sh check` — hash committed (the new source file participates in the hash, so this release triggers the usual one-time rescan).


ProspectOre:
@clawsweeper re-review

steipete:
Rebased onto the reviewed #1404 stack and force-updated with an exact lease.

Maintainer fixups:
- keep memo persistence portable on Linux by using the existing lock pattern instead of `OSAllocatedUnfairLock`
- normalize persisted observation IDs across process launches
- honor `forceRescan` by invalidating the selected trace-database memo before rebuilding it
- add relaunch, stale/corrupt artifact, and forced-cursor-discard regressions

Proof:
- `swift test --filter 'CostUsageScanner(Priority|CodexPriority)Tests'` (37 tests passed)
- `swift test --filter CostUsage` (200 tests passed)
- `make check` (parser hash current, SwiftFormat clean, SwiftLint 0 violations)
- structured autoreview against the #1404 parent: clean after fixing the force-rescan finding

ProspectOre:
Linux CI failure on `0874abaa` is the `canImport(SQLite3)` guard: `invalidateCodexPriorityTurnsMemo` is only defined inside the SQLite3 block, but the `forceRescan` call site in `CostUsageScanner.swift:2197` compiles unconditionally — needs a no-op stub in the `#else` branch alongside the load/persist stubs.

steipete:

Issue #1387, opened 2026-06-10
CodexBar app is experiencing severe UI lag, and freezing


eyaldr:

eyaldr:
CodexBar app is experiencing severe UI lag, freezing

isbkch:
started a few days ago. I was hoping yesterday's release would fix it, but nope

benshawuk:
Same issue here - beachballs within the app, but even freezes other running apps until it recovers..  Only started happening a couple of days ago.  Let me know if you want any logs/troubleshooting to be done.

mariusvniekerk:
so many beachballs, borderline unusable

hasanaydins:
+++

WinDaLone:
Same here

boundsj:
Same issue

kdu3142:
Same here, EXTREMELY laggy.

rkwap:
same here. feels so frustrated that i am looking to install an alternative now.

benshawuk:
> same here. feels so frustrated that i am looking to install an alternative now.

Yeah I just vibed my own! :) 

<img width="297" height="144" alt="Image" src="https://github.com/user-attachments/assets/005b7e05-f676-41d7-82b7-40629569de85" />

Simple but it works, without bloat! :)

hsiangfeng:
i same issue.

nRewik:
same here

ifussa:
same  issue

ProspectOre:
update to the just released 0.33

steipete:
CodexBar 0.33.0 is now available: https://github.com/steipete/CodexBar/releases/tag/v0.33.0

It includes the dedicated cost-scan executor and the recent menu tracking, hosted-view reuse, refresh scheduling, icon redraw, and provider-switch performance fixes associated with this report.

Please verify the issue on exactly 0.33.0. If it still reproduces, add:
- the provider/view and exact click or switch sequence;
- macOS version;
- an Activity Monitor Sample Process capture taken during the beachball;
- whether quitting/reopening CodexBar changes it.

That evidence will distinguish a remaining main-thread menu path from provider I/O, Keychain/browser access, or WindowServer behavior. Keeping this open for current-version proof.

Skeegan123:
Still early as I just downloaded the update but it is finally starting to feel better! Everything is feeling very responsive right now compared to before

steipete-oai:
[...]

PR #1422, opened 2026-06-11
Split the codex parser hash into session and priority-turns domains

ProspectOre:
> **Stacked on #1404 and #1421** — the diff includes their commits until they land; the change specific to this PR is the last commit. I'll rebase as the stack merges.

## Problem

`CodexParserHash` is one hash over every non-Claude source in `Vendored/CostUsage/`, and it producer-keys the codex session cost cache. Any change to codex **priority-turns** parsing therefore invalidates the entire session cache and forces every user to re-parse their full `~/.codex/sessions` corpus on update — even though session JSONL parsing didn't change a byte.

This isn't hypothetical: #1404 and #1421 each touch only `*CodexPriority*` sources, and each would trigger a full session re-parse for every user on release day. The 0.32.x "release-day lag wave" pattern (#1387) was exactly this class of invalidation.

## Change

Split the generated hash into two domains:

- `CodexParserHash.sessions` — non-Claude, non-priority-turns sources → producer-keys the codex session cost cache (`codex:cu:p…`)
- `CodexParserHash.priorityTurns` — `*CodexPriority*` sources → producer-keys the priority-turns memo artifact from #1421 (`codex:pt:p…`)

Correctness is preserved by machinery that already exists: when priority parsing changes, the priority artifact invalidates and re-derives, and the changed turn set flows through `codexPriorityTurnKeys` / `changedPriorityTurnIDs`, which already re-derives **only the session files whose cached `codexTurnIDs` intersect the changed turns** (`cachedCodexFileNeedsPriorityRescan`). Files untouched by priority changes keep their cache.

A session-parsing change still invalidates the session cache exactly as before.

## Runtime Proof (mechanism, live script run)

Editing a priority-turns source and regenerating: the sessions hash is byte-identical, only the priority domain moves — and reverting restores it:

```text
=== baseline (current sources) ===
    static let sessions = "91ab91ab2dd5246b"
    static let priorityTurns = "6d64628ee98b719a"
=== after editing a priority-turns source ===
    static let sessions = "91ab91ab2dd5246b"
    static let priorityTurns = "3db2184d58c628d5"
=== reverted ===
    static let sessions = "91ab91ab2dd5246b"
    static let priorityTurns = "6d64628ee98b719a"
```

Under the previous single-domain hash, that same edit changed the global `value` and would have producer-key-invalidated the session cost cache (observable in this repo's history: every `CodexParserHash` bump in #1404's commits was a forced full re-parse for all users).

## Migration cost

One-time: this PR changes the session producer key format itself, so the release that ships it triggers one final full re-parse — after which priority-only parser changes stop causing them.

## Validation

- `swift test --filter 'CostUsageCacheTests|CostUsageScannerCodexPriority'` — all pass, including the updated domain-key test asserting `codex:cu:p<sessions>` / `codex:pt:p<priorityTurns>`.
- `Scripts/regenerate-codex-parser-hash.sh check` — passes; check mode validates both domains.
- `make check` — clean.


ProspectOre:
Addressed the [P2] dependency-completeness finding in `1351fac6`:

**Fix**: the priority hash now covers **all** non-Claude sources (superset), while the sessions hash stays the non-priority subset. This is dependency-complete by construction — no hand-maintained dependency list to drift: any shared-helper change (timestamps, day ranges, scanner core) necessarily moves the priority hash. The asymmetry is deliberate: the session cache is the expensive artifact to rebuild, while a priority memo rebuild is one bounded background scan, so the memo conservatively invalidates on any scanner change. The original win is intact — priority-only edits leave the session cache untouched.

**Coverage for the three edit classes** (live script runs, restored to baseline after each):

```text
=== baseline ===
sessions="91ab91ab2dd5246b"  priorityTurns="a3e303f66e535cc6"
=== priority-only edit (CostUsageScanner+CodexPriority.swift) ===
sessions="91ab91ab2dd5246b"  priorityTurns="bab7fac12782e16b"   ← session cache survives
=== session-parsing edit (CostUsageScanner+CodexFastJSON.swift) ===
sessions="9067af907244a5f7"  priorityTurns="54a1b7c01bb19370"   ← both invalidate
=== shared-dependency edit (CostUsageScanner+Timestamp.swift) ===
sessions="07791ca9d8978ad1"  priorityTurns="551e3ea431b0d20e"   ← both invalidate (the finding's gap, now closed)
=== restored ===
sessions="91ab91ab2dd5246b"  priorityTurns="a3e303f66e535cc6"
```

On the rebase note: agreed — I'll rebase to the final small diff as soon as the prerequisite stack (#1404, #1421) lands.

@clawsweeper re-review

steipete:

PR #1423, opened 2026-06-11
Add CI regression gates for cost-usage scan storms

ProspectOre:
> **Stacked on #1404** (uses its memo test helpers and cursor accessor); contains its commits until it lands, then I'll rebase. Independent of #1421/#1422.

## Problem

The repo has shipped two distinct cost-usage scan-storm regressions: refreshes that re-parse an unchanged session corpus (#1387 release-day lag waves, #1392), and refreshes that re-run the full `logs_2.sqlite` table scan every tick (the pre-#1404 22–35 s/tick cost). Both were caught by users in production because nothing in CI measures scan work.

## Change

`CostUsagePerformanceGateTests` — two regression gates that run inside the existing `lint-build-test` job (+3.4 s, no workflow changes):

1. **`warm codex refresh over an unchanged session corpus must not re-parse it`** — synthesizes an 80-file / 64k-turn session corpus, does a cold scan, then an immediate warm refresh, and asserts the warm refresh is ≥5× faster. Re-parsing unchanged files on refresh fails the gate.
2. **`priority turns refresh must scan only appended trace rows`** — builds a 200k-row trace database, full-scans it, appends one row, refreshes, and asserts the refresh is ≥5× faster **and** that the memo cursor advanced by exactly one row. Falling back to full-table scans per refresh fails the gate.

Design notes:

- Gates assert **cold/warm work ratios, not absolute durations**, so they hold across runner speeds. The 5× floor sits far below measured ratios (~100–170×), keeping them deterministic under CI contention.
- The priority gate also has a fully deterministic assertion (cursor advanced by exactly 1) that doesn't depend on timing at all.
- Each gate prints its measured timings (`PERF-GATE …`) so CI logs double as a coarse perf-trend record.
- The suite is `.serialized` so the two timed gates don't contend with each other.

## Runtime Proof (local run, Apple Silicon)

```text
PERF-GATE codex-session-corpus: cold=3004ms warm=17ms ratio=173x
􁁛 Test "warm codex refresh over an unchanged session corpus must not re-parse it" passed after 3.196 seconds.
PERF-GATE priority-turns: full=25ms incremental=0ms ratio=95x
􁁛 Test "priority turns refresh must scan only appended trace rows" passed after 0.201 seconds.
􁁛 Test run with 2 tests in 1 suite passed after 3.398 seconds.
```

Negative test — the gate fires when the protection is removed. Forcing `state = nil` per refresh (the pre-memo full-scan behavior) and rerunning:

```text
PERF-GATE priority-turns: full=25ms incremental=24ms ratio=1x
Test "priority turns refresh must scan only appended trace rows" recorded an issue:
Expectation failed: (refreshDuration * 5 → 0.1189...) < (fullDuration → 0.0254...)
Suite CostUsagePerformanceGateTests failed after 0.186 seconds with 1 issue.
```

Reverting the sabotage restores `ratio=95x` and the gate passes.

## Validation

- `swift test --filter CostUsagePerformanceGate` — 2/2 pass.
- `make check` — 0 violations.


steipete:

PR #1465, opened 2026-06-12
Fix OpenAI Admin usage pagination

rohitjavvadi:
## Summary

OpenAI Admin usage responses can be paginated, but CodexBar was only reading the first page for both costs and completions usage. That could undercount organization usage whenever the API returned `has_more: true`.

This updates the OpenAI API usage fetcher to follow `next_page` cursors for both endpoints, preserve the existing range/project/group query parameters on each page request, and fail closed on malformed pagination state instead of silently returning partial totals.

## What changed

- Follow `has_more` / `next_page` pagination for OpenAI Admin costs and completions usage.
- Keep `project_ids`, `group_by`, time range, bucket width, and limit query params across page requests.
- Detect repeated cursors, missing cursors when `has_more` is true, and excessive pagination.
- Require top-level `data` and `has_more` in Admin usage page responses so malformed pages do not look like empty successful usage.
- Add regression coverage for multi-page costs/completions, repeated cursors, missing cursors, and malformed page responses.

## Live behavior proof

I verified the pagination path against a real OpenAI Admin API response with keys, cursors, project/org identifiers, and model names redacted. The probe used the same Admin costs and completions endpoints with `bucket_width=1d` and `limit=1` to force cursor traversal.

```text
OpenAI Admin live pagination proof, redacted. range=2026-05-14..2026-06-12 bucket_width=1d limit=1

costs:
  pages_fetched=30 final_has_more=false all_next_page_cursors_present_until_final=true
  first_page: buckets=1 results=0 cost_usd=0.000000 requests=0 input_tokens=0 output_tokens=0
  all_pages: buckets=30 results=17 cost_usd=0.000000 requests=0 input_tokens=0 output_tokens=0

completions:
  pages_fetched=30 final_has_more=false all_next_page_cursors_present_until_final=true
  first_page: buckets=1 results=0 cost_usd=0.000000 requests=0 input_tokens=0 output_tokens=0
  all_pages: buckets=30 results=7 cost_usd=0.000000 requests=22 input_tokens=16284907 output_tokens=292570
```

That shows the old first-page-only behavior would report zero completions usage for this range, while following all pages finds the later usage buckets.

## Testing

- `swift test --filter OpenAIAPIUsageFetcherTests --filter OpenAIAPICreditBalanceTests --filter OpenAIAPIMenuCardModelTests`
- `make check`

I also tried an aggregate `swift test --disable-sandbox` run after rebasing on latest `upstream/main`; it hit a timing-sensitive `MainThreadHangWatchdogTests` failure during the full run, and that same suite passed when rerun directly with `swift test --disable-sandbox --filter MainThreadHangWatchdogTests`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### Fixed
 - Provider switcher: keep localized tab titles visible by tightening outer insets only when equal-width segments would otherwise truncate.
+- OpenAI API: follow Admin usage pagination for costs and completions so multi-page organization usage totals are not undercounted (#1465). Thanks @rohitjavvadi!
 - Settings: slightly increase the window height so standard panes fit without clipping their final controls or helper text.
 - Menu bar: show immediate in-place feedback for manual refreshes, keep tracked-menu geometry stable, and coalesce repeated clicks until the active refresh succeeds or fails (#1458). Thanks @hhh2210!
 - Grok: recover web billing from status-7 credential failures by combining current browser sessions with non-expired CLI auth, accept raw protobuf responses, and render current zero-use periods (#1452). Thanks @bcharleson!
--- a/Sources/CodexBarCore/Providers/OpenAI/OpenAIAPIUsageFetcher.swift
+++ b/Sources/CodexBarCore/Providers/OpenAI/OpenAIAPIUsageFetcher.swift
@@ -38,6 +38,7 @@ public enum OpenAIAPIUsageFetcher {
         URL(string: "https://api.openai.com/v1/organization/usage/completions")!
 
     private static let maxDailyBucketLimit = 31
+    private static let maxPaginationPages = 100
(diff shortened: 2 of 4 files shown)


rohitjavvadi:
Added redacted live Admin API pagination proof to the PR body.

@clawsweeper re-review

steipete:
Landed as b9a22efa53fad1e9c64c3e728897365401bae76e.

Proof:
- OpenAIAPIUsageFetcherTests: 11 tests passed.
- Full local suite: 3,843 tests in 419 suites passed.
- make check: SwiftFormat clean; SwiftLint zero violations.
- Autoreview: clean, 0.84 confidence.
- Exact-head CI: macOS, Linux x64, Linux ARM64, and GitGuardian green at https://github.com/steipete/CodexBar/actions/runs/27418022701.

Thanks @rohitjavvadi!

PR #1468, opened 2026-06-12
Fix Ollama web guard scoping

### LINKED ISSUE #1466: OLLAMA_API_KEY can bypass web-source guard for non-Ollama providers
### Description

`OLLAMA_API_KEY` can bypass the CLI web-source platform guard for providers other than Ollama because the Ollama exemption condition mixes comma-separated `if` clauses with `||` without grouping the environment-token branch under `provider == .ollama`.

### Steps to reproduce

This is visible from the current source logic in `Sources/CodexBarCLI/CLIUsageCommand.swift`:

```swift
if provider == .ollama,
   sourceMode == .auto,
   settings?.ollama?.cookieSource == .off
   || environment.map({ ProviderTokenResolver.ollamaToken(environment: $0) != nil }) == true
{
    return false
}
```

1. Run the CLI on a non-macOS platform or another environment where web-backed source modes should be guarded.
2. Set `OLLAMA_API_KEY` in the environment.
3. Request usage for a provider other than Ollama whose `auto` fetch plan includes `.web`.
4. The `sourceModeRequiresWebSupport(...)` helper can return `false` because the environment-token branch is evaluated independently of `provider == .ollama`.

### Expected behavior

The Ollama-specific exemption should only apply when `provider == .ollama` and the source mode is Ollama's supported auto/token path.

For unrelated providers, `sourceModeRequiresWebSupport(...)` should still return `true` when `sourceMode == .web` or when `sourceMode == .auto` and the provider fetch plan includes `.web`.

### Actual behavior

When `OLLAMA_API_KEY` is present, the helper can skip the web-support requirement for unrelated providers because this branch is not fully scoped to Ollama:

```swift
environment.map({ ProviderTokenResolver.ollamaToken(environment: $0) != nil }) == true
```

### Environment

- Version: current `main` (`186c6de` observed)
- Surface: `CodexBarCLI usage`
- Provider: non-Ollama providers with auto/web source plans
- Install method: source / CLI

### Evidence

The affected logic is in `Sources/CodexBarCLI/CLIUsageCommand.swift`, inside `sourceModeRequiresWebSupport(...)`.

A minimal fix would group the Ollama exemption so the token check is only evaluated inside the Ollama branch, for example:

```swift
if provider == .ollama,
   sourceMode == .auto,
   settings?.ollama?.cookieSource == .off
       || environment.map({ ProviderTokenResolver.ollamaToken(environment: $0) != nil }) == true
{
    return false
}
```

or, more explicitly:

```swift
if provider == .ollama,
   sourceMode == .auto
{
    let hasEnvironmentToken = environment.map { ProviderTokenResolver.ollamaToken(environment: $0) != nil } == true
    if settings?.ollama?.cookieSource == .off || hasEnvironmentToken {
        return false
    }
}
```

A focused regression test for `sourceModeRequiresWebSupport(.auto, provider: <non-ollama-web-provider>, environment: ["OLLAMA_API_KEY": "..."])` would lock this down.

WadydX:
## Summary
- Scope the Ollama auto-source exemption so `OLLAMA_API_KEY` only affects the Ollama provider.
- Add a regression assertion that a non-Ollama web-capable provider still requires web support when `OLLAMA_API_KEY` is present.

Closes #1466.

## Test plan
- `git diff --check`
- `swift -e 'let providerIsOllama=false; let sourceModeAuto=true; let cookieOff=false; let hasOllamaToken=true; let result = providerIsOllama && sourceModeAuto && (cookieOff || hasOllamaToken); print(result)'`
- Attempted `swift test --filter CLIEntryTests/test_sourceModeRequiresWebSupportIsProviderAware` twice; both runs timed out after 10 minutes while SwiftPM was downloading/building the Sparkle binary artifact, before the focused test could execute.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Menu bar: move the highlighted Overview provider with trackpad or mouse-wheel scrolling while preserving native submenu and keyboard behavior (#1436). Thanks @joshuavial!
 
 ### Fixed
+- CLI: keep Ollama API credentials scoped to Ollama when deciding whether another provider requires macOS web support (#1466). Thanks @WadydX!
 - Provider switcher: keep localized tab titles visible by tightening outer insets only when equal-width segments would otherwise truncate.
 - OpenAI API: follow Admin usage pagination for costs and completions so multi-page organization usage totals are not undercounted (#1465). Thanks @rohitjavvadi!
 - Settings: slightly increase the window height so standard panes fit without clipping their final controls or helper text.
--- a/Sources/CodexBarCLI/CLIUsageCommand.swift
+++ b/Sources/CodexBarCLI/CLIUsageCommand.swift
@@ -549,11 +549,14 @@ extension CodexBarCLI {
             return false
         }
         if provider == .ollama,
-           sourceMode == .auto,
-           settings?.ollama?.cookieSource == .off
-           || environment.map({ ProviderTokenResolver.ollamaToken(environment: $0) != nil }) == true
+           sourceMode == .auto
         {
-            return false
+            let hasEnvironmentToken = environment.map {
+                ProviderTokenResolver.ollamaToken(environment: $0) != nil
+            } == true
+            if settings?.ollama?.cookieSource == .off || hasEnvironmentToken {
+                return false
+            }
         }
         return switch sourceMode {
         case .web:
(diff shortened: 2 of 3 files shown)


steipete:
Prepared on exact head `8c99fc2ec0a363e6c2f532e75deeae6b4efad92d`.

Proof:
- Rebased onto current `main`.
- `swift test --filter CLIEntryTests/test_sourceModeRequiresWebSupportIsProviderAware`
- `make check`
- Codex autoreview: no accepted/actionable findings; patch rated correct.
- Linux x64, Linux ARM64, and GitGuardian checks passed. macOS CI is running; squash auto-merge is gated on required checks.

The maintainer changelog entry credits @WadydX.

PR #1471, opened 2026-06-12
Memoize cookie cache lookups behind settings picker labels

ProspectOre:
## Problem

Opening Settings and switching to the Providers pane could freeze the app for ~10 seconds (beach ball).

`ProviderSettingsPickerRowView` evaluates each picker's `trailingText` closure inside `body`, and the cookie-source pickers resolve their "Cached: …" label through `CookieHeaderCache.load`, which pays a synchronous `SecItemCopyMatching` round-trip through securityd plus a CSSM decrypt per call. SwiftUI re-runs those bodies repeatedly within a single AppKit layout pass, so the layout pass degenerates into hundreds of redundant Keychain reads on the main thread — worst right after launch, while cost scans and probes are also contending for securityd.

A 17s `sample` of one Providers-pane click on 0.34.0 showed ~10s of a single layout pass bottoming out in Keychain reads:

```
10012 NSHostingView.layout() → … → ProviderSettingsPickerRowView.body.getter
 9834   static CookieHeaderCache.load(provider:scope:)
 9819     KeychainCacheStore.load<A>(key:as:)
 9816       SecItemCopyMatching → securityd → CSSM_DecryptDataFinal
```

## Fix

Display paths now use `CookieHeaderCache.loadForDisplay`, which serves a memoized snapshot and revalidates a stale one (30s) off the calling path, following the same stale-tolerant-cache approach as the menu-build `auth.json` fix. In-process `store`/`clear` update the snapshot synchronously, so only the first lookup per key pays the Keychain read. Strict `load` semantics are unchanged for probes and fetchers.
## Verification

Before: 17s `sample` of one Providers-pane click on the unpatched build shows ~10s of a single layout pass bottoming out in Keychain reads (9,834 `SecItemCopyMatching` samples / 9,835 `ProviderSettingsPickerRowView.body` samples on the main thread).

After: live human-driven session on the patched build — Settings opened from the menu bar, Providers pane opened and switched repeatedly, panes bounced — recorded as 28 consecutive 10s `sample` windows. Zero `SecItemCopyMatching` samples on the main thread in every window; worst window had ~0.8s of main-thread work (menu rebuild), all pane switches instant.

Includes unit tests for memoization, immediate store/clear snapshot updates, async revalidation, and (after review) deterministic race coverage for store/clear/clearAll versus an in-flight stale refresh. Full `swift test` passes (two pre-existing timing-flaky suites pass in isolation).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 0.34.1 — Unreleased
 
 - Menu bar: restore native macOS positioning for merged provider dropdowns while preparing current content before AppKit lays out the menu.
-- Settings: memoize cookie cache lookups behind the "Cached: …" picker labels so opening Settings and switching panes no longer pays a synchronous Keychain read per SwiftUI body evaluation, which froze the window for seconds on the Providers pane.
+- Settings: memoize cookie cache lookups behind the "Cached: …" picker labels so opening Settings and switching panes no longer pays a synchronous Keychain read per SwiftUI body evaluation, which froze the Providers pane for seconds (#1471). Thanks @ProspectOre!
 
 ## 0.34.0 — 2026-06-12
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.34.1 — Unreleased
 
 - Menu bar: restore native macOS positioning for merged provider dropdowns while preparing current content before AppKit lays out the menu.
+- Settings: memoize cookie cache lookups behind the "Cached: …" picker labels so opening Settings and switching panes no longer pays a synchronous Keychain read per SwiftUI body evaluation, which froze the window for seconds on the Providers pane.
 
 ## 0.34.0 — 2026-06-12
 
--- a/Sources/CodexBar/PreferencesDebugPane.swift
+++ b/Sources/CodexBar/PreferencesDebugPane.swift
@@ -524,9 +524,13 @@ struct DebugPane: View {
     }
 
     private func clearCookieCache() {
-        let cleared = CookieHeaderCache.clearAll()
-        if cleared > 0 {
-            self.cookieCacheStatus = "Cleared \(cleared) provider\(cleared == 1 ? "" : "s")."
+        let summary = CookieHeaderCache.clearAllDetailed()
+        if summary.failedCount > 0 {
+            self.cookieCacheStatus = "Cookie cache cleanup failed for \(summary.failedCount) "
+                + "operation\(summary.failedCount == 1 ? "" : "s")."
+        } else if summary.clearedCount > 0 {
+            self.cookieCacheStatus = "Cleared \(summary.clearedCount) "
+                + "provider\(summary.clearedCount == 1 ? "" : "s")."
         } else {
             self.cookieCacheStatus = "No cached cookies found."
         }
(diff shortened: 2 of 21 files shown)


ProspectOre:
Addressed the P1/P2 refresh-ordering risk in 9f63a991:

- Every in-process `store`/`clear` mutation bumps a per-key generation under the display-cache lock. Reads (both first population and stale revalidation, since both load the Keychain outside the lock) capture the generation they started from and only commit if it is unchanged; a losing read returns whichever newer snapshot won the race instead of overwriting it.
- `clearAll` bumps the generation of every known key before wiping, so an in-flight refresh cannot resurrect a wiped snapshot either.
- Added deterministic race tests for all three mutations (store, clear, clearAll) against a stale commit — no sleeps or task scheduling involved; they drive the commit path directly with a captured stale generation.

ProspectOre:
The `lint-build-test` failure is a pre-existing timing flake, unrelated to this diff:

- Failing test: `StatusMenuTests` › "closing menu during successful missing data refresh clears deferred retry" (`StatusMenuInstantOpenTests.swift:1315`, `await refreshGate.startCount == 1`). It waits on bounded `Task.yield()` spins, which is scheduler-sensitive on contended CI runners.
- The test exercises menu refresh gating and has zero overlap with this PR's surface (cookie display cache + settings picker labels); the test file references none of the touched symbols.
- It passes 8/8 consecutive local runs on this branch's code.
- The same `lint-build-test` job failed today on two unrelated branches (`feature/manual-provider-subscription-reminders`, `feat/italian-native-support`).

I don't have rerun permissions, so I pushed an empty commit to retrigger CI.

ProspectOre:
@clawsweeper re-review

steipete:
Landed in c1f9e3673a0c7561ddeb32e7dfcc46edb15bd2ba. Thanks @ProspectOre for the excellent profiling and fix direction.

Maintainer improvements:

- Kept Settings display reads memoized and moved legacy plaintext-cookie migration off the render path.
- Added generation guards so stale reads cannot overwrite stores or resurrect cleared snapshots.
- Preserved last-known display state across temporary Keychain failures and failed mutations.
- Added detailed cleanup failure reporting for provider, CLI, and Debug-pane clears.
- Serialized migration/store/clear with a cross-process file lock so a delayed app migration cannot undo `codexbar cache clear --cookies`.

Proof on exact head `a610c7aa08fd79cb93996e8948982c54e3ab5006`:

- `swift test --filter CookieHeaderCacheTests`: 24/24 passed.
- `make check`: SwiftFormat clean; SwiftLint 0 violations across 1,073 files.
- Full local suite exercised 3,898 tests; the only rerun was the unrelated watchdog timing test, which passed 6/6 in isolation.
- `./Scripts/compile_and_run.sh`: release bundle built, signed, launched, and remained running.
- Live Settings test: repeatedly switched OpenAI, Codex, Claude, Cursor, and MiniMax in the packaged app. A 12-second process sample contained zero `SecItemCopyMatching` frames on the main thread.
- Branch-wide autoreview against `origin/main`: no actionable findings, 0.90 confidence.
- Exact-head CI green: macOS lint/build/test, Linux x64, Linux ARM64, and GitGuardian. Linux x64 initially hit unrelated `ETXTBSY` runner contention; the failed-job rerun passed unchanged.

Issue #1427, opened 2026-06-11
Lag in merged icons mode when switching usage chart views


xingpz2008:
Thank you for providing the new build!

However, I can still observe lag in certain scenarios.

**Steps to reproduce:**

1. Enable merged icons mode.
2. Hover over a model.
3. In the newly opened window containing the usage chart, switch between the two view options: Session and Weekly.
4. After switching back and forth a few times, the lag starts to appear, usually around the third click.

**Observed behavior:**

The UI becomes noticeably laggy when switching between the Session and Weekly usage chart views.
The lag can be temporarily resolved by clicking elsewhere, as long as the usage chart window does not appear again.
One subtle behavior I noticed is that before the window disappears, it seems to block keyboard input in other windows. This might be useful for debugging.

**Additional note:**

The lag in the overview under merged icons mode still appears occasionally, but it has improved significantly in recent builds.

**Environment:**

* MacBook Pro 14-inch, 2021
* Apple M1 Pro
* 16 GB RAM
* macOS 26.5.1 (25F80)
* Codex 0.33.0 (80)

steipete:
@clawsweeper re-review

hhh2210:
Datapoint supporting the OS-level (WindowServer) hypothesis from #1399: I upgraded the same machine to the macOS 27 beta and the lag is completely gone with the **same CodexBar build**.

- Before: macOS 26.x, CodexBar 0.33.0 (81), merged icons ON — intermittent lag/freezes in the merged dropdown, even after the menu-rebuild optimizations that already landed (#1277 / #1286 / #1297 / #1314).
- After: macOS 27.0 (26A5353q), same CodexBar 0.33.0 (81), same providers and settings — fully smooth, no beachballs or dropped input.

Since nothing changed except the OS, this is consistent with #1399's conclusion that the remaining freezes are driven by macOS 26's WindowServer/SkyLight event handling during long NSMenu tracking sessions rather than by in-process work. Might be worth a note in the known issues that macOS 26 users hitting this can expect relief on 27.

steipete:
Fixed by #1472 (`f5567044db4c04719a36083e202385247877f68b`), which removed the menu-tracking run-loop reentry that left hosted submenus visible but unresponsive.

Current-main verification on macOS 26.5:
- packaged and launched the current app bundle;
- opened merged icons > Subscription Utilization;
- switched Session/Weekly 12 times in rapid succession;
[...]

Issue #1329, opened 2026-06-06
CodexBar menu bar icon hangs and remains memory-heavy


dhruv-anand-aintech:
## Environment
- App: `/Applications/CodexBar.app`
- Version: `0.32.4` (`CFBundleVersion: 79`, `CodexGitCommit: edf3d8da`)
- Bundle ID: `com.steipete.codexbar`
- Machine: `/Users/dhruvanand/Library/...`
- Observed on: `2026-06-07`

## Summary
Opening the CodexBar menu bar icon hangs and leaves the app instance in an unhealthy state with elevated resident memory.

## Live process snapshot
Captured while the issue was happening:

```text
pid 27497 /Applications/CodexBar.app/Contents/MacOS/CodexBar
cputime: 2:33.83
elapsed: 03:42:45
%cpu: 0.0
%mem: 1.0
rss: 170944 KB (~167 MB)
```

Related Codex-related processes still running and active in the same session:

```text
pid 90556 /Users/dhruvanand/.codex/packages/standalone/current/codex
  cputime: 91:01.02
  %mem: 1.1
  rss: 181744 KB (~177 MB)
```

## Evidence from macOS logs
Recent logs for PID 27497 show repeated LocalAuthentication/Keychain security activity and connection churn:

```text
2026-06-07 04:43:23.789022+0530  CodexBar: (Security) SecItemCopyMatching
2026-06-07 04:43:23.789038+0530  CodexBar: (Security) SecItemCopyMatching_ios
2026-06-07 04:43:23.793249+0530  CodexBar: (libxpc.dylib) ... connection cancelled by xpc_connection_cancel()
2026-06-07 04:43:23.793638+0530  CodexBar: (LocalAuthentication) ... Creating LAContext new
...
2026-06-07 04:43:24.662160+0530  CodexBar: (CFNetwork) Task ... done using Connection 486
```

I only collected local runtime logs and process state, not external diagnostics. The app seems to be stuck while interacting with auth/keychain and keeps the instance alive with high baseline RSS.

## Steps observed
1. Open CodexBar from menu bar.
2. Click the CodexBar icon to open UI / account area.
3. UI appears to hang and/or become unresponsive; menu interaction takes long and process state remains high-memory.

## Expected
- Menu opens quickly and interaction remains responsive.
- Process RSS should remain stable at normal idle baseline.

## Notes
I can provide the full command output snippets if needed.

dhruv-anand-aintech:
It was beachballing a while back
[...]

PR #1472, opened 2026-06-12
Keep switcher shortcut peeks from killing menu tracking sessions

ProspectOre:
## Problem

The merged status menu intermittently turns into a "zombie": it stays on screen but stops responding to clicks for tens of seconds while the cursor beach-balls. It happens most often right after opening the menu or after rapid ⌘-number provider switching, and clicking **Settings…** is the usual victim. This is the freeze behind #1387, #1427, and (at least partly) #1329.

## Evidence

Instrumented builds traced every event through a live repro:

- A click on Settings… sat at the event-queue head and was re-peeked **13 times over 21 seconds** while the user kept clicking; the menu never reacted.
- One menu stayed open and deaf for **92 seconds**. A live `sample` during that state showed the main thread idle in the **normal** run loop — the menu window was on screen with **no tracking session left at all**.
- The opening click's own mouseUp was already stuck the moment the menu opened, peeked during menu setup before any tracking session existed.
- A rapid ⌘-keypad switching burst (~8 presses/sec, each claimed and dequeued by the monitor) reproduced a 7-second deaf menu on demand.

## Cause

`ProviderSwitcherShortcutEventMonitor` peeks the queue with `NSApp.nextEvent(matching:until:inMode:dequeue:)` from a run-loop observer in `.eventTracking` mode. That call **re-enters the event loop in the given mode**, dispatching the menu session's own timers and sources mid-observer. When the reentry lands during menu setup — guaranteed by the HID-counter gate, since the opening click just advanced the counters — or amid rapid claimed key events, the tracking session dies and the menu zombifies.

## Why not a passive API

All passive delivery paths go dark during `NSMenu` tracking (verified empirically on macOS 26): local event monitors receive nothing, Carbon dispatcher handlers receive nothing, `RegisterEventHotKey` is suppressed (tracking pushes a hotkey-disable mode), and `menuHasKeyEquivalent` is not consulted for popup menus. The peek is the only delivery path, so the fix isolates it instead:

1. **Peeks run in a barren private run-loop mode** with no sources or timers registered — the queue is mode-agnostic so matching still works, but the reentry can no longer dispatch menu-session work.
2. **Peeks start only once the tracking session is pumping**, armed by a block queued in the tracking run-loop mode (these menus post no NSMenu tracking notifications).
3. **Mouse events are no longer monitored at all** — `ProviderSwitcherView` already handles switcher clicks via its `mouseDown`/`mouseUp` overrides, so the monitor can never dequeue a click meant for AppKit.

## Verification

Live user testing on the patched build: the previously ~every-time freeze is gone across repeated menu opens, rapid ⌘1/⌘2/⌘3 bursts (number row and keypad), provider clicks, overview scrolling, and Settings… opens — traced end-to-end with shortcuts claimed correctly, escape passing through, and every menu close immediate. Full `swift test` passes and lint is clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Menu bar: restore native macOS positioning for merged provider dropdowns while preparing current content before AppKit lays out the menu.
 - Settings: memoize cookie cache lookups behind the "Cached: …" picker labels so opening Settings and switching panes no longer pays a synchronous Keychain read per SwiftUI body evaluation, which froze the Providers pane for seconds (#1471). Thanks @ProspectOre!
 - Build: resolve packaged binaries from the bin path SwiftPM reports instead of assuming the legacy `.build/<arch>-apple-macosx/<conf>/` layout, so a stale directory left behind by the older build system no longer silently shadows fresh swiftbuild products in `package_app.sh`. Thanks @ProspectOre!
+- Menu bar: stop the provider-switcher shortcut monitor from killing the menu's event tracking session. Its event-queue peek re-entered the run loop in tracking mode, which could leave a zombie menu on screen that ignored clicks for tens of seconds (beach ball) — most often right after opening the menu or after rapid Cmd-number provider switching, with Settings… the usual victim. Peeks now run in a barren private run-loop mode, start only once the tracking session is pumping, and no longer touch mouse events. Thanks @ProspectOre!
 
 ## 0.34.0 — 2026-06-12
 
--- a/Sources/CodexBar/StatusItemController+ProviderSwitcher.swift
+++ b/Sources/CodexBar/StatusItemController+ProviderSwitcher.swift
@@ -71,34 +71,57 @@ final class ProviderSwitcherEventPeekGate {
     }
 }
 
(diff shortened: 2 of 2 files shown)


ProspectOre:
@clawsweeper re-review

ProspectOre:
@clawsweeper re-review

steipete:
Maintainer validation at exact head `29f8feafe0a4bf7d21da09e167cb1e450c64cfd8` after rebasing onto merged PR #1473:

- `make check`
- `swift test --filter 'ProviderSwitcherEventPeekGateTests|StatusMenuSwitcherTrackingTests|StatusMenuOverviewScrollTests'` — 19 tests passed
- structured autoreview against current `main` — clean, no actionable findings
- `./Scripts/compile_and_run.sh` — freshly packaged bundle launched and remained running
- live macOS 26.5.1 (25F80) stress: 30 open / Cmd-number switch / arrow navigation / overview scroll / close cycles completed in 85.07s; process remained responsive
- exact rebased bundle stress: 10 more cycles; bundle reported `CodexGitCommit=29f8feaf`, process remained alive and returned to 0.0% CPU

Waiting only for exact-head macOS CI before merge.

PR #1473, opened 2026-06-13
Fix package_app.sh preferring stale legacy build products over fresh swiftbuild output

ProspectOre:
## Problem

`package_app.sh` resolves built binaries by checking the legacy native build system layout (`.build/<arch>-apple-macosx/<conf>/`) first, falling back to `.build/<conf>/`. Newer toolchains default to the **swiftbuild** build system, which emits products to `.build/out/Products/<Conf>` (with `.build/<conf>` as a symlink into it).

When a stale `.build/<arch>-apple-macosx/` directory is left behind from an earlier toolchain run, the script silently packages those old binaries instead of the ones `swift build` just produced. The app looks freshly built — new Info.plist, new `CodexGitCommit` — but runs days-old code. This cost me a debugging session this week: "fresh" dev builds of the menu fix were actually running Jun 10 binaries until I noticed and `rm -rf`'d the legacy directory.

## Fix

Ask the toolchain where the products actually are via `swift build --show-bin-path -c <conf> --arch <arch>` (cached per arch) and prefer that location. The legacy layouts stay as fallbacks for toolchains where `--show-bin-path` is unavailable or the product is missing from the reported path, so existing flows are unchanged:

- **Native build system** (single arch or universal): `--show-bin-path --arch <a>` reports the per-arch `.build/<arch>-apple-macosx/<conf>/` directory the script already used — identical behavior.
- **swiftbuild, host arch**: resolves to `.build/out/Products/<Conf>`, immune to stale legacy directories.
- `install_binary`'s error message now lists every location searched instead of only the legacy path.

No associative arrays or other bash 4 features — works on macOS's stock bash 3.2.

## Proof

Function-level test (bash 3.2): with a stale `.build/arm64-apple-macosx/release/CodexBar` planted next to a fresh swiftbuild product, resolution now returns `.build/out/Products/Release/CodexBar`; with `swift build --show-bin-path` stubbed to fail (old-toolchain simulation), the legacy path still wins.

End-to-end: in a clean worktree I planted a distinguishable 20.8 MB binary (a copy of CodexBarCLI) at `.build/arm64-apple-macosx/release/CodexBar` alongside the fresh 32.9 MB swiftbuild product, then ran `CODEXBAR_SIGNING=adhoc ./Scripts/package_app.sh release`. Before this change the script packages the planted stale binary; with it, the packaged `CodexBar.app/Contents/MacOS/CodexBar` is the fresh 32.7 MB arm64 product (post re-sign), and the script exits 0 with the KeyboardShortcuts resource bundle and widget appex intact.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Menu bar: restore native macOS positioning for merged provider dropdowns while preparing current content before AppKit lays out the menu.
 - Settings: memoize cookie cache lookups behind the "Cached: …" picker labels so opening Settings and switching panes no longer pays a synchronous Keychain read per SwiftUI body evaluation, which froze the Providers pane for seconds (#1471). Thanks @ProspectOre!
+- Build: resolve packaged binaries from the bin path SwiftPM reports instead of assuming the legacy `.build/<arch>-apple-macosx/<conf>/` layout, so a stale directory left behind by the older build system no longer silently shadows fresh swiftbuild products in `package_app.sh`.
 
 ## 0.34.0 — 2026-06-12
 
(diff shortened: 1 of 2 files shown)


ProspectOre:
@clawsweeper re-review

PR #1478, opened 2026-06-13
fix: trust SwiftPM package product paths

steipete:
## Summary

- Treat SwiftPM's reported product directory as authoritative for app binaries, helpers, resources, and Sparkle.
- Stage each architecture's fresh products and dSYM before SwiftBuild can replace its shared output directory.
- Fail closed when a required reported product is missing instead of packaging a stale legacy-layout artifact.
- Add focused shell regression coverage to the standard lint gate.

## Context

Follow-up to #1473. The landed fix correctly introduced SwiftPM path discovery, but successful discovery could still fall back to stale legacy products, while Sparkle and release dSYM lookup remained tied to the legacy layout. Thanks @ProspectOre for identifying the original stale build-product selection.

## Proof

Exact candidate: `cc7c00edde1b8e29c454a23e6d3671418a574d96`

- Reproduced on landed `main`: removing the reported executable while retaining a stale legacy executable still produced a package; fresh app and legacy dSYM UUIDs differed; removing the legacy release symlink broke Sparkle lookup despite Sparkle existing under the reported path.
- Focused shell regression: pass, including reported native/SwiftBuild paths, staged precedence, query failure, and refusal of stale legacy binaries/frameworks/dSYMs.
- `make check`: pass, SwiftFormat and SwiftLint clean.
- Full isolated suite run: 429/429 suites passed.
- Native release package: reported arm64 path used; app, CLI, and widget arm64; Sparkle/resources present; app and staged dSYM UUID match; deep signing passed; isolated installed app remained alive for 8 seconds with zero launch-log bytes.
- Forced SwiftBuild universal package with the legacy release symlink absent: reported `out/Products/Release` path used; app, CLI, watchdog, and widget contain arm64 and x86_64; Sparkle/resources present; staged dSYMs contain the correct thin slices; merged dSYM UUIDs match both app slices; deep signing passed; isolated installed app remained alive for 8 seconds with zero launch-log bytes.
- Public Model Identifier Gate: PASS. Diff, test output, proof notes, and this PR body contain no non-public model identifiers or private payloads.

## Tradeoff

Missing required products now stop packaging. Older Swift toolchains that cannot report the product path will fail rather than silently selecting an unknown stale product; the repository's supported Swift toolchain provides the required query.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - Menu bar: keep cached provider content visible while switching merged tabs so the open menu no longer flickers through an empty state.
 - Menu bar: handle the global open-menu shortcut synchronously so repeated presses close the tracked menu instead of queueing a delayed reopen (#1470). Thanks @Zihao-Qi!
 - Settings: memoize cookie cache lookups behind the "Cached: …" picker labels so opening Settings and switching panes no longer pays a synchronous Keychain read per SwiftUI body evaluation, which froze the Providers pane for seconds (#1471). Thanks @ProspectOre!
-- Build: resolve packaged binaries from the bin path SwiftPM reports instead of assuming the legacy `.build/<arch>-apple-macosx/<conf>/` layout, so a stale directory left behind by the older build system no longer silently shadows fresh swiftbuild products in `package_app.sh`. Thanks @ProspectOre!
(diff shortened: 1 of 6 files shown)


steipete:
@clawsweeper re-review

PR #1519, opened 2026-06-14
fix: synchronize cookie cache test override

steipete:
## Summary

- serialize the testing-only legacy cookie-cache base URL override
- keep default legacy path construction outside the override lock
- cover concurrent override teardown and URL reads

## Root cause

The macOS x86 CI shard for #1496 exited with signal 11 while `CookieHeaderCacheTests` was transitioning between a scheduled legacy migration and test teardown. `legacyBaseURLOverride` stored a reference-typed `URL?` in `nonisolated(unsafe)` static state and was read by background migration work while tests assigned a new value or `nil` without synchronization.

## Proof

- `swift test --filter CookieHeaderCacheTests` (25 tests passed)
- `make check`
- Codex autoreview: clean, no accepted/actionable findings
- pre-fix stress observation: 30 suite runs passed on arm64; the original crash was observed on GitHub's x86 macOS shard, so the regression exercises the concurrent access directly rather than claiming deterministic local reproduction

### DIFF
--- a/Sources/CodexBarCore/CookieHeaderCache.swift
+++ b/Sources/CodexBarCore/CookieHeaderCache.swift
@@ -43,6 +43,7 @@ public enum CookieHeaderCache {
     }
 
     private static let log = CodexBarLog.logger(LogCategories.cookieCache)
+    private static let legacyBaseURLOverrideLock = NSLock()
     private nonisolated(unsafe) static var legacyBaseURLOverride: URL?
 
     private struct DisplaySnapshot {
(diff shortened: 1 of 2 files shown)


steipete:
Additional proof on exact head `5731fcf744f99aa6ff8caf04bc5189cbe7d9e525`:

- `swift test --sanitize thread --filter CookieHeaderCacheTests`
- 25 tests passed with no Thread Sanitizer findings, including the 5,000-operation concurrent override/read regression.

steipete:
Landed as `6c409f76`.

Proof:
- `swift test --filter CookieHeaderCacheTests` (25 tests passed)
- `swift test --sanitize thread --filter CookieHeaderCacheTests` (25 tests passed, no Thread Sanitizer findings)
- `make check`
- Codex autoreview clean; GitHub Codex no findings; ClawSweeper no repair needed
- Exact-head CI run https://github.com/steipete/CodexBar/actions/runs/27495345179 passed on macOS, Linux x64, and Linux ARM64

PR #1522, opened 2026-06-14
fix: retry transient Antigravity launch failures

steipete:
## Summary

- retry Antigravity PTY process creation only when `posix_spawn` returns `ETXTBSY`
- bound retries to three attempts with a short delay
- preserve immediate failure for all other launch errors
- add regression coverage for retry and fail-fast behavior

## Why

PR #1521 exposed an unrelated Linux x64 CI flake in `AntigravityProcessLauncherLinuxTests`: the temporary executable returned `Text file busy` while arm64 passed. The same transient can occur when the installed `agy` executable is being replaced during a refresh.

## Validation

- `swift test --filter AntigravityCLISessionTests` (48 passed)
- `make check`
- local Codex autoreview: clean


steipete:
@codex review

@clawsweeper re-review

PR #1535, opened 2026-06-14
Fix Kiro usage command pipe hangs

### LINKED ISSUE #1533: Kiro usage refresh can hang on process termination or inherited pipes
## Problem

Kiro's usage command runner has two unbounded waits in the provider refresh path:

- after timeout/idle termination it calls `waitUntilExit()` after only sending terminate
- after process exit it calls `readDataToEndOfFile()` for stdout/stderr, which waits indefinitely when a descendant retains either inherited pipe

That can hold a Kiro provider refresh open after the parent CLI has exited or ignored SIGTERM.

## Source reproduction

- a short parent prints output, launches a detached child inheriting stdout/stderr, then exits; the current final EOF reads wait for the child
- a process that ignores SIGTERM can leave the current timeout path blocked in `waitUntilExit()`

## Fix shape

- keep incremental stdout/stderr capture and idle-activity tracking
- bound TERM-to-KILL escalation and post-exit drains
- preserve completed no-output exit statuses for existing caller validation
- add regressions for inherited pipes, ignored SIGTERM, and completed no-output failures
- validate without live Kiro credentials or Keychain access

kiranmagic7:
## Summary

Fixes #1533.

Kiro usage refresh could still wait forever after the direct `kiro-cli` process exited or ignored termination, because the runner used an unbounded `waitUntilExit()` plus final EOF reads on stdout/stderr. This keeps the existing activity-based behavior, but bounds process cleanup and stops relying on EOF from inherited pipes.

The regression uses a fake `kiro-cli` that prints valid usage, starts a child that keeps stdout open, and exits. `fetch()` now returns from the parent output instead of waiting for the child.

## Proof

- `swiftc -swift-version 6 -parse Sources/CodexBarCore/Providers/Kiro/KiroStatusProbe.swift Tests/CodexBarTests/KiroStatusProbeTests.swift`

I also attempted the focused SwiftPM test locally, but SwiftPM hung while downloading Sparkle's binary artifact before it reached compilation. The added regression should run under normal CI package resolution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Provider probes: stop waiting indefinitely for inherited output pipes after subprocesses or CLI version checks exit (fixes #1531).
 - Menu bar: update visible usage values in place when a manual refresh completes instead of leaving the open provider card stale until the menu is reopened (fixes #1516).
 - Gemini: recognize the current `gemini-api-key` CLI auth setting so API-key sessions show the supported OAuth guidance instead of a misleading not-logged-in error (fixes #1511).
+- Kiro: keep usage refreshes bounded when CLI helpers retain output pipes, ignore termination, or are cancelled (fixes #1533). Thanks @kiranmagic7!
 - Xiaomi MiMo: cancel optional token-plan requests when the required balance request fails instead of delaying the error for up to 30 seconds.
 - Settings: make the cost history window directly editable by keyboard while preserving the existing stepper and 1–365 day bounds (fixes #1499). Thanks @kiranmagic7!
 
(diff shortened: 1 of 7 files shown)


kiranmagic7:
Follow-up pushed in `fba48abe` to clear the lint blockers on this branch.

What changed:
- Applied the repo SwiftFormat numeric-literal style in `KiroStatusProbe`.
- Reworked the fake CLI child-process fixture as a heredoc so SwiftLint line length stays green without changing the fixture behavior.

Local proof:

```text
$ ./Scripts/lint.sh lint
Done linting! Found 0 violations, 0 serious in 1098 files.

$ git diff --check
# no output
```

I also tried `swift test --filter KiroStatusProbeTests --skip-update`; SwiftPM stalled while downloading the Sparkle 2.9.1 binary artifact before compiling, so I am leaving focused test proof to the current CI run on `fba48abe`.

PR #1536, opened 2026-06-14
Fix Gemini fnm package discovery pipe hangs

### LINKED ISSUE #1534: Gemini fnm package discovery can hang on inherited output pipes
## Bug

Gemini OAuth client discovery can hang after an `fnm` helper exits when a descendant retains the helper's stdout pipe.

## Source reproduction

`GeminiStatusProbe.runProcess` waits for the direct process through a bounded semaphore, but after a successful exit it calls `readDataToEndOfFile()` without a deadline. A detached descendant inheriting stdout keeps the pipe open indefinitely even though the command itself has completed.

This path is used by fnm-managed Gemini installs for:

- `fnm current`
- `fnm exec --using <version> npm root -g`
- the Node `require.resolve` fallback

## Expected

Capture output incrementally and bound post-exit draining. Timeout cleanup should also escalate past SIGTERM so an uncooperative helper cannot remain alive.

## Regression proof

Add a test helper that prints a package-root-like value, launches a detached child that retains stdout, and exits. Package discovery must return within a short bound while preserving the completed command's output and status.

kiranmagic7:
## Summary

Fixes #1534.

Gemini fnm package discovery waited for stdout EOF after `fnm` exited. If a descendant inherited stdout, the direct helper could be done while discovery still hung. This captures stdout incrementally, drains stderr so helpers cannot block on a full pipe, and bounds timeout cleanup with SIGKILL escalation.

The regression keeps the existing expired-token fnm OAuth refresh path, but makes fake `fnm exec -- npm root -g` leave stdout open through a child process. The refresh now completes from the parent output instead of waiting for inherited EOF.

## Proof

- `swiftc -swift-version 6 -parse Sources/CodexBarCore/Providers/Gemini/GeminiStatusProbe.swift Tests/CodexBarTests/GeminiStatusProbeAPITests.swift Tests/CodexBarTests/GeminiTestEnvironment.swift`

I also attempted focused SwiftPM tests locally, but SwiftPM hung while downloading Sparkle's binary artifact before compilation. CI should run the focused regression under normal package resolution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 - Menu bar: update visible usage values in place when a manual refresh completes instead of leaving the open provider card stale until the menu is reopened (fixes #1516).
 - Gemini: recognize the current `gemini-api-key` CLI auth setting so API-key sessions show the supported OAuth guidance instead of a misleading not-logged-in error (fixes #1511).
 - Kiro: keep usage refreshes bounded when CLI helpers retain output pipes, ignore termination, or are cancelled (fixes #1533). Thanks @kiranmagic7!
+- Gemini: keep fnm package discovery bounded when helper descendants retain output pipes or ignore termination (fixes #1534). Thanks @kiranmagic7!
 - Xiaomi MiMo: cancel optional token-plan requests when the required balance request fails instead of delaying the error for up to 30 seconds.
 - Settings: make the cost history window directly editable by keyboard while preserving the existing stepper and 1–365 day bounds (fixes #1499). Thanks @kiranmagic7!
 
--- a/Sources/CodexBarCore/Providers/Gemini/GeminiStatusProbe.swift
+++ b/Sources/CodexBarCore/Providers/Gemini/GeminiStatusProbe.swift
@@ -577,100 +577,6 @@ public struct GeminiStatusProbe: Sendable {
         return nil
     }
 
-    private static func runProcess(
-        executable: String,
-        arguments: [String],
-        environment: [String: String],
-        timeout: TimeInterval) -> String?
-    {
-        let process = Process()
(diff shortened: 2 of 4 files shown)


kiranmagic7:
Follow-up pushed in `ccdd6a34` to clear the lint gate.

What changed:
- Moved `runProcess(...)` into a same-file `GeminiStatusProbe` extension so the primary struct stays under the SwiftLint `type_body_length` limit.
- Left the process runner behavior unchanged; this is a layout-only follow-up for the lint rule.

Local proof:

```text
$ ./Scripts/lint.sh lint
Done linting! Found 0 violations, 0 serious in 1098 files.

$ git diff --check
# no output
```

CI is running on `ccdd6a34`.

kiranmagic7:
Follow-up pushed in `d69a7e43` to make the fnm stdout-holder fixture portable on Linux.

What changed:
- Replaced the shell background `python3 ... &` fake with a foreground Python launcher that exits after spawning a detached sleeping child.
- The detached child still inherits stdout, so the regression continues to cover the held-pipe case, but the helper parent can be reaped promptly.

Local proof on `d69a7e43`:

```text
$ ./Scripts/lint.sh lint
Codex parser hash is current (1507fdd7e456ee03)
Package product path tests passed.
Release dSYM path tests passed.
Sparkle signing path tests passed.
Swift test sharding tests passed.
SwiftFormat: 0/1108 files require formatting.
SwiftLint: Found 0 violations, 0 serious in 1107 files.

$ git diff --check HEAD~1..HEAD
# no output
```

I also attempted `swift test --filter GeminiStatusProbeAPITests --skip-update`, but local SwiftPM stayed at the Sparkle 2.9.1 binary artifact download before compiling. Leaving the focused runtime suite to the current CI run on `d69a7e43`.

PR #1543, opened 2026-06-15
test: stabilize Antigravity readiness error assertion

hololee:
## Summary
- keep the Antigravity readiness-error test focused on preserving the latest fetch error
- avoid requiring more than one wall-clock polling attempt in CI

## Context
The retry behavior is already covered by separate Antigravity tests. This assertion can fail on faster CI/test-runtime combinations when the wait exits after a single fetch attempt, while still returning the latest readiness error correctly.

## Test plan
- `arch -x86_64 swift test --scratch-path .build/x86-antigravity-separate-pr-check --filter AntigravityCLIHTTPSFetchStrategyTests --skip-update`

### DIFF
--- a/Tests/CodexBarTests/AntigravityCLIHTTPSFetchStrategyTests.swift
+++ b/Tests/CodexBarTests/AntigravityCLIHTTPSFetchStrategyTests.swift
@@ -723,7 +723,7 @@ struct AntigravityCLIHTTPSFetchStrategyTests {
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


steipete:
Thanks @hololee. This CI flake is fixed on `main` by https://github.com/steipete/CodexBar/commit/3263f047, using an injected deterministic clock while preserving explicit retry coverage. That supersedes the assertion relaxation here, so I’m closing this PR as resolved by the stronger equivalent fix.

Issue #1425, opened 2026-06-11
No available fetch strategy for claude.


Oliverwqcwrw:
<img width="626" height="144" alt="Image" src="https://github.com/user-attachments/assets/acca90be-2650-4a8f-8092-e86036a61903" />

jessehui:
same here

steipete:
Retested this against freshly packaged current main `76e684cf`.

Live results on a configured Claude account:

- `codexbar diagnose --provider claude --format json --pretty` detected the installed Claude CLI; the planner did not reach the reported no-strategy state.
- `codexbar usage --provider claude --source web --json --web-timeout 30` succeeded with live session, weekly, model, and extra-usage data.
- `codexbar usage --provider claude --source auto --json --web-timeout 30` succeeded and selected the web source.
- Explicit OAuth correctly reported stale credentials and suggested `claude login`.
- Explicit CLI reached Claude Code but this account returned a subscription notice without session quota data, which CodexBar surfaced as a specific parse error rather than “No available fetch strategy.”

Current main also provides a **Re-login at claude.ai** recovery action when automatic web-session detection is unavailable.

I cannot reproduce the generic no-strategy error on current main. To identify the unavailable source without exposing credentials, please retest 0.34.0 or newer and attach:

```text
CodexBar.app/Contents/Helpers/CodexBarCLI diagnose --provider claude --format json --pretty
```

The diagnostic is intentionally redacted and omits tokens, cookies, emails, account IDs, organization IDs, and raw responses. Please also include the CodexBar version and whether Claude is set to Auto, OAuth, Web, or CLI.

steipete:
Closing as non-reproducible on current packaged main: configured Claude CLI/Web/Auto paths all produced source-specific results, and the requested redacted diagnostics/version/source details were not provided.

Please reopen with current CodexBar output from `CodexBarCLI diagnose --provider claude --format json --pretty` if the generic no-strategy error remains.

26tajeen:
I get this when using Safari. I logged in with Firefox (just a hunch it would work with FF) and it loaded fine.

minchik:
I encounter this issue frequently as well. It seems to be related to the background update. Manually pressing the Refresh button always resolves the problem.

gmkbenjamin:
[...]

PR #1547, opened 2026-06-15
fix: surface Claude recovery action when no fetch strategy is available

LeoLin990405:
## What

When Claude resolves to no available fetch strategy (OAuth, CLI, and web all unavailable), the menu surfaced only the generic developer-facing error `No available fetch strategy for claude.` with no next step. This adds a fallback so the menu offers the existing **Re-login at claude.ai** recovery action in that state.

## Why

Per the triage on #1425, the no-strategy state is the documented outcome for a fully unconfigured Claude provider — but the recommended direction was to *show per-source recovery guidance rather than only the generic empty-plan error*.

`loginMenuAction` already offers recovery for the web-session-error and OAuth-error cases. But when **all** sources are unavailable (nothing configured), `shouldOpenBrowserForWebSessionError` requires a `claude.web` attempt that may not exist, and `shouldOpenTerminalForOAuthError` finds no OAuth error → both fall through and the menu shows no action at all, leaving only the raw error string.

## Change

- `ClaudeProviderImplementation.loginMenuAction`: when neither the web nor OAuth recovery branch matches and the surfaced Claude error is `ProviderFetchError.noAvailableStrategy(.claude)`, fall back to the existing `("Re-login at claude.ai", .loginToProvider(url:))` action.
- Extracted a small pure static seam `isNoAvailableStrategyError(_:)` so the routing stays unit-testable without live provider state.

## Scope / safety

- **No detection or auth-boundary changes** — source availability resolution is untouched; this only changes what the menu shows once the result is already "no strategy".
- Reuses the existing `MenuAction.loginToProvider` case and the existing "Re-login at claude.ai" label → **no new user-facing strings, no `.lproj` changes needed**.
- No new public type, no `UsageStore` field.

## Tests

- New `ClaudeLoginRecoveryActionTests` (3 tests, in-memory store seam, no Keychain prompts):
  - fully unconfigured Claude → returns the Re-login recovery action
  - no Claude error → no forced action
  - `isNoAvailableStrategyError` matches only the generic empty-plan error
- Regression `ClaudeSourcePlannerTests` (6) still pass. `swift build` clean, `swiftformat` clean.

## Before / After

- **Before:** menu shows `No available fetch strategy for claude.` and nothing actionable.
- **After:** menu additionally offers **Re-login at claude.ai** so a fully unconfigured user can recover.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - Provider probes: stop waiting indefinitely for inherited output pipes after subprocesses or CLI version checks exit (fixes #1531).
 - Menu bar: update visible usage values in place when a manual refresh completes instead of leaving the open provider card stale until the menu is reopened (fixes #1516).
 - Gemini: recognize the current `gemini-api-key` CLI auth setting so API-key sessions show the supported OAuth guidance instead of a misleading not-logged-in error (fixes #1511).
+- Claude: offer the "Re-login at claude.ai" recovery action when no fetch strategy is available, instead of showing only the generic "No available fetch strategy" error (fixes #1425). Thanks @LeoLin990405!
 - Kiro: keep usage refreshes bounded when CLI helpers retain output pipes, ignore termination, or are cancelled (fixes #1533). Thanks @kiranmagic7!
 - Xiaomi MiMo: cancel optional token-plan requests when the required balance request fails instead of delaying the error for up to 30 seconds.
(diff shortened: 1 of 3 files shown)


LeoLin990405:
Closing this. Running the full test suite surfaced that the change conflicts with the recovery model `ClaudeWebRecoveryMenuTests` already pins down: `Re-login at claude.ai` is intentionally web-specific (web session errors, or an unavailable web attempt), while a generic `noAvailableStrategy` with no web attempt is meant to keep the `Add Account` action — which already gives a fully-unconfigured provider a way forward. My fallback overrode that purpose-built behavior, so the premise doesn't hold. Withdrawing — sorry for the noise.

PR #1566, opened 2026-06-16
[codex] Add remote Codex cost logs

raykr:
## Summary

- add optional Codex remote cost sources that mirror `sessions/` and `archived_sessions/` JSONL over SSH/rsync
- merge enabled remote scans with local Codex cost usage in the app, CLI `cost`, and `serve` payloads
- expose Codex provider settings and document SSH configuration
- add maintainer changelog credit for @raykr

## Hardening

- reject option-shaped SSH targets before invoking `ssh` or `rsync`
- reset each mirror only after a successful remote listing, removing deleted files and preventing old hosts or scan windows from contaminating totals
- preserve the previous mirror when listing itself fails
- keep per-source scan caches isolated and disable cached snapshot hydration while remote sources are active
- use batch-mode SSH, bounded subprocesses, and the shared 1 MiB output capture limit

## Verification

Exact head: `d839eec9885097acaa06f226f50fe90420dea7eb`

- 31 focused remote-source, CLI-cost, and provider-settings tests passed
- `make check` passed; SwiftFormat and SwiftLint clean
- initial review found option injection and stale-mirror contamination; both fixed with regressions
- final whole-branch autoreview clean (0.78), no accepted/actionable findings
- live disposable Ubuntu SSH/rsync E2E returned exactly 321 synthetic remote tokens as `source: local+remote`
- after the remote JSONL disappeared, a forced refresh returned an empty daily report instead of the stale 321 tokens
- an option-shaped target failed with `Invalid remote Codex SSH target` before any network process could run

## Privacy

Remote logs remain opt-in. CodexBar copies only JSONL session logs to its local cache and parses them on device; credentials stay in the user's existing SSH configuration or agent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.36.2 — Unreleased
 
 ### Added
+- Codex: optionally merge remote SSH session logs into local cost summaries. Thanks @raykr!
 - Cursor: show personal on-demand spend alongside the shared team pool. Thanks @yashiels!
 - Codex: show available manual rate-limit reset credits and their next expiry for signed-in OAuth accounts. Thanks @rogdex24!
 - Mistral: add Vibe monthly-plan usage and menu bar metric selection. Thanks @lfmundim!
--- a/README.md
+++ b/README.md
@@ -138,12 +138,12 @@ show an incident indicator.
 - Provider-specific usage meters with reset countdowns.
 - Optional Codex web dashboard enrichments (code review remaining, usage breakdown, credits history).
 - Inline spend and usage charts for API-backed providers such as OpenAI, Claude Admin API, OpenRouter, LiteLLM, z.ai, MiniMax, Mistral, and AWS Bedrock.
-- Configurable cost-usage scans for Codex + Claude, plus reused chart UI for supported provider histories.
+- Configurable cost-usage scans for Codex + Claude, including optional SSH-mirrored remote Codex logs, plus reused chart UI for supported provider histories.
 - Provider status polling with incident badges in the menu and icon overlay.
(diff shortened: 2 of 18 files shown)


steipete:
Thanks for the ambitious contribution, @raykr. I re-audited the current exact head and am closing this form because the remaining gaps require redesign across acquisition, scanning, caching, failure behavior, and privacy—not isolated thread fixes.

The blocking issues:

- path-date filtering drops older active JSONL sessions that receive new entries inside the requested window;
- remote scans use a separate cache root, so refreshed models.dev pricing can be missed;
- retained cache identity is not bound to the full connection/home signature, allowing previous-host aggregates after configuration changes;
- local and remote roots are scanned separately and totals are added, bypassing the existing session/file deduplication and double-counting mirrored sessions;
- remote listing/sync failure occurs before local scanning and aborts the complete result; destination deletion before rsync also destroys the last-good mirror;
- full Codex session JSONLs can persist after a source is disabled or fails, without an explicit retention/cleanup contract.

The SSH argv boundary itself looks reasonable; this is not a command-injection closure.

A narrower rewrite would be welcome: start with one explicitly invoked CLI remote source, use a private atomic temporary mirror, scan local and remote roots together through one scanner/pricing context, match the local modified-file lookback, delete raw logs after deriving usage, and define either all-or-nothing failure or local results plus a structured remote warning. Add persistent app/background behavior only after privacy and cleanup semantics are approved.

Required proof: cross-window active session, refreshed-pricing-only model, duplicate local/remote session counted once, connection/config isolation, listing and rsync failure atomicity, raw-mirror permissions/cleanup, exact full CI, and a disconnect/reconnect E2E.

PR #1574, opened 2026-06-16
Cache Pi session parsers

ProspectOre:
## Summary
- cache the Pi session filename regex instead of recompiling it for each filename
- reuse a locked pair of ISO8601DateFormatter instances for Pi session start parsing
- keep the change private to PiSessionCostScanner and preserve the same filename format and parse order

## Why
Pi session refreshes walk session directories and parse every candidate filename before deciding whether it falls inside the requested scan window. The previous path rebuilt the regular expression and ISO formatters for each filename, which adds avoidable overhead on larger session directories.

## Proof
Local validation, June 16, 2026 PT:

- `.build/lint-tools/bin/swiftformat Sources Tests --lint`
  - `0/1139 files require formatting`
- `.build/lint-tools/bin/swiftlint --strict`
  - `Found 0 violations, 0 serious in 1138 files`
- `swift test --filter PiSessionCostScannerTests`
  - `10 tests in 1 suite passed after 0.066 seconds`
- `swift build --build-tests`
  - `Build complete! (1.08 sec)`
- `git diff --check`
  - passed with no output

Benchmark harness was temporary and is not committed. Both runs used the same public `PiSessionCostScanner.loadDailyReport` path over 2,500 synthetic session filenames, with `swift test --skip-build --filter PiSessionParserBenchmarkTests` for the timed pass.

| Metric | Before | After | Change |
| --- | ---: | ---: | ---: |
| fastest scanner pass | 0.678561s | 0.323273s | 2.1x faster |
| command real time | 6.99s | 3.10s | 2.3x faster |
| user CPU | 2.94s | 1.68s | 1.8x lower |
| max RSS | 177,963,008 bytes | 121,094,144 bytes | 32.0% lower |
| peak memory footprint | 29,098,680 bytes | 28,672,696 bytes | 1.5% lower |
| checksum | 2500 | 2500 | unchanged |

## Risk
Low. NSRegularExpression is immutable after construction, and the shared ISO8601DateFormatter pair is protected by an NSLock, matching the existing scanner timestamp-parser pattern elsewhere in the codebase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Changed
 - Antigravity: use current backend quota labels in menus and widgets while preferring a usable quota lane over an exhausted one. Thanks @Yuxin-Qiao!
+- Pi: cache session filename and timestamp parsers to reduce cost-history refresh overhead. Thanks @ProspectOre!
 - Menu bar: reuse the icon-observation signature during provider refreshes instead of computing it twice. Thanks @abe238!
 - LiteLLM: show personal and team spend amounts directly on budget rows while suppressing duplicate budget sections. Thanks @hololee!
 
--- a/Sources/CodexBarCore/PiSessionCostScanner.swift
+++ b/Sources/CodexBarCore/PiSessionCostScanner.swift
@@ -1,5 +1,20 @@
 import Foundation
 
+private final class PiSessionISO8601FormatterBox: @unchecked Sendable {
+    let lock = NSLock()
+    let withFractional: ISO8601DateFormatter = {
+        let formatter = ISO8601DateFormatter()
+        formatter.formatOptions = [.withInternetDateTime, .withFractionalSeconds]
+        return formatter
+    }()
+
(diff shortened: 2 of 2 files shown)


steipete:
Maintainer proof for rebased head `eecfe892`:

- Rebased cleanly onto current `main` (`70f37afc`).
- `swift test --filter PiSessionCostScannerTests`: 10 tests passed.
- `make check`: passed (SwiftFormat, strict SwiftLint, locale and repository checks).
- Structured Codex autoreview: clean; no accepted/actionable findings (confidence 0.86).
- Diff remains limited to the private Pi parser cache plus maintainer changelog credit.
- Public model identifier gate: not applicable; no model identifiers changed.

The contributor's scanner-path benchmark remains the relevant performance proof: 2,500 synthetic session filenames, unchanged checksum, with the cached implementation reporting a 2.1x faster scanner pass. This maintainer pass verified correctness and integration on the rebased candidate rather than presenting a second incomparable benchmark environment.

PR #1572, opened 2026-06-16
Trim menu formatting work

ProspectOre:
## Summary
- Avoid building a formatter for token counts below 1,000, where grouping never changes the output.
- Replace repeated Bedrock billing-day date parsing with validated `yyyy-MM-dd` day-key comparison and direct month/day rendering.
- Add focused coverage for small token counts and Bedrock latest valid billing-day selection.

## Before / After
Temporary benchmark harness, not committed. Same checksums before/after.

Baseline: `main` at `ac01d736`
Branch: `5ab41dc1`
Command: `/usr/bin/time -l swift test --skip-build --filter FormatterHotPathBenchmarkTests`

| Probe | Before | After | Change |
| --- | ---: | ---: | ---: |
| `tokenCountString` small values, 2,000,000 calls | 19.073202s | 0.870158s | 21.9x faster |
| Bedrock menu model, 5,000 builds over 30 daily entries | 13.940430s | 1.064601s | 13.1x faster |
| Focused run wall time | 38.14s | 2.94s | 13.0x faster |
| Max resident set size | 299.7 MiB | 108.9 MiB | 63.7% lower |
| Peak memory footprint | 27.4 MiB | 27.2 MiB | effectively flat |

Checksums:
- `tokenCountStringSmallValues`: `5780000` before and after
- `bedrockMenuModel`: `175000` before and after

## Verification
- `swift test --filter UsageFormatterTests` passed: 34 tests
- `swift test --filter BedrockMenuCardTests` passed: 2 tests
- `swift build --build-tests` passed
- `swiftlint --strict` passed: 0 violations in 1138 files
- `git diff --check` passed

## Notes
- No new logging, persistence, network behavior, or secrets handling.
- The fresh `main` benchmark worktree needed the local Sparkle framework symlink workaround under `.build/out/Products/Debug/PackageFrameworks` before running the focused test binary.

PR #1576, opened 2026-06-16
Bound subprocess output capture

ProspectOre:
## Summary

- Cap each subprocess stdout/stderr capture at 1 MiB while continuing to drain the pipe.
- Keep the existing default call sites intact with an optional capture limit for tests or future callers.
- Add regression coverage for oversized stdout and stderr.

## Why

A noisy or wedged provider CLI could previously append unbounded output into memory. The status, version, credential, and login probes using this path expect small strings or JSON, so bounded diagnostics are enough; continuing to drain the pipe keeps oversized-output children from blocking.

## Impact

This is a shared subprocess guardrail, not a provider-specific change. It covers the common pipe-capture path used by provider version detection, Codex login, Claude direct usage, Gemini process probes, Kiro status probing, Antigravity process/port probing, Bedrock AWS credential export, Vertex `gcloud` token lookup, Amp usage, and Auggie account status. Normal status, token, credential, and version outputs are small; the cap is meant to stop pathological or noisy child processes from growing CodexBar memory without bound while still draining the pipe so the child cannot deadlock.

## Proof

- `swift test --filter SubprocessRunnerTests` passed, 10 tests.
- `.build/lint-tools/bin/swiftformat Sources Tests --lint` passed, 0/1139 files required formatting.
- `.build/lint-tools/bin/swiftlint --strict` passed, 0 violations.
- `swift build --build-tests` passed.
- `git diff --check` passed.
- Before/after harness compiled `ProcessPipeCapture` plus `BoundedTaskJoin` from `upstream/main` and this branch, then streamed `64,000,000` stdout bytes through `/usr/bin/python3`:
  - Before: captured `64,000,000` bytes; benchmark `phys_peak=68,256,272`; `/usr/bin/time -l` peak memory footprint `68,338,240`; max RSS `72,990,720`.
  - After: captured `1,048,576` bytes; benchmark `phys_peak=3,342,792`; `/usr/bin/time -l` peak memory footprint `3,441,144`; max RSS `9,781,248`.
- `make check` local blocker: Node v25.9.0 rejects `node --check docs/site.js` because the file is ESM while the package is not marked as a module. `upstream/main:docs/site.js` starts with the same ESM import, so this is not introduced by this patch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - LiteLLM: show personal and team spend amounts directly on budget rows while suppressing duplicate budget sections. Thanks @hololee!
 
 ### Fixed
+- Provider probes: cap captured subprocess output at 1 MiB per stream without dropping valid text at a truncated UTF-8 boundary. Thanks @ProspectOre!
 - Provider switcher: keep Codex quota rows visible when switching away and back during a manual refresh, including menus with usage-history sections. Thanks @Yuxin-Qiao!
 - Bedrock: ignore invalid billing dates when selecting the latest usage values. Thanks @ProspectOre!
 - Localization: improve Japanese terminology consistency and localize next-day reset times across all 21 app languages. Thanks @tukuyomil032!
--- a/Sources/CodexBar/CodexLoginRunner.swift
+++ b/Sources/CodexBar/CodexLoginRunner.swift
@@ -188,6 +188,6 @@ struct CodexLoginRunner {
     }
 
     private static func decode(_ data: Data) -> String {
-        String(decoding: data, as: UTF8.self)
+        ProcessPipeCapture.decodeUTF8(data)
     }
 }
--- a/Sources/CodexBar/CodexLoginRunner.swift
+++ b/Sources/CodexBar/CodexLoginRunner.swift
@@ -188,7 +188,6 @@ struct CodexLoginRunner {
     }
 
     private static func decode(_ data: Data) -> String {
-        guard let text = String(data: data, encoding: .utf8) else { return "" }
-        return text
+        String(decoding: data, as: UTF8.self)
     }
 }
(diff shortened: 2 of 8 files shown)


steipete:
Maintainer repair pushed at `32e7087d6ba2499e6b0204ad786a1d043be1154b`.

Changes:
- Keep the contributor's 1 MiB cap per stdout/stderr stream while continuing to drain both pipes.
- Decode captured process output lossily at the shared boundary so a cap that splits a multibyte UTF-8 scalar preserves the valid prefix instead of returning an empty string.
- Apply that decoder to SubprocessRunner, Kiro, Codex login, Gemini helper probes, and provider version detection.
- Add a three-byte-scalar boundary regression and simultaneous 4 MiB stdout/stderr stress coverage.
- Add changelog credit for @ProspectOre.

Proof:
- `swift test --filter 'SubprocessRunnerTests|CodexLoginRunnerTests|ProviderVersionDetectorTests|GeminiStatusProbeAPITests|KiroStatusProbeTests'` - 68 tests passed.
- Exact process E2E: both stdout and stderr produced 4,096,000 bytes concurrently; each captured string was exactly 1,048,576 bytes and the child exited without deadlock.
- UTF-8 E2E: a three-byte scalar split after its first byte retained the full ASCII prefix and emitted one replacement scalar at the boundary.
- `make check` - format, locale, and strict lint gates passed with zero violations.
- Full branch Codex autoreview - no accepted/actionable findings (0.83 confidence).
- Rebased onto `421bf576`; diff check clean.

steipete:
Post-#1572 rebase: exact candidate is now 8ba4b637616ce2cc58ae4c7bf9a1cb21182f6040.

No source conflict occurred. Diff check is clean. The earlier proof's abbreviated 32e7087d was correct, but its expanded hash was mistyped; the actual prior head was 32e7087dc25e29591a79766a5709740dc2433613.

Post-rebase focused result: all subprocess, Kiro, Codex login, and version tests passed. The combined run hit one concurrent Gemini helper flake; isolated GeminiStatusProbeAPITests immediately passed 18/18. CI restarted on the current head.

PR #1586, opened 2026-06-16
fix: don't highlight non-interactive menu card items on hover

elijahfriedman:
## Summary

  Avoid highlighting non-interactive menu card rows on hover by only marking
  `NSMenuItem`s enabled when they actually expose interaction.

## Implementation

  - Updated `Sources/CodexBar/StatusItemController+MenuCardItems.swift`
  - Replaced unconditional `item.isEnabled = true` with `item.isEnabled =
  submenu != nil || onClick != nil`
  - Applied the same rule in both the non-rendered menu path and the hosting-
  view path
  - Preserves submenu rows and clickable rows as enabled
  - Leaves informational rows disabled so AppKit does not treat them as hover-
  highlightable actions

## Why

  The previous implementation marked every menu card item as enabled, including
  rows that were purely informational. That caused AppKit to render them like
  interactive menu items on hover, which implied behavior that does not exist.

## Testing

  - Built the app locally
  - Tested the updated hover behavior in the app

## Screenshots
Before:

Screenshot 2026-06-16 at 3 26 11 PM
Screenshot 2026-06-16 at 3 25 57 PM
After:

Screenshot 2026-06-16 at 3 50 39 PM

### DIFF
--- a/Sources/CodexBar/StatusItemController+MenuCardItems.swift
+++ b/Sources/CodexBar/StatusItemController+MenuCardItems.swift
@@ -31,7 +31,7 @@ extension StatusItemController {
     {
         if !self.menuCardRenderingEnabledForController {
             let item = NSMenuItem()
-            item.isEnabled = true
+            item.isEnabled = submenu != nil || onClick != nil
             item.representedObject = id
             item.submenu = submenu
             if submenu != nil {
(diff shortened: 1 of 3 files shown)


elijahfriedman:
@codex review

PR #1589, opened 2026-06-17
fix: don't highlight non-interactive menu card items on hover

elijahfriedman:
## Summary

  Avoid highlighting non-interactive menu card rows on hover by only marking
  `NSMenuItem`s enabled when they actually expose interaction.

## Implementation

  - Updated `Sources/CodexBar/StatusItemController+MenuCardItems.swift`
  - Replaced unconditional `item.isEnabled = true` with `item.isEnabled =
  submenu != nil || onClick != nil`
  - Applied the same rule in both the non-rendered menu path and the hosting-
  view path
  - Preserves submenu rows and clickable rows as enabled
  - Leaves informational rows disabled so AppKit does not treat them as hover-
  highlightable actions

## Why

  The previous implementation marked every menu card item as enabled, including
  rows that were purely informational. That caused AppKit to render them like
  interactive menu items on hover, which implied behavior that does not exist.

## Testing

  - Built the app locally
  - Tested the updated hover behavior in the app
  - Ran `swift test --filter MenuCardViewRecyclingTests`
  - Swift Testing reported 17 tests in 1 suite passed

## Screenshots
Before:

Screenshot 2026-06-16 at 3 26 11 PM
Screenshot 2026-06-16 at 3 25 57 PM
After:

Screenshot 2026-06-16 at 3 50 39 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Fixed
 - Localization: improve Japanese terminology consistency and localize next-day reset times across all 21 app languages. Thanks @tukuyomil032!
+- Menu bar: stop informational usage-card rows from highlighting like clickable actions. Thanks @elijahfriedman!
 
 ## 0.36.1 — 2026-06-16
 
--- a/Sources/CodexBar/StatusItemController+CodexStackedMenu.swift
+++ b/Sources/CodexBar/StatusItemController+CodexStackedMenu.swift
@@ -34,7 +34,8 @@ extension StatusItemController {
                     id: "menuCard-\(cardIndex)",
                     width: context.menuWidth,
                     heightCacheScope: account.id,
-                    heightCacheFingerprint: model.heightFingerprint(section: "card")))
+                    heightCacheFingerprint: model.heightFingerprint(section: "card"),
+                    containsInteractiveControls: true))
                 cardIndex += 1
                 if account.id != section.accounts.last?.id {
                     menu.addItem(.separator())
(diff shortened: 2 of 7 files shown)


elijahfriedman:
@clawsweeper re-review

steipete:
Maintainer proof on improved head `b536e8de`:

- Added direct coverage that informational card rows are disabled while click and submenu cards remain enabled.
- `swift test --filter MenuCardViewRecyclingTests`: 18 tests passed.
- `make check`: passed.
- Autoreview: clean, no actionable findings.
- Built and launched the exact branch bundle with ad-hoc signing. Peekaboo successfully clicked the verified CodexBar status item frame. Live Accessibility inspection showed informational custom rows disabled and actionable/submenu rows enabled, including Buy Credits, Cost, Add Account, Usage Dashboard, Status Page, Refresh, Settings, About, and Quit.
- macOS Screen Recording permission is unavailable in this session, so the proof is live Accessibility state rather than a screenshot. Accessibility and event-synthesis permissions are granted.

Exact-head CI is now running on the pushed branch.

steipete:
Maintainer follow-up after autoreview:

- Preserved embedded controls on informational cards without restoring card-level hover highlighting.
- Added focused coverage for enabled embedded controls, suppressed card highlighting, recycling, and interaction-state behavior.
- `swift test --filter MenuCardViewRecyclingTests`: 19 tests passed.
- `make check`: passed with zero SwiftLint violations.
- Exact-head live proof on packaged `13d6983d`: launched an isolated LiteLLM configuration against a closed local endpoint, opened the menu through Accessibility, activated the error card's copy control, and verified the clipboard contained `Could not connect to the server.`.

Screen Recording permission is unavailable on this host, so the live proof used Accessibility inspection and event synthesis rather than a screenshot.

Issue #1126, opened 2026-05-23
Add provider-neutral usage confidence metadata for diagnostics


Yuxin-Qiao:
CodexBar fetches usage data from a variety of sources: official APIs, OAuth endpoints, browser cookies, local CLI session logs, PTY parsing, and dashboard scraping. Each source carries different guarantees about data precision.

When a user sees a percentage in the menu bar, they cannot tell whether it is:

- An exact value from an official API with precise reset timing
- An approximation derived from periodic sampling or log parsing
- A percentage-only display with no absolute token counts
- Something else entirely

This is a problem for both users making quota decisions and maintainers debugging unexpected behavior.

## Current State

- `UsageSnapshot` and `RateWindow` carry usage values but include no data quality classification.
- `UsageStore.isStale(provider:)` returns `true` only when a fetch error exists — it does not communicate whether data is precise or approximate when fetch succeeds.
- `ProviderFetchKind` and `ProviderFetchAttempt` track which strategy was used, but this is internal diagnostic state, not structured metadata accessible to the UI or notification logic.
- Provider-specific enums handle fetch routing internally, but there is no shared model that the UI or app logic can use without provider-specific switch statements.

## Proposed Solution

Introduce a provider-neutral `UsageDataConfidence` enum on `UsageSnapshot` that classifies data quality:

```swift
public enum UsageDataConfidence: Equatable, Sendable {
    /// Usage data is exact. Used/remaining values and reset timing are precise.
    case exact
    /// Usage data is approximate. Values are derived from sampling, log parsing, or PTY output.
    case estimated
    /// Only a percentage is available. No absolute token counts or precise reset time.
    case percentOnly
    /// Confidence level could not be determined.
    case unknown
}
```

**Important:** Confidence should be derived from the response fields present — not from which auth method was used. An API token fetch is not automatically `.exact`; a cookie-based fetch is not automatically `.estimated`. The actual response structure determines the classification.

Stale or unavailable are separate concerns from data confidence and are out of scope for this issue.

## Suggested First PR

Keep the initial change small and observational:

1. Add `UsageDataConfidence` to `UsageSnapshot`, defaulting to `.unknown`.
2. Surface it in debug/diagnostic output only — not in visible UI — so the values can be validated before committing to a UI design.
3. Wire it for one provider where the response structure is well-understood and the confidence mapping can be verified in tests. No assumptions about other providers until behavior is confirmed.
[...]

PR #1594, opened 2026-06-17
Add usage data confidence metadata

Yuxin-Qiao:
## Summary

- Add provider-neutral `UsageDataConfidence` metadata on `UsageSnapshot`.
- Default legacy, future, or unwired snapshots to `unknown` without re-encoding that default into cached snapshot JSON.
- Mark Codex OAuth usage as `exact` only when every advertised primary, weekly, and model-specific window decodes cleanly.
- Surface confidence in safe provider diagnostic JSON; no menu UI, fetch routing, fallback, or quota-calculation changes.
- Preserve decoding of legacy diagnostic JSON that predates the confidence field.

## Why

This is the first observational slice for #1126. `UsageSnapshot` is the shared diagnostic boundary, so confidence stays provider-neutral without leaking fetch-strategy details into UI or notification logic.

## Proof

- Combined Codex OAuth and provider diagnostic suites: 51 tests passed after the final current-main rebase.
- `make check` passed after the final current-main rebase.
- Local compatibility-fix autoreview: clean, confidence 0.86
- Full branch autoreview against `origin/main`: clean, confidence 0.82
- Deterministic OAuth fixtures prove `exact` for fully decoded windows and fail closed to `unknown` for malformed, future, and legacy data.
- Live authenticated Codex diagnostic returned source `openai-web`, usage present, no error, and `dataConfidence: unknown`. The account emitted no rate-limit event, so the branch reports the absence honestly instead of manufacturing positive `exact` proof.

Exact candidate: `3ac8e279cfed15077d92d6450901572722cc103b`. Its tree is identical to reviewed source parent `3de34dfebc02`; the final commit only retriggered CI.

Refs #1126

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Codex: show available manual rate-limit reset credits and their next expiry for signed-in OAuth accounts. Thanks @rogdex24!
 - Linux CLI: publish static musl release tarballs for x86_64 and aarch64. Thanks @Yuxin-Qiao!
 - Documentation: add safe troubleshooting for browser Keychain prompts that persist after uninstall. Thanks @Yuxin-Qiao!
+- Diagnostics: report provider-neutral usage confidence and mark fully decoded Codex OAuth windows exact. Thanks @Yuxin-Qiao!
 - Codex agents: add a read-only `codexbar` skill for bounded, redacted provider usage JSON. Thanks @coygeek!
 - Display: add a Hide critters option for plain menu bar quota capsules. Thanks @elijahfriedman!
 
--- a/Sources/CodexBarCore/Providers/Codex/CodexOAuth/CodexOAuthUsageFetcher.swift
+++ b/Sources/CodexBarCore/Providers/Codex/CodexOAuth/CodexOAuthUsageFetcher.swift
@@ -9,6 +9,7 @@ public struct CodexUsageResponse: Decodable, Sendable {
     public let credits: CreditDetails?
     /// Model-specific limits (e.g. GPT-5.3-Codex-Spark) that sit alongside the primary/weekly windows.
     public let additionalRateLimits: [AdditionalRateLimit]?
+    let additionalRateLimitsDecodeFailed: Bool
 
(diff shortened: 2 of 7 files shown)


Yuxin-Qiao:
Addressed the Codex review feedback in `2ab24e8`: Codex OAuth now reports `exact` confidence only when the core rate-limit windows decoded without a window decode failure. Partial OAuth responses that preserve one valid window while dropping a malformed sibling stay at `unknown` confidence.

Validation:
- `swift test --filter CodexOAuthTests --skip-update` (29 tests passed)
- `git diff --check`

steipete:
Final rebase proof for exact head `e2f93a36`:

- conflict was changelog-only; the landed Codex deadline implementation auto-merged with confidence propagation
- 50 focused Codex OAuth/diagnostic tests passed
- `make check` passed
- full branch autoreview clean (0.84)
- exact release CLI built successfully
- an authenticated Codex turn completed, but the current local CLI emitted no rate-limit event; exact provider output therefore correctly reported `no-rate-limit-event` instead of fabricating confidence
- prior reviewed-head live diagnostic proved provider usage with confidence `unknown`; exact deterministic fixtures prove fully decoded OAuth windows become `exact` and malformed windows fail closed to `unknown`
- public model identifier gate: N/A; no model-bearing content changed

Exact-head CI is running. Live quota proof remains account-state limited until the local Codex CLI emits a rate-limit event.

steipete:
Exact-head proof for `3ac8e279cfed15077d92d6450901572722cc103b`:

- Four macOS test shards, Linux x64/arm64, lint aggregate, and security are green.
- The final CI-trigger commit has the same tree as reviewed source parent `3de34dfebc02`.
- 51 focused OAuth/diagnostic tests and `make check` passed on the reviewed source.
- Compatibility-fix and full-branch autoreviews were clean (0.86 / 0.82).
- Live diagnostics honestly reported `unknown` confidence because the signed-in account emitted no rate-limit event; deterministic fixtures cover both `exact` and fail-closed `unknown`.

Merge-clean and ready to land.

PR #1599, opened 2026-06-17
Fix opencodego manual cookie usage on Linux CLI

### LINKED ISSUE #1596: opencodego fails on Linux even with manual cookie — platform gate blocks before pipeline
### Description

The `opencodego` provider does not work on Linux at all — even with `cookieSource: manual` and a valid cookie header configured. The error is:

```
Error: selected source requires web support and is only supported on macOS.
```

### Root Cause

There is a platform-level gate in `CLIUsageCommand.swift` (`#if !os(macOS)`) that runs **before** the fetch pipeline is even entered:

```swift
#if !os(macOS)
if Self.sourceModeRequiresWebSupport(effectiveSourceMode, provider: provider, ...) {
    return Self.webSourceUnsupportedOutput(...)
}
#endif
```

The check `sourceModeRequiresWebSupport` resolves to:

```swift
case .auto:
    descriptor.fetchPlan.sourceModes.contains(.web)  // true for opencodego
```

Since `opencodego` advertises `[.auto, .web]` as its source modes, it hits this gate and exits immediately — even when `cookieSource` is set to `.manual`.

### But Manual Cookie Doesn't Need WebKit

The manual cookie path in `OpenCodeWebCookieSupport.resolveCookieHeader` has no macOS dependency:

```swift
if let settings = context.settings, settings.cookieSource == .manual {
    if let header = self.requestCookieHeader(from: settings.manualCookieHeader) {
        return header  // purely reads string from config, no browser import
    }
}
```

With a manual cookie, the web strategy is just an HTTP GET to `opencode.ai/_server` — no Safari/Chrome cookie import, no WebKit. It would work fine on Linux.

Similarly, `OpenCodeGoLocalUsageFetchStrategy.cachedOrManualCookieHeader` already handles manual cookies cross-platform, but the gate prevents it from being reached.

### Suggested Fix

Rather than blanket-blocking all providers that list `.web` in `sourceModes`, the gate should account for the effective cookie source. For example, if `cookieSource == .manual`, or the provider has no actual macOS-only dependency given the current settings, allow the pipeline to run and let the strategies handle availability.

### Environment

- CodexBar 0.36.1 (Linux CLI)
- `cookieSource: manual` with valid auth cookie in config
- Platform: Linux (no WebKit)

Yuxin-Qiao:
Fixes #1596

## Root cause

- The non-macOS CLI web-support gate checked the effective source mode and provider catalog before OpenCode Go resolved its configured cookie source.
- OpenCode Go advertises web support for automatic mode, so Linux exited before `OpenCodeWebCookieSupport.resolveCookieHeader` could use a manual cookie.
- Manual cookie normalization needs no browser, WebKit, or Keychain access.

## Fix

- Keep the non-macOS web-support gate.
- Exempt only OpenCode Go with `cookieSource: manual`.
- Keep automatic/browser cookie import unavailable on non-macOS.
- Preserve the newer Codex and Ollama provider-specific source exceptions from current `main`.
- Add regression coverage for manual OpenCode Go, automatic OpenCode Go, and unrelated OpenCode settings.

## Maintainer validation

Exact candidate: `fca6adeb`

- `swift test --filter 'CLIEntryTests|OpenCodeGo|Zen'` - 22 CLI gate tests and 57 OpenCode Go/Zen provider tests passed after the final current-main rebase.
- `make check` - passed after the final current-main rebase; SwiftFormat clean and SwiftLint reported zero violations.
- Codex autoreview of the full branch against current `main` - clean, confidence 0.86.
- Fresh Ubuntu 24.04 x86_64 Crabbox proof built the static release CLI from the exact candidate with Swift 6.2.1. Explicit web mode plus a configured manual test cookie reached the real OpenCode Go provider path and returned an endpoint/auth response rather than the non-macOS platform gate. The lease was released after the passing run.
- macOS provider-path validation found no usable browser session cookie; no positive account-usage claim is made.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
 - Menu bar: show provider status markers only for the provider rendered in each icon. Thanks @Zihao-Qi!
 - Codex CLI: make automatic usage reads prefer OAuth and CLI sources instead of blocking on the optional web dashboard.
 - Codex web: apply `--web-timeout` to the full cookie import, account verification, retry, and dashboard fetch path.
+- OpenCode Go: allow configured manual cookies in the Linux CLI while keeping browser-cookie import gated to macOS. Thanks @Yuxin-Qiao!
 - Provider probes: cap captured subprocess output at 1 MiB per stream without dropping valid text at a truncated UTF-8 boundary. Thanks @ProspectOre!
 - Provider switcher: keep Codex quota rows visible when switching away and back during a manual refresh, including menus with usage-history sections. Thanks @Yuxin-Qiao!
 - Bedrock: ignore invalid billing dates when selecting the latest usage values. Thanks @ProspectOre!
--- a/Sources/CodexBarCLI/CLIUsageCommand.swift
+++ b/Sources/CodexBarCLI/CLIUsageCommand.swift
@@ -551,6 +551,11 @@ extension CodexBarCLI {
         if provider == .codex, sourceMode == .auto {
             return false
         }
+        if provider == .opencodego,
+           settings?.opencodego?.cookieSource == .manual
+        {
+            return false
+        }
         if provider == .ollama,
            sourceMode == .auto
         {
--- a/Tests/CodexBarTests/CLIEntryTests.swift
+++ b/Tests/CodexBarTests/CLIEntryTests.swift
@@ -326,6 +326,38 @@ final class CLIEntryTests: XCTestCase {
         XCTAssertFalse(CodexBarCLI.sourceModeRequiresWebSupport(.api, provider: .kilo))
         XCTAssertFalse(CodexBarCLI.sourceModeRequiresWebSupport(
             .auto,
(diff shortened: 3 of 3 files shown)


steipete:
Landed after exact-head verification at `fca6adebd7d3b1d34027432cc8f7a4e18a869f42`.

Proof:
- 22 CLI gate tests and 57 OpenCode Go/Zen provider tests passed.
- `make check` passed; SwiftFormat clean and SwiftLint reported 0 violations.
- Maintainer autoreview found no actionable defects (0.86).
- Fresh Ubuntu 24.04 x86_64 E2E built the static release CLI with Swift 6.2.1. Manual-cookie web mode reached the real OpenCode Go provider path and returned the expected endpoint/auth response instead of the non-macOS platform gate; the test box was released.
- Exact-head CI passed four macOS shards, Linux x64/arm64, aggregate lint, and security.

Thanks @Yuxin-Qiao.

Issue #1524, opened 2026-06-14
Linux CLI: ship fully-static (musl) builds so it runs on older glibc (Ubuntu 20.04 / Jetson) and NixOS


guhyun9454:
Follow-up to #1039 (Ubuntu 20.04) and #198 (musl / NixOS). I dug into this and have a **working fully-static aarch64 build**, so opening this to confirm direction before sending a PR.

## Problem

The Linux CLI release artifacts fail on any distro whose glibc is older than the CI builder's:

```
codexbar: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found
codexbar: /lib/aarch64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.30' not found
```

Repro: NVIDIA Jetson AGX (L4T R35.6.4 = Ubuntu 20.04, glibc 2.31, aarch64). Same class as #1039 on x86_64. Even the oldest published tarball (v0.15.0) requires `GLIBC_2.38` + `GLIBCXX_3.4.30`.

## Root cause (build environment, not code)

`release-cli.yml` / `ci.yml` build on `ubuntu-24.04` / `ubuntu-24.04-arm` with only `--static-swift-stdlib`. That statically links the *Swift* runtime, but `libc` / `libstdc++` / `libcurl` / `libsqlite3` stay **dynamically** linked against the 24.04 host, so the binary inherits a glibc 2.38 / GLIBCXX 3.4.30 floor.

## What works: fully-static musl build

Building `CodexBarCLI` with the **Swift Static Linux SDK (musl)** produces a fully static binary with **zero** dynamic dependencies, which runs unmodified on glibc 2.31:

```
swift sdk install <swift-6.2.1 static-linux SDK>
swift build -c release --product CodexBarCLI --swift-sdk aarch64-swift-linux-musl
```

Verified on the Jetson (glibc 2.31): `--version`, `config providers`, and a live `usage --provider codex --source oauth --json` (real OAuth refresh over TLS, returned usage windows) all work. `readelf -d` shows no `NEEDED` entries. This also solves NixOS (#198) and is immune to future SONAME bumps like the `libxml2.so.2 → .so.16` break in #1046.

## What it took (PR surface)

[...]

PR #1609, opened 2026-06-18
Add opt-in SQLite path for static Linux CLI builds

Yuxin-Qiao:
Refs #1524

Second step of the musl CLI stack, now rebased as one commit on the landed #1620 source-compatibility base.

## Summary

- add an opt-in `CODEXBAR_SQLITE3_LIB_DIR` Package.swift build contract
- pass that directory as a Linux-only linker search path for `CodexBarCore` and `CodexBarCLI`
- keep default builds unchanged when the variable is unset, empty, or whitespace-only

The Swift Static Linux SDK does not ship `libsqlite3.a`, while SweetCookieKit links SQLite. Release callers can now provide a target-architecture static library without changing normal macOS or glibc builds.

No release workflow or artifact-matrix changes are included here; #1610 consumes this hook.

## Verification

- exact `swift package dump-package` assertions for unset, blank, and a configured path containing spaces
- `swift test --filter CLIEntryTests` (22 tests)
- `make check`
- autoreview: clean, no actionable findings (0.87)
- isolated Ubuntu 24.04 x86_64 build with checksummed Swift 6.2.1 static Linux SDK and SQLite 3.53.2 amalgamation
- resulting CodexBarCLI: x86-64 static ELF; `--help` and `config validate --format json` both executed successfully

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Codex agents: add a read-only `codexbar` skill for bounded, redacted provider usage JSON. Thanks @coygeek!
 
 ### Changed
+- Linux CLI: accept an opt-in static SQLite library directory for musl builds. Thanks @Yuxin-Qiao!
 - Linux CLI: add musl source compatibility for static Linux SDK builds. Thanks @Yuxin-Qiao!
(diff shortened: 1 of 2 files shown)


Yuxin-Qiao:
@clawsweeper re-review

PR #1573, opened 2026-06-16
Cache cost usage date parsers

ProspectOre:
## Summary

- cache ISO 8601 and localized cost-report date formatters per thread
- preserve current timezone changes in cache keys
- preserve all existing parse formats and precedence
- add mixed-format concurrent parse coverage and changelog credit

## Proof

- `swift test --filter CostUsageDecodingTests` - 19 tests passed after the current-main rebase
- `swift test --filter OpenAIDashboardBrowserCookieImporterTests` - 13 tests passed
- 32 concurrent workers, 56,000 mixed parses: passed in about 2.5 seconds
- direct 14,000-parse benchmark: uncached 2.42 seconds, cached 0.59 seconds, about 4.1x faster
- `make check` passed after the current-main rebase
- whole-branch Codex autoreview clean, confidence 0.85
- exact CI shard 3/4 command passed all 120 groups locally while the Intel hosted shard completed successfully
- prior exact head passed Linux x64/arm64, three macOS shards, and security before its remaining old-base job was cancelled after #1570 made that head conflicting

Exact candidate: `c29f4e5b`.

PR #1639, opened 2026-06-18
Make cookie deadline timing tests saturation-safe

steipete:
## Summary

- measure when cookie-import timeout callbacks win instead of when a starved async test task resumes
- race callback deadlines against delayed completions so timeout enforcement remains covered
- verify timeout observers stay silent when the operation wins

## Proof

- `swift test --filter OpenAIDashboardBrowserCookieImporterTests`: 13 passed
- combined with #1573: 32 parser/cookie tests passed under the 32-worker parser stress
- `make check`
- final `autoreview --mode local`: clean, 0.87

## Context

#1573 exposed a test-only assumption: Swift's cooperative executor can delay the awaiting test task for roughly 1.9 seconds under CPU saturation even though the 50 ms timeout callback already won. The old wall-clock assertions measured caller resumption, not deadline delivery. The observer is nil by default and runs only inside the winning timeout completion guard; existing production callers are unchanged.

### DIFF
--- a/Sources/CodexBarCore/OpenAIWeb/OpenAIDashboardBrowserCookieImporter+Deadline.swift
+++ b/Sources/CodexBarCore/OpenAIWeb/OpenAIDashboardBrowserCookieImporter+Deadline.swift
@@ -44,6 +44,7 @@ extension OpenAIDashboardBrowserCookieImporter {
 
     nonisolated static func runBoundedCookieLoad<T: Sendable>(
         deadline: Date?,
+        timeoutObserver: (@Sendable () -> Void)? = nil,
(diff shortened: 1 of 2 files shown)


steipete:
Landed as `f2f9d8bc50bf2d17bf5f7ba55d2a855e28b099a4` from exact reviewed head `8cc0b974f3a4eff96f9643b99c9c0ab6ff00c0cb`.

Proof: all four macOS test shards, Linux x64/arm64 builds and tests, lint, aggregate, and security checks passed. Local saturation stress also passed with the timeout observer measuring the deadline independently of delayed Swift task resumption.

PR #1629, opened 2026-06-18
ci: shard macOS Swift tests

Yuxin-Qiao:
## Summary
- split macOS lint and Swift tests into separate CI jobs
- run macOS Swift test suite groups across 4 matrix shards
- keep the existing `lint-build-test` check as an aggregate gate for branch protection compatibility
- extend the sharding self-test to prove shard union equals the full test list

## Local validation
- `python3 -m py_compile Scripts/ci_swift_test_by_suite.py`
- `bash -n Scripts/test.sh`
- `bash -n Scripts/test_swift_test_sharding.sh`
- `bash Scripts/test_swift_test_sharding.sh`
- `git diff --check`

This is a draft PR to validate whether CI wall-clock time improves before requesting review.

PR #1643, opened 2026-06-19
ci: keep lint independent of macOS test shards

Yuxin-Qiao:
## Summary

- run the macOS lint job independently from the Swift test shard matrix
- keep the existing `lint-build-test` aggregate gate depending on both lint and the macOS shards
- avoid turning superseded/cancelled workflow runs into misleading aggregate failures

## Rationale

The shard split from #1629 reduced macOS test wall-clock time, but `lint` still had `needs: swift-test-macos`. That serialized lint behind the slowest shard and skipped lint feedback whenever the matrix failed or was cancelled. The aggregate gate is the right place to depend on both producers; lint itself should remain a fast, independent signal.

The aggregate job still runs on real dependency failures through `always()`, but now skips when the workflow itself is cancelled, which avoids the noisy `1 failed, 5 cancelled` pattern on runs superseded by a newer push.

## Validation

- `ruby -e "require 'yaml'; YAML.load_file('.github/workflows/ci.yml'); puts 'yaml ok'"`
- `bash Scripts/test_swift_test_sharding.sh`
- `git diff --check`
- Completed exact-head CI run: https://github.com/steipete/CodexBar/actions/runs/27802962340
  - `lint`: passed in 1m31s
  - `swift-test-macos (0, 4)`: passed in 16m56s
  - `swift-test-macos (1, 4)`: passed in 24m45s
  - `swift-test-macos (2, 4)`: passed in 21m29s
  - `swift-test-macos (3, 4)`: passed in 19m37s
  - `lint-build-test`: passed in 3s after lint and macOS shards completed
  - `build-linux-cli` x64/arm64 and GitGuardian also passed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,8 +13,6 @@ jobs:
   lint:
     runs-on: macos-15-intel
     timeout-minutes: 10
-    needs:
-      - swift-test-macos
     steps:
       - uses: actions/checkout@v6
 
@@ -86,7 +84,7 @@ jobs:
     needs:
       - lint
       - swift-test-macos
-    if: ${{ always() }}
+    if: ${{ always() && !cancelled() }}
     steps:
       - name: Verify macOS lint and test jobs
         run: |


steipete:
Exact-head maintainer proof for `e272cb3aa5a9c03313d4ffc488816db1560dc71a`:

- Hosted CI passed all four macOS shards, independent lint, Linux x64/arm64 builds and tests, aggregate gating, and security checks.
- Manual diff review is clean: `lint` no longer waits on the shard matrix, while `lint-build-test` still depends on and verifies both producers.
- The aggregate retains `always()` for dependency failures and adds `!cancelled()` only to avoid manufacturing a failure after the workflow itself is superseded or cancelled.
- YAML parsing, sharding-script validation, and `git diff --check` are recorded on the exact PR head.

Ready to land.

Yuxin-Qiao:
@clawsweeper re-review

Completed exact-head CI proof has been added to the PR body, and maintainer proof is posted in the thread. All checks are green, including independent lint, four macOS shards, Linux x64/arm64, GitGuardian, and the aggregate lint-build-test gate.

PR #1644, opened 2026-06-19
Fix Xiaomi MiMo fallback after auth redirect

### LINKED ISSUE #1548: Currently, the tool only supports reading Chrome cookies. I would appreciate it if support could be added for other browsers, like Safari and ChatGPT Atlas.

Yuxin-Qiao:
## Summary

- Treat Xiaomi MiMo API auth redirects as `loginRequired` instead of a terminal network error.
- Preserve browser import order while allowing a stale cached/Chrome session to fall through to a later valid imported session.
- Add focused regression coverage for redirects and stale-Chrome-to-Safari fallback.

## Root Cause

A stale session can make the MiMo API return a login redirect. Classifying that redirect as a terminal network error prevents the web strategy from clearing stale state and trying later browser sessions. Mapping it to login-required uses the existing bounded retry path.

## Proof

- `swift test --filter MiMoProviderTests`: 35 tests passed on the exact head.
- `make check`: format and strict lint clean on the exact head.
- Exact-head Codex autoreview clean, confidence 0.82.
- Exact CLI browser-cookie routing reached the real Chrome Safe Storage Keychain prompt; the read stopped there because unattended login-keychain entry was unavailable. No live account-success claim is made.

Exact candidate: `215584c29306bb016f35662a235fe48cf703ca7f`.

Fixes #1548.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - LiteLLM: show personal and team spend amounts directly on budget rows while suppressing duplicate budget sections. Thanks @hololee!
 
 ### Fixed
+- Xiaomi MiMo: retry another imported browser session when a stale session redirects API requests to login. Thanks @Yuxin-Qiao!
 - Kiro: keep parsed usage available when the optional account probe times out or fails. Thanks @Yuxin-Qiao!
 - Memory: release idle OpenAI WebViews under system pressure without blocking the main thread. Thanks @ProspectOre!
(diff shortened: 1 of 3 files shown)


steipete:
Maintainer proof override: accepting deterministic redirect/fallback coverage plus the focused MiMo suite for this change. Repository policy forbids unsolicited live credential and Keychain probes; the patch changes fallback classification, not credential storage.

Current proof: 35 focused tests green, make check green, full-branch autoreview clean (0.86). Exact-head CI is 7/8 green; macOS shard 1 remains queued.

Issue #1633, opened 2026-06-18
Show available Codex rate-limit reset credits


rogdex24:
Codex sometimes grants users manual rate-limit reset credits, but CodexBar does not currently show whether any resets are available or when they expire.

This means users may have reset credits on their account without knowing they exist, or may miss the chance to use them before they expire.

It would be helpful if CodexBar displayed this information somewhere in the Codex usage view.

## What I’d like to see
Show how many Codex rate-limit resets are currently available.

Show when the next available reset expires.

Keep the display simple and easy to understand.

#### Example:

Rate-limit resets: 2 available
Next expires: Jul 18

## Why this matters
Manual reset credits are useful when Codex usage limits are exhausted, but they are easy to miss if they are only visible through backend/account data. Showing them in CodexBar would make the app more complete as a Codex usage monitor.

## Notes
This should only be informational for now. The issue is about making the reset-credit data visible, not necessarily adding a button to redeem or use a reset.

rogdex24:
I've opened a PR addressing this issue! You can review the changes and UI integration here: https://github.com/steipete/CodexBar/pull/1638

Issue #1502, opened 2026-06-14
Feature request: utilize codex reset from codexbar


jtl06:
Hi, OpenAI recently added the ability to reset Codex limits on demand. It would be nice If I could reset my limits using Codexbar, rather than needing to use the codex app. Not sure how this could be implemented, but if it is possible, it would be a very nice UX improvement.

Thanks

aaamosh:
For anyone implementing this in CodexBar — the endpoint contract is small enough to fit in a comment. Verified live against a real Plus account on 2026-06-13 (we sent the actual POST and burned a real banked credit; weekly window went 100% → 0%, `windows_reset: 1` in the response).

Base: `https://chatgpt.com/backend-api`. Same `Authorization: Bearer <access_token>` + `ChatGPT-Account-Id: <account_id>` headers CodexBar already uses for `/wham/usage`.

```
GET  /wham/rate-limit-reset-credits           → { credits: [{id, reset_type:"codex_rate_limits", status:"available"|"redeemed", granted_at, expires_at, ...}], available_count }
POST /wham/rate-limit-reset-credits/consume   body {"credit_id": "<id>", "redeem_request_id": "<uuid4>"}
                                              → { code: "reset", credit:{status:"redeemed", redeemed_at}, windows_reset: 1 }
```

Edge case worth handling: in the `consume` response the server collapses `title`, `description`, `profile_image_url`, `profile_user_id` to `null` even though those fields are populated in the `GET /credits` listing. Be lenient on those in the response model.

Endpoints came from the official `openai.chatgpt` VS Code extension's webview bundle (`webview/assets/codex-api-*.js`). There's a tiny no-deps Python reference implementation at https://github.com/aaamosh/codex-reset if it helps as a sanity check — it's the same three calls plus a `before/after` pair on `/wham/usage` to demonstrate the reset took effect.

Soju06/codex-lb#1015 was an attempt at the same feature; it's currently closed pending rework but the consume client there ([code](https://github.com/Soju06/codex-lb/blob/30da9ec6/app/core/clients/rate_limit_reset_credits.py)) matched the payload exactly, so the contract above should be portable to whatever CodexBar wants to do with it.

steipete:
#1690 now contains the repaired, mechanically reviewed implementation at exact head `00da631f5`. A request stub verifies the consume path, selected-account header, credit ID, and unique request ID without performing a live provider-side write; focused suites, localization coverage, `make check`, packaged-app proof, and autoreview are green.

[...]

PR #1638, opened 2026-06-18
feat: Add Codex rate-limit reset credits to menu card

rogdex24:
## Summary

Show available Codex rate-limit reset credits and the next usable credit expiry in the app and CLI for signed-in OAuth accounts.

Fixes #1633.

## Changes

- Fetch the account-scoped reset-credit endpoint as bounded, best-effort OAuth enrichment with a 4-second timeout.
- Preserve positive reset-credit results even when the usage response has no rate-limit windows; empty reset-credit responses still surface missing usage.
- Render reset counts and expiry in menu cards and `codexbar usage`, respecting optional-usage and reset-time preferences.
- Ignore stale expired credits when selecting the next expiry.
- Add all 21 app localizations, height/recycling fingerprints, changelog credit, and focused request/parser/model/CLI regressions.
- Keep required usage authoritative and avoid sibling `async let` lifetime hazards.

## Verification

Exact head: `57170e008edbbfaa615802c44486251e6a2b3c26`

- `swift test --filter 'CodexOAuthTests|CodexRateLimitResetCreditsTests|CodexResetCreditsMenuCardTests|CLISnapshotTests|MenuCardHeightFingerprintTests'` (64 tests)
- `make check`
- Full-branch autoreview: no actionable findings
- Live OAuth CLI: exit 0, OAuth source, no provider error, two available reset credits returned
- Direct account-scoped endpoint: HTTP 200 with the expected credit schema
- Fresh release package built and launched; local menu screenshot was blocked by the macOS app-data privacy prompt, which was left untouched

## Screenshot

Codex reset credits in the usage card

PR #1647, opened 2026-06-19
Show Codex banked reset availability

davezfr:
## Summary

Adds read-only visibility for Codex banked rate-limit resets in the Codex menu.

This surfaces the number of available banked resets and their expiry dates using the Codex OAuth credentials already available to CodexBar. The feature is intentionally informational only; it does not redeem or consume resets.

Refs #1633.
Refs #1502.

## What changed

- Add Codex banked reset models and a read-only fetcher for `wham/rate-limit-reset-credits`.
- Extract a shared `loadRefreshed` credential path and route both Codex usage and banked resets through it.
- Add a shared ChatGPT backend base URL resolver used by Codex OAuth usage and banked resets.
- Store banked resets as independent Codex state with its own refresh task, error state, cache fallback, and account-scoped apply guard.
- Add banked resets to the Codex consumer projection.
- Show banked resets in the Codex menu section, including available count and expiry dates.
- Schedule a banked reset refresh when opening the Codex menu if the data has not been loaded yet.
- Include banked resets in the open-menu readiness signature so an open Codex menu refreshes when the data arrives.
- Fix closed-menu preparation so a newly attached menu is not forced to rebuild by a superseded required-rebuild version.
- Preserve Codex state when the same account is enriched from email-only identity to provider-account identity.
- Seed the initial menu localization signature to avoid a spurious localization-change refresh on the first settings evaluation.
- Document the banked resets endpoint and refresh strategy in `docs/codex.md`.

## Notes

This is read-only by design. Redeeming a reset is a separate write operation, consumes a scarce account resource, and is harder to test safely.

The decoder is conservative: malformed reset entries are skipped instead of failing the whole response, and unknown reset statuses are not counted as available.

## Testing

- `swift test --filter CodexBankedResetsModelsTests`
- `swift test --filter CodexBankedResetsFetcherTests`
- `swift test --filter CodexAccountScopedRefreshBankedResetsTests`
- `swift test --filter MenuSessionCoordinatorTests`
- `swift test --filter StatusMenuInstantOpenTests`
- `swift test --filter StatusMenuReadinessBaselineTests`
- `make format`
- `make check`
- `git diff --check`
- `git diff --cached --check`

### DIFF
--- a/Sources/CodexBar/MenuSessionCoordinator.swift
+++ b/Sources/CodexBar/MenuSessionCoordinator.swift
@@ -51,7 +51,12 @@ struct MenuSessionCoordinator<MenuID: Hashable> {
 
     func hasRequiredClosedPreparation(for menuIDs: some Sequence<MenuID>) -> Bool {
         guard self.latestRequiredRebuildVersion > 0 else { return false }
-        return menuIDs.contains { self.isRenderedVersion($0, olderThan: self.latestRequiredRebuildVersion) }
+        return menuIDs.contains { menuID in
+            guard let renderedVersion = self.renderedVersions[menuID] else {
+                return self.latestRequiredRebuildVersion == self.contentVersion
+            }
+            return renderedVersion < self.latestRequiredRebuildVersion
+        }
     }
 
     func closedPreparationPlan(for menuIDs: some Sequence<MenuID>) -> ClosedPreparationPlan {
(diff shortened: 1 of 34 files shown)


davezfr:
Closing this as superseded by #1638, which has already landed the read-only reset-credit display. If there are small hardening follow-ups, I’ll open them separately against current main.

PR #1663, opened 2026-06-19
test: stabilize Kiro inherited-pipe fixture

steipete:
## Summary

- replace the race-prone background `os.setsid()` fixture with `subprocess.Popen(start_new_session: true)`
- use a pipe handshake so the detached helper installs signal handlers and writes its PID before usage output is emitted
- keep the original regression contract: helper inherits output pipes, ignores termination, and must be cleaned up

## Why

The same Kiro test failed on two unrelated hosted x86_64 shards because the shell background job could already be a process-group leader, making `os.setsid()` exit with status 1. The fixture also used a one-second polling window. Production Kiro code was not implicated.

The first hosted repair run proved the status-1 race fixed, then exposed the old 5-second wall-clock assertion at 5.509 seconds. The final head allows 8 seconds, still below the runner's 10-second idle timeout and well below the 20-second command timeout or 30-second helper lifetime.

Failure proof:

- https://github.com/steipete/CodexBar/actions/runs/27840523813/job/82399169433
- https://github.com/steipete/CodexBar/actions/runs/27845424633/job/82413492650
- https://github.com/steipete/CodexBar/actions/runs/27848409400/job/82422430521

## Proof

- exact focused regression: 20 consecutive passes
- full `KiroStatusProbeTests`: 40/40
- `make check`
- autoreview: initial readiness finding fixed; final reviews clean, 0.86/0.87

### DIFF
--- a/Tests/CodexBarTests/KiroStatusProbeTests.swift
+++ b/Tests/CodexBarTests/KiroStatusProbeTests.swift
@@ -316,7 +316,7 @@ struct KiroStatusProbeTests {
 
         #expect(snapshot.planName == "KIRO FREE")
         #expect(snapshot.creditsUsed == 12.50)
-        #expect(elapsed < 5, "Kiro usage capture should not wait for inherited pipe EOF, took \(elapsed)s")
+        #expect(elapsed < 8, "Kiro usage capture should return before the 10s idle timeout, took \(elapsed)s")
 
         let childPIDText = try String(contentsOf: childPIDFile, encoding: .utf8)
         let childPID = try #require(pid_t(childPIDText.trimmingCharacters(in: .whitespacesAndNewlines)))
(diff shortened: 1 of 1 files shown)


steipete:
Exact-head proof for `3338146213b092b219a5e3afae8b379b81564a10`:

- CI run `27849344900`: lint, Linux x64/arm64, all four macOS shards, aggregate, and security passed.
- Intel macOS shard 2 executed the repaired Kiro inherited-pipe fixture successfully.
- Local: the focused regression passed 20 consecutive runs; the full Kiro suite passed 40/40; `make check` passed.
- Autoreview: clean after the readiness-handshake repair and hosted-latency bound adjustment (0.87 confidence).
- Cross-architecture check: the full x86_64 XCTest bundle compiled locally; this Apple Silicon host cannot load an x86-only test bundle, so hosted Intel CI is the execution proof.

Production code is unchanged; this stabilizes the fixture and preserves the timeout behavior under test.

PR #1700, opened 2026-06-21
fix: keep menu bar open on refresh

### LINKED ISSUE #1678: Manual Refresh closes the menu and loses in-flight feedback (regression from #1661, reverts #1458)
### Summary

After #1661 (shipped in 0.37.0), clicking the persistent **Refresh** row (⌘R) now closes the whole menu. Before that, #1458 kept the menu open and spun the row in place so you could watch the refresh land. So we're back to the "menu disappears, no in-flight feedback" behavior that #1458 was meant to fix.

### Steps to reproduce

1. Open the menu bar popup for any provider.
2. Click **Refresh** (or press ⌘R).
3. The menu closes right away.

**Expected:** the menu stays open and the Refresh row shows that a refresh is running (spinner), then settles into the updated card — like #1458 did.

**Actual:** the menu is dismissed on click, so you never see the refresh happen. You have to reopen the menu to check the result, which is the repeat-click loop #1458 was trying to avoid.

### Root cause

#1661 (`fix: align menu items and convert action rows to native NSMenuItems`) switched the persistent action rows from the custom view-backed `PersistentMenuActionItemView` back to plain `NSMenuItem`s.

A native `NSMenuItem` with a target/action dismisses the menu on click — that's just how AppKit handles it. The old custom view sidestepped that: it caught the click and called `refreshMenuProviderNow(in:)` without ending menu tracking, which is what kept the menu open.

Where this lives today:

- `Sources/CodexBar/StatusItemController+Menu.swift:825-830` — Refresh is now a plain `NSMenuItem`.
- `Sources/CodexBar/StatusItemController+Actions.swift:92-93` — `@objc func refreshMenuItem(_:)` is the native action, and firing it lets AppKit close the menu.
- `Sources/CodexBar/StatusItemController+PersistentMenuActions.swift:5-12` — the in-flight feedback is now just `item.isEnabled = !inFlight`, which you can't really see since the menu is already gone.

The `MenuCardRefreshMonitor` / `shouldShowRefreshingMenuCardIndicator` "Refreshing…" subtitle from #1458 is still in the tree, but for manual refresh it never gets a chance to render now.

### A bit of history

#1458 made the Refresh row view-backed on purpose, so it could spin in place and keep the menu open while still respecting the deferred parent-rebuild policy from #1001/#1196/#1325. #1661 was a structural cleanup (go native / align the items), and I think the refresh interaction got reverted as a side effect of that — I don't see the trade-off discussed in the PR, so it may just have slipped through.

### Question

Is the menu closing on refresh intended, or an accepted by-product of the native conversion? If keeping it open with in-flight feedback is still wanted, the Refresh item probably needs to stay special on top of the new native layout (keep it view-backed, or stop the item from ending tracking) rather than being a plain `NSMenuItem`. I'm glad to send a PR that restores it against the new structure if that's the direction you'd prefer.

elijahfriedman:
Fixes #1678 because of #1661

## Summary

- convert the **Refresh** menu item from a native `NSMenuItem` to a custom `PersistentMenuActionRowView` so clicking it no longer closes the menu — refresh happens in place while the menu stays open
- route the row's click through `performPersistentRefreshAction(in:)` keyed by the menu's `ObjectIdentifier`, so it refreshes only the scoped provider (or the overview) just like the old action did
- identify persistent refresh rows by `representedObject == persistentRefreshMenuItemID` instead of by `#selector`, since the custom row has no action selector — used in reconciliation, in-place enable/disable, and tracking
- keep the **⌘R** key equivalent working via `performKeyEquivalent`; only the visible row changed, not the shortcut
- leave the other persistent action items (Update, Settings, About, Quit) as native `NSMenuItem`s that close the menu as before
- place the refresh row above Settings, matching the prior ordering
- give the persistent refresh row a stable, content-based height-cache fingerprint (`persistentRefreshAction:<icon>|<title>`) instead of the version-scoped default, so its measured height survives `invalidateMenus` rather than being pruned by `pruneVersionScopedMenuCardHeightCache`

## Tests

- update `StatusMenuPersistentRefreshTests` to assert the refresh row is now a custom view (no action/target, `isPersistentRefreshItem` true) and sits above Settings
- assert remaining persistent action items stay native with actions and the install-update icon
- drive refresh via `performPersistentRefreshAction(in:)` in scoped/overview/⌘R regressions instead of `perform(action:)`
- verify enabled-state tracking still flips correctly and ignores rows whose `representedObject` no longer marks them as refresh
-   - update `StatusMenuMergedOverviewRefreshTests` to drive refresh via `performPersistentRefreshAction(in:)` and assert `isPersistentRefreshItem` instead of the item's `action`
  - update `StatusMenuTests` ("overview tab omits contextual provider actions") to assert the refresh row is a persistent item with no `keyEquivalent` (⌘R now lives at the menu level, not on the item)
  - `StatusMenuHeightCacheTests` ("menu card height cache is reused for stable card content") now passes with the stable fingerprint — the refresh row's height entry persists across `invalidateMenus`

## Verification

- built and ran locally; clicking Refresh keeps the menu open and refreshes in place
- ⌘R still refreshes the scoped provider without closing the menu
- Settings/Quit/About/Update still close the menu as before
- full `make test` suite green across all shards

## Screenshots
Before:
Screenshot 2026-06-21 at 1 30 00 PM
After:
Screenshot 2026-06-21 at 1 31 15 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 - Cost history: remove the redundant tooltip from submenu-backed Cost rows. Thanks @Zihao-Qi!
+- Menu refresh: keep the menu open and show in-place progress when Refresh is clicked. Thanks @elijahfriedman!
 - Menu: align provider usage-card spacing with the Overview layout. Thanks @Zihao-Qi!
 - Memory pressure: avoid actor-isolation crashes when system callbacks arrive on a utility queue. Thanks @Zihao-Qi!
 - Menu: remove extra separators and spacing around Storage, Cost, and Subscription Utilization rows. Thanks @elijahfriedman!
--- a/Sources/CodexBar/MenuContent.swift
+++ b/Sources/CodexBar/MenuContent.swift
@@ -162,6 +162,32 @@ struct MenuActions {
     let copyError: (String) -> Void
 }
 
+struct PersistentMenuActionRowView: View {
+    let title: String
+    let systemImageName: String?
+    @Environment(\.menuItemHighlighted) private var isHighlighted
+
+    var body: some View {
+        HStack(spacing: 8) {
+            if let systemImageName {
+                Image(systemName: systemImageName)
+                    .imageScale(.medium)
+                    .frame(width: 18, alignment: .center)
+                    .foregroundStyle(MenuHighlightStyle.secondary(self.isHighlighted))
+            }
+            Text(self.title)
+                .font(.body)
+                .foregroundStyle(MenuHighlightStyle.primary(self.isHighlighted))
+            Spacer(minLength: 0)
+        }
+        .padding(.horizontal, 18)
+        .padding(.vertical, 4)
+        .frame(maxWidth: .infinity, alignment: .leading)
+        .accessibilityElement(children: .combine)
+        .accessibilityLabel(self.title)
+    }
+}
+
 @MainActor
 struct StatusIconView: View {
     @Bindable var store: UsageStore
--- a/Sources/CodexBar/StatusItemController+Menu.swift
+++ b/Sources/CodexBar/StatusItemController+Menu.swift
@@ -10,6 +10,7 @@ extension StatusItemController {
     static let menuCardBaseWidth: CGFloat = 310
     private static let maxOverviewProviders = SettingsStore.mergedOverviewProviderLimit
     static let overviewRowIdentifierPrefix = "overviewRow-"
+    static let persistentRefreshMenuItemID = "persistentRefreshAction"
     private static let defaultMenuOpenRefreshDelay: Duration = .seconds(1.2)
     #if DEBUG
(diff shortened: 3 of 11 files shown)


steipete:
Maintainer verification on exact head `37afad562e687e8927f3b6f3296b88d7a26c09d0`:

- `swift test --filter MenuCardViewRecyclingTests` — 20/20 passed
- `swift test --filter StatusMenuPersistentRefreshTests` — 27/27 passed
- `make check` — passed (localization, scripts, SwiftFormat, strict SwiftLint)
- autoreview — clean after repairing the recycled-view accessibility role
- fresh bundle — built, launched, remained running, and rendered the persistent Refresh row without clipping
- exact-head CI — all required checks passed

The repair derives the accessibility button role from the current click action, so a recycled non-clickable row cannot retain stale button semantics. Ready to land.

PR #1719, opened 2026-06-23
Fix: "Today" cost shows current local day, not the latest historical bucket (#1705)

### LINKED ISSUE #1705: CodexBar Bug Report: "Today" Cost Shows Last Historical Usage Day
## Summary

CodexBar has a usage-accuracy bug where UI surfaces labeled **Today** can show the newest non-empty historical cost bucket instead of the actual current local day. If a user has no usage today, and also no usage yesterday, the app may keep showing the last day with usage as "Today" until a newer bucket appears.

Example failure:

- Current local date: `2026-05-18`
- Latest cost-history row: `2026-05-15`, `$1.50`, `300 tokens`
- Expected Today display: `$0.00 · 0 tokens`
- Actual buggy display: `$1.50 · 300 tokens`

This affects trust in cost/usage monitoring because the app appears to report current-day activity where none occurred.

## User-Visible Symptoms

When there is a gap after the latest recorded usage day, the menu card and inline dashboard can keep reporting that old usage as today's usage.

Likely affected displays:

- Token-cost menu card line: `Today: $X.XX · N tokens`
- Inline cost dashboard KPI titled `Today`
- OpenAI Admin API usage notes/submenus titled `Today`
- Claude Admin API usage notes/submenus titled `Today`
- Any future provider that reuses the same "latest bucket" helper under a "Today" label

The cost-history chart itself should still preserve historical rows. The bug is specifically about using historical rows as current-day summary values.

## Root Cause

Several code paths use "latest available daily bucket" semantics while the UI text says "Today".

That conflates two different concepts:

- **Latest historical bucket**: the newest day in the data set, useful for "Latest" labels and historical charts.
- **Current local day bucket**: entries whose date maps to the user's current local day, required for "Today" labels.

When the data set has no row for the current local day, selecting the latest row is incorrect for "Today". The result is stale historical usage presented as current usage.

The critical local-cost path was in `CostUsageFetcher.tokenSnapshot(...)`. The old behavior selected the newest parseable day:

```swift
// Old behavior, simplified:
let currentDay = daily.data
    .compactMap { parse(entry.date) }
    .max { lhs.date < rhs.date }?
    .entry

sessionTokens = currentDay?.totalTokens
sessionCostUSD = currentDay?.costUSD
```

This should not be used for fields named `sessionTokens` / `sessionCostUSD` when those fields feed UI text labeled "Today".

## Correct Semantics

For displays titled **Today**:

- Match against the actual current local day.
- If the data set has historical rows but none for today, report zero.
- Preserve the historical totals, such as Last 30 days, separately.

For displays titled **Latest**:

- Continue using the newest valid historical bucket.
- This is appropriate for providers where current-day semantics are unavailable or not intended.

Recommended invariant:

> A UI label of "Today" must never be backed by a "latest non-empty historical bucket" helper.

### 1. Add explicit current-day helpers to cost snapshots

Add helpers on `CostUsageTokenSnapshot`:

```swift
public func currentDayEntry(calendar: Calendar = .current) -> CostUsageDailyReport.Entry? {
    Self.entry(in: self.daily, forLocalDayContaining: self.updatedAt, calendar: calendar)
}

public static func latestEntry(in entries: [CostUsageDailyReport.Entry]) -> CostUsageDailyReport.Entry? {
    // Existing latest-day tie-breaking behavior lives here.
}

public static func entry(
    in entries: [CostUsageDailyReport.Entry],
    forLocalDayContaining date: Date,
    calendar: Calendar = .current
) -> CostUsageDailyReport.Entry? {
    let todayKey = localDayKey(from: date, calendar: calendar)
    return entries.first { entry in
        let rawDate = entry.date.trimmingCharacters(in: .whitespacesAndNewlines)
        if rawDate == todayKey { return true }
        guard let parsed = CostUsageDateParser.parse(rawDate) else { return false }
        return localDayKey(from: parsed, calendar: calendar) == todayKey
    }
}
```

The important part is that date comparison should happen at the local calendar-day level, not by "newest timestamp".

### 2. Make `CostUsageFetcher.tokenSnapshot(...)` choose current day by default

Update `CostUsageFetcher.tokenSnapshot(...)` so the default path uses the current local day:

```swift
let sessionEntry = useCurrentLocalDayForSession
    ? CostUsageTokenSnapshot.entry(in: daily.data, forLocalDayContaining: now)
    : CostUsageTokenSnapshot.latestEntry(in: daily.data)

let hasHistoricalRows = !daily.data.isEmpty

return CostUsageTokenSnapshot(
    sessionTokens: sessionEntry?.totalTokens ?? (hasHistoricalRows ? 0 : nil),
    sessionCostUSD: sessionEntry?.costUSD ?? (hasHistoricalRows ? 0 : nil),
    last30DaysTokens: last30DaysTokens,
    last30DaysCostUSD: last30DaysCostUSD,
    historyDays: historyDays,
    daily: daily.data,
    updatedAt: now
)
```

This keeps two important cases distinct:

- Empty history: `nil` still means "no known data".
- Non-empty history but no current-day row: `0` means "known zero today".

### 3. Keep explicit latest-day semantics where the UI says "Latest"

Add a parameter like:

```swift
useCurrentLocalDayForSession: Bool = true
```

Then call with `false` only for providers/surfaces that intentionally show "Latest".

In the local patch, Bedrock keeps latest semantics because its dashboard text uses "Latest" rather than "Today".

### 4. Add current-day summaries for Admin API snapshots

OpenAI and Claude Admin API snapshots have rolling summaries and a `latestDay` summary. They also need a `currentDay` summary:

```swift
public var currentDay: Summary {
    self.summary(forLocalDayContaining: self.updatedAt)
}

public func summary(forLocalDayContaining date: Date, calendar: Calendar = .current) -> Summary {
    let dayKey = Self.localDayKey(from: date, calendar: calendar)
    let selected = self.daily.filter { bucket in
        if bucket.day == dayKey { return true }
        return Self.localDayKey(from: bucket.startTime, calendar: calendar) == dayKey
    }
    return Summary(
        costUSD: selected.reduce(0) { $0 + $1.costUSD },
        ...
    )
}
```

Then change UI sites labeled "Today" to use `currentDay`, not `latestDay`.

## Edge Cases to Cover

- History has no rows: Today should remain unavailable/unknown, not necessarily zero.
- History has rows but none for current local day: Today should be zero.
- History has a current local day row with zero cost/tokens: Today should be zero.
- History has current local day plus older rows: Today should use only current local day; Last 30 days should include all rows.
- Date strings may be plain `yyyy-MM-dd`, ISO timestamps, or legacy month/day strings; matching should normalize through a local day key.
- Invalid future calendar date strings, such as `2026-06-31`, should not win latest-day selection.
- Providers intentionally labeled `Latest` should keep newest-bucket behavior.

## Why This Fix Is Low Risk

The fix separates semantics instead of changing all historical aggregation:

- Last 7/30/90 day totals continue to sum over the history window.
- Cost-history charts continue to use the available daily data.
- Latest-bucket behavior remains available through `latestEntry` / `latestDay`.
- Only UI fields labeled "Today" move to current-local-day semantics.

The biggest behavior change is intentional: users with historical usage but no current-day usage now see `$0.00 · 0 tokens` under Today.

LeoLin990405:
Fixes #1705.

## Problem

UI surfaces labeled **Today** showed the newest non-empty *historical* cost bucket instead of the actual current local day. With a gap after the latest recorded usage day, the menu card, inline dashboard, and OpenAI/Claude Admin + Poe summaries kept reporting an old day's usage as "today".

## Fix

Invariant: **a "Today" label is always the current local day — `$0.00 · 0 tokens` when there is no usage today — never a latest-bucket helper.**

- **`CostUsageTokenSnapshot`**: add `localDayKey`, `entry(in:forLocalDayContaining:)`, `latestEntry(in:)`, `currentDayEntry()`.
- **`tokenSnapshot`**: session (Today) fields come from the current local day — today's row → its values; history but no today row → **zero**; no history → nil. A `sessionDay` mode keeps **Bedrock** on *Latest billing day* semantics (AWS Cost Explorer lags), and Mistral keeps its existing latest path.
- Every Today-labeled consumer updated to the current local day: `MenuDescriptor` (OpenAI/Claude Admin, Poe), `InlineUsageDashboardContent` (usage notes + dashboards + cost-history KPI), OpenAI `toCostUsageTokenSnapshot`. New `currentDay()` accessors on the OpenAI/Claude/Poe snapshots (zero when no bucket for today).
- The historical chart rows and "Last 30 days" totals are untouched.

## Tests

New `CostUsageTodayBucketTests` (drives `tokenSnapshot`: today-zero, today-populated, empty→nil, `latestEntry`, boundary); existing fixtures pinned to deterministic days. `swift build` + `swift test` green (pre-existing timing-flaky suites aside).

## Known limitation (follow-up)

The OpenAI/Claude Admin and Poe provider APIs return usage **pre-aggregated by UTC day**. `currentDay()` resolves the local day from each bucket's absolute `startTime` (Poe compares its `yyyy-MM-dd` string), so the multi-day "stale bucket as Today" bug is fixed, but a perfectly local "Today" around UTC/local midnight boundaries would require those fetchers to pull hourly granularity and re-aggregate by `Calendar.current` — out of scope here. The local cost-usage path (the issue's subject) is fully local-correct.


steipete:
Superseded by #1710, which now covers every Today surface using the user’s local calendar day and includes UTC-boundary regression coverage. Landed in 929d55aa. Thanks @LeoLin990405 for the contribution.

PR #1728, opened 2026-06-23
[codex] Add Sakana AI usage provider

LeoLin990405:
## Summary
- add Sakana AI as a manual web/cookie provider with `SAKANA_COOKIE` and config `cookieHeader` support
- fetch `https://console.sakana.ai/billing` and parse the billing page's plan label plus 5-hour and Weekly usage limit windows
- register Sakana across provider metadata, app settings, docs, icon resources, widgets, and unsupported cost/debug fallbacks

## OpenCLI notes
- inspected the Sakana AI console with opencli; the overview/billing data is rendered in the console HTML/Next payload rather than exposed by a clean documented JSON API
- the billing page renders the plan card and usage-limit rows with percent-used and reset text for 5-hour and Weekly windows
- this PR intentionally does not add browser-cookie auto-import; users must paste an explicit Cookie header/config value

## Validation
- `make check`
- `swift test --filter 'Sakana|ProviderConfigEnvironmentTests|ConfigurationDocsProviderIDTests|ProviderIconResourcesTests'`

## Risk
- draft because the implementation parses console HTML, so future Sakana UI markup changes may require parser updates.


LeoLin990405:
Addressed the current P2 in `0bad8fdd`:

- Kept Sakana usage percentages in `usedPercent` only and stopped putting them into `resetDescription`.
- Added a regression for an unparsed reset date to ensure the UI does not render a fake reset line like `Resets 92% used`.

Validation:
- `swift test --filter 'Sakana'`
- `make check`

Proof note: I checked local env/config presence without printing secret values. No Sakana cookie is available here, so I still cannot produce a real billing-page fetch proof from this machine.

@codex review
@clawsweeper re-review

LeoLin990405:
Addressed the live Sakana proof gap in `f23516ec`:

- The real Sakana billing page can omit a 5-hour reset line while still rendering the 5-hour percentage.
- Updated the parser so a window only requires the label plus the following `% used` value; `Resets on ...` is optional and still maps to `resetsAt` when present.
- Added a regression for a window without reset text so the UI still gets the 5-hour window and keeps `resetDescription` nil.

### Redacted live behavior proof

Ran a temporary local proof harness against the current PR head (`f23516ec`) using the Sakana provider fetcher with a Cookie header imported from the local Chrome profile. The harness was removed after the run; no proof-only code is part of this PR.

No cookie values, account identifiers, invoice URLs, raw HTML, raw response bodies, IP addresses, API keys, or private endpoints are included below.

```json
{
  "provider": "sakana",
  "fetchSource": "web",
  "billingURL": "https://console.sakana.ai/billing",
  "cookieSource": "Chrome profile",
  "cookieHeaderPresent": true,
  "cookieCount": 3,
  "planNamePresent": true,
  "priceLabelPresent": true,
  "fiveHourRendered": true,
  "fiveHourWindowMinutes": 300,
  "fiveHourUsedPercent": 0,
  "fiveHourResetPresent": false,
  "fiveHourResetDescriptionNil": true,
  "weeklyRendered": true,
  "weeklyWindowMinutes": 10080,
  "weeklyUsedPercent": 32,
  "weeklyResetPresent": true,
  "weeklyResetDescriptionNil": true,
  "identityLoginMethodPresent": true,
  "timestamp": "2026-06-24T12:56:24Z"
}
```

Validation:
- `swift test --filter SakanaUsageFetcherTests`
- `swift test --filter Sakana`
- `make check`

Maintainer sign-off still requested: this provider depends on Sakana console HTML/Next markup rather than a documented API, and uses a manual Cookie header path.

@codex review
@clawsweeper re-review

LeoLin990405:
Addressed the two new P2s in `84a1c067`:

- Linux CLI/manual Cookie: `sourceModeRequiresWebSupport` now allows Sakana `auto`/`web` on non-macOS when `SAKANA_COOKIE` or config-derived `cookieHeader` resolves to a valid Cookie header, so the request reaches `SakanaWebFetchStrategy` instead of returning the generic web-unsupported error.
- Quota parsing: Sakana window parsing now first bounds extraction to the current quota block before reading `% used` / reset text. If a window label is present but its own `% used` paragraph is missing, that window stays `nil` instead of borrowing the next window's percentage/reset.

Validation:

- `swift test --filter SakanaUsageFetcherTests`
- `swift test --filter CLIEntryTests/test_sourceModeRequiresWebSupportIsProviderAware`
- `make check`

@codex review
@clawsweeper re-review

steipete:
Superseded by merged integration PR #1774. Your original commits and implementation were preserved there, and the squash commit credits you as co-author. Thank you for adding Sakana support.

PR #1737, opened 2026-06-24
fix: distinguish sub-one usage percentages

### LINKED ISSUE #1734: [MiMo] Usage percentage truncated to 0% for values under 1%
## Bug Description

MiMo token plan usage percentage is displayed as **0% used** in the menu bar when the actual usage is between 0% and 1% (e.g., 0.96%). The Xiaomi MiMo console correctly shows **1%** (rounded up), but CodexBar truncates to integer.

## Root Cause

In UsageFormatter.usageLine(), the format string "%.0f%% %@" truncates the percentage to zero decimal places. For MiMo's 38B token plan, 0.96% usage (363M / 38B) is displayed as % used instead of 1% used or .96% used.

`swift
// Current code in UsageFormatter.swift
return String(format: "%.0f%% %@", clamped, suffix)
`

## Expected Behavior

- Values under 1% should show at least 1 significant digit
- The Xiaomi console rounds 0.96% to 1%, which is the expected UX

## Suggested Fix

Option A: Round to nearest integer:
`swift
let rounded = clamped.rounded()
return String(format: "%d%% %@", Int(rounded), suffix)
`

Option B: Show one decimal place when under 10%:
`swift
let format = clamped < 10 ? "%.1f%%" : "%.0f%%"
return String(format: "\(format) %@", clamped, suffix)
`

## Environment

- Provider: Xiaomi MiMo
- Cookie source: Manual
- Token plan: Pro (38B tokens/month)
- Actual usage: ~0.96%, displayed as 0%

steipete:
## Summary

- add one shared percentage-label rule for menu, menu-card, and menu-bar surfaces
- keep true zero as `0%` while rendering every positive value below one percent as `<1%`
- preserve localized percent placement for provider-cost cards and cover MiMo plus shared display paths

Fixes #1734.

## Root cause

The affected surfaces independently rounded percentages to zero decimal places. Values below the rounding threshold therefore became indistinguishable from true zero. The reported `0.96%` sample would normally round to `1%`; the verified shared defect is the broader positive sub-one range, reproduced at `0.49%`.

## Validation

- focused Swift tests: 109 tests across `UsageFormatterTests`, `StatusItemAnimationTests`, `MiMoProviderTests`, and `CursorMenuCardModelTests`
- deterministic full matrix: all 511 non-account-scoped selections passed in 43 shards
- account-scoped refresh suite: all 76 cases passed as individual selections, avoiding the existing aggregate harness deadlock without omitting coverage
- `make check`: package/script/docs guards, SwiftFormat, and strict SwiftLint passed with zero violations
- autoreview: clean after preserving localized provider-cost formatting

## Packaged-app proof

Ad-hoc debug package built from `1196f9e1ed0b7ea5a67f0640975d914a79eceef7`; bundle commit metadata matched and `codesign --verify --deep --strict` passed. App executable SHA-256: `1d20295dacb6d9624f4ae2391fe6d4b85501969c32cda432fa55763b1d2c19eb`.

Using a controlled local CLI fixture through the packaged app:

- `0.49%` used rendered the status item as `99.51 · <1%` and the real menu card as `<1% used`
- paired true-zero input rendered the status item as `100 · 0%` and the same menu card as `0% used`
- both menu cards were rendered at the actual 310 x 162 card size; temporary preferences and fixture state were restored after proof

No live account data or credentials were used or recorded.

## Safety

Public Model Identifier Gate: PASS. The exact diff introduces no model identifiers, the packaged binary's existing model catalog is unchanged, and tests, screenshots, logs, and this public proof contain no private account data, cookie values, credentials, identifying profile paths, or non-public model identifiers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - z.ai: support saved token-account team usage with account-scoped organization and project metadata. Thanks @zqbake!
 
 ### Fixed
+- Usage percentages: distinguish positive usage below 1% from true zero across menu and menu-bar labels. Thanks @Max0633!
 - Mistral: restore Vibe monthly-plan usage by forwarding only required console session cookies. Thanks @lfmundim!
 
 ## 0.37.3 — 2026-06-23
--- a/Sources/CodexBar/MenuBarDisplayText.swift
+++ b/Sources/CodexBar/MenuBarDisplayText.swift
@@ -5,8 +5,7 @@ enum MenuBarDisplayText {
     static func percentText(window: RateWindow?, showUsed: Bool) -> String? {
         guard let window else { return nil }
         let percent = showUsed ? window.usedPercent : window.remainingPercent
-        let clamped = min(100, max(0, percent))
-        return String(format: "%.0f%%", clamped)
+        return UsageFormatter.percentLabel(percent)
     }
 
     static func paceText(pace: UsagePace?) -> String? {
--- a/Sources/CodexBar/MenuCardView+Costs.swift
+++ b/Sources/CodexBar/MenuCardView+Costs.swift
@@ -304,8 +304,21 @@ extension UsageMenuCardView.Model {
             title: title,
             percentUsed: percentUsed,
             spendLine: "\(periodLabel): \(used) / \(limit)",
-            percentLine: String(format: L("%.0f%% used"), min(100, max(0, percentUsed))),
+            percentLine: self.localizedUsedPercentLine(percentUsed),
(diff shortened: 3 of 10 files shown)


steipete:
@codex review

steipete:
@clawsweeper review

steipete:
Superseded by #1759, landed in `c09a3d4a`. That PR shipped the shared sub-percent formatter, focused regression coverage, full `make check` / 43-shard `make test` proof, and changelog credit for #1734.

Closing this older conflicting implementation in favor of the landed fix.

PR #1746, opened 2026-06-25
Check local links across docs pages

kiranmagic7:
## Summary

- expand the documentation link checker to scan local markdown/image/html links inside docs pages, not only README and the provider index
- resolve docs-page links relative to the file that contains them, so nested links such as `../claude.md` are validated correctly
- include the source file in future missing-target or missing-anchor errors

## Why

`Scripts/check-documentation-links.mjs` already runs in `lint-linux`, but it only covered README links and the provider detail list. The docs tree has additional relative links between pages, including nested docs under `docs/refactor` and `docs/solutions`, that were not part of the guard.

Current main has no broken docs-page links; this change makes the existing lint lane catch future drift.

## Validation

- `node --check Scripts/check-documentation-links.mjs`
- `node Scripts/check-documentation-links.mjs` -> `documentation links OK: 134 local links`
- `git diff --check`
- `./Scripts/lint.sh lint-linux`

## Compatibility

No app behavior changes. The only risk is that future docs-only changes may fail lint when a local docs-page link points outside `docs/`, has a missing target, or references a missing markdown heading anchor.

Review focus: whether docs-page links should continue to be constrained to targets inside `docs/`.

### DIFF
--- a/Scripts/check-documentation-links.mjs
+++ b/Scripts/check-documentation-links.mjs
@@ -4,6 +4,12 @@ import path from "node:path";
 import { fileURLToPath } from "node:url";
 
 const repoRoot = path.resolve(path.dirname(fileURLToPath(import.meta.url)), "..");
+const approvedRootDocumentation = new Set([
+  "README.md",
+  "CHANGELOG.md",
+  "LICENSE",
+  "VISION.md",
+].map((relativePath) => path.join(repoRoot, relativePath)));
 
 const readme = readText("README.md");
 const readmeLinks = [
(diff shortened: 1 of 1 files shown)


steipete:
@clawsweeper re-review

Exact head: `d0e327f720646bd2376b15a7d3e0c733d7b40033`.

Synced current main and resolved the policy question: recursive links may target `docs/**` plus the public root docs `README.md`, `CHANGELOG.md`, `LICENSE`, and `VISION.md`; every other repository escape remains rejected. The real 134-link scan, `lint-linux`, `make check`, the full 43-shard suite, and local autoreview pass. A temporary fixture also proved an approved root link succeeds while `../Package.swift` fails.

steipete:
@clawsweeper re-review

Exact head: `8b6e02273e66ad53bd9a9897d42bc17d6fe24cb2`.

Fragment-only Markdown and HTML links now resolve against their containing Markdown file before anchor validation. A temporary fixture proved a valid same-page anchor passes and a missing one fails with a source-labelled diagnostic; the real 134-link scan, `lint-linux`, `make check`, the full 43-shard suite, and local plus whole-branch autoreview pass.

PR #1748, opened 2026-06-25
serve: bound /usage per provider so one slow provider can't stall the response

enieuwy:
## Problem

`codexbar serve`'s `/usage` handler (`serveUsage`) collected providers in a **sequential loop with no per-provider timeout**. A single slow or hung provider (e.g. a CLI/web fetch that never returns) blocked the entire handler.

The only backstop was the outer request deadline (`serveResponseWithDeadline`, default 30s), which returns a **504 with an empty body and discards every provider already collected**. Worse, the last-known-good merge (`CLIServeResponseCache.mergeLastGoodUsageItems`) only runs when `response.status == .ok`, so the 504 path produces **zero data** — no fresh providers, no cached fallback.

Net effect: one stuck provider makes the whole endpoint return nothing, which pushes shell/Zellij consumers (e.g. showy-quota) onto degraded CLI fallback. Observed live: a wedged serve held port 8080 with `/health` returning `ok` while `/usage` timed out for the full 30s.

## Fix

- Collect providers **concurrently**, bounding each with `BoundedTaskJoin` at a budget **strictly below** the outer request deadline (`serveProviderTimeout(requestTimeout:)` = `requestTimeout * 0.8`, or 25s when the deadline is disabled/non-finite).
- A provider that exceeds its budget contributes its **own provider error row** instead of blocking the others, so the response still renders **every healthy provider** with fresh data.
- Each provider's timeout clock starts **when its task is spawned**, so a hung provider can't serialize the others' deadlines.
- Results merge in **caller-provided provider order** regardless of completion order.
- The serve usage context's `webTimeout` is aligned to the per-provider budget (was a fixed 60s that exceeded the 30s request deadline).
- The outer request deadline is retained as a last-resort safety net.

### Last-known-good behavior (scope clarification)

Per-account error rows that carry a cache key are still merged with last-known-good by `CLIServeResponseCache` exactly as before. A **timeout** row is account-agnostic (no `cacheAccountKey`) and is intentionally **not** reconstructed from last-good — this matches the existing cache rule that "a timeout cannot prove which account is currently active" (`staleResponse` already refuses `usage:` keys, and tests assert unkeyed timeout rows are not reconstructed). This PR's win is that a hung provider no longer takes the healthy providers down with it; it does not claim to restore the hung provider's own last value.

## Real behavior proof

Built the debug CLI and ran the fixed `codexbar serve` on a spare port (8092, leaving the app's 8080 untouched). On this machine `claude` and `cursor` are genuinely wedged — the same providers that triggered the original incident.

**Normal `--request-timeout 30` (per-provider budget 24s):** `/usage` returns in **25.3s** with the healthy providers intact:

```
codex        DATA   source=oauth
claude       ERROR  msg="claude usage timed out" kind=provider
cursor       ERROR  msg="cursor usage timed out" kind=provider
antigravity  DATA   source=cli
```

Pre-fix, the claude/cursor hang would have produced an empty 504 and lost `codex` + `antigravity` too. Post-fix they render fresh.

**Small `--request-timeout 2` (per-provider budget 1.6s):** `/usage` returns in **1.70s** — proving the per-provider budget, not the outer deadline, is the cutoff:

```
codex        DATA   source=oauth
claude       ERROR  msg="claude usage timed out" kind=provider
cursor       ERROR  msg="cursor usage timed out" kind=provider
antigravity  ERROR  msg="antigravity usage timed out" kind=provider
```

(`elapsed` measured with `curl`; provider values/account labels redacted.)

## Tests

`CLIServeRouterTests`:
- `serveProviderTimeout` budget mapping, including the strict-below-deadline invariant for sub-second timeouts and the disabled-deadline case.
- A hung provider (30s simulated) with a 0.1s budget: collection returns in ~0.1s, healthy providers render in caller order, and the hung one yields a single provider error row that is account-agnostic (`cacheAccountKey == nil`), documenting the deliberate non-reconstruction.

Verified locally: `swift build` clean (StrictConcurrency), `swift test --filter CLIServeRouterTests` green, `make check` (SwiftFormat + SwiftLint) 0 violations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Fixed
 - Mistral: restore Vibe monthly-plan usage by forwarding only required console session cookies. Thanks @lfmundim!
+- CLI server: collect providers concurrently with a per-provider timeout when serving `/usage` so one slow or hung provider degrades to its own error row while the others still return fresh data, instead of stalling the whole response into an empty timeout. Thanks @enieuwy!
 
 ## 0.37.3 — 2026-06-23
 
--- a/Sources/CodexBarCLI/CLIServeCommand.swift
+++ b/Sources/CodexBarCLI/CLIServeCommand.swift
@@ -419,6 +419,14 @@ private enum CLIServeArgumentError: LocalizedError {
     }
 }
 
+private struct CLIServeProviderTimeoutError: LocalizedError {
+    let provider: UsageProvider
+
+    var errorDescription: String? {
+        "\(self.provider.rawValue) usage timed out"
+    }
+}
+
 extension CodexBarCLI {
     static let defaultServeRequestTimeout: TimeInterval = 30
 
@@ -561,7 +569,8 @@ extension CodexBarCLI {
                 await Self.serveUsage(
                     provider: provider,
                     config: snapshot.config,
-                    refreshInterval: runtime.refreshInterval)
+                    refreshInterval: runtime.refreshInterval,
+                    requestTimeout: runtime.requestTimeout)
             }
         case let .cost(provider):
             let snapshot: CLIServeConfigSnapshot
@@ -691,7 +700,8 @@ extension CodexBarCLI {
     private static func serveUsage(
         provider rawProvider: String?,
         config: CodexBarConfig,
-        refreshInterval: TimeInterval) async -> CLILocalHTTPResponse
+        refreshInterval: TimeInterval,
+        requestTimeout: TimeInterval) async -> CLILocalHTTPResponse
     {
         let selection: ProviderSelection
         do {
@@ -710,6 +720,12 @@ extension CodexBarCLI {
             return Self.serveError(status: .internalServerError, message: error.localizedDescription)
         }
 
+        // Bound each provider strictly below the outer request deadline so a
+        // single slow/hung provider degrades to its own error row while every
+        // other provider still returns fresh data, instead of stalling the whole
+        // response and being discarded by the deadline's empty 504.
+        let providerTimeout = Self.serveProviderTimeout(requestTimeout: requestTimeout)
+
         let browserDetection = BrowserDetection()
         let command = UsageCommandContext(
             format: .json,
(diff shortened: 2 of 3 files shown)


enieuwy:
Thanks both — addressed in `94a53fb3`.

**On the cache-key finding (Codex P2 / ClawSweeper P2):** correct, and the PR body over-claimed. I took the "explicitly narrow the behavior and tests" option rather than keying timeout rows. Reconstructing a timeout row from last-good would require re-resolving accounts (including Codex's `visibleCodexAccounts()`) *outside* the per-provider budget — reintroducing the blocking call this PR removes — and it contradicts the existing rule that "a timeout cannot prove which account is currently active" (`staleResponse` already refuses `usage:` keys, and the current tests assert unkeyed timeout rows are not reconstructed). So a timeout row is intentionally account-agnostic and not restored; per-account error rows that carry a cache key still merge with last-good exactly as before. The code comments, CHANGELOG, and PR body are corrected, and the test now asserts the timeout row has `cacheAccountKey == nil` / `account == nil`.

**On real behavior proof (P1):** added to the PR body. Ran the fixed `codexbar serve` on a spare port; on this machine `claude`/`cursor` are genuinely wedged. With `--request-timeout 30` (24s budget) `/usage` returns in 25.3s with `codex` + `antigravity` rendering real data and `claude`/`cursor` as timeout error rows — pre-fix that hang produced an empty 504 that lost the healthy providers too. With `--request-timeout 2` (1.6s budget) `/usage` returns in 1.70s, proving the per-provider budget is the cutoff.

@clawsweeper re-review

steipete:
@clawsweeper re-review

Exact head: `2c3454f1a2bb0143ca0488f919879f3e9d3bc226`.

Synced current main and preserved the documented `--request-timeout 0` contract: it adds no serve-level provider join bound, while finite deadlines still isolate hung providers, share the outer 24-hour cap, and reject non-finite input. The focused 41-test serve suite, `make check`, the full 43-shard suite, and local plus whole-branch autoreview pass.

PR #1753, opened 2026-06-25
Hide personal info without placeholder text

Zihao-Qi:
## Summary

- Removes the `"Hidden"` placeholder when `Hide personal information` is enabled, so redacted account/email fields render as empty instead.
- Omits account rows that would otherwise show only an empty redacted value.
- Makes LiteLLM team-budget redaction explicit by stripping team aliases before email redaction side effects can affect matching.
- Adds regression coverage for normal workspace aliases and email-shaped LiteLLM team aliases.

## Why

Users enabling `Hide personal information` expect private values to disappear, not be replaced with visible placeholder text. Showing `"Hidden"` still draws attention to where personal data exists and creates noisy UI.

The LiteLLM cleanup also makes the behavior easier to reason about and prevents future changes in email redaction formatting from accidentally changing team-budget display.

## Validation

- `swift test --filter LiteLLMMenuCardModelTests`
- `swift test --filter 'MenuCardModelTests|LiteLLMMenuCardModelTests|UsageStoreCoverageTests|CodexPresentationCharacterizationTests'`
- `make check`

## Screenshot

Before
Screenshot 2026-06-25 at 09 24 37

After
Screenshot 2026-06-25 at 09 24 19


steipete:
@clawsweeper re-review

Exact head: `a2ded707527c0af14deb9236f19cae737b433413`.

Synced current main and audited every `PersonalInfoRedactor` call site: the empty placeholder has no hidden consumer, account rows suppress empty redactions, and LiteLLM aliases remain removed before display. The focused 86-test menu/redaction suite, `make check`, the full 43-shard suite, and whole-branch autoreview pass; the contributor's before/after screenshots remain representative of the unchanged UI path.

PR #1754, opened 2026-06-25
Fix menu bar pace showing signed zero (+0%/-0%)

devYRPauli:
`MenuBarDisplayText.paceText` builds its label as `"\(sign)\(deltaValue)%"` where `deltaValue = Int(abs(pace.deltaPercent).rounded())`. When `|deltaPercent| < 0.5` the value rounds to `0`, but the sign is still prepended, so the menu bar shows `+0%` or `-0%` for a state that is effectively on pace. The expanded detail view already renders "On pace" for the same input, so the compact status item is inconsistent with it.

This returns `0%` when the rounded delta is zero. Non-zero values and their sign are unchanged.

Companion to #1737, which cleaned up sub-one-percent display for `percentText` / `usageLine` but did not touch `paceText`.

### Test

Added `MenuBarPaceTextTests`:
- sub-half-percent deltas (`+0.3`, `-0.3`) now render `0%`
- non-zero deltas keep their sign (`+3%`, `-3%`)

`swift test --filter MenuBarPaceTextTests` fails before the change (`+0%` / `-0%`) and passes after.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Codex: show enterprise monthly credit limits across OAuth, CLI, menu, and widget surfaces. Thanks @ChenZiHong-Gavin!
 - Codex: avoid launching monthly-credit CLI enrichment during usage-only OAuth refreshes.
 - Usage display: keep positive values below one percent visible instead of rounding them to zero. Thanks @Max0633!
+- Menu bar: show pace as `0%` instead of a signed `+0%`/`-0%` when the pace delta rounds to zero. Thanks @devYRPauli!
 - Menu: match the persistent Refresh row to native menu styling without replacing keyboard-navigable Settings, About, and Quit actions. Thanks @Zihao-Qi!
 - Claude: stop installed-version checks from invoking a login shell and triggering unwanted Keychain prompts. Thanks @enieuwy!
 - Localization: reject blank translated values and restore the affected Vietnamese provider prompts. Thanks @kiranmagic7!
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,6 @@
 - Codex: show enterprise monthly credit limits across OAuth, CLI, menu, and widget surfaces. Thanks @ChenZiHong-Gavin!
 - Codex: avoid launching monthly-credit CLI enrichment during usage-only OAuth refreshes.
 - Usage display: keep positive values below one percent visible instead of rounding them to zero. Thanks @Max0633!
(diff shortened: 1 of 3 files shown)


kiranmagic7:
CI note: the Linux x64 failure does not look related to this diff. It fails before compilation while SwiftPM is downloading Sparkle’s binary target:

```text
failed downloading https://github.com/sparkle-project/Sparkle/releases/download/2.9.3/Sparkle-for-Swift-Package-Manager.zip
NSURLErrorDomain Code=-1001
```

The Linux ARM64 build, lint, macOS Swift shards, aggregate `lint-build-test`, and GitGuardian checks all passed on head `30685e77`. A rerun of the Linux x64 job should be enough for the CI side; the remaining contributor-facing item is the requested runtime/menu-bar proof for the `0%` pace label.

devYRPauli:
Rebased onto current main (the `0.37.4` sub-percent work landed cleanly alongside this; `percentText` and `paceText` don't overlap).

Re: the runtime proof request - here is the actual `paceText` behavior. I ran the exact `main` logic and this PR's logic over a range of pace deltas and printed what the menu bar string resolves to:

```
deltaPercent  main     this PR
------------------------------
0.3           +0%      0%
-0.3          -0%      0%
0.0           +0%      0%
0.49          +0%      0%
-0.49         -0%      0%
3.0           +3%      +3%
-3.0          -3%      -3%
```

So on main any pace delta with `|deltaPercent| < 0.5` rounds the magnitude to 0 but the sign is still prepended, producing `+0%` / `-0%` in the menu bar. This PR returns a plain `0%` in that range and leaves every non-zero pace untouched.

The committed `MenuBarPaceTextTests` exercise the real `MenuBarDisplayText.paceText` and lock both halves (zero -> `0%`, non-zero keeps its sign). Local run on Swift 6.3.2:

```
Suite MenuBarPaceTextTests passed
Test "paceText drops the sign when the rounded delta is zero" passed
Test "paceText keeps the sign for non-zero deltas" passed
2 tests in 1 suite passed
```

The red Linux x64 check is unrelated - it fails while SwiftPM downloads Sparkle's binary target (NSURLErrorDomain -1001), before any compilation; a rerun should clear it.

steipete:
@clawsweeper re-review

Synced current main and removed the contributor-owned changelog entry. `swift test --filter MenuBarPaceTextTests`, `make check`, the full 43-shard suite, and local autoreview pass.

steipete:
@clawsweeper re-review

Exact head: `253e380553c216ea68bf12d51970d28dd129fd7d`.

Synced current main and removed the contributor-owned changelog entry. Focused pace, Kiro PTY, and Antigravity session tests pass; `make check`, the full 43-shard suite, and local autoreview also pass.

PR #1765, opened 2026-06-26
Stabilize refresh row alignment across provider menus

Zihao-Qi:
## Summary

- Tune the custom Refresh row metrics to better match native AppKit menu rows.
- Use a smaller semibold Refresh symbol to match neighboring menu icons.
- Change the Usage Dashboard icon to `chart.xyaxis.line`.

## Reasoning

Follow-up to [#1729](https://github.com/steipete/CodexBar/pull/1729).

The Refresh row stays custom because clicking it or pressing `⌘R` should refresh while keeping the menu open. A native key equivalent would let AppKit handle the shortcut and dismiss the menu.

The earlier fix used one fixed Refresh inset based on the menus measured then. Later testing showed there are two native layout cases: menus with the Usage Dashboard row, and menus like Overview/Antigravity that may not have it. AppKit can place native rows slightly differently between those cases, so changing only the Refresh padding is not enough.

The previous Dashboard icon was wide enough to affect AppKit’s native image/text columns. Switching it to the narrower `chart.xyaxis.line` avoids triggering that native row drift, while still fitting the dashboard action. With the native rows stabilized, the Refresh row metrics are tuned to the final visually tested values.

## Testing

- `swift test --filter StatusMenuPersistentRefreshTests` passed.
- `make test`
- `make check`

## Screenshots

Bef
Screenshot 2026-06-25 at 09 32 36
ore:

After:
Screenshot 2026-06-26 at 13 02 07


steipete:
@clawsweeper re-review

Exact head: `15a622827ac5c1cea9d4c80a45203bff7241ddff`.

Synced current main. The focused 31-test menu suite, `make check`, the full 43-shard suite, and local autoreview pass. A freshly packaged, isolated app also confirms the Dashboard, Refresh, and Settings icon/title columns align on screen.

PR #1770, opened 2026-06-28
Fix Codex archived usage dedupe

vincent-peng:
## Summary

- replace coarse Codex session-id dedupe with row-level usage keys that include `turnID`
- persist raw Codex usage rows so warm-cache scans can re-evaluate active/archive overlap when archived files appear later
- preserve full Codex cached-token counts instead of clamping cached reads to input-token deltas
- add regression coverage for active stubs, active/archive partial overlap, same-sized distinct turns, warm-cache archived overlap, and persisted cached-token rows

Refs #1768.

## Validation

- `swift test --filter CostUsageScannerBreakdownTests`
- `swift test --filter CostUsageFetcherTests`
- `make check`
- `git diff --check`
- synthetic CLI proof:
  - active-only A+B reported `inputTokens: 30`, `outputTokens: 6`, `totalTokens: 36`, `totalCost: 0.000195`
  - after adding archived duplicate row B, totals stayed `inputTokens: 30`, `outputTokens: 6`, `totalTokens: 36`, `totalCost: 0.000195`


vincent-peng:
@clawsweeper re-review

steipete:
Closing as superseded by #1768, which has now landed with the archived-session row-level dedupe, cache-read pricing correction, cache migration safeguards, and expanded warm/narrow-cache regression coverage. Thanks @vincent-peng for independently investigating the same root cause and proposing a fix.

PR #1794, opened 2026-06-30
Align refresh menu icon

micnem:
## Summary
- Align the custom persistent Refresh menu icon with native AppKit menu action rows.
- Update the metric regression test for the new icon sizing and alignment values.

## Root Cause
The persistent Refresh row uses a custom view so it can stay active while menu tracking is in progress. Its icon metrics used a smaller slot and symbol size than the surrounding native menu items, which made the refresh glyph appear visually offset.

## Impact
The Refresh command now lines up with Settings, About, and Quit in the status menu while keeping the persistent refresh row behavior intact.

## Validation
- `swift test --filter StatusMenuPersistentRefreshTests`
- `make check`
- `make test`


steipete:
Landed the refresh-icon alignment via #1804 in merge commit 735a91ac3c25a07b8917f4d88d4b6bcf4a6f8ae7.

The original branch could not be updated because maintainer edits were disabled, so I recreated the focused change on a maintainer branch. Your original commit authorship is preserved, and the changelog thanks @micnem.

Proof at exact head 3ed62ab6e1b921a850bbfd2e0ae8d6415a374043:

- all 10 GitHub checks terminal green, including all four macOS shards
- focused refresh-row tests, `make check`, and the full 44-shard `make test` suite passed
- packaged-app validation passed on macOS 26.5 for both standard and compact menus; the Refresh title/shortcut column stayed fixed while the symbol became larger and aligned with native action rows
- bundled autoreview returned no findings

Thank you for the fix, @micnem. Enabling **Allow edits by maintainers** on future PRs lets us apply these final validation/changelog updates directly to the contributor branch.

PR #1804, opened 2026-07-01
fix: align refresh menu icon

steipete:
## Summary

- expand the persistent Refresh row's icon frame and symbol to the native 16-point action size
- use regular symbol weight while preserving the existing title and shortcut columns
- credit @micnem in the changelog

This is the maintainer branch for #1794 because the contributor branch does not allow maintainer edits. Michael's implementation commit remains authored by Michael Nemni.

## Layout rationale

The title origin is unchanged: the previous `17 + 14 + 4.5` and new `15 + 16 + 4.5` both resolve to `35.5` points. The change only expands the symbol two points to the left and matches the 16-by-16 image frames used by the native Settings, About, and Quit rows.

## Proof

- `swift test --filter StatusMenuPersistentRefreshTests` — 31 tests passed
- `make check` — SwiftFormat clean; SwiftLint 0 violations
- `make test` — all 44 test shards passed
- bundled autoreview — no accepted/actionable findings
- packaged Debug app built from exact head `3ed62ab6`; bundle commit metadata matched and `codesign --verify --deep --strict` passed
- macOS 26.5 Parallels VM, independently captured from the host: standard Codex menu and narrow Antigravity menu both rendered the 16-point regular Refresh symbol with the title column unchanged and aligned to native action rows

Closes #1794 after merge.


steipete:
Landed in 735a91ac3c25a07b8917f4d88d4b6bcf4a6f8ae7 after the exact head `3ed62ab6e1b921a850bbfd2e0ae8d6415a374043` reached 10/10 terminal-green checks with zero queued or active rows.

Final proof: focused refresh-row coverage, `make check`, full 44-shard `make test`, clean bundled autoreview, strict packaged-app signature verification, and accepted macOS 26.5 packaged-app captures for both standard and compact menus. Contributor authorship is preserved and the changelog thanks @micnem.

Zihao-Qi:
Thanks for working on this. I made a local comparison to understand the sizing better:

1. current custom Refresh row from this PR
2. native `NSMenuItem` Refresh row
3. previous custom Refresh row

On my machine, the native row appears closer to the previous custom metrics than to the new 16pt symbol version. The new row keeps the title column aligned, but the visible Refresh glyph looks larger than the native row, and it also appears slightly farther left on my MacBook.

Could this difference be caused by a system setting? I could not find a setting that explains the difference. I attached the screenshot in case it helps compare the rendering.

Screenshot (order: current, native, previous)

Screenshot 2026-07-01 at 11 19 28

Issue #1282, opened 2026-06-03
Support storing UI preferences in config.json/preferences.json for dotfiles syncing


kiankyars:
Currently, CodexBar stores provider enablement/toggles and CLI tokens in `~/.codexbar/config.json`, which is great for syncing under version control (dotfiles).

However, general UI preferences and application settings (such as `refreshFrequency`, `launchAtLogin`, `mergedOverviewSelectedProviders`, and various display/visual settings) are stored in the macOS User Defaults binary plist (`~/Library/Preferences/com.steipete.codexbar.plist`). 

Because plist files are heavily cached by `cfprefsd` and contain dynamic/device-specific data (like window frames/coordinates and keychain tokens), they cannot be easily or safely tracked in dotfiles.

### Proposal
Could we support storing these application/UI preferences either:
1. Directly in `~/.codexbar/config.json` (e.g., under a `"preferences"` block), or
2. In a separate `~/.codexbar/preferences.json` file?

This would make the entire configuration of CodexBar declarative, transparent, and easy to version-control and sync across multiple Macs via dotfiles.

PR #1817, opened 2026-07-01
Prototype portable UI preferences contract

steipete:
## Summary

- prototype a sparse, versioned portable UI-preference schema and file store
- keep portable preferences in a separate `preferences.json` beside the resolved config, with a `CODEXBAR_PREFERENCES` override
- document explicit-key precedence plus portable, device-local, privacy, and security boundaries
- intentionally leave `SettingsStore` runtime wiring out until the product choice is approved

## Product decision

Recommended contract:

- separate preference file rather than the existing provider config
- opt in by creating the file; no automatic creation or migration
- explicit valid file key > local `UserDefaults` value > code default
- once enabled, UI changes dual-write the preference file and `UserDefaults`

The existing config can contain API keys, cookie headers, and other credentials. Putting UI preferences there would encourage users to sync or commit a secret-bearing file. The alternatives are a preference block inside that config (simpler, unsafe sharing affordance) or export/import only (safer, but not continuous sync).

## Scope

The prototype includes presentation and behavior preferences that can safely follow a user between Macs. It excludes credentials, account routing, login items, terminal paths, window state, discovery/debug state, privacy toggles, histories, notifications, and storage policy.

## Proof

- `swift test --filter CodexBarPortablePreferencesTests` (8 tests)
- `make check`
- `make test` (44 shards)
- `./Scripts/package_app.sh`
- autoreview clean after fixing sparse-version and invalid-language handling

No screenshot: this decision branch deliberately adds no live UI or runtime behavior.

Refs #1282

### DIFF
--- a/Sources/CodexBarCore/Config/CodexBarPortablePreferences.swift
+++ b/Sources/CodexBarCore/Config/CodexBarPortablePreferences.swift
@@ -1,0 +1,206 @@
+import Foundation
+
+public struct CodexBarPortablePreferences: Codable, Equatable, Sendable {
+    public static let currentVersion = 1
+
+    public var version: Int?
+    public var refreshFrequency: String?
+    public var usageBarsShowUsed: Bool?
+    public var resetTimesShowAbsolute: Bool?
+    public var providerChangelogLinksEnabled: Bool?
+    public var menuBarDisplayMode: String?
+    public var menuBarShowsBrandIconWithPercent: Bool?
+    public var menuBarHidesCritters: Bool?
+    public var menuBarShowsHighestUsage: Bool?
+    public var mergeIcons: Bool?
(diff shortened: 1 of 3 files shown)


steipete:
Closing this stale prototype rather than publishing its current file contract.

Dry-applying the three-file change to current `main` still passes `CodexBarPortablePreferencesTests` (8/8), but the hard-coded allowlists already reject current refresh/language values, synthesized decoding drops unknown fields, and no `SettingsStore` integration or file-watching path exists. That makes the documented forward-compatibility promise unsafe and leaves the feature incomplete.

#1282 remains open. The next implementation should start from current settings state, preserve unknown fields losslessly (or refuse unsafe writes), and include live import/export diagnostics.

PR #1789, opened 2026-06-30
[codex] Add Codex pace headroom hint

astuteprogrammer:
## Owner decision — product-tone hold

**Decision requested:** Should the menu proactively tell a user `Try 1.5x!` when the conservative headroom model says their remaining quota can support it, and should the minimum historical-week threshold be user-configurable in Advanced settings?

**Tradeoff:** The wording turns safe headroom into an immediate, useful action and the setting exposes a real confidence threshold. The copy can feel prescriptive or celebratory in a utility menu, while the setting adds another expert-facing control.

**Recommendation:** **Land as written** with the conservative gate and current copy for one release. Keep this PR unmerged only for the owner's tone/settings acceptance and exact-head CI; if the tone is declined, change the copy to `1.5x headroom` without changing the model.

## Summary

- Show `Try 1.5x!` when Codex usage has more than 15% reserve and remaining quota supports at least 1.5x the projected remaining burn.
- Add an Advanced setting for the minimum number of complete Codex weeks used to enable historical pace, defaulting to 3 and allowing 2-8.
- Keep Ollama cookie diagnostics out of this PR so the pace/settings decision stays independently reviewable.

## Behavior and tradeoff

The hint appears only when the quota is projected to last until reset, reserve exceeds 15%, and the safe remaining-burn multiplier is at least 1.5. Linear and workday pace use remaining capacity versus usage projected across the remaining window; historical pace uses remaining capacity versus the remaining expected historical curve.

The current `Try 1.5x!` wording is intentionally prominent and actionable, but it is a product-tone choice. Recommendation: keep the conservative gate and use the current copy for one release; if it feels too celebratory in the menu, change only the copy to `1.5x headroom` without changing the model.

## Maintainer preparation

- Split the unrelated Ollama access-denied change from this branch.
- Fixed late-window headroom math and added linear, historical, UI, and CLI regression coverage.
- Fixed Italian localization and repository formatting failures present on the contributor head.
- Packaged the debug app and verified the Advanced setting in the live bundle.
- Generated a redacted synthetic three-state card fixture for ahead/on-track/behind comparison. No real account usage is represented by that fixture.

## Validation

- `make check`
- `make test` (43 shards)
- focused pace, historical pace, CLI snapshot, localization, and Ollama regression suites
- `CODEXBAR_SIGNING=adhoc ./Scripts/package_app.sh debug`
- autoreview: clean, no accepted/actionable findings

[...]

Issue #1299, opened 2026-06-04
Feature request: predictive pace warning notifications


CodingCanuck:
CodexBar already computes and displays a pace projection for usage windows through UsagePace, including whether a window is expected to last until reset, ETA to run-out, and modeled run-out probability. Today the app's notifications are still mostly reactive: quota warnings fire when remaining percent crosses configured thresholds, and session quota notifications fire at 0% / restored.

I'd like to add an opt-out proactive notification for Codex and Claude windows when the existing pace projection says the current window is trending to run out before it resets. The goal is to warn while the user can still ease off or drop fast mode, instead of only notifying once a static remaining-percent threshold or 0% has already been crossed.

Proposed behavior:

- Evaluate session and weekly windows for Codex and Claude when a UsagePace is available.
- Fire when UsagePace.willLastToReset is false and an ETA exists.
- If the model supplies runOutProbability, require it to be at least 0.5 before notifying.
- Rate-limit by provider/window with a one-hour cooldown.
- Clear cooldown once the pace recovers so a later relapse can alert again.
- Add a Preferences toggle, defaulting on, alongside the existing notification controls.

I searched existing issues/PRs and found #617 / #852 for threshold notifications and related notification work, but did not find one specifically for proactive pace-projection warnings. Is this wanted upstream, and is the proposed behavior aligned with how you want CodexBar notifications to evolve?

steipete:
I reviewed this against the current pace model, notification paths, `VISION.md`, and active PR #1789. I kept implementation out of the overlapping UI/model lane and opened #1818 as a decision-ready spec.

Recommendation: default off; Codex/Claude session and weekly windows only; one-hour cooldown per provider/account/window; successful recovery clears cooldown; missing/incomplete data preserves state; in-memory state only. The explicit alternative is one alert per relapse with no hourly repeat.

Validation: `make check` passed and focused structured autoreview found no actionable issues. Keeping this issue open for product sign-off before implementation.

CodingCanuck:
My user product perspective, in case it helps: I've been working on a local fork with pace notifications added, and I find it helpful. I use them to manually rate-limit my use of fast mode, or parallel agent sessions, or to switch between Codex and Claude.

But it's the 5h window notifications that are especially helpful for professional use, since I hit the 5h limits semi-often but almost never hit the weekly limits at work (where most of my use is during the day on workdays).

[...]

PR #1818, opened 2026-07-01
docs: define predictive pace warning decision

steipete:
## Decision

Define a bounded, opt-in predictive pace warning before any implementation work. This PR intentionally contains only the reviewed decision spec; it does not change runtime behavior.

## Recommendation

- Default off.
- Codex and Claude only.
- Session and weekly windows only.
- One-hour cooldown per provider/account/window while risk persists.
- A successful recovery clears cooldown; missing/incomplete data does not.
- In-memory state only.
- Reuse existing notification delivery, sound, localization, and privacy redaction paths.

## Tradeoff

Hourly repetition keeps sustained risk visible but is noisier than one alert per relapse. The spec therefore makes the feature opt-in, fixes the cooldown rather than adding more settings, and offers one-alert-per-relapse as the explicit alternative for maintainer sign-off.

## Scope boundary

PR #1789 owns pace presentation, historical confidence settings, and localization. This proposal does not alter those surfaces and calls for a separate implementation only after approval.

## Validation

- `make check`
- Focused structured autoreview: clean; patch correct (0.89)

Refs #1299.


steipete:
Maintainer closeout on exact head `6e8a8301eb5960bc3ac52693130ef2cf02fcae5e`:

- Accepted the default-off Codex/Claude session-and-weekly predictive warning boundary.
- Chose one notification per provider/account/reset-window risk episode. Continuous risk never repeats merely because an hour passed.
- Re-arm requires a successful authoritative observation with `willLastToReset == true`; low-confidence, missing, incomplete, and failed refreshes preserve episode state rather than manufacturing recovery.
- New reset-window identities start independently and expired keys must be pruned. State remains memory-only.
- Proof: `make check` passed; exact-head CI is green (both Linux builds, lint, aggregate gate, GitGuardian; macOS correctly skipped for docs-only scope).

This lands the product/state-machine decision only; runtime implementation remains separate.

PR #1827, opened 2026-07-01
fix(sakana): parse weekly/5-hour reset time as UTC, not device-local

### LINKED ISSUE #1826: Sakana AI: weekly/5-hour quota reset time is off by the viewer's UTC offset
## Summary

`SakanaUsageFetcher` (added in #1774) parses the reset time shown on `console.sakana.ai/billing` ("Resets on <date> at <time>") as if that text were expressed in the device's local timezone (`TimeZone.current`). It isn't — the raw server-rendered HTML always renders that string in **UTC**, and the browser only corrects it to the viewer's local time client-side, after JS hydration. Since CodexBar fetches raw HTML (no JS execution), it always sees the UTC-rendered string, then mis-interprets it as local time — so every non-UTC user gets a systematically wrong reset countdown, off by exactly their UTC offset.

## Repro

Fetching `https://console.sakana.ai/billing` directly (`curl`, real cookie, no JS) on an account with an active weekly quota window returns:

```
Resets on July 6, 2026 at 12:00 AM
```

The same account's billing page, viewed in a real browser (IST, UTC+5:30) at the same moment, shows:

```
Resets on July 6, 2026 at 5:30 AM
```

Both are the same instant (`2026-07-06T00:00:00Z`) — July 6 2026 is a Monday, matching Sakana's own copy elsewhere on the page: *"Weekly usage resets every Monday at 00:00 UTC."* The raw HTML response also embeds `"timeZone":"UTC"` in its hydration payload, confirming the server always renders in UTC before the client corrects it.

`SakanaUsageFetcher.parseResetDate` parses the raw "12:00 AM" string using `TimeZone.current`. On an IST machine this produces `2026-07-05T18:30:00Z` — **5.5 hours before the true reset time**. On a US-Pacific machine it would instead land *after* the true reset time. Verified by building `main` (`05475115`) into the `codexbar` CLI and running `codexbar usage --provider sakana --format json` against a live account: the tool's own JSON output (`resetsAt`) matched this miscalculation exactly.

## Impact

Any Sakana user not physically in UTC gets a wrong "resets in Xh" countdown for both the 5-hour and weekly quota windows — early for positive UTC offsets, late for negative ones. The percent-used figures and plan label are unaffected (verified correct against a live account).

## Fix

`parseResetDate` (and the `parseWindow`/`parseBillingHTML` `timeZone:` parameter that feeds it) should parse the captured "Resets on …" text as UTC unconditionally, not as the device's local timezone. Opening a follow-up PR.

## Related files
- `Sources/CodexBarCore/Providers/Sakana/SakanaUsageFetcher.swift` (`parseResetDate`, `parseWindow`, `parseBillingHTML`)
- `Tests/CodexBarTests/SakanaUsageFetcherTests.swift` (existing fixtures assert the *current*, incorrect local-timezone interpretation and will need updating alongside the fix)

ss251:
Fixes #1826.

## Summary
`console.sakana.ai/billing` always server-renders "Resets on <date>" in UTC — confirmed by the raw HTML's embedded `"timeZone":"UTC"` hydration marker and Sakana's own "Weekly usage resets every Monday at 00:00 UTC" copy elsewhere on the page. The browser only corrects that string to the viewer's local timezone client-side, after JS hydration, which `SakanaUsageFetcher` (an HTML-only fetcher) never runs.

`parseResetDate` parsed the raw string using `TimeZone.current` instead of UTC, so every non-UTC user got a reset countdown off by exactly their UTC offset.

- Removes the `timeZone` parameter from `parseBillingHTML`/`parseWindow`/`parseResetDate` entirely — there's no legitimate caller-supplied value here, the source text is always UTC — and hardcodes UTC in the formatter.
- Updates the existing fixture tests, which previously asserted the *buggy* local-timezone interpretation (via a `shanghaiTimeZone` test parameter) as if it were correct.
- Adds a regression test pinned to a literal Unix timestamp, independent of the date-building test helper, so a future refactor of the helper itself can't silently reintroduce the bug.

## Verification
I built this branch (and, for comparison, unmodified `main`) into the `codexbar` CLI and ran both against a real Sakana account (Standard plan, active weekly window) with a live cookie, comparing to the same account's billing page rendered in a real browser at the same moment:

| | Browser (ground truth) | `main` (before) | This branch (after) |
|---|---|---|---|
| Weekly reset | July 6, 2026, 5:30 AM IST (`2026-07-06T00:00:00Z`) | `2026-07-05T18:30:00Z` (5.5h early) | `2026-07-06T00:00:00Z` ✓ |

Plan name, 5-hour %, and weekly % were already correct in both and are untouched by this change.

## Test plan
- [x] `swift build --target CodexBarCore` / `swift build --target CodexBar` — clean
- [x] `./Scripts/lint.sh format` — 0 files reformatted
- [x] `./Scripts/lint.sh lint` (swiftlint --strict) — 0 violations
- [x] `swift test --filter SakanaUsageFetcherTests` — 12/12 pass
- [x] Verified against a live Sakana account via the built `codexbar` CLI before and after the fix (table above)


ss251:
Addressed both P3 findings from clawsweeper's review:

- **Regression test now actually pins the fix.** The original test would have passed on a UTC CI runner even against the pre-fix `TimeZone.current` code (since `TimeZone.current == UTC` there). Now forces `NSTimeZone.default` to UTC+14 for the test's duration (same pattern as `BedrockUsageStatsTests`). Verified locally by temporarily reverting the fix and confirming the test fails with the expected 5.5h-off result, then restored it.
- **docs/sakana.md updated** — it still said reset dates are parsed using the device's local time zone (`TimeZone.current`), exactly the behavior this PR removes. Now describes the actual UTC parsing and links #1826.

12/12 Sakana tests pass, lint clean.

ss251:
@clawsweeper re-review

ss251:
Addressed the P2 finding from the re-review: marked `SakanaUsageFetcherTests` `@Suite(.serialized)`, matching the existing `BedrockUsageStatsTests` precedent for the same `NSTimeZone.default` mutation pattern. Confirmed serialized execution locally (tests now run strictly one-at-a-time instead of all starting concurrently). 12/12 pass, lint clean.

@clawsweeper re-review

steipete:
Landed as https://github.com/steipete/CodexBar/commit/8fb8c74259dd7561395cb1d2439f0a947faf78a0.

Proof:
- Exact contributor head `0834fa5382f7f315bcfd5644e3ac35a3e1b609ea` passed all 10 required checks.
- Focused Sakana suite passed 12/12 tests, including a raw UTC fixture verified under a non-UTC device timezone.
- `make check` passed with 0 SwiftFormat/SwiftLint violations; `make test` completed all 45 shards.
- Reviewed the parser boundary: server-rendered reset timestamps are now interpreted as UTC before conversion for display.

Thanks @ss251 for the strong reproduction and focused fix!

PR #1812, opened 2026-07-01
Document Claude multi-account architecture

steipete:
## Summary

- audit the current Claude OAuth, token-account, snapshot, and status-item architecture
- recommend an opt-in, read-only `claude-swap` v0.15 adapter for durable multi-subscription usage
- define a provider-neutral account snapshot and bounded per-account status-item behavior
- include a rendered UI decision mock and redacted packaged synthetic sign-in proof

This is a decision document only. It intentionally does not add credential storage, account switching, an external dependency, or product behavior.

## Recommendation

1. Approve display-only `cswap --list --json` integration as Phase 1.
2. Defer `--switch-to` until a separate explicit-action product/auth review.
3. Normalize account usage before status-item work.
4. Make selected account items opt-in, capped at four, and mutually exclusive with Merge Icons.

## Architecture and overlap evidence

- #1800 and #1776 are merged and improve single-active-Claude-account correctness; neither enumerates subscriptions.
- #1707 is closed and should not be revived because it mixed credential discovery, routing, menus, and prompt behavior.
- The current `ProviderTokenAccount` model has one token and no refresh/expiry fields.
- `StatusItemIdentity` and status-item storage remain provider-scoped.
- claude-swap v0.15.0 documents a credential-free, versioned `--list --json` output contract.

## Validation

- `make check`
- structured Codex autoreview: clean, no findings
- SVG XML validation and local render inspection
- packaged synthetic-account screenshot inspection; no real credential, browser session, or provider call

Refs #1756
Refs #1268
Follow-up to #1811

Issue #1756, opened 2026-06-25
Support multiple Claude accounts OR support account swapping with claude-swap


robd003:
For some reason CodexBar only appears to track a single Claude account.

It would be great if it could load Claude accounts using the default preferences file that the `claude` process reads from.

Being able to switch claude accounts with [claude-swap](https://github.com/realiti4/claude-swap) would be great.

jroth1111:
+1 — I use claude-swap daily (the `cswap` command, v0.14.0 via `uv tool`) to manage multiple Claude subscription accounts, and I'd love CodexBar to display usage across all of them.

To be specific about the ask: I'm looking for **multi-subscription account support via OAuth** — i.e., tracking several consumer Claude Pro/Max subscriptions side by side — **not** API keys or token accounts. CodexBar already covers the API-key / Admin-API / token-account paths; the actual gap is multiple OAuth subscription accounts shown together.

Worth noting why the existing tokenAccount path can't back this: `ProviderTokenAccount` (`Sources/CodexBarCore/TokenAccounts.swift`, current main) holds a single `token` string — no `refreshToken`, no `expiresAt`. Anthropic OAuth access tokens are short-lived (~1h), so an `sk-ant-oat` token placed in a tokenAccount expires with no way to refresh and silently goes stale within hours. Durable multi-subscription OAuth needs the account model extended (refresh-token + expiry), **or** CodexBar delegating to a switcher that already handles per-account OAuth refresh.

claude-swap already exposes what CodexBar would need, without CodexBar touching credentials:
- `cswap --list --json --token-status` — machine-readable list of all managed accounts, including OAuth token expiry state
- `cswap --status --json` — the currently active account
- `--strategy best` picks the account with the most remaining 5h/7d quota, so claude-swap already computes the per-account quota figures CodexBar would surface

A low-risk integration: CodexBar shells out to `cswap --list --json --token-status` to enumerate managed accounts and show their quota/usage in the menu — CodexBar never handles credential storage/switching, since claude-swap owns that. This sidesteps both the refresh problem above and the secret-storage boundary.

Related: #1715 (the in-app "add second Claude account" flow currently hangs on "waiting for browser" and never creates a stored account). Happy to share the JSON schema or test a build.

steipete:
Decision-ready architecture and UI proof is in #1812.

[...]

Issue #81, opened 2025-12-28
Multiple Accounts Support


steipete:
Currently we get data from oauth tokens or cookies (or clis). That will not work if we ever wanna support multiple accounts.

Just opening this ticket as a brainstorming session.

I'd like that too but it'd be messy. Login often is gated by anti-bot measures/captchas.
Supporting multiple accounts (one per browser) would work, but seems messy and limiting.

Don't have a good solution for this yet.

dalisoft:
Maybe looking at source code of https://github.com/router-for-me/CLIProxyAPI may help?

steipete:
@aavetis oh I don't like this either. Just starting a discussion.

deksden:
multiply accounts can be done via multiply profiles on chrome. 
I do have several profiles to switch my 5 team accounts. 

Maybe controlled browser can be used to scrape info from multiply accounts. Not sure how we can login via headless browser.

aavetis:
for the first time ever, i bought credits to top up my Codex usage. i don't think this will be sustainable. i'd like to bring another account in the mix hopefully soon!! 😁

t8or:
Just adding that I would love to have my multiple accounts shown here. Ive got work and personal claude cursor and chatgpt. Would be nice to see the status of all of them at once.

Cerise1:
> I created an app based off CodexBar that does exactly this https://github.com/DiogoDuart3/claude-multi-account

Does your method would also work for codex ?

setchy:
It would be nice to support multiple Copilot accounts.  

This should be feasible since its using OAuth Device Code flow.  

I have a corporate copilot account and a personal copilot account that I'd like to track quotas for separately.

Techie5879:
> Would love to have what [@Techie5879](https://github.com/Techie5879) here in upstream!

Makes sense. I've myself tested this for about a week so I feel a bit more comfortable that the implementation actually works. I'll have to clean some stuff up + revert a little bit of the changes - since I didn't care about making it usable for others - initially I just told codex to not worry about the existing UI and just make the UI I wanted - which might lead to it getting unusable for others. 

I'm gonna take a few days to kind of clean it up and do some audits myself before opening a PR bc I dont feel comfortable shipping changes I dont understand a 100% lmao

DiogoDuart3:
Just commenting here to say that I stopped using the Claude Multi Account and started using this project https://github.com/kaitranntt/ccs

gianpaj:
[...]

PR #1829, opened 2026-07-01
feat(claude): read OAuth usage per CLAUDE_CONFIG_DIR account

LawyZheng:
## Summary

Makes Claude OAuth credential resolution `CLAUDE_CONFIG_DIR`-aware so a single CodexBar process can read usage for **multiple Claude subscription accounts** — without an external dependency or manual token pasting.

## Background

Claude Code isolates each `CLAUDE_CONFIG_DIR` into its own macOS Keychain entry and credentials file:

| Config dir | Keychain service | File |
|---|---|---|
| `~/.claude` (default) | `Claude Code-credentials` | `~/.claude/.credentials.json` |
| custom `CLAUDE_CONFIG_DIR` | `Claude Code-credentials-<hash>` | `$CLAUDE_CONFIG_DIR/.credentials.json` |

`<hash>` is the first 8 hex chars of the config directory path's SHA-256. Verified against a real machine: `sha256("/Users/…/.claude-work")[0..<8]` → `da6d47a8` → Keychain entry `Claude Code-credentials-da6d47a8`.

Previously `claudeKeychainService` and the credentials file path were hardcoded to the default account, so CodexBar could only ever read one Claude subscription (#81, #1756). Because Claude Code creates a **distinct** Keychain entry per config dir, deriving the service name from the active `CLAUDE_CONFIG_DIR` is enough to read any account's usage.

## What changed

- **Keychain service** — `claudeKeychainService` becomes a computed property derived from the active `CLAUDE_CONFIG_DIR`. Making the single constant computed means all 14 existing references (candidate probes, preflight, the `security find-generic-password -s` reader, and log labels) resolve to the correct account with **no other call-site changes**.
- **Credentials file** — `defaultCredentialsURL()` resolves `$CLAUDE_CONFIG_DIR/.credentials.json` for the file fallback.
- **Cache isolation** — the in-memory and Keychain credential caches are scoped per account so two accounts never cross-read. The default account's cache key is unchanged (byte-for-byte backward compatible); `KeychainCacheStore.Key.oauth` gains an optional `accountScope` following the existing `cookie(…scopeIdentifier:)` pattern.
- **Injection** — the config dir is injected as a `TaskLocal` at the `loadRecord` entry point (the `environment` dict already flows through the whole read path), so the entire synchronous read sees it.

## How it enables multi-account

A caller reads a specific account's usage by supplying its `CLAUDE_CONFIG_DIR` in the fetch environment, e.g. `["CLAUDE_CONFIG_DIR": "~/.claude-work"]`. The default (empty / `~/.claude`) path is completely unchanged.

This PR is **infrastructure only** — it does not add UI, account enumeration, or credential storage. It's the credential-layer primitive any multi-account approach (menu items, `claude-swap`, etc.) needs.

## Tests

`ClaudeOAuthConfigDirAccountTests`:
- service-name derivation (default → suffix-less; custom → `-<hash>`; trailing-slash normalization; hash == sha256 first 8 hex)
- reading a credentials file from a custom config dir
- the in-memory cache not leaking tokens between two accounts

The hash/normalization logic was also cross-checked independently against the real Keychain entry on macOS.

> Note: `swift build` is clean. I couldn't run `swift test` locally because a transitive `KeyboardShortcuts` SwiftUI `#Preview` macro fails to load under my local Swift 6.3 toolchain (unrelated to this change) — CI should run the suite normally.

## Related

- Closes the enabling half of #1756 / #81 (multi-account) at the credential layer.
- Complementary to the direction in #1812; this provides the zero-dependency, Claude-Code-native primitive.

### DIFF
--- a/Sources/CodexBarCore/KeychainCacheStore.swift
+++ b/Sources/CodexBarCore/KeychainCacheStore.swift
@@ -607,7 +607,12 @@ extension KeychainCacheStore.Key {
         return Self(category: "cookie", identifier: identifier)
     }
 
-    public static func oauth(provider: UsageProvider) -> Self {
-        Self(category: "oauth", identifier: provider.rawValue)
+    public static func oauth(provider: UsageProvider, accountScope: String? = nil) -> Self {
+        let identifier: String = if let accountScope, !accountScope.isEmpty {
+            "\(provider.rawValue).\(accountScope)"
+        } else {
+            provider.rawValue
+        }
+        return Self(category: "oauth", identifier: identifier)
     }
 }
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeOAuth/ClaudeOAuthCredentials+Hashing.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeOAuth/ClaudeOAuthCredentials+Hashing.swift
@@ -15,4 +15,21 @@ extension ClaudeOAuthCredentialsStore {
         return nil
         #endif
     }
+
+    /// Suffix that Claude Code appends to the Keychain service name for a custom
+    /// `CLAUDE_CONFIG_DIR`. Claude Code derives it from the lowercased hex SHA-256
+    /// of the config directory's absolute path, truncated to the first 8 characters.
+    ///
+    /// Example: `/Users/alice/.claude-work` → `da6d47a8` →
+    /// Keychain service `"Claude Code-credentials-da6d47a8"`.
+    static func claudeConfigDirHash8(_ path: String) -> String? {
+        #if canImport(CryptoKit)
+        let digest = SHA256.hash(data: Data(path.utf8))
+        let hex = digest.compactMap { String(format: "%02x", $0) }.joined()
+        return String(hex.prefix(8))
+        #else
+        _ = path
+        return nil
+        #endif
+    }
 }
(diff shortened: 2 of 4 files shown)


steipete:
Thanks @LawyZheng for investigating this and for adding focused tests. I am closing this implementation while keeping #1756 open for the product decision.

This is an auth/storage feature, so VISION requires owner sign-off. The current decision record in #1812 explicitly holds native credential reads and recommends Phase 1 as an opt-in, read-only `cswap --list --json` adapter with no Claude credential or Keychain access.

The audited head `a82985cf24a170bbfc12b93a64ee4dd54bdb5286` is also not safe as a native-account primitive:

- Claude Code 2.1.197 derives the Keychain suffix from the NFC-normalized raw config-directory value. The patch canonicalizes the path and collapses values Claude Code treats as distinct, including an explicit default directory and trailing-slash variants, so it can query the wrong service.
- The TaskLocal scope wraps only `loadRecord`. Availability checks, invalidation, delegated-refresh sync, persistent-reference matching, probes, fingerprints, prompt/failure state, and throttling can still use default or global state while a custom account is requested.
- The tests exercise the wrapped load path and use global invalidation, so they do not expose those unscoped paths.
- This does not provide multi-account product behavior: there is no account enumeration or UI, while usage storage, menu/status items, and widgets remain provider-keyed.

There is no independently useful narrow seam to retain here: the cache-key overload has no approved account consumer, and the remaining changes encode the held native-Keychain design. If native credential discovery is approved later, it should start from an explicit account context threaded through every credential/cache/probe/refresh path and an account-aware usage/UI model, rather than this TaskLocal patch.

Canonical path forward: #1812 and #1756. Thanks again for contributing the investigation.

PR #1837, opened 2026-07-02
feat(claude): weekly Fable usage tracking across menu, CLI, and widget

chrisw1005:
## What & why

Adds a first-class **weekly "Fable"** rate-limit window to the Claude provider, parallel to the existing all-models and model-specific (Sonnet/Opus) weekly buckets, and surfaces it everywhere Claude usage is shown: the menu card, the menu descriptor, the CLI renderer, **and the macOS widget**, plus a dedicated plan-utilization history series.

Motivation: the newest Claude usage payload exposes model-scoped weekly limits in a new `limits` array, and the current code has no way to show the Fable weekly bucket.

## Key correctness detail — the real payload uses `limits[]`, not `seven_day_fable`

On the newest response from the official Anthropic OAuth usage endpoint (`GET https://api.anthropic.com/api/oauth/usage`), the legacy top-level `seven_day_opus` / `seven_day_sonnet` / `seven_day_fable` keys are **null**, and the model-scoped weekly limits live in a `limits` array, identified by `scope.model.display_name`:

```json
{ "kind": "weekly_scoped", "group": "weekly", "percent": 59,
  "resets_at": "…", "scope": { "model": { "id": null, "display_name": "Fable" } } }
```

So this PR parses the `limits` array (OAuth + Web) and maps:
- `weekly_scoped` with `display_name == "Fable"` → the new Fable window.
- other `weekly_scoped` entries → the existing model-specific (Sonnet/Opus) slot — which **also fixes** that slot on the new payload shape, where it currently shows nothing.

The legacy `seven_day_fable` key is kept as a fallback. The CLI TTY path parses the confirmed `Current week (Fable)` panel.

<details>
<summary>Evidence — trimmed real response from the official <code>/api/oauth/usage</code> endpoint</summary>

Unreleased per-model codename keys and the `*_dollars` fields are omitted; the shape of the relevant keys is unchanged.

```json
{
  "five_hour":       { "utilization": 100, "resets_at": "2026-07-02T07:10:00Z" },
  "seven_day":       { "utilization": 34,  "resets_at": "2026-07-04T11:00:00Z" },
  "seven_day_opus":   null,
  "seven_day_sonnet": null,
  "seven_day_fable":  null,
  "limits": [
    { "kind": "session",       "group": "session", "percent": 100,
      "resets_at": "2026-07-02T07:10:00Z", "scope": null, "is_active": true },
    { "kind": "weekly_all",    "group": "weekly",  "percent": 34,
      "resets_at": "2026-07-04T11:00:00Z", "scope": null, "is_active": false },
    { "kind": "weekly_scoped", "group": "weekly",  "percent": 59,
      "resets_at": "2026-07-04T10:59:59Z",
      "scope": { "model": { "id": null, "display_name": "Fable" } }, "is_active": false }
  ]
}
```

Here the account has no `weekly_scoped` Sonnet/Opus entry, so only Session / Weekly / Fable are shown — the model-specific (opus) slot correctly stays empty.

</details>

## Design

Plumbs a `quaternary` window through the shared stack so it renders on all four surfaces **including the widget**:
- `UsageSnapshot` (manual `Codable`, `replacing()`, `backfillingResetTimes`, `hasRateLimitWindows`)
- `WidgetSnapshot.ProviderEntry` + widget UI rows
- `ProviderMetadata.fableLabel` / `supportsFable` (defaulted, so the other ~40 providers compile and behave unchanged)
- the three `rateWindowLabels` render surfaces (extracted to a shared `RateWindowLabels` struct to stay within the `large_tuple` lint limit)
- a `.fable` plan-utilization history series

The lighter `extraRateWindows` approach was considered but does not reach `WidgetSnapshot`, so it can't give widget parity.

When Fable data is absent the window is simply `nil` at every layer (no rows emitted, widget filters the `nil`-percent row, history skips the sample, `Codable` encodes `null`) — identical to how an account without an Opus/Sonnet weekly behaves today. No force-unwraps on the new field.

## Honesty / review notes

- **OAuth path is confirmed** against the real response from the official Anthropic `/api/oauth/usage` endpoint (Fable at 59% via `limits[].scope.model.display_name`; see the evidence block above).
- The **Web `limits` fallback** mirrors OAuth by analogy (same backend shape) but was **not** verified against a real web (`claude.ai/api/.../usage`) payload.
- The CLI probe-JSON keys `week_fable` / `week_fable_only` are CodexBar-internal (no external producer); added by analogy to the existing `week_sonnet` handling.

## Tests

`swift build`, `swift test`, `swiftformat --lint`, and `swiftlint --strict` all pass. New tests cover the real OAuth `limits` payload, the Sonnet+Fable split, the web `limits` fallback, the `Current week (Fable)` TTY panel, and the fixture contract.


chrisw1005:
Updated the description with a trimmed real response from the official `/api/oauth/usage` endpoint as evidence.

On the newest payload the top-level `seven_day_opus` / `seven_day_sonnet` / `seven_day_fable` keys are `null`, and the model-scoped weekly limits (including Fable) now arrive through the `limits` array via `scope.model.display_name`. This PR reads that (keeping `seven_day_fable` as a fallback) and, as a side effect, also restores the Sonnet/Opus weekly slot on the new payload shape.

`swift build`, `swift test`, `swiftformat --lint`, and `swiftlint --strict` all pass. Happy to switch to the lighter `extraRateWindows` route instead of the first-class `quaternary` window if you'd prefer — the reason I went with `quaternary` was widget parity (extra windows don't reach `WidgetSnapshot`).

steipete:
Thanks for the broader implementation and proof, @chrisw1005. We landed the narrower shared scoped-weekly mapping in #1851 (`21e3beb3`), covering both Claude OAuth and web usage responses without introducing separate widget/CLI parsing paths.

Closing this conflicted variant as superseded. If widget presentation still needs a dedicated adjustment after the shared snapshot propagation, that can be a focused follow-up.

PR #1851, opened 2026-07-02
Claude: surface per-model weekly quota carve-outs (e.g. Fable 5) from limits[]

konon4:
## Summary

Anthropic's OAuth (`/api/oauth/usage`) and web (`claude.ai/api/organizations/{org}/usage`)
responses now report weekly quota carve-outs via a `limits` array, e.g.:

```json
{
  "kind": "weekly_scoped", "group": "weekly", "percent": 5,
  "resets_at": "2026-07-08T09:00:00Z",
  "scope": { "model": { "id": null, "display_name": "Fable" }, "surface": null },
  "is_active": false
}
```

This is how Anthropic exposes the July 2026 Fable 5 promotional access window ("up to 50%
of your weekly limit on Fable 5") — it's already shown as a third bar ("Fable only") on
Anthropic's own web/mobile Usage screen, but CodexBar didn't surface it. The existing flat
`seven_day_sonnet`/`seven_day_opus` fields are `null` once this shape is present, and
neither CodexBar fetch path decoded `limits` at all.

- Add `OAuthLimitEntry`/`OAuthLimitScope`/`OAuthLimitScopeModel` to decode the array in
  `ClaudeOAuthUsageFetcher.swift`, and a matching dict-based parser in
  `ClaudeWebExtraRateWindowParser.swift` for the web fetch path (both paths need it — Claude
  auto-selects web over OAuth when a browser session is available).
- Surface any active `weekly_scoped` limit as a `NamedRateWindow`, titled `"{model} only"`,
  keyed off `scope.model.display_name` rather than hardcoding `"Fable"` — the promo ends
  2026-07-07 and this shape will likely be reused for a different model afterwards.
- No UI changes needed: the menu already renders `extraRateWindows` unconditionally for
  Claude (same mechanism as the existing "Daily Routines" window).

## Test plan

- [x] `swift test --filter "ClaudeOAuthTests|ClaudeWebUsageExtraWindowTests"` — 37/37 pass,
      including two new tests per fetch path using the real `limits` payload shape.
- [x] `swiftformat --lint` / `swiftlint --strict` on changed files — 0 violations.
- [x] Verified against a live Claude Max account: CLI JSON output for both `--source oauth`
      and `--source web` (the default `auto` path) reports
      `{"id": "claude-weekly-scoped-fable", "title": "Fable only", "usedPercent": 5}`,
      matching the native Claude Usage screen's "Fable only 5% used" bar exactly.
- [x] Built and ran the macOS app locally; the "Fable only" bar renders in the Claude menu
      alongside Session/Weekly/Daily Routines.


steipete:
Landed as `21e3beb3`.

Maintainer closeout on exact head `46cd1fbcc5a6f36b22156e15b88f0364b47c8306`:

- 42 focused Claude OAuth/web/scoped-window tests passed.
- `make test`: 46/46 groups passed.
- `make check`: passed.
- Autoreview: clean; no accepted/actionable findings.
- Exact-head CI: 10/10 checks passed.
- Contributor authenticated proof confirmed the scoped Fable quota response shape.

The landed version shares one mapper across OAuth and web sources, uses stable model IDs, rejects non-finite utilization, trims/deduplicates labels, and keeps unrelated weekly windows unchanged. Thanks @konon4!

PR #1867, opened 2026-07-03
Show OpenCode Go monthly pace

joeVenner:
## Summary

- show pace status and projected run-out details on the OpenCode Go monthly row
- reuse the existing billing-cycle pace calculation shared with Cursor
- keep OpenCode Go's 5-hour row unchanged, limiting the fix to the reported monthly gap
- add deterministic menu-card model coverage

Fixes #1863.

## After-fix visual proof

This image is rendered from the production `UsageMenuCardView` on this branch with a deterministic OpenCode Go fixture. It shows the unchanged 5-hour/Weekly rows and the new Monthly pace detail:

OpenCode Go monthly pace after fix

The Monthly row now shows `25% in deficit`, `Runs out in 5d`, and the expected pace marker. The capture path does not read browser cookies, Keychain data, or a live provider account.

## Testing

- `swift test --disable-sandbox --filter OpenCodeGoMenuCardModelTests` (4 passed)
- `swift test --disable-sandbox --filter UsagePaceTextTests` (21 passed)
- `swift test --disable-sandbox --filter MenuBarPaceTextTests` (2 passed)
- production SwiftUI proof render passed (`OpenCodeGoProofCaptureTests`, temporary uncommitted capture harness)
- `make check` (0 formatting or lint violations)
- `make test` (sandbox module-cache failure during test discovery in the isolated checkout; all focused changed-area suites pass)

PR #1870, opened 2026-07-03
Fix monthly reset-window pace

### LINKED ISSUE #1863: OpenCode Go monthly bar should show on pace, runs out in etc
similar to #1336 

In opencode go, currently it only shows "on pace", "deficit", "runs out in" etc in weekly but not in monthly. It should also show that in monthly and potentially 5-hour too (5-hour might be less needed).

Image

Zihao-Qi:
## Summary

Fix monthly reset-window pace details for OpenCode Go and related providers.

- Generalize the old Cursor-only billing-cycle pace helper into an explicit reset-window pace helper.
- Enable monthly pace details for OpenCode Go, Doubao, Alibaba Coding Plan, and Alibaba Token Plan.
- Infer actual monthly duration from `resetsAt` so 28/29/31-day cycles do not blank or skew pace.
- Keep the helper provider/window allowlisted to avoid affecting unrelated tertiary rows.
- Add focused menu-model tests for the supported providers and month-length edge cases.

## Why

OpenCode Go exposes a monthly quota window, but the menu only showed pace details for Cursor billing-cycle windows. The same monthly reset-window shape also exists for Doubao and Alibaba plans, so keeping the logic Cursor-specific would miss equivalent provider behavior.

The fix keeps the behavior narrow: only known monthly reset-window providers opt in, and unrelated daily, credit-balance, or arbitrary tertiary rows continue using existing rendering.

This overlaps with #1867, but covers the broader monthly reset-window family and adds 28/31-day month coverage.

Fixes #1863.

## Validation

- `swift test --filter 'OpenCodeGoMenuCardModelTests|DoubaoMenuCardModelTests|AlibabaCodingPlanMenuCardModelTests|AlibabaTokenPlanMenuCardModelTests|CursorMenuCardModelTests'`
- `make check`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Xiaomi MiMo: require authoritative cadence evidence before showing reserve or deficit projections, avoiding guesses from plan dates or names.
 
 ### Fixed
+- Monthly quota pace: show reserve, deficit, and run-out estimates for OpenCode Go, Doubao, and Alibaba monthly reset windows using their calendar-cycle length. Thanks @Zihao-Qi and @joeVenner!
 - Claude: block background delegated CLI OAuth refresh when the keychain holds MCP-only state (`mcpOAuth` without `claudeAiOauth`) while preserving explicit Refresh recovery (#1844). Thanks @Yuxin-Qiao!
 
 ## 0.38.0 — 2026-07-03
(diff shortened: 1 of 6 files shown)


Zihao-Qi:
@clawsweeper re-review

steipete:
Landed as a broader, allowlisted monthly-pace fix.

Proof on exact head `fe0717496da52a18406d1cc68e6b0e217579c6d1`:

- 14 focused monthly pace tests across OpenCode Go, Doubao, and both Alibaba plan providers
- `make check`
- `make test` (45/45 shards)
- exact-head CI: 10/10 checks green

The implementation derives real 28/29/30/31-day calendar cycles from reset timestamps and keeps Cursor's existing behavior unchanged. Thanks @Zihao-Qi and @joeVenner!

PR #1878, opened 2026-07-04
[codex] Fix Claude keychain prompt toggle

### LINKED ISSUE #1877: Claude "Avoid Keychain prompts" can trigger Keychain prompts
## Problem

In CodexBar 0.38.0 (94), enabling the Claude provider setting **Avoid Keychain prompts** can cause repeated macOS Keychain prompts for the Claude credential. Turning the setting off stops the prompts.

That is inverted from the expected behavior and makes the Claude provider noisy for users who explicitly chose the no-prompt path.

## Repro

1. Use CodexBar 0.38.0 (94) with the Claude provider configured via Claude OAuth.
2. Open Claude provider settings.
3. Enable **Avoid Keychain prompts**.
4. Let a background refresh run, or refresh Claude usage manually.
5. Observe a macOS Keychain prompt for the Claude credential.
6. Disable **Avoid Keychain prompts**.
7. Observe that the prompts stop.

## Expected behavior

When **Avoid Keychain prompts** is enabled, Claude OAuth credential reads should use the non-interactive Keychain policy and fail closed when macOS would require user interaction. The setting should not switch to a read path that can show prompts.

## Actual behavior

The setting appears to select the experimental `/usr/bin/security` CLI reader. That reader shells out to `security find-generic-password -w`, which can still display macOS Keychain prompts. With the setting disabled, the Security.framework path is used instead, where CodexBar can apply its no-UI policy.

## Suggested fix

- Treat stored `securityCLIExperimental` preferences as legacy and migrate them back to `securityFramework`.
- Make **Avoid Keychain prompts** set the Claude OAuth prompt policy to `never` rather than selecting the experimental CLI reader.
- Keep the Claude Keychain prompt policy control visible as the source of truth.
- Update the setting copy so it describes prompt prevention, not the CLI read strategy.
- Cover the migration, toggle mapping, and provider descriptor behavior with focused tests that do not touch the real Keychain.

## Labeling

Suggested labels from `docs/ISSUE_LABELING.md`:

- `bug`
- `priority:high`
- `area:auth-keychain`
- `provider:claude`

gmkbenjamin:
## Summary

- Migrate legacy Claude `securityCLIExperimental` keychain read preferences back to the Security.framework reader.
- Make **Avoid Keychain prompts** control the Claude OAuth prompt policy (`never`) instead of selecting `/usr/bin/security`.
- Keep the Claude Keychain prompt policy visible and update the setting copy to describe prompt prevention.
- Add focused settings/provider descriptor coverage and document the fix in the changelog.

Fixes #1877.

## Root Cause

The **Avoid Keychain prompts** toggle selected the experimental `/usr/bin/security find-generic-password -w` reader. That CLI path can still trigger macOS Keychain prompts, so enabling the no-prompt setting inverted the expected behavior. The standard Security.framework path is the one where CodexBar can apply its no-UI policy.

## Validation

- `swift test --filter SettingsStoreCoverageTests`
- `swift test --filter ProviderSettingsDescriptorTests`
- `make check`
- `git diff --check`

`make test` was started but intentionally interrupted after a macOS Keychain prompt appeared. Per `AGENTS.md`, full-suite validation paths that can display Keychain prompts should not be used for this change; coverage here is through focused parser/settings/provider tests and lint/checks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 
 ### Fixed
 - Branding: replace the malformed Poe icon and use Poe's official purple consistently across the app, widget, and website. Thanks @garethpaul!
+- Claude: make the "Avoid Keychain prompts" setting use the no-prompt policy instead of the experimental `security` CLI reader.
 - Monthly quota pace: show reserve, deficit, and run-out estimates for OpenCode Go, Doubao, and Alibaba monthly reset windows using their calendar-cycle length. Thanks @Zihao-Qi and @joeVenner!
 - Localization: translate the Default Terminal setting across every supported app language. Thanks @Zihao-Qi!
 - Settings: recover collapsed sidebars and undersized saved window frames when reopening Settings. Thanks @ProspectOre!
--- a/Sources/CodexBar/Providers/Claude/ClaudeProviderImplementation.swift
+++ b/Sources/CodexBar/Providers/Claude/ClaudeProviderImplementation.swift
@@ -67,7 +67,7 @@ struct ClaudeProviderImplementation: ProviderImplementation {
         let subtitle = if context.settings.debugDisableKeychainAccess {
             "Inactive while \"Disable Keychain access\" is enabled in Advanced."
         } else {
-            "Use /usr/bin/security to read Claude credentials and avoid CodexBar keychain prompts."
+            "Never allow Claude OAuth credential reads to show macOS Keychain prompts."
         }
 
         let promptFreeBinding = Binding(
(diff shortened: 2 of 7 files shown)


steipete:
Thanks @gmkbenjamin — your implementation and contributor credit landed in https://github.com/steipete/CodexBar/pull/1879 at b435e27b429f167cc7dce532be896a1929e69b3b.

I recreated the branch because this fork did not allow maintainer edits, then added the migration, regression, and exact built-app proof needed to land it safely. For future PRs, enabling Allow edits by maintainers lets us improve the original branch directly.

PR #1890, opened 2026-07-04
Guard Mistral credit balances

joeVenner:
## Summary

- reject Mistral credit responses whose components or computed available balance are non-finite
- prevent public credit snapshots from formatting an infinite balance
- preserve the existing zero floor for negative available balances
- add regression coverage for finite inputs whose sum overflows

## Production-path proof

A temporary standalone executable linked against `CodexBarCore` called the public `MistralUsageFetcher.fetchCredits` with an in-memory HTTP fixture. The temporary target was removed after the run.

```text
$ swift run --disable-sandbox --scratch-path .build MistralCreditProof
Build of product MistralCreditProof complete!
overflowing-credit-fixture=rejected reason=Invalid credit amount
```

This exercises the shipping credit fetch/decode path without accessing browser cookies or a live account.

## Testing

- `swift test --disable-sandbox --filter MistralUsageParserTests` (21 tests passed across matched Mistral suites)
- `make check` (SwiftFormat clean; SwiftLint 0 violations; later host plist write failed because the sandbox cannot write the system cache directory)
- `git diff --check`

### DIFF
--- a/Sources/CodexBarCore/Providers/Mistral/MistralModels.swift
+++ b/Sources/CodexBarCore/Providers/Mistral/MistralModels.swift
@@ -287,7 +287,8 @@ public struct MistralCreditsSnapshot: Codable, Equatable, Sendable {
     }
 
     public var availableAmount: Double {
-        max(0, self.walletAmount + self.creditNotesAmount - self.ongoingUsageBalance)
+        let amount = self.walletAmount + self.creditNotesAmount - self.ongoingUsageBalance
+        return amount.isFinite ? max(0, amount) : 0
     }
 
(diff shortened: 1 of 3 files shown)


joeVenner:
@clawsweeper re-review

steipete:
Landed as `2e325b86`.

Verified the exact PR head `676cd2a6` with:

- `swift test --filter MistralUsageParserTests` — 21 tests passed
- `make check` — clean
- Codex autoreview — no actionable findings
- GitHub CI — all 10 exact-head checks green

Added the 0.39.1 changelog entry in `f837c33b`. Thanks @joeVenner!

PR #1895, opened 2026-07-04
Fix language picker localization

Zihao-Qi:
## Summary

- Keep every language-picker option readable in its own stable/native language, independent of the current UI language.
- Make System follow macOS language preferences instead of inheriting a stale CodexBar override.
- Cache per-language bundle lookups used by the picker.
- Preserve unrelated external `AppleLanguages` overrides; migrate only the exact legacy override CodexBar previously owned.

## Why

The old picker localized language names into the currently selected language. A mistaken selection could therefore make recovery difficult. Explicit app choices were also mirrored into `AppleLanguages`, so returning to System could retain the previous CodexBar language.

The picker now uses stable native names. CodexBar keeps its explicit choice in `appLanguage`, and startup removes the old app-domain `AppleLanguages` value only when it exactly matches that stored choice. System mode and unrelated external overrides remain untouched.

## Maintainer improvements

- Merged current `main` before validation.
- Replaced unconditional startup deletion with an ownership-aware migration guard.
- Added regressions for System mode, matching legacy cleanup, and unrelated external overrides.
- Used readable Swift Testing sentence names.

## Validation

- `swift test --filter 'LocalizationLanguageCatalogTests|PreferencesPaneSmokeTests'` — 34 tests passed.
- `make check` — 0 format/lint violations.
- `make test` — all 47 shards passed.
- Full-branch autoreview — no actionable findings.
- Fresh ad-hoc debug bundle, isolated temporary bundle identifier:
  - System mode + external German `AppleLanguages` retained the override and rendered German menus (`Ablage`, `Bearbeiten`, `Darstellung`, `Fenster`, `Hilfe`).
  - Japanese `appLanguage` + matching legacy `AppleLanguages` removed the legacy key and rendered Japanese menus (`ファイル`, `編集`, `表示`, `ウインドウ`, `ヘルプ`).

## Screenshots

| Before | After |
|---|---|
| Language picker before | Language picker after |


steipete:
Exact head `9470a32988663b222941c501099cc8c3da66380c` verified: focused language migration suite (34 tests), `make check`, full 47-shard `make test`, isolated live app proof for externally managed German and app-selected Japanese, and all required CI checks green. The maintainer rewrite preserves external `AppleLanguages` overrides and only removes the exact legacy override owned by CodexBar. Thank you, @Zihao-Qi!

PR #1901, opened 2026-07-04
Settings window layout

Zihao-Qi:
## Summary

This PR improves the Settings window layout and sidebar behavior.

- Replaces the unstable resizable `NavigationSplitView` sidebar with a fixed-width custom layout.
- Keeps the sidebar visible and non-draggable, so it can no longer collapse, disappear, or become extremely wide.
- Polishes the sidebar visual treatment with a rounded translucent panel, lighter/darker wash for light and dark mode, and adjusted spacing.
- Keeps the detail content constrained so settings rows do not stretch awkwardly across the full window.
- Preserves normal window resizing: an 800-point minimum, the 880-point default, and wider layouts all keep the sidebar stable while the detail column stays readable.
- Reduces unnecessary window appearance refresh work when only the selected pane title changes.

## Why

The previous Settings window used SwiftUI `NavigationSplitView`, but the underlying split divider was still draggable. That caused two bad states:

- dragging left could collapse the sidebar until it disappeared
- dragging right could make the sidebar much wider than intended

Trying to repair this from AppKit was fragile because SwiftUI still owned the split layout. A fixed custom layout better matches the intended behavior: a stable System Settings-style sidebar that is always visible and not user-resizable.

The visual updates make the sidebar feel closer to macOS Settings while keeping the existing app structure and behavior.

## Validation

Ran:

- `swift test --filter SettingsWindowAppearanceTests`
- `make check`
- `git diff --check`
- exact release bundle: resized live to 800×572, 880×652, and 1500×900; verified the sidebar remains fixed, the detail column remains capped, and the minimum layout scrolls instead of clipping controls

### Before

| Issue | Screenshot |
|---|---|
| Dragging left could collapse the sidebar until it disappeared. | Screenshot 2026-07-03 at 21 56 27 |
| Dragging right could make the sidebar much wider than intended. | Screenshot 2026-07-03 at 21 56 13 |

### after
Screenshot 2026-07-04 at 13 59 17

### DIFF
--- a/Sources/CodexBar/CodexbarApp.swift
+++ b/Sources/CodexBar/CodexbarApp.swift
@@ -100,7 +100,7 @@ struct CodexBarApp: App {
                 })
         }
         .defaultSize(width: SettingsPane.windowWidth, height: SettingsPane.windowHeight)
-        .windowResizability(.contentSize)
+        .windowResizability(.contentMinSize)
     }
 
     private func openSettings(pane: SettingsPane) {
--- a/Sources/CodexBar/PreferencesAboutPane.swift
+++ b/Sources/CodexBar/PreferencesAboutPane.swift
@@ -80,6 +80,7 @@ struct AboutPane: View {
         }
         .formStyle(.grouped)
         .toggleStyle(.switch)
+        .scrollContentBackground(.hidden)
         .onAppear {
             guard !self.didLoadUpdaterState else { return }
             // Align Sparkle's flag with the persisted preference on first load.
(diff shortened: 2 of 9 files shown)


steipete:
Landed in 17347d84 from exact head c979cdd8 after 10/10 checks passed. Focused Settings appearance tests and `make check` passed; the release bundle was verified at 800×572, 880×652, and 1500×900 with a fixed visible sidebar, capped readable detail content, and scrolling rather than clipping at the minimum size. Changelog follow-up: 68f5f915. Thanks @Zihao-Qi!

Issue #1886, opened 2026-07-04
Claude account switch not reflected on 0.38.0 — #1800's Keychain-identity history key collides for same-dir `/login` accounts (regression of #1785)


kes02:
## Problem

On CodexBar 0.38.0, switching the active Claude account is not reflected: the menu bar and usage history keep showing the previous account. This is a regression introduced by #1800 (the fix for #1785). #1800 scopes Claude OAuth history by a Keychain-derived identity key — but when multiple accounts share the default `~/.claude`, Claude Code's `/login` overwrites the **same** Keychain item in place, so the derived key is identical for different accounts and their history collapses into a single bucket.

Two independent live reproductions exist on 0.38.0:

1. **Komunikuji's report** (in #1785): Teams ↔ Max 5x via VS Code `/switch-account`. On 0.38.0 it reproduces in **both directions**, the earlier "switch a second time" and full-restart workarounds **no longer help**, and the `Claude Code-credentials` Keychain prompt fires repeatedly despite *Always Allow*.
2. **This report**: two personal (non-org) accounts on default `~/.claude`, switched via CLI `/login`. Details below.

Filing a fresh issue because #1785 is closed and cannot be reopened without write access (ClawSweeper declined a re-review on the closed thread for that reason).

## Environment

- CodexBar **0.38.0 (94)**, installed via Homebrew cask
- macOS 26.5.1
- Two personal Claude accounts (**A** and **B**) on the default `~/.claude`
- Switched with the Claude Code CLI `/login` (not VS Code `/switch-account`)

## Reproduction

1. On 0.37.2, switch personal↔personal with `/login` → CodexBar follows within ~20–30s (baseline: history fully `unscoped`, `accounts: []`, `preferredAccountKey: null`).
2. Upgrade to 0.38.0.
3. `/login` from Account A to Account B.
4. Watch CodexBar / `history/claude.json`.

## Actual result

Polling the on-disk state (`history/claude.json` + the `ClaudeOAuthClaudeKeychainFingerprintV2` pref) every 3s across a real `/login` (A → B):

| Signal | before (A) | after (B) | |
|---|---|---|---|
| CLI account (`~/.claude.json`) | A | B | ✅ actually switched |
| `persistentRefHash` (`…FingerprintV2`) | `ebfc0365…` | `ebfc0365…` | ❌ identical |
[...]

PR #1903, opened 2026-07-04
Fix Claude same-dir account-switch history mis-attribution (#1886)

ss251:
## Summary

Addresses the history-isolation portion of #1886.

Claude Code can update the shared Keychain item and `~/.claude.json` during `/login` while `~/.claude/.credentials.json` remains stale. A background refresh could then attribute a successful OAuth sample to the wrong account-history lifecycle.

## Final design

- Derives an opaque history owner from the exact OAuth credential used for the successful fetch; raw tokens are never persisted.
- Brackets the fetch with prompt-free credential/account observations and rejects evidence that changes mid-fetch.
- Distinguishes exact Keychain match, mismatch, proven absence, and unavailable comparison.
- Confirms new owner-to-account bindings with two stable exact-match observations; existing bindings quarantine stale samples after an observed account switch.
- Keeps supported file-only and default background refreshes recording into their credential-owned bucket when no authoritative binding exists.
- Preserves explicit/environment credentials and proven refresh-token rotation lineage.
- Stores only hashed account identities and opaque credential-owner identifiers.

This intentionally does not change visible menu snapshot selection. #1886 should remain open for that remaining behavior.

## Proof

- `swift test --filter 'ClaudeOAuthHistoryCredentialRoutingTests|UsageStorePlanUtilizationClaudeIdentity'` — 37 tests passed across 3 suites.
- `make check` — clean; SwiftFormat and strict SwiftLint passed.
- `make test` — all 47 shards passed.
- `/Users/steipete/Projects/agent-skills/skills/autoreview/scripts/autoreview --mode branch --base origin/main` — clean; no accepted/actionable findings.

Exact reviewed head: `bfe9ccda2ba7c07db6648ba246e1a63eed2cfe99`

## Notes

- No UI change; no screenshots required.
- No dependency or Keychain-prompt policy change.
- Contributor live reproduction established the original same-directory `/login` behavior; maintainer hardening and isolated regression tests cover the final evidence boundary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 - Claude CLI: skip the identity probe after terminal usage errors or loading stalls, cutting failed refresh latency and subprocess churn.
 - OpenCode web: search Dia after Chrome for automatic cookie import, with Keychain preflight scoped to the candidate browser (fixes #1822). Thanks @zeajose!
 - Claude: make the "Avoid Keychain prompts" setting use the no-prompt policy instead of the experimental `security` CLI reader. Thanks @gmkbenjamin!
+- Claude: keep per-account plan-utilization history separate after an in-place account switch in the shared `~/.claude` directory, so background polls no longer attribute the new account's usage to the previous account's history (fixes #1886). Thanks @ss251!
 
 ## 0.38.1 — 2026-07-04
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,6 @@
 - Claude CLI: skip the identity probe after terminal usage errors or loading stalls, cutting failed refresh latency and subprocess churn.
 - OpenCode web: search Dia after Chrome for automatic cookie import, with Keychain preflight scoped to the candidate browser (fixes #1822). Thanks @zeajose!
 - Claude: make the "Avoid Keychain prompts" setting use the no-prompt policy instead of the experimental `security` CLI reader. Thanks @gmkbenjamin!
-- Claude: keep per-account plan-utilization history separate after an in-place account switch in the shared `~/.claude` directory, so background polls no longer attribute the new account's usage to the previous account's history (fixes #1886). Thanks @ss251!
 
 ## 0.38.1 — 2026-07-04
 
--- a/Sources/CodexBar/UsageStore+CodexResetCredits.swift
+++ b/Sources/CodexBar/UsageStore+CodexResetCredits.swift
@@ -119,7 +119,8 @@ extension ProviderFetchOutcome {
                 strategyKind: result.strategyKind,
                 claudeOAuthKeychainPersistentRefHash: result.claudeOAuthKeychainPersistentRefHash,
                 claudeOAuthHistoryOwnerIdentifier: result.claudeOAuthHistoryOwnerIdentifier,
-                claudeOAuthKeychainCredentialMismatch: result.claudeOAuthKeychainCredentialMismatch)),
+                claudeOAuthKeychainCredentialMismatch: result.claudeOAuthKeychainCredentialMismatch,
+                claudeOAuthKeychainCredentialUnavailable: result.claudeOAuthKeychainCredentialUnavailable)),
             attempts: self.attempts)
     }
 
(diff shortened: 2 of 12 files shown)


ss251:
Thanks for the reviews. Addressed the P2 (empty-map first-sighting / map poisoning) flagged by both Codex and ClawSweeper in `8bf824dc`:

- The first-sighting `owner → accountUuid` binding is now created **only on a user-initiated refresh** (`ProviderInteractionContext.current == .userInitiated`), where the keychain freshness sync guarantees the served owner reflects the active account. Background polls **honor** an existing binding (quarantine on mismatch) but never **create** one, and an unmapped background read fails open with no persisted binding — so the map can no longer be poisoned. New regression test `background first sighting does not poison the account map`, verified to fail if the trustworthy-read gate is removed.
- Documented the accepted warm-up limitation (ClawSweeper maintainer option 2): the safeguard arms after the first user-initiated sample on an account; the narrow window before that fails open (same as pre-fix) but never poisons the map.

**Real-behavior proof** (now in the PR body): ran the patched `recordPlanUtilizationHistorySample` against **real** identity data on a two-account machine, prompt-free (plain-file reads only) — real owner (app-formula SHA256 of the real refresh token, never printed), active account live-read from the real `~/.claude.json`, account B's real `accountUuid` captured from an actual `/login` switch. The stale background sample after the switch is **not** written to the old bucket (history stays `[[30.0]]`, `unscoped` empty), and the empty-map upgrade path creates no binding.

`make test` (47 shards) + `make check` green; all Claude-identity + plan-utilization tests pass.

@clawsweeper re-review

ss251:
@codex review

ss251:
Re: the follow-up P2 on the empty-map upgrade window — this is a deliberate, documented tradeoff rather than an oversight, and I want to lay out why fail-open is the default (with the quarantine variant ready if maintainers prefer it).

The genuine regression this PR introduced — **map poisoning** (binding a stale owner to the new UUID) — is fixed: background polls never create a binding. What remains is the narrow warm-up window where an upgraded user switches accounts *before* the map is warmed by any user-initiated refresh; there the background sample fails open, identical to pre-fix behavior.

I considered the suggested quarantine-on-known-scoped-owner, but it has a broad downside: `shouldRecordPlanUtilizationHistory` returns `true` for `.claude` unconditionally, so **background polls record Claude history**. Quarantining background first-sightings that already have a bucket would stall plan-utilization history for every user who relies on the menu-bar text and rarely opens the menu (their owner stays unmapped, so every post-first background sample would be dropped) — turning a rare, self-healing contamination into a broad ongoing history gap. That's a worse default for a usage-history feature.

So: fail-open keeps history continuous and never poisons the map; the residual contamination is bounded to the pre-warm-up window and self-heals on the first user-initiated refresh. If the maintainers would rather guarantee no cross-account contamination at the cost of that background-history gap, the quarantine variant is a ~10-line follow-up (drop background first-sightings whose owner already has a scoped bucket) — happy to add it. That's the product call framed as ClawSweeper's option 1 vs 2.

ss251:
@clawsweeper re-run

ss251:
**clawsweeper's option 1 vs 2, condensed** (full rationale in the PR description):

- **Fail-open over quarantine:** background polls also record plan-utilization history (`shouldRecordPlanUtilizationHistory`), so quarantine would stall history for menu-bar-only users. A broad ongoing gap vs. a rare one-shot mis-write.
- **Residual risk is bounded:** only an upgraded user who switches accounts before any user-initiated refresh — worst case one write to the old bucket, self-healed on the first refresh. Never worse than today, strictly better after warm-up.
- **On the two merge-risk labels:** no keychain gate touched (#1848 boundary intact — only `~/.claude.json` is read, prompt-free); the UserDefaults map is purely additive (absent map = current behavior).

If you prefer guaranteed no-contamination, the quarantine variant (option 1) is ~10 lines — I can push it same-day.

steipete:
@clawsweeper re-review

Maintainer repair is now on current `main` at exact head `70dc617a4ee5bd88090225ee50191d80daac34de`. The owner map now arms only from an exact current-Keychain credential match; user interaction alone cannot bind stale ownership. Focused tests, all 47 local shards, `make check`, and AutoReview are green.

steipete:
Landed as [`9a002844`](https://github.com/steipete/CodexBar/commit/9a0028442e3508d9d2a9728bcd88ef68e7217135), with the release-note closeout in [`cda65e70`](https://github.com/steipete/CodexBar/commit/cda65e70).

Exact-head proof for `bfe9ccda`:

- GitHub CI: all 10 checks green, including four macOS shards, both Linux CLI jobs, lint, aggregate gate, and GitGuardian.
- Focused history/identity proof: 37 tests across `ClaudeOAuthHistoryCredentialRoutingTests` and `UsageStorePlanUtilizationClaudeIdentity*` passed.
- Full local suite: all 47 shards passed.
- `make check`: passed; SwiftFormat clean and SwiftLint reported 0 violations.

This lands the history-isolation portion only. #1886 remains open for the separate visible-menu account-selection problem.

Thanks @ss251!

Issue #1758, opened 2026-06-25
Show all Codex reset-credit expiration times, not only the next one


simon-ami:
## Problem

CodexBar currently shows only the next expiring manual Codex reset credit in the menu, e.g. `Next expires in 1d`.

That is useful, but incomplete when multiple reset credits are available. Users can see how many manual resets they have, but not when the other available resets expire.

## Current behavior

When multiple Codex reset credits are available, the menu shows:

* `Limit Reset Credits`
* e.g. `2 manual resets available`
* e.g. `Next expires in 1d`

The other available reset credits' expiration times are not visible from the menu.

## Expected behavior

CodexBar should make all available reset-credit expiration times visible while keeping the default menu compact.

Preferred UI:

* Keep the current compact row unchanged by default.
* Make the `Limit Reset Credits` row expandable/clickable, or add a small `Show details` affordance.
* When expanded/clicked, show all available reset credits sorted by expiration time.

Example expanded content:

```text
Manual reset expirations:
1. Expires in 1d — Jan 12, 14:30
2. Expires in 6d — Jan 17, 09:15
3. Expires in 13d — Jan 24, 18:00
```

Alternative UI:

* If it better matches existing CodexBar menu conventions, show the same information via hover/help text.
* However, click-to-expand may be preferable because hover-triggered popovers/tooltips can become noisy when moving the mouse through the menu.

The exact formatting can follow the existing reset-time display preference:

* countdown style if reset times are configured as countdowns
* absolute clock/date style if reset times are configured as absolute values
* ideally include enough date context to distinguish resets expiring on different days

## Why this matters

Manual reset credits can expire independently. Showing only the next expiration helps with the most urgent one, but it does not answer:

* whether all resets expire close together,
* whether later resets can be saved safely,
* how much time remains before each individual reset expires.

This makes the reset-credit count less actionable, especially for users who have more than one available reset.

## Possible implementation direction

The existing model already appears to have enough source data: the reset-credit snapshot contains the full list of credits, while the menu currently displays only the next expiring available credit.

A possible implementation path:

1. Add a model field for a full reset-credit expiration summary, e.g. `codexResetCreditsDetailItems`, `codexResetCreditsExpandedText`, or similar.
2. Build it from available, non-expired reset credits sorted by `expiresAt`.
3. Keep the normal menu row compact by continuing to show only the next expiration inline.
4. Show the full list only on click/expand, or as hover/help text if that fits the app better.
[...]

PR #1907, opened 2026-07-05
Opt-in expiry runway for Codex reset credits (addresses #1758)

cleanerzkp:
## Summary

An **opt-in** "expiry runway" view for the Codex reset-credits menu row: a horizontal timeline with one dot per credit, so all expiries are visible at a glance instead of clipped behind `+N` in the compact list. Addresses #1758 (show all reset-credit expirations, not only the soonest).

**Default is unchanged** — this ships behind a new `CreditExpiryDisplayStyle` picker (Display settings) defaulting to `.list`, so nobody's current row changes unless they opt in. Happy to flip the default or drop the setting entirely — see the open question below.

## Why

The 0.39.x compact list (#1902) reads well at 2–3 credits but at 4+ it clips (`7d 7h · 13d 5h · 22d 4h · +2`) and asks you to place four countdowns on a mental timeline. The runway makes *position* carry the timing: a far-right cluster reads "nothing expires for weeks" pre-attentively, and the nearest dot turns red under 48h. Hovering any dot reveals its exact countdown + absolute date (mirroring the `ZaiHourlyUsageChartMenuView` hover), so the runway is a strict **superset** of the list — no information is lost, and the row's `.help()` tooltip + accessibility label are untouched.

## Design notes (kept native)

- **Data-derived horizon** (`max(expiresAt − grantedAt)` across the inventory) — no hardcoded window; credits with no expiry stay counted/in the tooltip but off the track.
- **Theming via `MenuHighlightStyle`** (`progressTrack` / `progressTint` / `error`), mirroring `UsageProgressBar`, so the runway stays legible on the highlighted (blue) row rather than clashing with raw `.accentColor`/`.red`.
- **Reset-time style** is respected for free (the nearest label reuses `compactExpiryText`); hover shows both forms.
- **Reduce Motion**: hover scale/fade gate on `accessibilityReduceMotion` (per `QuotaWarningAlertOverlayController`).
- **Height fingerprint**: the runway is a distinct fixed-height row, so `CodexResetCreditsPresentation.heightFingerprint` gets a `runway` discriminator (the honest cost of gating — two heights).

## Open question for you

Straight **replacement**, or keep the **gate** (this PR)? Replacement is defensible since hover makes the runway a superset of the list and it drops the second height/settings surface; the gate keeps your fresh default untouched. I went with opt-in to be conservative on a surface you just built — flipping the default or removing the picker is a one-line change, your call.

Not in scope (declared): single-point resets (session/weekly/monthly, `CodexCreditLimitSnapshot.resetsAt`) stay as text — one timestamp needs no axis. A natural *future* adopter is OpenAI API credit grants (per-grant `expiresAt` exists but the snapshot currently collapses to `nextGrantExpiry`, so it'd need fetcher plumbing first) — happy to follow up if this direction lands.

## Commands run

- `swift build` — green
- `swift test --filter CodexResetCreditRunwayTests --filter CodexResetCreditsMenuCardTests --filter MenuCardHeightFingerprintTests` — green
- `make check` — SwiftFormat + SwiftLint + locale gate green

## Tests

`CodexResetCreditRunwayTests` covers marker placement over the data-derived horizon, the nearest-credit flag, `<48h` urgency, no-expiry exclusion, empty→list fallback, the horizon label, and that `make(…)` only builds runway data for `.runway`.

### DIFF
--- a/Sources/CodexBar/MenuCardHeightFingerprint.swift
+++ b/Sources/CodexBar/MenuCardHeightFingerprint.swift
@@ -121,7 +121,15 @@ extension UsageMenuCardView.Model.TokenUsageSection {
 
 extension CodexResetCreditsPresentation {
     fileprivate var heightFingerprint: String {
-        MenuCardHeightFingerprint.join([
+        // The runway is a fixed-height component (track + axis) distinct from the single-line list,
+        // so the style choice is the only height-relevant contributor once runway is active.
+        if self.showsRunway {
+            return MenuCardHeightFingerprint.join([
+                MenuCardHeightFingerprint.field("text", self.text),
+                MenuCardHeightFingerprint.field("expiry", "runway"),
+            ])
+        }
+        return MenuCardHeightFingerprint.join([
             MenuCardHeightFingerprint.field("text", self.text),
             MenuCardHeightFingerprint.field("expirySummary", self.expirySummaryText),
         ])
--- a/Sources/CodexBar/MenuCardView+CodexResetCreditRunway.swift
+++ b/Sources/CodexBar/MenuCardView+CodexResetCreditRunway.swift
@@ -1,0 +1,158 @@
+import CodexBarCore
+import SwiftUI
+
+/// One expiring reset credit placed on the runway timeline.
+struct CodexResetCreditRunwayMarker: Equatable, Identifiable {
+    let id: String
+    /// 0…1 along the horizon: 0 = now, 1 = a freshly granted credit at the far edge.
+    let position: Double
+    let isNearest: Bool
+    /// Nearest credit expiring within 48h.
+    let isUrgent: Bool
+    /// Resting label shown above the track for the nearest credit only.
+    let restingLabel: String?
+    /// Exact values revealed on hover for every credit (countdown · absolute date).
+    let hoverText: String
+}
+
+extension CodexResetCreditsPresentation {
+    private static let runwayUrgencyWindow: TimeInterval = 48 * 3600
+
+    /// Places each dated, still-valid credit on a data-derived horizon (the longest credit
+    /// lifetime in the inventory), so a credit's position drifts left predictably as it ages.
+    /// Credits with no expiry stay in the count and tooltip but off the track.
+    static func runwayMarkers(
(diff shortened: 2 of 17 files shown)


steipete:
Thanks for the thoughtful implementation and especially the attention to highlight styling, accessibility, and focused tests.

I am closing this as superseded by #1902 (`c5cf41c94411bd783aaf0ebf16320eb65eaf63f6`). That landed the product direction we want: one compact expiry summary in the menu, with the complete sorted inventory retained in tooltip and accessibility text. The optional runway would add a 17-file persistent Display setting for a low-frequency row immediately after we simplified Settings, so the extra configuration and rendering surface is not justified here.

The current CI failures are not the reason for closing; the core issue is product/scope fit. Thank you for exploring the alternative so thoroughly.

cleanerzkp:
Status update on the two failing macOS shards from the first run (both now understood):

- **`swift-test-macos (1, 4)`** — real bug on my side: the new `credit_expiry_style_*` picker strings were added to English + the strict locales (ar/fa/th) but not the fully-maintained catalogs, so `LocalizationLanguageCatalogTests` (Galician/Turkish/Ukrainian/Korean exact-key-match) failed. **Fixed in `57dd27c0`** — the keys are now in all 23 catalogs with translations; `LocalizationLanguageCatalogTests` (22/22) and `make check` (23 locales) pass locally.
- **`swift-test-macos (3, 4)`** — CI infra flake, not the code: the job failed at 40s in `swift test list` with `fatal: cannot change to .../org.swift.swiftpm/repositories/...: No such file or directory` for swift-crypto/sparkle/swift-log/etc. — a corrupted SwiftPM cache on that runner. A fresh run should clear it.

Also folded in a small `/simplify` pass (route the runway horizon/hover labels through the existing `compactExpiryText`).

On your review questions: product direction is **opt-in by default** (picker defaults to the current list; the runway changes nothing unless selected) — happy to make it replacement or drop the gate, your call. The new run needs a maintainer approval to start; `@clawsweeper re-review` once it's green. Built-app light/dark screenshots to follow.

cleanerzkp:
CleanShot 2026-07-05 at 11 05 46@2x
[*]

PR #1915, opened 2026-07-05
Reject non-finite Kimi K2 usage values

joeVenner:
## Summary

- reject non-finite Kimi K2 credit and average-token values from JSON or headers
- keep missing/malformed values on the existing zero-or-nil fallback path
- allow an injected HTTP transport for deterministic production-path validation
- cover NaN, infinity, and overflow numeric strings

## Production-path proof

The regression test calls the public fetcher with an in-memory HTTP response containing non-finite JSON values and a non-finite `X-Credits-Remaining` fallback. It verifies the request authorization header and the resulting zero-or-nil fallback. No live credentials are used.

## Testing

- `swift test --disable-sandbox --filter KimiK2UsageFetcherTests` (7 passed)
- `make test` (all 48 shards passed)
- `make check` (SwiftFormat clean; SwiftLint 0 violations)
- autoreview (clean, 0.98 confidence)
- `git diff --check`
- exact head: `909b415a751eb7e34af141a8a8db80e989442219`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Fixed
 - Codex menu: hide error-only optional Credits and OpenAI web setup diagnostics while keeping them visible in provider Settings.
 - Codex cost history: reuse cached aggregate pricing and one pricing catalog across daily and project reports, carry fresh cache state across launches, and treat unpriced models as migrated, avoiding repeated row scans, filesystem work, and duplicate background scans on large local histories.
+- Kimi K2: reject non-finite credit and token values before they reach menus, CLI output, or widgets. Thanks @joeVenner!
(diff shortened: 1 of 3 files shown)


steipete:
Landed as `b00797537dbffe0e0bd469544003fd5769b75542`. Thanks @joeVenner!

Maintainer verification:
- `swift test --disable-sandbox --filter KimiK2UsageFetcherTests` — 7 passed
- `make test` — all 48 shards passed
- `make check` — SwiftFormat and SwiftLint clean
- autoreview — no actionable findings; patch-correct confidence 0.98
- `git diff --check` — clean
- exact-head CI for `909b415a751eb7e34af141a8a8db80e989442219` — all 10 checks passed: https://github.com/steipete/CodexBar/actions/runs/28777491318

The regression uses an in-memory transport through the public fetch path; no live credentials or Keychain access were used.

PR #1946, opened 2026-07-06
Show <1% for all sub-1% usage, not just under 0.5%

devYRPauli:
## Summary

- render every positive sub-1% usage value as `<1%` instead of rounding values above 0.5% to `1%`
- preserve zero, exact-one, and capped-at-100 behavior
- preserve Traditional Chinese suffix-first ordering and lock it with regression coverage
- add changelog credit for @devYRPauli

`percentText` previously formatted with `%.0f` before trying to rewrite `0%`. Values in `[0.5%, 1%)` therefore rounded to `1%` before the rewrite could apply. The fix handles the sub-1 range before integer formatting, matching `percentString`.

## Behavior proof

Before on `main`:

```text
0.6% used  -> 1% used
0.75% used -> 1% used
0.99% used -> 1% used
0.75% left -> 1% left
```

After on this branch:

```text
0.6% used  -> <1% used
0.75% used -> <1% used
0.99% used -> <1% used
0.75% left -> <1% left
```

A source-blind executable linked against the public `CodexBarCore` API passed 9/9 probes covering zero, one, the 100% cap, values below and above the old rounding boundary, used/left output, and suffix-first localization.

## Testing

- `swift test --disable-sandbox --filter UsageFormatterTests` (42 passed)
- source-blind public formatter contract (9/9 passed)
- `make check` (SwiftFormat clean; SwiftLint 0 violations)
- `make test` (48/48 shards passed)
- autoreview clean (0.98 confidence)
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Fixed
 - Amp: open the current Usage page from the menu dashboard action. Thanks @3kh0!
 - Settings: keep Language, Default Terminal, and Refresh cadence selectors interactive on macOS 27.
+- Usage formatting: show every positive sub-1% value as `<1%` instead of rounding values above 0.5% up to `1%`. Thanks @devYRPauli!
 - Codex menu: hide error-only optional Credits and OpenAI web setup diagnostics while keeping them visible in provider Settings.
 - Codex quotas: show the session quota as unavailable while an exhausted weekly limit is still binding, including menu-bar icons and widgets. Thanks @Yuxin-Qiao!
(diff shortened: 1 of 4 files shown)


steipete:
Maintainer verification on exact head `5e8373e06a0ccaad0f3180af876d2d57506e6f94`:

- Rebased onto current `main`; original contributor authorship preserved, with maintainer changelog/test follow-up and `@devYRPauli` credit.
- Focused formatter suite: 42 passed.
- Source-blind public formatter contract: 9/9 cases passed, including zero, exact one, cap, positive sub-1% used/remaining values, and suffix-first localization.
- `make check`: passed.
- Exact-head hosted checks: Linux x64 (7m15s), Linux arm64 (5m53s), lint, change detection, and GitGuardian all passed.

The added regression coverage also locks the traditional Chinese suffix-first localization so the new `<1%` contract cannot silently regress there.

PR #1947, opened 2026-07-06
Roll Claude reset date forward across the year boundary

devYRPauli:
## Summary

- Roll yearless Claude date-and-time quota resets into the next year when their current-year interpretation is already in the past.
- Apply the behavior to both minute-bearing and hour-only reset formats, matching the existing Codex parser contract.
- Add focused year-boundary coverage and changelog credit for @devYRPauli.

## User-visible effect

On December 31, a Claude reset such as `Jan 2, 3:15am` now remains an upcoming reset. Previously it parsed as January of the current year, causing countdown surfaces to show `now` and active-window checks to treat the quota as expired.

## Verification

- Focused status-probe selection: 253 passed.
- Source-blind public API contract: 5/5 passed for both changed formats, same-year preservation, downstream countdown text, and invalid input.
- `make check`: passed.
- Autoreview: clean, 0.99 confidence.
- Public Model Identifier Gate: PASS; no model identifiers changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Fixed
 - Amp: open the current Usage page from the menu dashboard action. Thanks @3kh0!
+- Claude: keep yearless reset dates in the upcoming year when a quota crosses New Year's Day. Thanks @devYRPauli!
 - Settings: keep Language, Default Terminal, and Refresh cadence selectors interactive on macOS 27.
 - Usage formatting: show every positive sub-1% value as `<1%` instead of rounding values above 0.5% up to `1%`. Thanks @devYRPauli!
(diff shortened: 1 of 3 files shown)


steipete:
Maintainer verification on exact head `85183acd3821cdf43268814b9f2b0b19f335842f`:

- Expanded the regression to cover both changed parser branches: minute-bearing and hour-only yearless reset dates.
- Focused status-probe selection: 253 passed.
- Source-blind public API contract: 5/5 passed, covering both changed formats, same-year preservation, downstream countdown text, and invalid input.
- `make check`: passed.
- Clean autoreview: 0.99 confidence, no actionable findings.
- Exact-head hosted checks: Linux x64 (7m15s), Linux arm64 (6m10s), lint, change detection, and GitGuardian all passed.

The black-box validation confirms the reset remains future-facing and yields `in 1d 4h` instead of the stale `now` presentation.

PR #1948, opened 2026-07-06
Keep fractional Claude web utilization instead of dropping it to 0%

devYRPauli:
## Summary

- Preserve fractional utilization from Claude web's five-hour and seven-day quota windows.
- Reuse the established fractional parser already used by adjacent Claude quota lanes.
- Verify both decoded values plus the downstream session window's used and remaining percentages.
- Add changelog credit for @devYRPauli.

## User-visible effect

Claude web values such as `45.5` no longer fall through an integer-only cast and appear as a false `0% used` session or an unavailable weekly quota.

## Verification

- Focused placeholder/window suite: 8/8 passed.
- Source-blind compiled parser contract: 6/6 passed, including fractional and integer values, reset preservation, live-session classification, and the null-session fallback.
- `make check`: passed.
- Autoreview: clean, 0.98 confidence.
- Public Model Identifier Gate: PASS; no model identifiers changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Fixed
 - Amp: open the current Usage page from the menu dashboard action. Thanks @3kh0!
 - Claude: keep yearless reset dates in the upcoming year when a quota crosses New Year's Day. Thanks @devYRPauli!
+- Claude web: preserve fractional session and weekly utilization instead of displaying it as zero or unavailable. Thanks @devYRPauli!
 - Settings: keep Language, Default Terminal, and Refresh cadence selectors interactive on macOS 27.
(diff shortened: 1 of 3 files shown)


steipete:
Maintainer verification on exact head `511d0e33fa2c42e9d522d5987d275864e3e7847b`:

- Strengthened the regression to verify fractional session and weekly values plus downstream session used/remaining percentages.
- Focused window/placeholder suite: 8/8 passed.
- Source-blind compiled parser contract: 6/6 passed, covering fractional and integer values, reset preservation, live-session classification, and the null-session fallback.
- `make check`: passed.
- Exact-head hosted checks: Linux x64 (7m34s), Linux arm64 (6m24s), lint, change detection, and GitGuardian all passed.

The parser now keeps decimal utilization exact instead of turning a live session into `0% used` or dropping the weekly lane.

PR #1950, opened 2026-07-06
Do not fabricate a 0% Gemini primary window when there is no Pro tier

devYRPauli:
## Summary

- Keep Gemini's primary Pro window absent when the account reports no Pro quota.
- Let existing metric selection fall back to the real Flash window instead of a fabricated `0% used` Pro window.
- Cover Flash-only and Pro-present snapshot mapping plus automatic, explicit-primary, and average menu-bar selection.
- Add changelog credit for @devYRPauli.

## User-visible effect

A Flash-only account that is 95% used now shows that real usage in the menu bar. It no longer shows a phantom `0% used` primary lane.

## Verification

- Focused snapshot/resolver suites: 25/25 passed.
- Source-blind public parser contract: 5/5 passed for Flash-only and Pro-present stats tables.
- `make check`: passed.
- Autoreview: clean, 0.98 confidence.
- Public Model Identifier Gate: PASS; all test identifiers are established released Gemini 2.5 identifiers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Amp: open the current Usage page from the menu dashboard action. Thanks @3kh0!
 - Claude: keep yearless reset dates in the upcoming year when a quota crosses New Year's Day. Thanks @devYRPauli!
 - Claude web: preserve fractional session and weekly utilization instead of displaying it as zero or unavailable. Thanks @devYRPauli!
+- Gemini: use the real Flash quota in menu-bar metrics when an account has no Pro quota. Thanks @devYRPauli!
 - Settings: keep Language, Default Terminal, and Refresh cadence selectors interactive on macOS 27.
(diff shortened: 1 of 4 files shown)


steipete:
Maintainer verification on exact head `8fa4ed8347b5bc3f5383ea02966ab63c219e1a9f`:

- Added resolver-level regression coverage so automatic, explicit-primary, and average Gemini metrics all fall back to Flash when Pro is absent.
- Focused snapshot/resolver suites: 25/25 passed.
- Source-blind public parser contract: 5/5 passed for Flash-only and Pro-present stats tables.
- `make check`: passed.
- Exact-head hosted checks: Linux x64 (7m13s), Linux arm64 (6m3s), lint, change detection, and GitGuardian all passed.
- Hosted macOS capacity remains incomplete. Landing under the documented capacity-only policy because the exact candidate has full local 48-shard proof and both Linux jobs are green.
- Public Model Identifier Gate: PASS. All model identifiers in the tests are established released Gemini 2.5 identifiers.

This locks the actual menu-bar result: a Flash-only account at 95% used can no longer be hidden by a fabricated `0% used` Pro lane.

PR #1966, opened 2026-07-07
Add Wayfinder Provider (Local vs Hosted Router)

tcballard:
## Summary

Adds [Wayfinder](https://github.com/itsthelore/wayfinder-router) as a CodexBar provider.

Wayfinder is a self-hosted, deterministic local/cloud LLM router — its gateway runs on the user's machine and scores each prompt offline (no model call) to route it to a cheap/local or dearer/cloud tier. This provider polls the gateway's read-only JSON API and surfaces gateway health, the local/cloud routing split, savings versus always-cloud, and average routing decision latency.

## Why

Users running Wayfinder currently have no way to see its routing activity from the menu bar alongside their other AI tool usage — they'd have to open Wayfinder's own dashboard separately. CodexBar already aggregates 50+ providers into one place; Wayfinder fits the same pattern as the existing ClawRouter integration (a router with a budget/usage split), so adding it here means one less place users have to check.

## What Changed

- Added the Wayfinder provider: descriptor, fetcher, and settings reader in `Sources/CodexBarCore/Providers/Wayfinder/`, following `docs/provider.md`.
- Fetcher polls exactly four read-only gateway endpoints — `/healthz`, `/router/models`,`/v1/savings?period=30d`, `/metrics` — never the chat endpoints, and never reads or transmits prompt text.
- No credentials required: the gateway's read endpoints are unauthenticated on loopback, so the only setting is an optional Gateway URL (defaults to `http://127.0.0.1:8088`), validated with the same loopback-HTTP policy already used by the CrossModel provider.
- Added app-side settings field, menu card lines, CLI render lines, and a template icon (`Sources/CodexBar/Providers/Wayfinder/`, `Sources/CodexBar/Resources/ProviderIcon-wayfinder.svg`).
- Wired the new provider through the existing registration points: `UsageProvider`/`IconStyle` enum cases, both provider registries, a new `UsageSnapshot.wayfinderUsage` field, and `ProviderConfigEnvironment` base-URL projection. All existing exhaustive switches over `UsageProvider` (widget colors/names, debug-log messages, cost validation) were extended for the new case — the compiler enforces this, so nothing else can silently break.
- Added `docs/wayfinder.md` and a README/`docs/providers.md` entry.
- Added tests: `TestsLinux/WayfinderProviderLinuxTests.swift` (16 tests) and a registry resolution test in `Tests/CodexBarTests/WayfinderProviderTests.swift`.
- The provider is opt-in (`defaultEnabled: false`) — zero effect on existing behavior unless a user explicitly enables it.

## Proof

Ran the actual `wayfinder-router` gateway locally (two stand-in OpenAI-compatible upstreams for a local and cloud tier, priced config) and sent it real prompts — easy ones stayed on the local tier, hard reasoning/math/constraint-heavy prompts escalated to cloud under Wayfinder's lexical routing. 14 requests total: 10 local / 4 cloud, 61.5% saved. Fetched through the real provider code via the CLI:

**Before:** no Wayfinder provider exists in CodexBar; a user running the gateway has to check its own dashboard separately.

**After:**

```
$ codexbar usage --provider wayfinder --source api
== Wayfinder (api) ==
Gateway: ok · 2 models
Routed: local 10 · cloud 4
Saved: <$0.01 · 61.5% vs always-cloud
Avg decision: 0.1 ms
Plan: Local Gateway
```

Failure path (gateway not running), showing the error is actionable:

```
$ WAYFINDER_GATEWAY_URL=http://127.0.0.1:9999 codexbar usage --provider wayfinder --source api
Error: Could not reach the Wayfinder gateway. Start it with `wayfinder-router serve`
(default http://127.0.0.1:8088) or fix the Gateway URL in Settings.
```

*Menu card screenshot: to be added — see Risk section.*

## Validation

**Automated tests:**
- `swift test --parallel` — full suite green: 107 tests across 18 suites, including 16 new Wayfinder tests plus a registry-resolution test. The Wayfinder tests use the live gateway responses above as fixtures (not hand-written JSON): full/empty/degraded/unpriced response variants, `UsageSnapshot` mapping, Prometheus metrics parsing (missing/garbage/labeled lines), URL normalization, loopback-HTTP endpoint policy (accepts `127.0.0.1`/`localhost`, rejects other plain-HTTP hosts, accepts any HTTPS), config→environment projection, Codable round-trip, an endpoint-set pinning test (proves the fetcher only ever calls the 4 documented paths), HTTP error mapping, same-origin/redirect rejection, and CLI text rendering.
- `swiftformat Sources Tests --lint` (0.61.1, matching this repo's pinned CI version) — clean, 0 files need formatting.
- `swift build -c release --product CodexBarCLI` — clean.
- Portable lint checks from `Scripts/lint.sh` (generated-hash check, package/product paths, repository size, documentation links) — all pass.

**Manual testing:** exercised the provider end-to-end against a real, locally running Wayfinder gateway (not mocked) via the CLI, on Linux, including both the healthy path and the gateway-unreachable failure path shown above.

**Not yet verified:**
- `swiftlint --strict` — could not be run as CI would in the environment this was developed in (no outbound access to fetch the pinned release binary, and a from-source build of the same version crashed on any input there — confirmed unrelated to this diff since it crashed identically on pre-existing files too). The diff was reviewed by hand against every rule in
  `.swiftlint.yml` with no findings, but this repo's CI run will be the first real confirmation.
- The macOS app target (`CodexBar`) itself — only the cross-platform `CodexBarCore` and `CodexBarCLI` targets were built and tested, on Linux. Every macOS-only exhaustive switch was extended for the new case, but hasn't been compiled/run on macOS yet.

## Risk

- **Auth/credentials:** none — the gateway's read-only endpoints are unauthenticated on loopback, so this provider introduces no new credential handling.
- **Provider routing:** none — this only reads Wayfinder's own gateway; it never calls `/v1/chat/completions` or anything that could route a prompt.
- **Persistence/migration:** none — one new optional field on `UsageSnapshot` (`wayfinderUsage`), additive and `Codable`-compatible; no schema migration needed.
- **Settings:** adds one new optional config field scoped only to the `.wayfinder` provider (Gateway URL); no existing provider's settings are touched.
- **Existing workflows/UI:** additive only — new enum case, new registry entries, new renders lines gated on `snapshot.wayfinderUsage != nil`. Every existing exhaustive switch was extended (compiler-enforced), so no other provider's behavior changes. The provider defaults to disabled.
- **Known limitations:** no macOS screenshot yet, and `swiftlint --strict` hasn't had a real run (see Validation). Opening as a draft for that reason — will flip to ready once verified on macOS.

## Notes for Review

The three files worth actually reading:
- `Sources/CodexBarCore/Providers/Wayfinder/WayfinderUsageFetcher.swift` — the core fetch/parse logic and the `UsageSnapshot` mapping.
- `Sources/CodexBarCore/Providers/Wayfinder/WayfinderSettingsReader.swift` — the loopback-HTTP validation policy (mirrors CrossModel's).
- `TestsLinux/WayfinderProviderLinuxTests.swift` — the live-gateway-fixture tests.

Everything else in the diff is mechanical wiring into existing registration points (enum cases, registries, exhaustive switches) — shouldn't need line-by-line attention.


tcballard:
Pushed a fix for the route-summary assumption (commit 4cbf4ab).

**Root cause:** `routedSummary` read `models.models.first?.name` and treated array position as
"the local tier." `/router/models` has no field asserting that — route names are just whatever the user named their own endpoints.

**Fix:** removed the local/cloud guess entirely. `routedSummary` now renders the real per-route breakdown by name (e.g. `groq-8b: 10 · openai-o1: 4`), matching the existing ClawRouter pattern
in this repo instead of swapping in a different fragile heuristic (endpoint-host sniffing has
the same problem — a local model isn't guaranteed to be on loopback).

**Tests added** (`TestsLinux/WayfinderProviderLinuxTests.swift`): heavier-traffic route
configured second (no positional reliance), and arbitrary route names asserting the summary
never contains "local"/"cloud". Docs updated to match. 109/109 tests green, formatting clean.

**Still open** (unchanged by this push, already noted in the PR body): no macOS screenshot,
and `swiftlint --strict` hasn't had a real run in my dev environment (sandbox network restriction; a from-source build crashes on any input there, confirmed unrelated to this diff).

This push should trigger CI on macos-15-intel, which will give a real answer on both.

tcballard:
@clawsweeper re-review please

steipete:
Landed the reviewed and hardened implementation in #2006, merge commit `c2d0cd1a1ff7ce90f8a4e4dabdb615af94efe10c`.

Thank you, @tcballard, for the provider implementation. Your commits and co-author credit were preserved, and the changelog credits you. I used a maintainer branch because this PR did not allow maintainer edits; enabling **Allow edits by maintainers** makes future fixups possible directly on the contributor PR.

Closing this superseded PR in favor of the landed implementation.

tcballard:
First time contributor, thank you!

PR #2002, opened 2026-07-09
feat(antigravity): detect CLI listening ports via /proc when lsof is unavailable

junmo-kim:
## Problem
On minimal Linux hosts without `lsof` installed, Antigravity CLI usage fails with `Antigravity language server not detected`. `AntigravityStatusProbe.listeningPorts` resolves the spawned `agy` language server's ports only through `lsof`, so when it is absent the lookup throws even though `agy` starts and begins listening normally. This affects both the fresh-spawn and warm-reuse paths, since both go through `listeningPorts`.

## Solution
Keep `lsof` as the primary resolver (the common denominator across macOS and Linux) and add a Linux-only fallback when it is missing: collect the process's socket inodes from `/proc/<pid>/fd` and match them against `/proc/net/tcp{,6}` LISTEN entries. This mirrors the existing `/proc/<pid>/cwd` fallback in `LocalAgentSessionScanner.cwdByPID`. macOS behavior is unchanged.

The `/proc/net/tcp` parsing lives in a small platform-independent `ProcNetTCPListeningPortParser` so it can be unit tested on any host, matching the existing `LSOFCWDOutputParser` pattern. Port detection is kept in a dedicated `AntigravityStatusProbe+PortDetection.swift` (following the existing `+Deadline`/`+LocalEndpoints` extension-file layout) so the probe file and type stay within the SwiftLint `file_length`/`type_body_length` limits.

## Tests
Added `ProcNetTCPListeningPortParserLinuxTests` covering owned listening ports, foreign inodes, non-listening sockets, and fd symlink inode parsing.

## Verification (Linux without `lsof`)
Built this branch (`swift build -c release --product CodexBarCLI --static-swift-stdlib`) and ran it on a host with no `lsof`:

```console
$ . /etc/os-release && echo "$PRETTY_NAME $(uname -m)"
Debian GNU/Linux 13 (trixie) x86_64
$ command -v lsof || echo "lsof: not installed"
lsof: not installed

$ ./CodexBarCLI usage --provider antigravity --source cli --format json --log-level debug
# debug: the /proc fallback recovers the spawned agy's listening ports without lsof
… com.steipete.codexbar.antigravity: … pid=1332864 ports=45559,46517 …
[{"source":"cli","provider":"antigravity","usage":{
  "identity":{"accountEmail":"<redacted>","loginMethod":"Google AI Pro","providerID":"antigravity"},
  "primary":{"usedPercent":48.26,"windowMinutes":10080,"resetsAt":"2026-07-10T18:38:24Z"},
  "secondary":{"usedPercent":33.57,"windowMinutes":10080,"resetsAt":"2026-07-12T14:50:24Z"},
  "extraRateWindows":[{"title":"Gemini weekly", …},{"title":"Claude/GPT weekly", …}]
}}]
```

Without this change the same invocation fails with `Antigravity language server not detected`.


steipete:
Maintainer proof for exact head `ed9a3016a0dd1386c59da2304a4d20f41f0816d8`:

- Deep review found the original fallback read the caller network namespace. The final patch reads `/proc/<pid>/net/tcp{,6}` for the target Antigravity process, validates port ranges, and keeps the existing `lsof` path primary.
- Focused port-detection suite: 8 tests pass, including IPv4/IPv6, headerless/malformed tables, non-listening sockets, fd symlinks, deduplication, and a conflicting caller-vs-target namespace fixture.
- Local exact-head gates: `make check` passes; all 49 `make test` shards pass; autoreview is clean.
- Authenticated live evidence: the contributor reproduced successful Antigravity discovery on Debian 13 without `lsof`; exact-head Linux x64 and arm64 builds are green.
- CI: changes, lint, both Linux jobs, and GitGuardian are green. Hosted macOS shards have no failures and remain capacity-limited; landing follows the repository's documented capacity-only policy backed by the exact local full suite.
- Public Model Identifier Gate: PASS. Dependency freshness: unchanged; no dependency added or updated.

Thanks for the solid Linux fallback, @junmo-kim!

PR #2009, opened 2026-07-09
Improve provider quota warning settings

Zihao-Qi:
## Summary
- Redesign provider quota warning rows around a compact per-window `Global | Custom | Off` control with inline `Warning` and `Critical` threshold fields.
- Route provider threshold edits through the dirty draft save path so no-op blur, Return, outside clicks, or teardown do not create or restore explicit provider overrides.
- Preserve extra stored thresholds and show them in inherited summaries, while localizing the new `Global` picker label across app catalogs.

## Why
The provider quota warning settings needed the same hardening as the global quota warning layout: the old editor could churn provider refreshes on no-op lifecycle events and risk freezing inherited thresholds as explicit overrides. The new layout also keeps the rows aligned, makes inherited/custom/off state clearer, and avoids hiding historical/manual threshold values beyond the two fields exposed in the UI.

## Validation
- `swift test --filter SettingsStoreTests`
- `swift test --filter PreferencesPaneSmokeTests`
- `swift test --filter LocalizationLanguageCatalogTests`
- `git diff --check`
- `make check`

## Screenshot
before
Image 7-7-26 at 22 35
after
Screenshot 2026-07-09 at 10 59 32 AM

### DIFF
--- a/Sources/CodexBar/QuotaWarningSettingsViews.swift
+++ b/Sources/CodexBar/QuotaWarningSettingsViews.swift
@@ -86,7 +86,7 @@ struct ProviderQuotaWarningSettingsView: View {
                 .frame(minHeight: Self.windowRowMinHeight, alignment: .center)
                 .gridColumnAlignment(.leading)
 
-            Picker("", selection: self.overrideModeBinding(for: window)) {
+            Picker(window.localizedCapitalizedDisplayName, selection: self.overrideModeBinding(for: window)) {
                 Text(L("quota_warning_global")).tag(ProviderQuotaWarningOverrideMode.global)
                 Text(L("Custom")).tag(ProviderQuotaWarningOverrideMode.custom)
                 Text(L("quota_warning_off")).tag(ProviderQuotaWarningOverrideMode.off)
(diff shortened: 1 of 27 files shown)


steipete:
Maintainer verification on exact head `fa383a4e393e6531659ee65e3e94919ba3c32c06`:

- Focused settings/localization suite: 119 tests passed; the final `PreferencesPaneSmokeTests` rerun passed 19/19.
- Full `make test`: all 50 shards passed.
- `make check`: SwiftFormat clean, SwiftLint 0 violations, repository/docs/localization gates passed.
- Autoreview: clean, patch correct at 0.84 confidence.
- Signed packaged app: Developer ID `Y5PE65HELJ`; deep codesign and Gatekeeper accepted. App SHA-256 `707dde648a6d7fc9e1e8f858a1c1f5feb28a05cb8e1679713503772ebfacc26e`; bundled CLI SHA-256 `0aa58bf39ab3c08fb79245676e1661e32569b1cc650c55ee9cb1e9c56955ec74`.
- Live Settings proof in the exact packaged app: Global rendered inherited 50/20 thresholds; Custom exposed inline Warning/Critical editors; editing Warning to 73, switching to Off, then back to Custom restored 73/20. The original user config was restored byte-for-byte after QA.
- Added regression coverage for Off → Custom threshold restoration and Global teardown not committing a stale editor.

Public model identifier gate: PASS (model-free diff). Dependency freshness: no dependency changes.

PR #2026, opened 2026-07-10
ci: batch macOS test selections

Yuxin-Qiao:
## Summary

- Batch macOS Swift test selections in groups of four while preserving the existing four-way hosted shard split.
- Build test targets once during `swift test list`, then run each selected group with `--skip-build` and `--no-parallel`.
- Keep the historical local retry for non-timeout failures, but make hosted CI fail immediately so a real test failure cannot retry green.
- Isolate timed-out batches one selection at a time and report accurate discovery, selected-selection, first-pass, retry, recovery, timeout, and timing metrics.

## Correctness contract

- Selection counts are computed after sharding and `--limit-groups`.
- Hosted non-timeout failures preserve their first exit code without a retry.
- Singleton timeouts report one timed-out group and exit 124.
- Batched timeouts retry individual selections and report recovery only when every isolated selection passes.
- A local non-timeout retry that then times out falls through to the same isolated-selection path.
- Operator summaries use one-based shard labels consistently.

## Observed performance

The contributor's first hosted group-size-four run passed all checks. Its macOS Swift Test steps took approximately 13m22s to 18m17s despite runner queueing. The final candidate also removes repeated SwiftPM build planning after discovery; exact-head hosted timing is tracked by the per-shard summaries added here.

## Exact-head proof

Head: `85a0924e7a7851e0dbcfed6ded9f9ee75187f906`

- `DEVELOPER_DIR=/Applications/Xcode-26.6.app/Contents/Developer make check`
- `DEVELOPER_DIR=/Applications/Xcode-26.6.app/Contents/Developer make test`: 594/594 selections, 50/50 groups, zero first-pass failures, retries, recoveries, or timeouts
- Real strict group-size-four run: 4/4 selected selections passed through `--skip-build`
- `bash Scripts/test_swift_test_sharding.sh`: strict hosted failure, local retry, shard completeness/counts, singleton timeout, timeout recovery, retry-to-timeout isolation, skip-build, and discovery-timing cases passed
- Final branch autoreview: clean, no accepted/actionable findings
- Public Model Identifier Gate: PASS; no model-bearing public data in this CI-only diff
- Dependency freshness: N/A; no dependencies changed

Exact-head hosted CI/security is running.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,6 +128,7 @@ jobs:
         run: |
           CODEXBAR_TEST_GROUP_SIZE=4 \
             CODEXBAR_TEST_SUITE_TIMEOUT=120 \
+            CODEXBAR_TEST_RETRY_NON_TIMEOUT_FAILURES=0 \
             CODEXBAR_TEST_SHARD_INDEX=${{ matrix.shard-index }} \
             CODEXBAR_TEST_SHARD_COUNT=${{ matrix.shard-count }} \
             ./Scripts/test.sh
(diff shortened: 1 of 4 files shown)


steipete:
Exact-head maintainer proof for `85a0924e7a7851e0dbcfed6ded9f9ee75187f906`:

- Rebased onto current `main`; contributor's later `--skip-build` optimization preserved and reviewed.
- `make check` passed under Xcode 26.6.
- Full local suite passed: 594 selections in all 50 groups, with zero failures, retries, or timeouts.
- A real strict group-size-four invocation passed 4/4 selections using the built test bundle.
- The deterministic harness covers hosted fail-fast behavior, legacy local retry, group-size-four shard completeness/counts, singleton timeout accounting, batch-timeout recovery, retry-then-timeout isolation, discovery timing, and `--skip-build` wiring.
- Final autoreview is clean.
- Public Model Identifier Gate: PASS. Dependency freshness: N/A.

Hosted CI/security is now validating this exact head; merge remains gated on that result.

PR #2048, opened 2026-07-11
Fix stranded menu scroll after a manual refresh clears a provider error

### LINKED ISSUE #2046: Menu stays scrolled after in-menu Refresh clears a provider error; viewport can't return to top until reopened
## Symptom

When a provider (Claude, in my case) is in an error state, the menu card header shows the red multi-line error banner at the top. Clicking the persistent **Refresh** row clears the error and the open menu updates in place — but the tracked menu's viewport is left anchored mid-list. Scrolling never brings the menu back all the way to the top, so the refreshed header (the part you actually want to check after a refresh) stays out of reach. The only recovery is closing the menu and reopening it.

## Steps to reproduce

1. Enough providers/sections enabled that the menu is taller than the screen (AppKit menu scrolling engaged).
2. Get a provider into a failed-refresh state so the red error subtitle renders in the top card (e.g. Claude OAuth usage fetch failing).
3. Open the menu, scroll down to the persistent **Refresh** row, click it.
4. Refresh succeeds; the error banner collapses and content is reconciled in place.
5. Try to scroll back to the top of the menu.

**Expected:** the viewport can return to the top (or re-anchors there after the user-initiated refresh completes).
**Actual:** the menu never scrolls all the way up; close + reopen is required.

## Analysis

The manual-refresh path (`startManualRefresh` → `performStoreRefresh(refreshOpenMenusWhenComplete: true)` → `refreshOpenMenusAfterExplicitStoreAction()` → `rebuildOpenMenuIfStillVisible` → `populateMenu`/smart update) reconciles the tracked menu in place, and nothing re-anchors the scrolled viewport afterwards. When the reconcile changes row heights (a recovered error banner collapsing is the biggest single jump: the reserved up-to-4-line error layout in `UsageMenuCardHeaderView` disappears on rebuild), AppKit's private menu scrolling is left with a stale offset.

I probed the view-based menu implementation on macOS 26.2 with a standalone repro (tall menu of custom-view rows, mutated while tracking):

- A tracked menu is `NSPopupMenuWindow` → container → `NSMenuScrollView` (a real `NSScrollView`) → `NSClipView` → `NSTableViewBackedMenuRepresentation`, plus two arrow `NSImageView` strips.
- Row views are **recycled**: once an item scrolls offscreen, `item.view` is detached from the window (`enclosingScrollView` goes nil), so any viewport repair has to resolve the scroll view through a currently-attached row.
- The clip view's offset is reachable and correctable with public API (`view.enclosingScrollView`, `NSClipView.scroll(to:)`, `reflectScrolledClipView`) — no private symbols needed.

## Proposed fix

After a **user-initiated** manual refresh completes and the open-menu rebuild lands, re-anchor the tracked menu's viewport at the top (no-op when the menu isn't scrolled or isn't oversized). Background/scheduled refreshes stay untouched so in-place data ticks never yank the viewport while someone is reading the bottom of the menu.

This both repairs the stale scroll state and matches the intent of clicking Refresh on an error: you want to see the recovered header. Happy to send a PR — have it working locally.

Seen on CodexBar 0.41.0, macOS 26.2.

ss251:
Fixes #2046.

## Summary

An explicit refresh can collapse a multi-line provider error while an oversized AppKit menu is being tracked, leaving the viewport stranded below the provider header. This change:

- Arms a one-shot restore only for persistent Refresh/Command-R and scopes it to the initiating menu, tracking session, interaction generation, provider/account selection, and refresh generation.
- Waits for clip/document geometry to settle through the existing event-tracking-aware scheduler, distinguishing AppKit layout corrections from deliberate viewport movement.
- Cancels restoration after scrolling, highlighting or selection changes, child or hosted submenu activity, provider/account switches, close/reopen, cancellation, shutdown, or replacement by a newer refresh.
- Leaves background refreshes untouched and restores observer state during teardown.

The implementation integrates the manual-refresh, menu-session, switcher/submenu, interaction, and shutdown paths. It adds 39 focused viewport tests, adjacent lifecycle coverage, a deterministic root-menu assertion for asynchronous submenu cleanup, and a changelog entry thanking @ss251.

## Verification

- `make check`
- `swift test --filter 'StatusMenu(ViewportRestore|OpenRefresh|PersistentRefresh|SwitcherTracking)Tests'` — 114 tests across 4 suites passed.
- `swift test --filter StatusMenuViewportRestoreTests` — 39 tests passed.
- Hosted-submenu cleanup regression — 30/30 isolated repetitions passed.
- `CODEXBAR_TEST_RETRY_NON_TIMEOUT_FAILURES=0 make test` — 602 selections; all 51 groups passed first attempt with no retries, timeouts, or failures.
- Isolated release build passed.
- Autoreview reported no actionable findings; the final test-only race hardening was independently reviewed.
- A signed debug bundle from the exact candidate passed strict deep signature verification and Gatekeeper assessment. In an isolated deterministic failure-to-success run, one Command-R produced one provider invocation; the same native popup remained open, the error cleared, fresh-update state appeared, and the provider header returned into view without close/reopen.

No authenticated-provider behavior is claimed; live proof used deterministic local responses with credential access disabled. No dependencies changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Claude: prevent CodexBar's passive CLI probes from starting background Claude Code updates, avoiding repeated partial downloads when a probe exits before an update completes. Thanks @PG2047!
 - Codex cost history: bound malformed session-metadata lines and release read chunks promptly, preventing metadata pre-scans from retaining memory in proportion to oversized JSONL records. Thanks @Yuxin-Qiao!
 - Widgets: add Cursor to configurable and switcher widgets with accurate legacy Requests and current Total, Auto, and API quota labels (#2040). Thanks @Zihao-Qi!
+- Menus: return oversized tracked menus to the provider header after a manual refresh without moving background updates, highlighted rows, open submenus, or newer menu/provider sessions (#2046). Thanks @ss251!
 
 ## 0.42.0 — 2026-07-11
 
--- a/Sources/CodexBar/MenuSessionCoordinator.swift
+++ b/Sources/CodexBar/MenuSessionCoordinator.swift
@@ -12,6 +12,7 @@ struct MenuSessionCoordinator<MenuID: Hashable> {
     private(set) var renderedVersions: [MenuID: Int] = [:]
     private(set) var deferredUntilNextOpen: Set<MenuID> = []
     private(set) var parentRebuildsDeferredDuringTracking: Set<MenuID> = []
+    private(set) var pendingViewportRestores: Set<MenuID> = []
 
     @discardableResult
     mutating func invalidate(
@@ -92,16 +93,29 @@ struct MenuSessionCoordinator<MenuID: Hashable> {
         self.parentRebuildsDeferredDuringTracking.contains(menuID)
     }
 
+    /// One-shot viewport restore armed by a completed user-initiated manual refresh and
+    /// consumed when that refresh's open-menu rebuild lands.
+    mutating func armViewportRestore(_ menuID: MenuID) {
+        self.pendingViewportRestores.insert(menuID)
+    }
+
+    @discardableResult
+    mutating func consumeViewportRestore(_ menuID: MenuID) -> Bool {
+        self.pendingViewportRestores.remove(menuID) != nil
+    }
+
     mutating func removeMenu(_ menuID: MenuID) {
         self.renderedVersions.removeValue(forKey: menuID)
         self.deferredUntilNextOpen.remove(menuID)
         self.parentRebuildsDeferredDuringTracking.remove(menuID)
+        self.pendingViewportRestores.remove(menuID)
     }
 
     mutating func clearMenuTracking() {
         self.renderedVersions.removeAll(keepingCapacity: false)
         self.deferredUntilNextOpen.removeAll(keepingCapacity: false)
         self.parentRebuildsDeferredDuringTracking.removeAll(keepingCapacity: false)
+        self.pendingViewportRestores.removeAll(keepingCapacity: false)
     }
 
     #if DEBUG
--- a/Sources/CodexBar/MenuSessionCoordinator.swift
+++ b/Sources/CodexBar/MenuSessionCoordinator.swift
@@ -12,7 +12,8 @@ struct MenuSessionCoordinator<MenuID: Hashable> {
     private(set) var renderedVersions: [MenuID: Int] = [:]
     private(set) var deferredUntilNextOpen: Set<MenuID> = []
     private(set) var parentRebuildsDeferredDuringTracking: Set<MenuID> = []
-    private(set) var pendingViewportRestores: Set<MenuID> = []
+    private var nextViewportRestoreToken = 0
+    private(set) var pendingViewportRestores: [MenuID: Int] = [:]
 
     @discardableResult
     mutating func invalidate(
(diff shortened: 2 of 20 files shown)


ss251:
Thanks — the P1 finding was correct, and rev 2 (`b1f8cbba`) restructures the fix around it.

**What changed:** the pending-intent mechanism is gone. The restore now executes directly from `startManualRefresh`'s completion (one main-actor turn after the unfrozen card layout settles), scoped to open, non-hosted menus the refresh left dirty. Since explicit store actions intentionally defer visible parent rebuilds while the menu is tracking, nothing waits on a rebuild anymore, no state can linger, and an unrelated later rebuild can never jump the viewport. The production diff is now a single call in the manual-refresh completion plus the self-contained restore extension — `StatusItemController+MenuTracking.swift` and `MenuSessionCoordinator.swift` are untouched relative to `main`.

**End-to-end coverage:** `manual refresh completion restores the open dirty menu viewport end to end` drives `performPersistentRefreshAction(in:)` (the entry the persistent Refresh row and ⌘R use) on an open tracked menu and asserts the restore fires after completion **without any parent rebuild landing** — it also asserts the menu is still dirty afterwards, pinning that the deferral behavior from `StatusMenuOpenRefreshTests` is preserved. Negative coverage: a clean menu is not restored, a background rebuild never restores, and a menu closed during an in-flight refresh gets no restore.

Commands run: `make test` (full suite, 0 failed / 0 timed-out groups), `make check` (0 violations), `swift test --filter StatusMenuViewportRestoreTests` (8/8), `swift test --filter StatusMenuOpenRefreshTests` (31/31).

**On the fresh-bundle recording:** attempted and deliberately stopped — this machine is a daily driver with a live CodexBar install, and a debug bundle resolves the real user home for provider detection, cost logs, and Codex auth, so an "isolated" instance surfaces real accounts even under an overridden `$HOME`, and seeding `~/.codexbar/config.json` with synthetic providers would reconfigure the live install too. I'd rather not post a recording built on my real usage data. The PR body has a one-shot repro recipe for a clean user account (debug pane → Error Simulation → tall menu → ⌘R); happy to set up a separate macOS user and attach the video if you want it before merge.

@clawsweeper re-review

ss251:
Good catch again — the cross-menu jump was real. Rev 3 binds the restore to the initiating menu.

**What changed:** `refreshMenuProviderNow` now captures the initiating menu's `ObjectIdentifier` when the manual refresh starts, and completion restores **only that menu** if it is still open, non-hosted, and dirty. The all-open-menus scan is gone, so a menu opened and scrolled while another menu's refresh is in flight can never be yanked by shared invalidation. Refreshes that don't originate from a menu carry no identity and never restore anything.

**Regression coverage for the exact scenario in the finding:** `menu opened during another menus in-flight refresh is never restored` starts a gated refresh from menu A via `performPersistentRefreshAction(in:)`, closes A, opens menu B mid-flight, completes the refresh, and asserts B is dirty from shared invalidation yet never restored. The closed-initiator and clean-initiator negatives are also covered, and the end-to-end restore test still passes through the production entry unchanged.

Commands run: `make test` (full suite, 0 failed / 0 timed-out groups), `make check` (0 violations), `swift test --filter StatusMenuViewportRestoreTests` (9/9), `swift test --filter StatusMenuOpenRefreshTests` (31/31).

On the remaining proof ask: the clean-user fresh-bundle recording is the one item I can't produce from this machine without exposing real account data (details in the PR body); I'll set up a separate macOS user for it if wanted before merge — or a maintainer-side run of the recipe in the body (debug pane → Error Simulation on Claude → tall menu → scroll to bottom → ⌘R) demonstrates the collapse-and-re-anchor in under a minute.

@clawsweeper re-review

ss251:
@clawsweeper re-run

steipete:
Maintainer exact-head proof for `a38674daf0b9547dd280ebfc8e0dddb1f0383e7d` (tree `9816f39a60e2ef4a55f44c11279d2477552d8818`):

- Public Model Identifier Gate: PASS for the introduced diff and this public text.
- Scope: 20 files, +2612/-82. No dependency, credential, permission, network, or build-pipeline changes.
- `make check`: PASS.
- Adjacent focused suites: 114 tests across 4 suites, all PASS.
- Viewport suite: 39/39 PASS.
- Hosted-submenu cleanup regression: 30/30 isolated repetitions PASS; the assertion now waits for the root-menu rebuild it claims instead of racing a separate hosted-submenu callback.
- Full no-retry suite: 602 selections; 51/51 groups passed first attempt, with 0 retries, 0 timeouts, and 0 failures.
- Isolated release build: PASS.
- Autoreview: clean on the product patch; no accepted or actionable findings. The final test-only race hardening had two independent read-only reviews.
- Exact signed debug bundle: embedded commit matched this head; strict deep signature verification and Gatekeeper assessment passed.
- Live behavior: an oversized menu was scrolled until its header was absent. One controlled Command-R produced exactly one provider invocation. After completion, the identical popup-window identity remained on-screen, the error cleared, fresh-update state appeared, and the provider header was visible again without close/reopen. The run used isolated deterministic responses with credential access disabled; it does not claim authenticated-provider behavior.
- Dependency freshness: unchanged; no dependency review required.

@clawsweeper re-review

Issue #1986, opened 2026-07-08
Codex (OpenAI): spurious limit-reset confetti + gauge shows 100% when no reset actually happened


stijnhoste:
### Summary

For the **OpenAI / Codex** provider specifically, CodexBar fires the "limit reset" confetti animation and shows the limit at **100%** (i.e. full / freshly reset) when **no actual reset occurred**. The Codex usage limit had not actually rolled over.

### What happens

- Confetti animation plays as if a Codex usage limit just reset.
- The limit gauge jumps to 100% (full).
- In reality no reset happened — the limit was not actually replenished.

This appears to be a **false/spurious reset detection** on the Codex provider: the app briefly reads the limit as reset (100%) and celebrates, then presumably corrects itself.

### Expected

Confetti + 100% should only appear on a genuine Codex limit reset. A transient/erroneous read should not trigger the celebration or show the limit as full.

### Frequency / impact

- Happened **multiple times in quick succession** just now.
- Reproduced independently by a **colleague** at the same time, so it does not appear to be a single-machine fluke.
- Provider affected: **OpenAI / Codex** (oauth, Pro 20x plan).

### Environment

- CodexBar **0.41.0 (100)**
- macOS **15.1.1 (24B91)**
- Codex provider, `source: oauth`, `loginMethod: pro`, Plan `Pro 20x`

### CLI diagnostics

Current `codexbar` CLI snapshot (captured just now — account email redacted). Note the Codex **Session** reads `100% left` (i.e. `usedPercent: 0` / full) while the **Weekly** window is fully consumed, and a `remoteControl/status/changed` notify line prints on each refresh:

```
[codex notify] remoteControl/status/changed
== Codex (oauth) ==
Session: 100% left [============]
Pace: 5% in reserve | Expected 5% used | Lasts until reset
Resets in 4h 45m
Weekly: 0% left [------------]
Resets in 1d 12h
Credits: 0 left
Limit Reset Credits: 0 available
Account: <redacted>
Plan: Pro 20x
```

Corresponding `--json` for the Codex entry (trimmed to the relevant fields; long `usageBreakdown` history and email removed):

```json
{
  "provider": "codex",
  "source": "oauth",
  "usage": {
    "primary":   { "usedPercent": 0,   "windowMinutes": 300,   "resetsAt": "2026-07-08T10:09:40Z" },
    "secondary": { "usedPercent": 100, "windowMinutes": 10080, "resetsAt": "2026-07-09T18:12:13Z" },
[...]

PR #2056, opened 2026-07-11
fix(codex): require weekly reset boundary before confetti (#2054)

### LINKED ISSUE #2054: Codex 0.42.0: weekly reset confetti still fires on transient zero / same-email multi-account samples
### Summary

CodexBar 0.42.0 can still briefly show a Codex weekly limit as 100% remaining and play reset confetti even though the account did not reset.

This is narrower than #1986 and #1987: v0.42.0 protects Codex **session** reset celebrations from transient zero samples and isolates authenticated HTTP transports, but the **weekly** reset detector still accepts a below-threshold sample without requiring the weekly reset boundary to advance.

### Environment

- CodexBar 0.42.0 (102), official signed release
- Codex OAuth, multiple visible accounts
- Weekly-limit confetti enabled; session-limit confetti disabled

### Observed behavior

1. The currently selected Codex account has not reset and its real weekly remaining value is below 100%.
2. CodexBar briefly renders 100% weekly remaining.
3. Weekly reset confetti plays.
4. A subsequent refresh restores the correct non-100% value.

No emails, account IDs, workspace IDs, auth fingerprints, local paths, or raw credentials are included here.

### Source-level evidence

In `UsageStore+PlanUtilization.swift`, the reset-boundary guard applies only to session samples:

```swift
let resetBoundaryAllowsPost = descriptor.seriesName != .session
    || Self.limitResetBoundaryAdvanced(
        previous: previousState?.resetBoundary,
        current: observation.resetBoundary)
```

For `.weekly`, `resetBoundaryAllowsPost` is therefore always true. Any transition from `usedPercent > 1` to `usedPercent <= 1` can post a weekly reset event even when `resetsAt` has not advanced.

There is a second multi-account edge case: when a concrete token-account object is unavailable, Codex detector ownership can fall back to the Codex plan-utilization identity key, which is email-based. In the affected setup, three distinct OAuth accounts produce only two persisted Codex detector keys because two accounts have the same normalized sign-in email but distinct workspace/account identities. That lets one account seed the above-threshold baseline and another low/transient sample satisfy the crossing condition.

Relevant code:

- [`postLimitResetCelebrationIfNeeded`](https://github.com/steipete/CodexBar/blob/v0.42.0/Sources/CodexBar/UsageStore%2BPlanUtilization.swift#L461-L509)
- [`planUtilizationIdentityAccountKey`](https://github.com/steipete/CodexBar/blob/v0.42.0/Sources/CodexBar/UsageStore%2BPlanUtilization.swift#L798-L839)
- [`limitResetAccountIdentifier`](https://github.com/steipete/CodexBar/blob/v0.42.0/Sources/CodexBar/UsageStore%2BPlanUtilization.swift#L868-L898)

### Expected behavior

- Weekly confetti should require a proven weekly reset boundary advance, just like the new Codex session protection.
- Reset detector state should remain isolated by stable account/workspace/auth identity when multiple accounts share the same email.
- A transient `usedPercent == 0` response should not temporarily render a false full gauge or trigger a celebration.

### Suggested regression coverage

- Codex weekly sample changes from above threshold to zero while `resetsAt` is unchanged: no event.
- Same-email Codex accounts with distinct account/workspace identities: detector baselines do not cross.
- A real weekly reset with an advanced boundary still posts exactly once.

### Related

- #1986 fixed the transient-zero **session** reset path.
- #1987 / #2019 fixed authenticated transport/cache isolation, but not detector-key or weekly-boundary isolation.

Yuxin-Qiao:
## Summary

- Require Codex weekly limit celebrations to advance `resetsAt`, matching the existing session transient-zero guard from #1986.
- Preserve established boundaries and above-threshold baselines across transient, missing, or regressed reset metadata.
- Scope reset-detector state to stable provider, managed-account, profile, workspace, or final-fallback auth identity instead of mutable plan labels.
- Resolve the active visible Codex account for every refresh path, including the default non-stacked layout.

## Improvements in review

- Replaced the large live-proof harness with focused deterministic ownership and boundary tests.
- Preserved a known weekly boundary when an above-threshold sample temporarily omits reset metadata.
- Prefer stable managed-account and profile identity over credential fingerprints so token refreshes retain detector continuity.
- Added coverage for same-email workspace isolation, default refresh routing, auth-only fallback isolation, plan changes, token refreshes, and real resets after intermittent metadata.

## Test plan

- [x] `swift test --filter UsageStorePlanUtilizationTests` (62 tests)
- [x] `make check`
- [x] `make test` (603 selections, 51 groups)
- [x] Autoreview: no accepted/actionable findings

Closes #2054


Yuxin-Qiao:
@clawsweeper re-review

Yuxin-Qiao:
@clawsweeper re-review

Issue #1062, opened 2026-05-20
Codex cost scanner may overcount repeated token_count rows when total_token_usage is unchanged


BarryYangi:
## Summary

CodexBar 0.27.0 appears to overcount Codex local cost usage for some JSONL logs.

The local Codex JSONL can contain repeated `token_count` events where:

- `total_token_usage` is unchanged from the previous event
- `last_token_usage` is still non-zero

If `last_token_usage` is counted as a fresh delta in this case, the scanner appears to double count already-reported usage. Using `total_token_usage` deltas gives a much lower and internally consistent result for the same logs.

This may be related to closed issues #578 and #602, but this report includes a more specific suspected cause.

## Environment

- CodexBar: 0.27.0
- Codex CLI: 0.130.0
- Command: `codexbar cost --provider codex --refresh`
- Source: local Codex JSONL logs
- Account/billing comparison: not using provider billing as the source of truth here; this report is about internal consistency of local log scanning.

## What I observed

For one local Codex session JSONL file, I found:

- final `total_token_usage`: about 204.15M tokens
- sum of `last_token_usage`: about 582.15M tokens
- CodexBar counted about 582.15M tokens for that file
- an independent scanner using `total_token_usage` deltas counted about 204.15M tokens

For privacy, I am not attaching the full session log, but I can share a reduced/sanitized fixture if helpful.

Across my local Codex logs, I also found:

- 230 files containing `token_count`
- 36,898 rows where both `total_token_usage` and `last_token_usage` were present
- 10,668 rows where `total_token_usage` was unchanged from the previous token row while `last_token_usage` was non-zero
- those unchanged-total rows alone contributed about 1.339B extra tokens if `last_token_usage` is summed
- 0 rows where `total_token_usage` decreased/reset in this sample

That makes me suspect `last_token_usage` is sometimes repeated with status/snapshot events, while `total_token_usage` remains the monotonic counter.

## Minimal example

```jsonl
{"timestamp":"2026-05-13T00:00:00Z","type":"event_msg","payload":{"type":"token_count","info":{"last_token_usage":{"input_tokens":100,"cached_input_tokens":90,"output_tokens":10},"total_token_usage":{"input_tokens":100,"cached_input_tokens":90,"output_tokens":10}}}}
[...]

PR #2066, opened 2026-07-11
Contain Ultra interleaved token overcounting (#2037)

### LINKED ISSUE #2037: Ultra-mode Terra and Sol sessions can overcount forked context
CodexBar appears to overstate usage for `gpt-5.6-terra` and especially `gpt-5.6-sol` when they run in Ultra mode.

The pricing is correct: models.dev reports the correct input, cached-input, output, and long-context rates for these models. The inflation seems to be coming from usage accounting. Ultra sessions can fork multiple agents and replay large portions of the parent context while interleaving cumulative token snapshots.

In one local Sol session, the raw log reached roughly 268M cumulative input tokens, while CodexBar attributed 3.29B input tokens and about $4,000 of standard cost. A single forked turn contributed more than 3B tokens across hundreds of rows. The cache marked the file as having divergent totals, which is consistent with multiple cumulative-token lineages being mixed together.

Potential fixes include:

- Keep cumulative-token baselines scoped per task/turn or fork lineage.
- Detect fork/replay boundaries and avoid treating replayed parent context as new usage.
- Prefer `last_token_usage` deltas only when cumulative totals are unavailable or trustworthy; otherwise use cumulative deltas per lineage.
- Add a guardrail that rejects implausibly large deltas relative to the reported per-turn usage.

The goal is to preserve genuine Ultra usage while preventing replayed fork context from being counted repeatedly. This is may be related to #1062 and #1138, but appears to be a distinct current failure involving interleaved Ultra-mode task counters.

Zihao-Qi:
## Summary

- Ultra-mode Terra/Sol sessions can interleave cumulative `total_token_usage` snapshots from multiple lineages in one JSONL file. A single file-global baseline then recounts the high/low gap on every flip — turning ~268M real input tokens into ~3.29B (~$4,000) in a reported session.
- Phase 1 makes inflation **provably bounded**: a never-lower watermark latches interleaved mode on any component drop; post-latch deltas use a dedicated containment helper (preserving #968 counted-baseline recovery) and `min(adjustedLast, containedTotalDelta)` so `last` cannot grow usage when the contained totals delta is zero.
- **Single-lineage and pre-latch fork behavior is preserved** (including #1164 totals-only fork replay handling). The slow JSON path and fork-parent snapshot builder share the same policy.
- Cache persists watermark + interleaved flag (`seenRawTotals` is optional precision only); incomplete critical state forces a full rescan. Parser hash invalidation rebuilds old caches once.
- **Smaller-lineage undercount is intentional** and deferred to fixture-gated Phase 2. Latched files are a conservative estimate, not claimed true per-lineage usage.

Fixes #2037. Related: #968, #1062, `45b68c34` (fork replay).

Spec: `docs/issue-2037-ultra-fork-overcount-spec.md`

## Behavior changes

- Interleaved (Ultra) files: no multiplied-gap inflation; after latch, counted totals stay within the containment bound and `last` cannot exceed the contained totals delta.
- Single-lineage / never-latched files: unchanged counting semantics.
- Pre-latch resolved forks: still totals-only (replay-safe).
- Post-latch: below-watermark smaller-lineage usage may be dropped until Phase 2.

## Test plan

- [x] Focused `CostUsageScannerBreakdownTests` / `CostUsageCacheTests` (containment, caps, eviction, property bound, cache gates)
- [x] Existing #968 / #1062 / fork replay tests
- [x] `make check`
- [x] `make test` (full sharded suite; 0 first-pass failed groups)
- [ ] Optional follow-up: rescan a real Sol/Ultra log for plausibility (not a merge blocker)

## Notes for reviewers

- Design is lineage-ID-free; value-based multi-run recovery is Phase 2 behind a sanitized Ultra fixture.
- `seenRawTotals` is **not** load-bearing after the post-latch min-cap; missing it must not force a rescan.
- A real Ultra log is valuable validation, not required to merge Phase 1 containment.


Zihao-Qi:
Follow-up draft for the cross-file / missing-parent provenance slice: https://github.com/steipete/CodexBar/pull/2082

Please treat **#2082** as the ordinary fork-family / missing-parent review path. This PR (#2066) can stay draft as the interim file-local Ultra containment experiment. Neither should close #2037 yet.

steipete:
Superseded by #2082, which incorporates this containment work plus the fork-family provenance fixtures, missing-parent fail-open repair, regression coverage, and signed Parallels CLI proof. Landed as c852c135e10cfdfb047895bd13c1a4425ded4afb.

Thanks @Zihao-Qi for carrying this accounting work through the safer consolidated path. The broader #2037 remains open for exact cross-file provenance.

PR #2086, opened 2026-07-12
Isolate plan history loading in tests

steipete:
## Summary

- keep `.testing` startup deterministic when a custom plan-history store is injected
- preserve explicit gated load coverage for load-before-record ordering
- verify that an in-flight record merges and persists both the loaded and new samples

## Proof

- `swift test --filter UsageStorePlanUtilizationAsyncLoadTests` — 11/11
- `make check`
- `make test` — 633/633 selections, 53/53 groups, no retries
- autoreview — clean

The production startup path remains enabled and was already exercised by the signed Tahoe VM proof for #2084. This follow-up changes only `.testing` startup semantics, so no additional provider or UI live boundary applies. No changelog entry: internal test isolation and regression coverage only.

### DIFF
--- a/Sources/CodexBar/UsageStore.swift
+++ b/Sources/CodexBar/UsageStore.swift
@@ -406,8 +406,7 @@ final class UsageStore {
         })
         self.startPlanUtilizationHistoryLoad(
             gate: planUtilizationHistoryLoadGateForTesting,
-            enabled: self.startupBehavior.automaticallyStartsBackgroundWork ||
-                planUtilizationHistoryStore != nil)
+            enabled: self.startupBehavior.automaticallyStartsBackgroundWork)
         self.sessionLimitResetDetectorStates = Self.loadLimitResetDetectorStates(
             from: settings.userDefaults,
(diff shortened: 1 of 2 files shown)


steipete:
Landing proof for exact head `7627b855c9a0ca3bf8995674400d264313344eff`:

- focused plan-history async suite: 11/11
- `make check`: pass
- full `make test`: 633/633 selections; 53/53 groups; zero retries
- autoreview: clean, no actionable findings
- GitGuardian: pass
- dependencies: unchanged; no freshness action required
- Public Model Identifier Gate: PASS; the exact diff, tests, workflow surfaces, and this comment contain no model identifiers or model-bearing artifacts

Live boundary: this follow-up only prevents `.testing` from implicitly starting a custom history-store load. Production startup remains the already-landed #2084 path. Exact landed main was separately exercised as a signed, Gatekeeper-trusted Tahoe VM build across three runs: history decoding stayed off the main thread in all runs, the AppKit run loop remained active, fixture hashes were unchanged, and the production process was untouched. No additional provider or UI live boundary applies to this test-isolation follow-up.

PR #2095, opened 2026-07-12
Fix stale cost/token cache pinned by non-forced refreshes

### LINKED ISSUE #2089: Bug: Automatic cost refreshes can silently serve a stale scan and pin it for the full token-cost TTL
**Affected version:** v0.42.1 (build 103)
**Component:** cost usage scanning (`CodexBarCore`)
**Type:** correctness / stale data

## Summary

`CostUsageFetcher.loadDailyReport` zeroes the scanner debounce (`refreshMinIntervalSeconds`) **only when `forceRefresh` is true**. Every non-forced fetch — the hourly token-cost timer, the first fetch after app launch, and cost-scope/settings changes — keeps the default 60-second debounce. When such a fetch lands within 60 seconds of *any* prior scan, the scanner skips file inspection entirely and returns the previous cached report. `UsageStore` then treats that stale snapshot as fresh for the full `tokenFetchTTL` (60 minutes), so a single throttled fetch pins out-of-date cost/token totals in the menu bar for up to an hour.

The 60-second debounce is reasonable as a scanner-internal rate limit, but the app-side caller (`UsageStore`) already rate-limits itself with its own TTL — so at the app layer the debounce never saves work; it only ever manifests as silent staleness that the TTL then amplifies.

## Mechanism (code walk, all references at tag `v0.42.1`)

1. **The force-only gate** — `Sources/CodexBarCore/CostUsageFetcher.swift:175-184`:
   
   ```swift
   if forceRefresh {
       options.refreshMinIntervalSeconds = 0
   }
   ...
   if forceRefresh {
       resolvedPiOptions.refreshMinIntervalSeconds = 0
   }
   ```
   
   The default is `refreshMinIntervalSeconds: TimeInterval = 60` (`Sources/CodexBarCore/Vendored/CostUsage/CostUsageScanner.swift:29`, and identically for the pi-session scanner at `Sources/CodexBarCore/PiSessionCostScanner.swift:22`).

2. **The throttle skips file inspection entirely** — `CostUsageScanner.makeCodexRefreshPlan` computes (`CostUsageScanner.swift:2395-2406`):
   
   ```swift
   let shouldRefresh = options.forceRescan
       || ...
       || refreshMs == 0
       || cache.lastScanUnixMs == 0
       || nowMs - cache.lastScanUnixMs > refreshMs
   ```
   
   When `shouldRefresh` is false, `loadCodexDaily` bypasses the whole scan block — no directory listing, no file-metadata (size/mtime) checks, no incremental append — and goes straight to `buildCodexReportFromCache` (`CostUsageScanner.swift:2440` and `:2590`). Rows appended to session JSONL files since the stamped scan are invisible. The Claude scanner has the same gate (`CostUsageScanner+Claude.swift:662`), as does `PiSessionCostScanner` (`PiSessionCostScanner.swift:103`).

3. **The throttle timestamp is cross-process** — `cache.lastScanUnixMs` is persisted in the on-disk cache at `~/Library/Caches/CodexBar/cost-usage/<provider>-v<N>.json` (`CostUsageCache.swift:20-31`). The menu-bar app, the `codexbar cost` CLI, and any other consumer using the default cache root share it. A scan by *another process* arms the throttle against the app's next non-forced fetch.

4. **The staleness gets pinned for an hour** — `UsageStore.refreshTokenUsage` (`Sources/CodexBar/UsageStore.swift:1508-1517`) stamps `lastTokenFetchAt[provider] = now` *before* the fetch and gates non-forced refreshes on `tokenFetchTTL = 60 * 60` (`UsageStore.swift:336`). There is no detection that the fetch was served from a throttled cache, and no earlier retry. The hourly timer path is non-forced (`UsageStore+TokenRefreshSequence.swift:10-24`: `startTokenTimer` → `scheduleTokenRefresh` → `startTokenRefreshSequence(force: false, scope: .all)`), and so is the first post-launch fetch (in-memory `lastTokenFetchAt` starts empty, but the *on-disk* `lastScanUnixMs` survives relaunch).

### User-level

1. Start an active Codex (or Claude Code) session so the session JSONL under `~/.codex/sessions/...` is growing.
2. Run `codexbar cost` (or otherwise trigger a scan). This stamps `lastScanUnixMs` in the shared on-disk cache.
3. Within 60 seconds, launch CodexBar.app (or have a cost-scope/settings change trigger a non-forced fetch).
4. The app's fetch takes the throttled path and displays totals from step 2's scan. Usage produced after step 2 is missing, and the display does not correct itself until the hourly timer fires again or the user forces a refresh — which is why the symptom presents as "token/cost totals lag until I manually refresh."

### Deterministic, test-level

The following test passes with the proposed fix and, by inspection of `makeCodexRefreshPlan` (both `loadTokenSnapshot` calls share the same `now`, so `nowMs - cache.lastScanUnixMs = 0 ≤ 60_000` and `shouldRefresh` is false on the second call), the second assertion fails on stock v0.42.1 with `refreshed.daily.first?.totalTokens == 110`:

```swift
@Test
func `non-forced refresh keeps incremental cost cache current`() async throws {
    // 1. Write a codex session JSONL containing one token_count row (110 total tokens)
    //    into a temp codex home, with a temp cacheRoot.
    // 2. loadTokenSnapshot(provider: .codex, now: day, ...) — non-forced.
    //    #expect(first.daily.first?.totalTokens == 110)   // fresh scan, passes
    // 3. Append a second token_count row to the same file (total becomes 176).
    // 4. loadTokenSnapshot again with the same `now` — non-forced, inside the 60 s window.
    //    #expect(refreshed.daily.first?.totalTokens == 176)
    //    stock v0.42.1: returns 110 (throttled cache); fixed: returns 176
}
```

(Full runnable version available — it uses only existing test fixtures: a temp `CostUsageScanner.Options(codexSessionsRoot:cacheRoot:)` and JSONL `token_count` events.)

## Expected vs. actual

- **Expected:** a non-forced refresh reflects all usage present in the session logs at fetch time, or at minimum the scanner notices that file metadata changed and performs its (already-implemented) cheap incremental append.
- **Actual:** within 60 s of any prior scan by any process, the fetch silently returns the previous report without looking at the filesystem, and the app then displays it as fresh for up to 60 minutes.

## Suggested directions (either resolves it)

1. **Zero `refreshMinIntervalSeconds` for app-driven refreshes** (not just forced ones). `UsageStore` already serializes and TTL-gates its fetches, so the scanner debounce is redundant on this path: set `options.refreshMinIntervalSeconds = 0` and `resolvedPiOptions.refreshMinIntervalSeconds = 0` unconditionally in `CostUsageFetcher.loadDailyReport`. The full test suite passes with this change applied to v0.42.1 (634 test selections, 0 failures).
2. Alternatively, **make the throttled path metadata-aware**: when within `refreshMinIntervalSeconds`, still stat the known session files and fall through to the incremental-append path if sizes/mtimes changed. Keeps the debounce for true no-change polls while eliminating the stale window.

Option 1 is a two-line change; the scan itself remains cheap because the incremental cache (`keepCachedCodexFileIfFresh` / `appendCodexFileIncrementIfPossible`, `CostUsageScanner.swift:2327-2333`) already avoids re-reading unchanged files.

thanthi-del:
## Summary
Fixes #2089.

- `CostUsageFetcher.loadDailyReport` only zeroed the scanner's `refreshMinIntervalSeconds` debounce (default 60s) when `forceRefresh` was true. Every non-forced fetch — the hourly token-cost timer, the first fetch after launch, cost-scope/settings changes — kept the default debounce.
- A non-forced fetch landing within 60s of *any* prior scan (including one from another process sharing the on-disk cache, e.g. `codexbar cost`) skipped file inspection entirely and returned the previous cached report.
- `UsageStore` then treats that snapshot as fresh for the full `tokenFetchTTL` (60 min), so a single throttled fetch pins stale cost/token totals in the menu bar for up to an hour.
- `UsageStore` already TTL-gates its own callers, so the scanner-level debounce never saved real work at the app layer on this path — it only ever manifested as this staleness window. This applies option 1 from the issue: zero `refreshMinIntervalSeconds` unconditionally instead of only on `forceRefresh`. The scanner's incremental cache (`keepCachedCodexFileIfFresh` / `appendCodexFileIncrementIfPossible`) still keeps re-scans of unchanged files cheap.

## Test plan
- [x] `swift build --target CodexBarCore` builds clean
- [x] Added `non-forced refresh keeps incremental cost cache current` to `CostUsageFetcherTests.swift`, mirroring the existing `force refresh keeps incremental cost cache` test but without `forceRefresh`: appends a second `token_count` row to a session log within the same debounce window and asserts the non-forced refresh picks it up (110 → 176 tokens)
- [ ] Could not execute `swift test` in my environment — `AdaptiveReplayKitTests` fails to build here with "no such module 'Testing'" (pre-existing, unrelated to this change; same failure on a pristine checkout of `main`). Would appreciate CI or a maintainer confirming the new test passes.

### DIFF
--- a/Sources/CodexBarCore/CostUsageFetcher.swift
+++ b/Sources/CodexBarCore/CostUsageFetcher.swift
@@ -172,16 +172,18 @@ public struct CostUsageFetcher: Sendable {
         } else if provider == .claude {
             options.claudeLogProviderFilter = .excludeVertexAI
         }
-        if forceRefresh {
-            options.refreshMinIntervalSeconds = 0
-        }
+        // Always disarm the scanner's own debounce here, not just for forceRefresh: UsageStore
+        // already TTL-gates its own callers (tokenFetchTTL, 60 min), so this debounce never saves
+        // real work at the app layer. Left conditional, it silently pins a stale scan — from this
+        // process or any other sharing the on-disk cache (the CLI, a second scan) — for the full
+        // TTL window whenever a non-forced fetch lands within refreshMinIntervalSeconds of a prior
+        // scan. The scanner's incremental cache still keeps re-scans cheap. See GitHub issue #2089.
+        options.refreshMinIntervalSeconds = 0
         var resolvedPiOptions = overridePiScannerOptions ?? PiSessionCostScanner.Options()
         if resolvedPiOptions.cacheRoot == nil {
             resolvedPiOptions.cacheRoot = options.cacheRoot
         }
-        if forceRefresh {
-            resolvedPiOptions.refreshMinIntervalSeconds = 0
-        }
+        resolvedPiOptions.refreshMinIntervalSeconds = 0
         let piOptions = resolvedPiOptions
 
         try Task.checkCancellation()
--- a/Tests/CodexBarTests/CostUsageFetcherTests.swift
+++ b/Tests/CodexBarTests/CostUsageFetcherTests.swift
@@ -820,6 +820,91 @@ struct CostUsageFetcherTests {
         #expect(refreshed.daily.first?.totalTokens == 176)
     }
 
+    // Regression test for GitHub issue #2089: a non-forced refresh must reflect usage appended
+    // to the session log since the prior scan, even when it lands inside the scanner's internal
+    // refreshMinIntervalSeconds debounce window. UsageStore already TTL-gates its own callers, so
+    // that debounce should never cause the app layer to silently pin a stale scan.
+    @Test
+    func `non-forced refresh keeps incremental cost cache current`() async throws {
+        let env = try CostUsageTestEnvironment()
+        defer { env.cleanup() }
+
+        let day = try env.makeLocalNoon(year: 2026, month: 4, day: 11)
+        let iso0 = env.isoString(for: day)
+        let iso1 = env.isoString(for: day.addingTimeInterval(1))
+        let iso2 = env.isoString(for: day.addingTimeInterval(2))
+        let model = "openai/gpt-5.4"
+
+        let turnContext: [String: Any] = [
+            "type": "turn_context",
+            "timestamp": iso0,
+            "payload": ["model": model],
+        ]
+        let firstTokenCount: [String: Any] = [
+            "type": "event_msg",
+            "timestamp": iso1,
+            "payload": [
+                "type": "token_count",
+                "info": [
+                    "model": model,
+                    "total_token_usage": [
+                        "input_tokens": 100,
+                        "cached_input_tokens": 20,
+                        "output_tokens": 10,
+                    ],
+                ],
+            ],
+        ]
+        let fileURL = try env.writeCodexSessionFile(
+            day: day,
+            filename: "session.jsonl",
+            contents: env.jsonl([turnContext, firstTokenCount]))
+
+        let nativeOptions = CostUsageScanner.Options(
+            codexSessionsRoot: env.codexSessionsRoot,
+            claudeProjectsRoots: [env.claudeProjectsRoot],
+            cacheRoot: env.cacheRoot)
+        let piOptions = PiSessionCostScanner.Options(
+            piSessionsRoot: env.piSessionsRoot,
+            cacheRoot: env.cacheRoot)
+
+        let first = try await CostUsageFetcher.loadTokenSnapshot(
+            provider: .codex,
+            now: day,
+            scannerOptions: nativeOptions,
+            piScannerOptions: piOptions)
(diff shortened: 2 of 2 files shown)


steipete:
Superseded by #2096, which landed the same stale cost-refresh repair in `28504b9c551490880d1d73d85f07f6a240329816` with the bypass scoped to app-owned refreshes so direct library callers retain debounce semantics. The landed version includes focused stale-cache coverage, full local checks/tests, and a green Linux/macOS CI matrix. Thanks for reporting and working on this failure mode.

Issue #867, opened 2026-05-08
Clicking the status bar icon to switch from Overview to sub-API provider may not work. 点开状态栏图标从Overview切换到子api提供商点击可能会失效


Karl-Dai:
Clicking the status bar icon to switch from Overview to sub-API provider may not work. 
[

https://github.com/user-attachments/assets/5248e4d0-f7e4-4d32-aa79-dcf77aab5ca3

](url)

PR #2097, opened 2026-07-12
Fix dropped overview menu-card row clicks (NSMenu tracking loop)

### LINKED ISSUE #2090: Bug: Overview menu-card row clicks are dropped — `NSClickGestureRecognizer` inside NSMenu tracking, same failure class as #867
**Affected version:** v0.42.1 (build 103); behavior present since the clickable overview rows were introduced
**Component:** merged-menu overview rows (`Sources/CodexBar/StatusItemController+MenuPresentation.swift`)
**Type:** intermittent UI input loss
**Prior art:** issue #867 and its fix `9907f749` ("fix(menu): route switcher clicks through parent view") — the same failure mode, fixed for the provider switcher but not for menu-card rows

## Summary

Clickable hosted-SwiftUI menu rows (`MenuCardItemHostingView`) deliver their `onClick` through an `NSClickGestureRecognizer` installed on the hosting view (`StatusItemController+MenuPresentation.swift:194-199` at `v0.42.1`). Gesture recognizers on custom views inside a tracking `NSMenu` are subject to the same event-delivery loss that upstream already diagnosed and fixed for the provider switcher's `NSButton`s in `9907f749`:

> "NSMenu's tracking run loop occasionally drops NSButton target-action dispatch when the menu is rebuilt under the cursor (e.g. after switching back from a provider tab to Overview)." — comment added by `9907f749` in `StatusItemController+SwitcherViews.swift`

The switcher fix routes clicks through the **parent view's** `hitTest` + `mouseDown`/`mouseUp` so the click "never has to round-trip through NSButton's tracking loop." `MenuCardItemHostingView` still uses the pre-#867 pattern — a recognizer that needs the tracking loop to deliver and complete a gesture — so overview row clicks (which select that provider's tab, `StatusItemController+Menu.swift:580-583`) can intermittently do nothing.

## Symptom

With merged icons / the overview menu: clicking a provider's card row sometimes has no effect — the menu stays on the overview and no selection happens. It is timing-dependent and most likely when menu content is being rebuilt or refreshed while the menu is open under the cursor (the same trigger described in #867). A second click usually works, which makes it easy to dismiss as a mis-click.

## Why the recognizer path is fragile here

- `NSMenu` runs its own tracking loop; custom-view event delivery inside it is not the normal window responder chain.
- `MenuCardItemHostingView` instances are **reused and re-wired** while the menu is open: `prepareForReuse(rootView:allowsMenuHighlight:onClick:)` swaps `rootView` and `onClick` in place on refresh (`StatusItemController+MenuPresentation.swift:169-176`). A gesture that began before a rebuild can complete against a view whose recognizer state was disturbed by the in-place SwiftUI diff.
- A gesture recognizer requires the full begin→end state machine to run inside the tracking loop; parent-view `mouseDown`/`mouseUp` routing does not.

Upstream's own #867 analysis covers this class of failure; the menu-card rows are the remaining user of the fragile pattern. (A grep at `v0.42.1` shows `NSClickGestureRecognizer` in `MenuCardItemHostingView` is the only recognizer-based click path left inside menu content.)

## Reproduction

Intermittent by nature — this is a race with the menu tracking loop, so there is no deterministic step list; #867 was the same kind of report. Best-known trigger sequence:

1. Enable merged icons so the overview menu with per-provider card rows is shown.
2. Open the menu and leave the cursor over a provider row while a background refresh rebuilds/reuses menu content (easiest during active provider refresh cycles; switching to a provider tab and back to Overview increases churn, mirroring #867).
3. Click the row. Occasionally the click is swallowed: no provider tab opens.

## Suggested fix — apply the #867 technique to `MenuCardItemHostingView`

Mirror `9907f749`: remove the `NSClickGestureRecognizer` and route clicks through the hosting view itself —

- `hitTest(_:)`: when `onClick != nil`, return `self` for hits on descendants, so child SwiftUI content never owns the click;
- `mouseDown(with:)`: record pressed state if the click began inside bounds;
- `mouseUp(with:)`: fire `onClick` only if the press began and ended inside bounds (preserves click-cancel-by-drag-out semantics).

Applied to v0.42.1, this change passes the full test suite (634 test selections, 0 failures). Note the accessibility path is unaffected: `accessibilityPerformPress()` already invokes `onClick` directly and keeps working.

A regression test in the same shape as #867's `StatusMenuSwitcherClickTests` is straightforward: expose a DEBUG-only `_test_simulateRuntimeClick()` on `MenuCardItemHostingView` that drives the production click path without a recognizer, and assert `onClick` fires:

```swift
@Test
func `routes runtime click without gesture recognizer`() {
    var clicked = false
    let view = MenuCardItemHostingView(
        rootView: Text("Overview row"),
        highlightState: MenuCardHighlightState(),
        allowsMenuHighlight: true,
        onClick: { clicked = true })
    #expect(view._test_simulateRuntimeClick())
    #expect(clicked)
}
```

## Evidence index (at tag `v0.42.1`)

- Recognizer installation: `Sources/CodexBar/StatusItemController+MenuPresentation.swift:194-199` (`installClickRecognizer`), invoked from `init` (:159-161) and `prepareForReuse` (:172-174)
- Click handler requiring gesture completion: `:218-221` (`handlePrimaryClick`, `guard recognizer.state == .ended`)
- User-visible consumer: overview provider rows, `Sources/CodexBar/StatusItemController+Menu.swift:580-583` (`onClick` → `selectOverviewProvider`)
- Upstream precedent: commit `9907f749` ("fix(menu): route switcher clicks through parent view", fixes #867) — adds `hitTest`/`mouseDown`/`mouseUp` routing plus `StatusMenuSwitcherClickTests` to `ProviderSwitcherView`, with in-code rationale describing the dropped-dispatch failure mode

thanthi-del:
## Summary
Fixes #2090.

- `MenuCardItemHostingView` delivered `onClick` through an `NSClickGestureRecognizer`, which needs its full begin→end state machine to run inside `NSMenu`'s tracking loop to fire.
- That loop can drop the dispatch when the row is rebuilt under the cursor — the same failure class already diagnosed and fixed for the provider switcher's `NSButton`s in #867 (commit `9907f749`), and made worse here by `prepareForReuse` swapping `rootView`/`onClick` in place on a view whose recognizer state a gesture-in-progress could be tied to.
- Applies the #867 technique: route the click through the view itself instead of a gesture recognizer.
  - `hitTest` swallows hits on the hosted SwiftUI content so it never owns the click.
  - `mouseDown`/`mouseUp` track a simple `isPressed` flag and fire `onClick` only when both the press and release land inside bounds (preserving click-cancel-by-drag-out semantics).
- `accessibilityPerformPress()` already calls `onClick` directly and is unaffected.

## Test plan
- [x] Added `Tests/CodexBarTests/MenuCardItemHostingViewClickTests.swift` with a `#if DEBUG _test_simulateRuntimeClick()` helper (mirrors `CodexAccountSwitcherView`'s existing one) and three tests:
  - basic click fires `onClick`
  - click still routes correctly after `prepareForReuse` (the actual reuse-under-refresh scenario from the report)
  - mouse-up outside bounds cancels the click
- [x] `swiftc -parse` on both changed files (syntax-only, no diagnostics)
- [ ] Could not run a full `swift build`/`swift test` in my environment — `KeyboardShortcuts`' `#Preview` macro plugin isn't found there (pre-existing, unrelated to this change; same failure on a pristine checkout of `main`), which blocks the `CodexBar` target specifically. Would appreciate CI or a maintainer confirming the new tests pass.

### DIFF
--- a/Sources/CodexBar/StatusItemController+MenuPresentation.swift
+++ b/Sources/CodexBar/StatusItemController+MenuPresentation.swift
@@ -134,7 +134,7 @@ final class MenuCardItemHostingView<Content: View>: NSHostingView<Content>, Menu
     let highlightState: MenuCardHighlightState
     private(set) var allowsMenuHighlight: Bool
     private var onClick: (() -> Void)?
-    private var hasClickRecognizer = false
+    private var isPressed = false
 
     override var allowsVibrancy: Bool {
         true
@@ -156,9 +156,6 @@ final class MenuCardItemHostingView<Content: View>: NSHostingView<Content>, Menu
         self.allowsMenuHighlight = allowsMenuHighlight
         self.onClick = onClick
         super.init(rootView: rootView)
-        if onClick != nil {
-            self.installClickRecognizer()
-        }
     }
 
     /// Reuses this hosting view for a rebuilt card with the same identity: the replaced
@@ -169,9 +166,6 @@ final class MenuCardItemHostingView<Content: View>: NSHostingView<Content>, Menu
         self.rootView = rootView
         self.allowsMenuHighlight = allowsMenuHighlight
         self.onClick = onClick
-        if onClick != nil, !self.hasClickRecognizer {
-            self.installClickRecognizer()
-        }
     }
 
     /// `NSMenu` tracking consumes keyboard events before they reach a menu item's custom view, so
@@ -190,13 +184,6 @@ final class MenuCardItemHostingView<Content: View>: NSHostingView<Content>, Menu
         return true
     }
 
-    private func installClickRecognizer() {
-        let recognizer = NSClickGestureRecognizer(target: self, action: #selector(self.handlePrimaryClick(_:)))
-        recognizer.buttonMask = 0x1
-        self.addGestureRecognizer(recognizer)
-        self.hasClickRecognizer = true
-    }
-
     required init(rootView: Content) {
         self.highlightState = MenuCardHighlightState()
         self.allowsMenuHighlight = false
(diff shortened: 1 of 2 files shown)


steipete:
Thanks for tackling this, @thanthi-del. We landed the same issue through #2093 in commit 7465c6a11674d4cacb97b8f023dd9ad8fc70e446, including nested-control preservation, drag-cancel behavior, native submenu drag tracking, focused regression coverage, and signed live menu proof. Closing this duplicate in favor of the landed fix.

PR #2102, opened 2026-07-12
Fix Claude no-prompt Keychain repair

### LINKED ISSUE #1975: Claude "Avoid Keychain prompts": account switch leaves ~/.claude/.credentials.json missing → "OAuth credentials not found" while Keychain token is valid
**TL;DR:** With **"Avoid Keychain prompts"** enabled (`debugDisableKeychainAccess = true`), CodexBar reads Claude OAuth from the plaintext `~/.claude/.credentials.json` instead of the Keychain. But after switching Claude accounts, Claude Code sometimes writes the new token **only to the Keychain and not to `~/.claude/.credentials.json`** — the file goes missing entirely. CodexBar then shows **"Claude OAuth credentials not found. Run `claude` to authenticate."** even though a valid, non-expired token with the right scopes is sitting in the Keychain.

Every existing keychain/credentials issue I found is about the *Keychain item* being wrong or absent (#1844, #1823, #1161). This one is the opposite: the Keychain is fine; the **file** the avoid-keychain path depends on is the thing that disappears on account switch.

## Environment
- CodexBar 0.40.0 (build 99)
- macOS 26.5.2 (25F84)
- Claude Code CLI 2.1.197, Claude Max, logged in and working
- Claude provider setting: **Avoid Keychain prompts = ON** (`defaults read com.steipete.codexbar debugDisableKeychainAccess` → `1`)

## Repro
1. Enable **Avoid Keychain prompts** for the Claude provider (so CodexBar reads `source=credentialsFile`). Confirm usage loads.
2. Switch Claude accounts in Claude Code (e.g. `/login` as a different account).
3. Observe the Claude tab shows **"Claude OAuth credentials not found. Run `claude` to authenticate."**

## Smoking gun
At the point of failure:

```
$ ls -la ~/.claude/.credentials.json
ls: /Users/…/.claude/.credentials.json: No such file or directory     # file is GONE

$ security find-generic-password -s "Claude Code-credentials" -a "$USER" -w | python3 -c '...'
has accessToken: True
scopes: ['user:file_upload', 'user:inference', 'user:mcp_servers', 'user:profile', 'user:sessions:claude_code']
expiresAt: 1783487829153 -> expired? False                            # Keychain token is VALID
```

So the Keychain has a healthy `claudeAiOauth` token; the plaintext file simply does not exist. Because keychain access is disabled, CodexBar has nothing to read and fails closed with "credentials not found."

`codexbar usage --provider claude -v` after recreating the file confirms the same token reads fine:
```
allowKeychainPrompt=false … hasUserProfileScope=true isExpired=false … source=memoryCache … credentials loaded for usage
```

## Manual recovery (workaround)
Regenerate the file from the Keychain, then Refresh (⌘R):
```bash
security find-generic-password -s "Claude Code-credentials" -a "$USER" -w > ~/.claude/.credentials.json && chmod 600 ~/.claude/.credentials.json
```

## Suggested fix
When **Avoid Keychain prompts** is enabled and `~/.claude/.credentials.json` is missing/unreadable, fall back to a **non-interactive** Keychain read (the Security.framework no-UI policy from #1877) before surfacing "credentials not found" — i.e. read-only, never prompting. That would make the avoid-keychain path resilient to Claude Code writing the token to only one of the two stores after an account switch, without reintroducing prompts.

(Reported via a debugging session; happy to grab additional logs.)

OfficialAbhinavSingh:
Fixes #1975.

## Summary

When Claude's **Avoid Keychain prompts** preference was set to Never, CodexBar blocked every Claude Keychain read. That also blocked guaranteed no-UI Security.framework reads, so a missing `~/.claude/.credentials.json` could not be repaired from an otherwise valid Claude Code Keychain item.

This change makes the policy boundary explicit:

- Never still blocks interactive Security.framework reads.
- Never also blocks the experimental `/usr/bin/security` reader because that subprocess can prompt.
- Reads protected by `KeychainNoUIQuery` remain available and fail closed instead of showing UI.
- Advanced → Disable Keychain access remains the hard global kill switch and blocks every Keychain path.

The patch also updates the existing experimental-reader expectations and removes the proposed live verification script. Routine proof now uses isolated fixture stores only; it never changes real defaults or touches the login Keychain.

## Verification

- `swift test --filter 'ClaudeOAuthCredentialsStoreTests|ClaudeOAuthCredentialsStoreSecurityCLITests|ClaudeOAuthCredentialsStoreNeverPromptCacheTests|ClaudeOAuthCredentialsStoreIsolatedSecurityCLITests'` — 59 tests passed.
- Source-blind behavior validation — 5/5 contracts passed: no-UI repair, zero security CLI calls, interactive denial, absent-item fail-closed, global disable.
- `make check` — SwiftFormat and SwiftLint clean; repository/docs/package gates passed.
- `CODEXBAR_TEST_RETRY_NON_TIMEOUT_FAILURES=0 make test` — 652/652 selections, 55/55 groups passed first attempt, zero retries and zero timeouts.
- Maintainer autoreview — no actionable findings.

No live Keychain, defaults, browser, network, or account credentials were used during validation.

Thanks @OfficialAbhinavSingh for the report and original fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - ZenMux: add Management API usage with five-hour and weekly quotas, subscription expiry, and USD PAYG balance. Thanks @kays0x!
 
 ### Fixed
+- Claude: recover a missing credentials file from a valid Claude Code Keychain item without showing Keychain UI when Never prompt is selected (#1975). Thanks @OfficialAbhinavSingh!
 - Codex cost usage: invalidate cached fork totals when the parent session appears, changes, or resolves to a different file, preventing stale inherited baselines. Thanks @xx205!
 - Cursor: bind interactive account login to one readable browser, preserve the active session on cancellation or failure, and prevent background refreshes from replacing the selected account. Thanks @chapati23!
 - Menu bar: prevent duplicate provider items when usage updates re-enter initial status-item setup (#2162). Thanks @ss251!
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeOAuth/ClaudeOAuthCredentials+SecurityCLIReader.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeOAuth/ClaudeOAuthCredentials+SecurityCLIReader.swift
@@ -42,7 +42,8 @@ extension ClaudeOAuthCredentialsStore {
     }
 
     /// Attempts a Claude keychain read via `/usr/bin/security` when the experimental reader is enabled.
-    /// - Important: `interaction` is diagnostics context only and does not gate CLI execution.
+    /// - Important: `interaction` is diagnostics context only. The stored Never policy still blocks the CLI because
+    ///   `security` can prompt.
     static func loadFromClaudeKeychainViaSecurityCLIIfEnabled(
         interaction: ProviderInteraction,
         readStrategy: ClaudeOAuthKeychainReadStrategy = ClaudeOAuthKeychainReadStrategyPreference.current())
@@ -92,6 +93,7 @@ extension ClaudeOAuthCredentialsStore {
         -> Data?
     {
         guard self.shouldPreferSecurityCLIKeychainRead(readStrategy: readStrategy) else { return nil }
+        guard ClaudeOAuthKeychainPromptPreference.storedMode() != .never else { return nil }
         let interactionMetadata = interaction == .userInitiated ? "user" : "background"
 
         do {
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeOAuth/ClaudeOAuthCredentials.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeOAuth/ClaudeOAuthCredentials.swift
@@ -613,7 +613,9 @@ public enum ClaudeOAuthCredentialsStore {
             self.context.run {
                 #if os(macOS)
                 let mode = ClaudeOAuthKeychainPromptPreference.current()
-                guard ClaudeOAuthCredentialsStore.shouldAllowClaudeCodeKeychainAccess(mode: mode) else { return false }
+                guard ClaudeOAuthCredentialsStore.shouldAllowClaudeCodeKeychainAccess(
+                    mode: mode,
+                    allowKeychainPrompt: false) else { return false }
                 if ClaudeOAuthCredentialsStore.loadFromClaudeKeychainViaSecurityCLIIfEnabled(
                     interaction: ProviderInteractionContext.current) != nil
                 {
@@ -621,7 +623,10 @@ public enum ClaudeOAuthCredentialsStore {
                 }
 
                 let fallbackPromptMode = ClaudeOAuthKeychainPromptPreference.securityFrameworkFallbackMode()
-                guard ClaudeOAuthCredentialsStore.shouldAllowClaudeCodeKeychainAccess(mode: fallbackPromptMode) else {
+                guard ClaudeOAuthCredentialsStore.shouldAllowClaudeCodeKeychainAccess(
+                    mode: fallbackPromptMode,
+                    allowKeychainPrompt: false)
+                else {
                     return false
                 }
                 if ProviderInteractionContext.current == .background,
@@ -707,7 +712,8 @@ public enum ClaudeOAuthCredentialsStore {
         {
             #if os(macOS)
             let mode = ClaudeOAuthKeychainPromptPreference.current()
-            guard ClaudeOAuthCredentialsStore.shouldAllowClaudeCodeKeychainAccess(mode: mode) else { return nil }
+            guard ClaudeOAuthCredentialsStore
+                .shouldAllowClaudeCodeKeychainAccess(mode: mode, allowKeychainPrompt: false) else { return nil }
             if ClaudeOAuthCredentialsStore.isPromptPolicyApplicable,
                respectKeychainPromptCooldown,
                !ClaudeOAuthKeychainAccessGate.shouldAllowPrompt(now: now)
@@ -786,7 +792,8 @@ public enum ClaudeOAuthCredentialsStore {
         {
             #if os(macOS)
             let mode = ClaudeOAuthKeychainPromptPreference.current()
-            guard ClaudeOAuthCredentialsStore.shouldAllowClaudeCodeKeychainAccess(mode: mode) else { return nil }
+            guard ClaudeOAuthCredentialsStore
+                .shouldAllowClaudeCodeKeychainAccess(mode: mode, allowKeychainPrompt: false) else { return nil }
 
             if ClaudeOAuthCredentialsStore.shouldShowClaudeKeychainPreAlert() {
                 return nil
@@ -897,7 +904,9 @@ public enum ClaudeOAuthCredentialsStore {
             self.context.run {
                 #if os(macOS)
                 let mode = ClaudeOAuthKeychainPromptPreference.current()
-                guard ClaudeOAuthCredentialsStore.shouldAllowClaudeCodeKeychainAccess(mode: mode) else { return false }
+                guard ClaudeOAuthCredentialsStore.shouldAllowClaudeCodeKeychainAccess(
+                    mode: mode,
+                    allowKeychainPrompt: false) else { return false }
 
                 if let data = ClaudeOAuthCredentialsStore.loadFromClaudeKeychainViaSecurityCLIIfEnabled(
                     interaction: ProviderInteractionContext.current),
@@ -916,7 +925,10 @@ public enum ClaudeOAuthCredentialsStore {
                 }
 
                 let fallbackPromptMode = ClaudeOAuthKeychainPromptPreference.securityFrameworkFallbackMode()
-                guard ClaudeOAuthCredentialsStore.shouldAllowClaudeCodeKeychainAccess(mode: fallbackPromptMode) else {
+                guard ClaudeOAuthCredentialsStore.shouldAllowClaudeCodeKeychainAccess(
+                    mode: fallbackPromptMode,
+                    allowKeychainPrompt: false)
+                else {
                     return false
                 }
 
@@ -1641,7 +1653,8 @@ public enum ClaudeOAuthCredentialsStore {
             return .value(override)
         }
         #endif
-        guard self.shouldAllowClaudeCodeKeychainAccess(mode: mode) else { return .unavailable }
+        guard self.shouldAllowClaudeCodeKeychainAccess(mode: mode, allowKeychainPrompt: false)
+        else { return .unavailable }
         if self.isPromptPolicyApplicable,
            ProviderInteractionContext.current == .background,
            !ClaudeOAuthKeychainAccessGate.shouldAllowPrompt()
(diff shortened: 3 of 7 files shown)


OfficialAbhinavSingh:
Thanks for the thorough review. I dug into the P1 ("Block the experimental CLI reader for no-prompt callers") and I don't think it holds as stated — this PR doesn't newly route a no-prompt caller through `/usr/bin/security`. Walking through it:

**Under the default `securityFramework` read strategy** (the strategy in the #1975 report), the CLI reader is inert: `readRawClaudeKeychainPayloadViaSecurityCLIIfEnabled` early-returns at `guard shouldPreferSecurityCLIKeychainRead(...)`, which is only true for `.securityCLIExperimental`. So the `.never` missing-file repair this PR enables reaches the token exclusively through the enforced no-UI `KeychainNoUIQuery` path — no subprocess, no prompt.

**Under the `.securityCLIExperimental` strategy**, the experimental reader already runs under `.never` on `main`, independent of this PR: `ClaudeOAuthKeychainPromptPreference.current()` resolves through `effectiveMode(readStrategy:)`, which returns `.always` for any non-`securityFramework` strategy (`isApplicable` is `false`). So the background silent paths (`hasClaudeKeychainCredentialsWithoutPrompt`, repair, sync) see `.always` at the gate and invoke the CLI reader on `main` today. This is covered and asserted by an existing test — `experimental reader ignores prompt policy and cooldown for background silent check` — which sets `.securityCLIExperimental` + `.never` + `.background` and expects the security-CLI read to succeed. So the experimental reader deliberately ignoring the `.never` policy is intended, pre-existing behavior, not something this PR introduces.

I did initially try the suggested "skip the CLI reader when `allowKeychainPrompt == false`" change, but it broke 7 tests in `ClaudeOAuthCredentialsStoreSecurityCLITests` (including the one above) precisely because it suppressed that intended experimental-reader behavior. I've reverted that part; the PR now contains only the central gate change (`.never` permitting guaranteed no-UI reads) plus the `securityFramework` no-UI repair fix and its regression tests.

If you'd still like the experimental reader's `.never` behavior tightened, I think that's a separate product decision from this bug (it would change shipped, tested behavior for opted-in experimental users) — happy to open a follow-up if a maintainer wants it. Otherwise the security-boundary risk here should be resolved. Could you take another look?

OfficialAbhinavSingh:
@clawsweeper re-review — the P1 (experimental CLI reader under `.never`) doesn't hold: this PR leaves that reader untouched, and its `.never` behavior is pre-existing/by-design (the `experimental reader ignores prompt policy and cooldown for background silent check` test). I trialed the suggested "skip when `allowKeychainPrompt == false`" change and it broke 7 `ClaudeOAuthCredentialsStoreSecurityCLITests`, so I reverted it. Reasoning in the comment above; PR body updated with the design rationale and before/after.

OfficialAbhinavSingh:
## Verification update — real-macOS evidence + honest scope

I ran this on real hardware (arm64e, macOS 14) and want to lay out precisely what is proven and where the boundary is, rather than overclaim.

### 1. Deterministic before/after (the core of the fix)

The regression test reproduces the reporter's exact failure and the patch fixes it:

- On `main`: `never mode repairs a missing credentials file …` fails with `ClaudeOAuthCredentialsError.notFound` — the same *"Claude OAuth credentials not found"* the issue reports.
- With this change: it passes, resolving `source = .claudeKeychain`, `owner = .claudeCLI`, and logging `Claude keychain credentials loaded without prompt` — no UI.

This ran live (not just CI): the full `ClaudeOAuthCredentialsStoreSecurityCLITests` (17) and `…MCPOnlyGuardTests` suites stay green alongside it, so the experimental reader is unaffected.

### 2. Real-`.never` end-to-end confirms the gate change fires

With the **real** `claudeOAuthKeychainPromptMode = never` preference set (via `defaults`, in a real login session), a missing `~/.claude/.credentials.json`, and `securityFramework` strategy, the debug flow shows the no-UI read actually executing:

```
… allowKeychainPrompt=false interactive=false … service=Claude Code-credentials  Claude keychain legacy data read start
… interactive=false … service=Claude Code-credentials status=… Claude keychain legacy data read result
```

On `main` this read **never starts** — `shouldAllowClaudeCodeKeychainAccess(.never)` returns `false`, so `loadClaudeKeychainLegacyData` returns at its guard before any `SecItemCopyMatching`. Seeing `legacy data read start` under real `.never` is the whole behavioral change: `.never` now means *"no interactive prompts"*, not *"no Keychain access at all"*, while the interactive readers stay fully blocked.

### 3. Where I stopped, and why (transparency)

I did **not** produce a full synthetic end-to-end *recovery* PASS against a real Keychain, for two honest reasons:

- The reader's query searches the default keychain by service. On the test machine a **real** `Claude Code-credentials` item is present, so my verify script (`Scripts/verify_1975_live.sh`) deliberately **aborts before planting/deleting** anything rather than risk a real credential — it never touches real OAuth material.
- Reading that real item end-to-end requires the **signed** app: its ACL is scoped to the `claude` CLI's code signature, so an unsigned local `swift build` CLI gets `errSecInteractionNotAllowed` on it regardless of this change. That's a code-signing boundary, not a logic gap — and only reproducible cleanly in a signed/notarized context where the user has granted access.

`Scripts/verify_1975_live.sh` stands the scenario up safely (disposable HOME, `.never` set-and-restored, a synthetic `-A` item planted only when no real one exists, restored on exit). Run from the signed app context it should show the full green recovery; here it correctly reports `SKIPPED` rather than go near real credentials.

Happy to adjust the test shape or add coverage if you'd prefer a different proof surface.

OfficialAbhinavSingh:
@clawsweeper re-review — the verification comment above adds real-macOS evidence (deterministic before/after + the no-UI read now executing under real `.never`) and explains the code-signing boundary on full end-to-end recovery. Please re-assess the `needs proof` gate.

OfficialAbhinavSingh:
@steipete — a quick note on where this stands, since the remaining gate is now on your side rather than mine.

ClawSweeper's one open ask is *"complete real signed-runtime before-and-after recovery."* I've taken that as far as an external contributor can:

- **Deterministic before/after:** the repair test fails on `main` with the reporter's exact `.notFound` and passes with this change (`source = .claudeKeychain`, no prompt).
- **Real macOS:** under the real `.never` preference, the no-UI Keychain read now *executes* (`allowKeychainPrompt=false → legacy data read start → SecItemCopyMatching`), where `main` returns at the gate before any read.

The last step — end-to-end recovery from a real `Claude Code-credentials` item — I can't reproduce from a fork: an unsigned `swift build` CLI hits `errSecInteractionNotAllowed` on that item because its ACL is bound to a code signature I don't have. That's a signing boundary, not a logic gap, so it's only reproducible in your signed context.

If it's useful, `Scripts/verify_1975_live.sh` produces exactly that signed before/after in one command. It's non-destructive: disposable `HOME`, sets `.never` and restores it, and **aborts rather than touch a real credential** — it only plants a synthetic `-A` item when no real one exists, and deletes it on exit. Run from the signed app/context it should show the full green recovery.

One other thing on your side: as a first-time contributor here, CI is gated behind your workflow approval (only GitGuardian has run so far).

Happy to reshape the test or proof surface however you'd prefer — just let me know.

devYRPauli (approved):
LGTM. I read through the post-change source at each call site rather than the description, and the `.never` no-prompt boundary holds.

The key invariant - under `.never`, no interactive Keychain read is ever permitted - is preserved because the three interactive entry points (`loadFromClaudeKeychain`, `loadFromClaudeKeychainWithPromptIfAllowed`, and the security-framework fallback decision) keep the default `allowKeychainPrompt: true`, so `shouldAllowClaudeCodeKeychainAccess` returns `!true == false` for them under `.never` and they stay blocked. Only the guaranteed no-UI paths (missing-file repair, freshness sync, fingerprint/candidate probes) pass `allowKeychainPrompt: false`. The `guard self.keychainAccessAllowed` kill switch is untouched and still short-circuits before the mode switch.

What makes it safe rather than just plausibly safe is that the same `allowKeychainPrompt` drives both the policy check and the actual query: `loadClaudeKeychainData` applies `KeychainNoUIQuery.apply(&query)` when `!allowKeychainPrompt` (`ClaudeOAuthCredentials.swift` ~L2071), which sets `kSecUseAuthenticationUI = kSecUseAuthenticationUIFail`. So passing `false` produces a query that fails instead of prompting - the policy gate and the query construction can't drift apart.

I built the PR head and ran the new suite locally: all 3 tests pass, including `never mode still blocks an interactive Keychain read even with a valid item present`, which covers the negative direction of the boundary (a valid item is present, prompt requested, still throws `.notFound` under `.never`) - not just the happy-path repair. The logic diff is minimal (+40/-17 in one file, a defaulted parameter plus mechanical call-site threading); the rest is the test file and a standalone verification script.

Two small notes, neither blocking:
- `Scripts/verify_1975_live.sh` temporarily writes the real `com.steipete.codexbar` defaults domain (`claudeOAuthKeychainPromptMode`) during a live run and restores it via a trap. Non-destructive, but worth flagging in case you'd rather a verification script not touch real user prefs.
- This is really the security-framing decision that was flagged for maintainer sign-off: the implementation correctly keeps interactive reads blocked under `.never` and only newly permits guaranteed no-UI reads, so the behavior change is "avoid prompts" continuing to mean "never show a prompt" rather than "never touch the Keychain." That reading looks right to me, but the policy call is yours.

OfficialAbhinavSingh:
Thanks @devYRPauli — appreciate you reading it call-site by call-site and running the suite rather than trusting the description. You've restated the invariant exactly: the three interactive entry points keep `allowKeychainPrompt: true` and stay blocked under `.never`; only the guaranteed no-UI paths pass `false`; and since the same flag drives both the policy check and `KeychainNoUIQuery.apply`, the gate and the query can't drift apart. That last part was the design goal, so it's good to see it read that way independently.

On your two notes:

1. **The real-prefs touch is structural for the *live* check, unfortunately.** `ClaudeOAuthKeychainPromptPreference` reads the mode via `UserDefaults(suiteName: "com.steipete.codexbar")`, which cfprefsd serves per-uid — `CFFIXED_USER_HOME` can't redirect it, so a fixture-home write isn't seen by the spawned CLI. That's why the script writes the real domain and restores the prior value (or deletes the key if it was unset) via an `EXIT` trap. If you or @steipete would rather a casual run never touch real prefs, I'm happy to gate that Phase-2 write behind an explicit opt-in flag (e.g. `CODEXBAR_1975_LIVE_E2E=1`) so it's off by default — just say the word and I'll push it.

2. Agreed the crux is the `.never` = "never show a prompt" vs "never touch the Keychain" reading, and that the policy call is @steipete's. The PR body isolates exactly that as the one thing to sign off, and the `keychainAccessAllowed` kill switch is left untouched ahead of the mode switch so "disable Keychain access" still hard-blocks everything.

Thanks again for the thorough pass.

steipete:
Maintainer proof for exact head `ad3a67cfea6f9aace45cfc12039df8ebe68f657f`:

- Reworked the change so **Never** blocks every prompt-capable path, including `/usr/bin/security`, while retaining only the Security.framework query guarded by `kSecUseAuthenticationUIFail`.
- Added fixture contracts for no-UI recovery, interactive-read rejection, absent-item failure, global-disable behavior, and the experimental CLI-reader boundary: 5/5 passed.
- Focused affected suites: 59/59 passed.
- Full local suite: 652/652 selections passed first attempt; 0 retries, failures, or timeouts.
- `make check`: passed (SwiftFormat, SwiftLint, package/docs/repository gates).
- Autoreview: clean after centralizing the prompt-policy guard.
- Hosted exact-head CI: https://github.com/steipete/CodexBar/actions/runs/29491895280

Thanks @OfficialAbhinavSingh for finding and fixing this recovery gap.

PR #2106, opened 2026-07-13
fix(cost): align token usage refresh cadence with global settings

### LINKED ISSUE #2103: Widget token/cost data should follow the global refresh frequency instead of a fixed hourly TTL
## Summary

The usage/quota values and the token/cost rows shown by CodexBar do not currently share the same automatic refresh cadence.

When the global refresh frequency is configured to a shorter interval (for example, one minute), the main provider usage refresh follows that setting, but the token/cost data written into the widget snapshot remains governed by a separate fixed one-hour TTL. As a result, a widget can show recently refreshed quota bars together with token/cost values that are up to an hour old.

This is distinct from the WidgetKit presentation-staleness issue in #2091: the request here concerns when CodexBar refreshes the token/cost data that it puts into the snapshot, not whether WidgetKit renders a newly supplied timeline.

## Environment

- CodexBar: 0.42.1 (103), official signed release
- Global refresh frequency: one minute
- Widget: CodexBar Usage

No account names, email addresses, usage amounts, credentials, device identifiers, usernames, or local paths are included.

## Reproduction

1. Enable token/cost usage.
2. Set CodexBar's global refresh frequency to one minute.
3. Add or view a CodexBar Usage widget that shows both quota and token/cost data.
4. Continue using the provider so both values should change.
5. Observe that quota data refreshes according to the configured cadence, while the token/cost timestamp and values can remain unchanged for roughly one hour.

## Expected behavior

The widget's token/cost data should follow the same global refresh-frequency setting as the rest of CodexBar.

For a one-minute setting, CodexBar should make token/cost data eligible for refresh every minute and publish the resulting snapshot to the widget. Existing in-flight coalescing and provider sequencing can continue to prevent overlapping work.

Manual refresh should continue to force an immediate refresh.

## Actual behavior

Token/cost refresh uses a separate fixed hourly TTL, so the widget can combine fresh quota data with stale token/cost data.

## Current source behavior

On current main at commit `7465c6a11674`:

- `UsageStore.tokenFetchTTL` is fixed at `60 * 60`.
- `startTokenTimer()` sleeps for `tokenFetchTTL`.
- Automatic app refresh calls `scheduleTokenRefresh()`, but non-forced token fetches are skipped while the fixed TTL has not elapsed.
- Settings copy explicitly says `Auto-refresh: hourly`.

Relevant files:

- `Sources/CodexBar/UsageStore+TokenRefreshSequence.swift`
- `Sources/CodexBar/PreferencesMenuPane.swift`

## Requested behavior

Use the configured global refresh frequency as the token/cost refresh eligibility interval, including for widget snapshot updates, rather than maintaining an independent fixed one-hour cadence.

If the hourly interval is intentional to protect expensive local scans, a useful alternative would be an explicit separate token/cost refresh setting. The key requirement is that the UI should not imply one global cadence while silently applying a different fixed cadence to part of the widget data.

Yuxin-Qiao:
### Summary
Fixes #2103.

Aligns the token/cost auto-refresh cadence with the configured global refresh settings instead of keeping them governed by a separate, hardcoded hourly TTL.

### Background / Problem
Previously, when the global refresh frequency was configured to a shorter interval (e.g., every minute), the main provider usage refresh followed that setting, but the token/cost data remained governed by a separate fixed `tokenFetchTTL` of 1 hour. This caused the widget/UI to display fresh quota data alongside stale token/cost metrics, creating a mismatch in auto-refresh frequency.

### Proposed Changes
- **Dynamic tokenFetchTTL (`Sources/CodexBar/UsageStore.swift`)**: Replaced the hardcoded hourly constant `tokenFetchTTL` with a dynamic property that retrieves the nominal interval from `settings.refreshFrequency`.
- **Timer and Observer Sync (`Sources/CodexBar/UsageStore.swift`)**: Updated the settings observation loop to call `startTokenTimer()` upon configuration changes, ensuring that updating the refresh frequency immediately applies to the background token refresh timer.
- **Dynamic Preferences UI (`Sources/CodexBar/PreferencesMenuPane.swift`)**: Replaced the static "hourly" auto-refresh text under cost settings with a localized string formatted dynamically using the user's active global frequency label, preventing UI text and execution behavior divergence.
- **Regression Tests (`Tests/CodexBarTests/UsageStoreTokenTTLReproductionTests.swift`)**: Added comprehensive tests validating dynamic TTL logic, verification of cooldowns, and setting transitions.

### Verification Results
- Added new test suite `UsageStoreTokenTTLReproductionTests` and ran successfully on macOS.
- Local lint checks (`swiftlint` and `swiftformat`) all passed cleanly.


Yuxin-Qiao:
Addressed the adaptive-mode review finding.

- token/cost eligibility now uses the same live adaptive delay as the refresh loop
- fixed and Manual behavior remains unchanged
- added adaptive TTL regression coverage for idle and recently-opened-menu states
- focused validation: `swift test --filter AdaptiveRefreshHeuristicsTests` — 10 tests passed

@codex review
@clawsweeper re-review

Yuxin-Qiao:
Triggered a fresh CI run with empty retry commit 29614ae9. The prior arm64 failure was unrelated to this PR: `PlatformGatingTests.claudeAutoPipeline_skipsUnsupportedWebAndUsesCLI` failed with `noAvailableStrategy(claude)` in the Linux runner. The PR code itself has passed lint and x64 build; the retry is to obtain a clean CI result. @clawsweeper re-review

Yuxin-Qiao:
Ran the PR branch at `29614ae` through the compiled production `UsageStore` suite:

```console
$ env HOME="$PWD/.tmp-home" CLANG_MODULE_CACHE_PATH="$PWD/.tmp-clang-module-cache" swift test --filter AdaptiveRefreshHeuristicsTests
✔ Test "adaptive token fetch TTL follows the live adaptive decision delay" passed
✔ Test "adaptive resolves to the live adaptive decision delay" passed
✔ Test "adaptive cadence schedules a reset-boundary refresh through the refresh pipeline" passed
✔ Suite AdaptiveRefreshHeuristicsTests passed
✔ Test run with 10 tests in 1 suite passed
```

This is runtime evidence from the actual `UsageStore` production consumers: the token-fetch TTL, adaptive decision delay, and reset-boundary refresh pipeline all pass together. No UI screenshot is involved.
@clawsweeper re-review

Yuxin-Qiao:
Fixed both scheduling findings on `34d35f8`:

- canceled token timer sleeps now return before they can enqueue a stale refresh;
- timed-out scans start their TTL cooldown at completion, so a one/two-minute cadence cannot immediately restart a ten-minute scan.

Validation: `swift test --filter 'UsageStoreTokenTTLReproductionTests|UsageStoreTokenRetryPolicyTests|AdaptiveRefreshHeuristicsTests'` — 13 tests passed.\n\n@clawsweeper re-review

Yuxin-Qiao:
Follow-up: CI exposed only SwiftLint size thresholds after the functional fix. Commit `3439162` keeps the same behavior while reducing the function/file count by one executable line.

Verified locally with the same command used by CI:
- `./Scripts/lint.sh lint-linux` — 0 violations across 1,426 Swift files

The prior focused regression suite remains green (13 tests): cancellation no longer schedules a stale timer, and timeout cooldown starts at scan completion.

@clawsweeper re-review

Yuxin-Qiao:
Maintainer decision needed to close the remaining product-policy blocker.

This branch deliberately makes token/cost scan cadence follow the global refresh setting, including `.oneMinute` and Adaptive mode. The implementation protects against stale timer work and starts cooldown from scan completion; the focused production `UsageStore` suite already covers the fixed/manual/adaptive scheduling transitions without relying on UI screenshots.

@steipete: is it acceptable for potentially expensive token/cost scans to inherit those global cadences, or should CodexBar enforce a separate minimum interval for them?

Once that policy is confirmed, I will either keep this behavior or implement the requested explicit floor. @clawsweeper re-review

Yuxin-Qiao:
Closing pending a maintainer product decision on whether Manual/Adaptive modes should run background local cost scans. The issue is real, but the cadence, battery, and performance policy should be agreed before reopening a smaller implementation PR.

PR #2113, opened 2026-07-13
Partially fix inherited token accounting across Codex session forks

### LINKED ISSUE #2037: Ultra-mode Terra and Sol sessions can overcount forked context

iam-brain:
## Summary

Partially fixes #2037 by improving how the usage scanner resolves parent sessions and subtracts token totals inherited by forked sessions.

Forked rollout files can retain copied `session_meta` from an ancestor, which made metadata-only lookup unreliable and allowed the same inherited usage to be counted repeatedly. This change:

- treats the session UUID in a canonical rollout filename as authoritative
- prevents copied metadata from impersonating a missing ancestor
- fails open when legacy metadata ownership is ambiguous
- lazily searches outside the normal 30-day window when resolving a referenced long-lived parent
- keeps the broader historical search limited to ancestry resolution rather than expanding the standard scan

Testing against local session data from #2037 reduced the calculated total from roughly 2.23B tokens to 1.24B, compared with approximately 618M reported by OpenAI for the same period. This is still not an exact match, and more work is needed to understand the remaining counter behavior across related and unrelated sessions. However, it closes a substantial part of the gap without introducing speculative family-wide or global suppression that could undercount valid usage.

## Testing

- `swift test --filter CodexSessionFileIndexTests\|Issue2037`
- `make test`
- `make check`
- `git diff --check`

All checks pass.


steipete:
Thank you for identifying the inherited-token direction early; this was directionally right and helped frame the later scanner work.

This implementation now predates and conflicts with #2118, #2208, and the copied-prefix fix in #2228, so I’m closing it as superseded. If the post-#2228 measurement leaves a specific, reproducible residual, a freshly rebased and narrowly scoped PR for that remaining case would be very welcome.

PR #2118, opened 2026-07-13
Fix Codex fork parent cache invalidation

xx205:
## Summary

- persist the parent-session file dependency used by each cached Codex fork
- rebuild a cached child when its parent appears, changes, or resolves to another file
- bind the stored dependency key to the exact stable parent snapshot used during parsing
- add regression coverage for warm-cache invalidation and concurrent parent/child growth

## Why

Fork accounting depends on the parent totals at the fork timestamp, but cache freshness previously tracked only the child file. A child cached while its parent was missing—or before the parent changed—could keep stale inherited totals indefinitely.

Current `main` already contains the leaf-session identity correction from overlapping work. This branch is intentionally narrowed to the remaining parent-cache dependency bug.

## Safety

The cache field is optional. Existing Codex caches rotate through the generated parser producer key; the on-disk schema and artifact layout remain compatible. Parent snapshots are fingerprinted before and after parsing. If the parent changes during both read attempts, the unresolved result is deliberately not reusable on the next refresh.

## Testing

- `swift test --filter 'CostUsageScannerBreakdownTests|CostUsageCacheTests'` — 98 tests passed
- `make check` — passed
- `make test` — 651 selections in 55 groups passed; zero retries or failures
- pre-commit and exact-commit autoreview — no actionable findings
- source-blind synthetic CLI validation was inconclusive because the debug CLI stayed alive before emitting JSON; child fixture checksum remained unchanged and no live credentials or account data were used

Co-authored-by: Xu Xiang <xx205@outlook.com>

PR #2121, opened 2026-07-13
fix: resolve Codex fork duplication when parent is unresolved (#2108)

### LINKED ISSUE #2108: Codex real fork corpus: 5,427 replayed parent events across rewritten timestamps
## Summary

This is a redacted real-world corpus report for the remaining cross-file work in #2037, not a separate root-cause claim.

A high-fanout Codex session produced dozens of child rollout files. The children replay parent `token_count` history with rewritten timestamps. In one verified parent/child pair, the child starts with **5,427 consecutive `last_token_usage` vectors** that exactly match a contiguous section of the parent's immutable token stream. The copied prefix spans about **1.98 seconds** in the child, so a same-second replay heuristic cannot remove it.

## Environment

- macOS 26.2 (25C56), Apple Silicon
- Codex CLI 0.143.0
- CodexBar 0.42.0
- Local timezone: Asia/Shanghai
- Affected workload: Codex Ultra-style high-fanout subagents/forks

## Redacted corpus shape

For one local day:

- 111 rollout JSONL files
- 22 root sessions
- 88 files with an explicit parent reference
- largest parent family: 46 child files
- second-largest parent family: 24 child files

For one sanitized parent/child pair:

```text
parent token_count events:          12,273
child token_count events:            5,716
exact copied child-prefix events:     5,427
matched parent stream start index:    2,449
child replay duration:                1.983 seconds
```

The first and last matching vectors were compared using only the numeric fields in `last_token_usage`:

```json
{
  "input_tokens": 170483,
  "cached_input_tokens": 167680,
  "output_tokens": 244,
  "reasoning_output_tokens": 87,
  "total_tokens": 170727
}
```

```json
{
  "input_tokens": 54031,
  "cached_input_tokens": 52992,
  "output_tokens": 199,
  "reasoning_output_tokens": 0,
  "total_tokens": 54230
}
```

No prompt, response, account, session ID, filename, or local path was retained in this report.

## Observed result

On this corpus, the released CodexBar scanner reported approximately:

```text
Last 2 days: $7,634.61 / 21B tokens
```

The dollar value is included only as a symptom. I am not claiming a corrected bill amount here. The directly verified defect is that thousands of parent events appear again at the beginning of child files with rewritten timestamps.

## Reproduction method

1. Read only the first `session_meta` record from each rollout.
2. Map children to parents using `forked_from_id` or `source.subagent.thread_spawn.parent_thread_id`.
3. Read only `event_msg.payload.type == "token_count"` records.
4. Normalize each event to its numeric `last_token_usage` vector, excluding the envelope timestamp.
5. Compare the child prefix with contiguous sections of the referenced parent's immutable stream.
6. The pair above produced an exact 5,427-event match before the child's new suffix began.

## Expected behavior

- Copied parent history should have one owner across the fork family.
- A child should contribute only genuinely new usage after the copied prefix.
- If the parent is unavailable or ownership is ambiguous, the result should be marked incomplete/estimated rather than silently multiplying copied usage.
- Equal token vectors from distinct events must not be globally collapsed without provenance.

## Relation to current main

I understand that #2082 adds bounded file-local containment and fixtures, while exact cross-file fork-family provenance remains open in #2037. This report is intended to provide the affected real-setup evidence requested there.

I can provide a generated numeric-only fixture derived from this pair, or run an instrumented CodexBar build against the frozen local corpus and return initial-scan plus cached repeat-scan summaries.

Yuxin-Qiao:
This PR resolves #2108.

### Problem
When a child session is forked from a parent, the child session logs often replay the parent session's token usage events at the beginning of the file.
If the parent session is unresolved/missing, and the logs only contain `last_token_usage` (without `total_token_usage`), CodexBar previously had no watermark and no way to filter out the replayed parent events. Sibling child sessions would thus repeat-count the parent prefix, leading to duplicate token usage counting.

### Solution
Use the parent's session ID (`forkedFromId`) instead of the child's `sessionId` as the key-namespace for row deduplication keys when `forkedFromId` is present.
This ensures that any replayed parent turns (which have the same turn ID in all sibling child sessions) are recognized as duplicates of the parent session and are counted exactly once across all sibling sessions, rather than being multiplied for each child session.


Yuxin-Qiao:
@clawsweeper re-review

Yuxin-Qiao:
@clawsweeper re-review

Yuxin-Qiao:
Addressed the cache-ownership review finding.

- when a cached fork owner disappears, remaining siblings in the same fork family are invalidated and reparsed
- added regression coverage for initial scan, owner removal, and repeat scan
- focused validation: `swift test --filter CostUsageScannerBreakdownTests` — 81 tests passed

@codex review
@clawsweeper re-review

Yuxin-Qiao:
Follow-up pushed as c15ef6e6: regenerated `Sources/CodexBarCore/Generated/CodexParserHash.generated.swift` with the repository script. The prior lint failure was the stale generated hash; local `./Scripts/lint.sh lint-linux` now passes with 0 violations. @clawsweeper re-review

Yuxin-Qiao:
Follow-up pushed as f2331160: regenerated the parser hash again after the explicit lint annotation changed the hashed parser source. The latest CI log identified only this generated-artifact mismatch. @clawsweeper re-review

Yuxin-Qiao:
Rechecked the current PR head `f2331160`: the focused Codex fork/cache regression suite remains green with `81` tests passed.

The tests cover the unresolved-parent and fork-family cache transitions in the production scanner/cache path. I also checked the local machine for a real fork-transition corpus, but it does not contain a usable fork metadata sample, so I am not claiming a live corpus reproduction or attaching private session data. The remaining proof boundary is specifically a public/redacted unresolved-parent corpus supplied by the maintainer or CI environment.
@clawsweeper re-review

Yuxin-Qiao:
Fixed the four cache-transition findings on `53ea045`:

- clean unresolved-fork children now reuse their cache instead of reparsing every refresh;
- untagged child rows stay in the child session namespace while provenance-backed turns use the parent scope;
- fork-family invalidation no longer removes retained siblings outside the active scan window;
- cached fork rows are filtered against earlier parent/sibling contributors before reuse.

Validation: `swift test --filter 'CostUsageScannerBreakdownTests|Issue2037ScannerIntegrationTests'` — 84 tests passed.\n\n@clawsweeper re-review

Yuxin-Qiao:
Follow-up: commit `827cbfb` refreshes the repository-required `CodexParserHash.generated.swift` after the scanner cache changes.

Verified locally:
- `Scripts/regenerate-codex-parser-hash.sh --check` — current (`08ed55552fb30dfe`)
- focused fork scanner suites — 84 tests passed (reported in prior update)

The fixture is the committed, redacted 4d90→52bf fork family with a documented scanner oracle (deduped 52,185,847 vs naive 100,916,095 units); it is executable evidence, not UI imagery.

@clawsweeper re-review

Yuxin-Qiao:
Current-head follow-up: `27980d8` resolves the Linux CI race by serializing only `PlatformGatingTests`, whose Claude test seams otherwise overlap under `swift test --parallel`. Production code is unchanged.

Validated locally: `swift test --filter PlatformGatingTests` — 10 tests passed.

Please review the current head together with the prior redacted fork-corpus evidence.

@clawsweeper re-review

Yuxin-Qiao:
Follow-up pushed in 2aec564 for the remaining fork-cache correctness threads.

Addressed behavior:
- regenerate `CodexParserHash` so pre-fix cache entries are invalidated instead of silently preserving the old row-key semantics;
- treat a `turn_context` turn ID as the active provenance, so child-only work is not accidentally parent-scoped;
- when a cached parent disappears, schedule its fork children for reparse;
- rebuild the sibling-rescan state from retained cache rows, preserving parent/earlier row identities while reparsing only the affected children.

Validation:
- `swift test --filter 'CostUsageScannerBreakdownTests'` — 82 tests, 0 failures\n- `Scripts/regenerate-codex-parser-hash.sh --check`\n- `git diff --check`\n\nAdded regression coverage for `turn_context`-provided turn identity.\n\n@clawsweeper re-review

Yuxin-Qiao:
Closing this duplicate candidate to reduce competing Codex fork-accounting contracts. The underlying replay-overcount report remains important in #2108 / canonical #2037.\n\nThe safer merge-ready landing vehicle is #2118: it uses canonical leaf-session identity plus parent-file cache dependencies, includes focused cold/warm cache regressions and frozen-corpus evidence, and avoids this branch's unresolved missing-parent and stale-turn-ID undercount risks.\n\nPlease consolidate any unique coverage only if it is demonstrably absent from #2118; otherwise maintaining one canonical implementation will make review and release validation substantially clearer.

PR #2137, opened 2026-07-13
Stop background Claude keychain discovery

### LINKED ISSUE #2115: Constantly reoccuring keychain prompt for "Claude Code-credentials"
# Summary
Many times a day, the same keychain prompt appears from macOS. They often appear even when I don;t actively click on CodexBar from the menu. It doesn't just show upon regular background refresh, too, since it occurs less frequently than my set 2/5 minute reset.

Clicking "Always Allow" prevents the prompt from occurring every time I open the app from the menu, but it does not prevent the prompt from occurring throughout each day at various points.

## Similar issues
This issue is _not_ the same symptom as #1823, which reports an error whereas my issue is a regular request from macOS. #1161 reports needing to relogin to Claude, but this issue only requires macOS password. #1844 addresses CodexBar opening the browser, which I have not experienced.

## Environment
Claude usage source: Auto
Macbook Pro 2024, 24gb RAM, M4
CodexBar 0.42.1 (and for at least a week prior)
Claude Code version: numerous

Image

m-rokai:
## Summary

- prevent background MCP-only credential discovery from reading `Claude Code-credentials` through Security.framework when the prompt policy is `onlyOnUserAction`
- preserve MCP-only detection for explicit user refreshes and the opt-in `always` policy
- add regression coverage for both interaction modes

## Root cause

The MCP-only payload detector bypassed the normal Claude prompt-policy gate and issued a Security.framework “no UI” read during background availability checks. Legacy Keychain ACLs can still display the macOS password dialog for these reads, causing recurring prompts even though the call was marked non-interactive.

Fixes #2115.

## User impact

Background refreshes no longer touch the Claude Code Keychain item under the default `onlyOnUserAction` policy. Explicit refreshes can still inspect it when needed.

## Validation

- `swift test --filter ClaudeOAuthCredentialsStoreMCPOnlyGuardTests`
- `make check`
- `make test` *(blocked by an unrelated current-main `AdaptiveRefreshTimerTests` baseline failure: “manual mode performs the initial refresh but no recurring ticks” observes 2 refreshes instead of 1; reproduced when running that suite alone)*

No live Keychain read was used for validation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Kimi: reuse fresh signed-in Kimi Code CLI credentials in Auto mode without refreshing or rewriting CLI-owned authentication state. Thanks @Leechael!
 
 ### Fixed
+- Claude: keep MCP-only credential discovery out of background Security.framework reads when Keychain prompts are limited to user actions, preventing recurring macOS password dialogs (#2115).
 - Claude: cache successful CLI version probes for 30 minutes while invalidating on executable changes, avoiding repeated PTY launches without retaining failed or stale wrapper results. Thanks @Yuxin-Qiao!
 - Linux CLI: bootstrap the configured IANA timezone before Foundation startup on non-FHS systems, preventing SIGILL on NixOS (#2127). Thanks @xikhar!
 - Ollama: release temporary dashboard network sessions after each fetch, preventing repeated refreshes from retaining delegates and URL-cache resources. Thanks @astuteprogrammer!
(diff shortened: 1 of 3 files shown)


steipete:
Thanks @m-rokai for finding the MCP-only discovery path that could trigger the legacy Keychain prompt during background work. Your diagnosis identified this path first and the proposed interaction-aware guard was sound.

PR #2191 has now merged as `5f136af0899f9867165bae5b23b899065b2793f0`. It contains the near-identical MCP-only discovery guard plus the broader background-probe policy fix for the opaque Claude CLI and `/usr/bin/security` paths. I’m closing this PR as superseded by that merged fix, with credit to you for the original MCP-only diagnosis.

PR #2150, opened 2026-07-14
Feature/Fix: Add a reversible lineage accounting selector

iam-brain:
## Stack dependency

This is slice 11 of 15. It follows [Feature/Fix: Define lineage promotion gates](https://github.com/steipete/CodexBar/pull/2149); review and merge that PR first.

## Summary
- add internal legacy, shadow, and lineage accounting modes with legacy as the default
- isolate caches by accounting mode and schema for safe rollback/rebuild
- run shadow and lineage through the two-pass scanner with debug-only diagnostics
- combine primary ledger rows with permanent family-scoped containment without double counting
- fail lineage mode closed before cache publication when evidence is incomplete or cancellation occurs

## Validation
- `swift test --jobs 4 --filter CodexLineageAccountingSelectorTests` (7 passed)
- `make check`
- `env SDKROOT=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX26.5.sdk make test` (54 groups, no failures or retries)

Related investigation: [steipete/CodexBar#2037](https://github.com/steipete/CodexBar/issues/2037).

PR #2151, opened 2026-07-14
Feature/Fix: Validate lineage accounting on frozen local histories

iam-brain:
## Stack dependency

This is slice 12 of 15. It follows [Feature/Fix: Add a reversible lineage accounting selector](https://github.com/steipete/CodexBar/pull/2150); review and merge that PR first.

## Summary

- add an opt-in, aggregate-only frozen-history replay for comparing legacy, ledger, and private OpenAI UTC totals without exposing rollout contents, identities, paths, models, or account totals
- stop treating retained metadata IDs shared by fork siblings as physical ancestry edges
- key contained-family fallback by physical owner and scope so sibling rollouts cannot envelope one another
- keep legacy removal explicitly blocked pending repeat validation, ordinary-day reference totals, and acceptable performance

## Real-world result

On the frozen July 9–11 UTC history, ledger selection reduced aggregate absolute error by about 65% relative to the legacy scanner and finished about 3.1% below the OpenAI aggregate.

The remaining evidence is not strong enough to promote the ledger yet. July 9 remains 10.4% low and unchanged from legacy, which points to discovery, parsing, or reference semantics outside lineage deduplication. July 11 is 4.4% high. July 5–8 diverge enough from legacy that they need OpenAI daily totals before they can be classified as corrections or regressions. The replay therefore emits `supportsLegacyRemoval: false`.

The frozen replay covered 819,779 observations across 37 families. It also showed that the current lineage path is materially slower than legacy scanning on this corpus, so performance remains a promotion blocker.

## Tests

- `swift test --jobs 4 --filter 'CodexLineage(Ledger|AccountingSelector|LocalValidation)Tests'` (32 tests passed)
- `make check`
- opt-in frozen local replay
- `make test` (647 selections in 54 groups; 54 passed, 0 failures, 0 retries)

Part of the planned scanner rearchitecture.

Related investigation: [steipete/CodexBar#2037](https://github.com/steipete/CodexBar/issues/2037).


iam-brain:
Superseded by the consolidated four-PR review stack. This slice is now included in #2158; closing this draft to keep the upstream review surface manageable.

PR #2155, opened 2026-07-14
Feature/Fix: Add a shadow lineage ledger for Codex token accounting

iam-brain:
## Summary

Adds the first isolated component of a planned Codex scanner rearchitecture: a typed lineage ledger plus an adapter from the existing fast JSONL parser.

The ledger treats rollout files as overlapping physical views of transitive session families. It now carries a copy-stable logical event identity derived from the turn and token-event ordinal. Physical copies collapse by that identity, unchanged ancestor/descendant re-emissions remain deduplicated, and independent equal-looking sibling events remain additive.

This PR does not change production totals. It establishes the accounting boundary used by the three follow-up PRs.

## Stack

1. **This PR:** ledger foundation and parser adapter
2. [Discovery, accounting dimensions, and shadow execution](https://github.com/steipete/CodexBar/pull/2156)
3. [Conservative evidence, bounded processing, and promotion controls](https://github.com/steipete/CodexBar/pull/2157)
4. [Real-world validation, diagnostics, and performance](https://github.com/steipete/CodexBar/pull/2158)

## Validation

- focused lineage ledger/parser tests, including copied and independent equal-looking events
- redacted frozen-history replay summarized in PR #2158
- `make check`


steipete:
Holding this lineage-ledger stack (#2155–#2158) open as Phase-2 design input for #2037. #2228 is now the bounded baseline fix; we should measure the residual on current `main` before deciding whether the scanner-wide ledger cost is warranted.

The stack’s own validation reported roughly -5% to +9% day-level residuals, so its cost/benefit needs to be reassessed against the new post-#2228 baseline. No merge decision yet; keeping the stack open pending those measurements.

steipete:

PR #2156, opened 2026-07-14
Feature/Fix: Integrate lineage discovery and shadow accounting

iam-brain:
## Stack dependency

This is PR 2 of 4. It follows [PR #2155](https://github.com/steipete/CodexBar/pull/2155); review and merge that PR first.

## Summary

Integrates bounded lineage discovery, token-accounting dimensions, and shadow execution with the existing scanner.

- pulls exceptional long-lived parents outside the normal scan window only when referenced by an included descendant
- indexes physical filename owners for parent discovery; retained metadata aliases cannot suppress the actual parent
- preserves UTC/local-day, model, cached-input, output, pricing, and physical-owner dimensions
- runs privacy-safe lineage comparison after refreshed legacy scans
- keeps the legacy report and cache authoritative
- contains non-cancellation shadow failures without changing production output

## Validation

- focused discovery, ledger, and shadow suites, including retained-parent metadata regression coverage
- redacted frozen-history replay summarized in PR #2158
- `make check`


steipete:

PR #2157, opened 2026-07-14
Feature/Fix: Add conservative lineage safety and promotion controls

iam-brain:
## Stack dependency

This is PR 3 of 4. It follows [PR #2156](https://github.com/steipete/CodexBar/pull/2156); review and merge that PR first.

## Summary

Adds the conservative safety and promotion layer needed before lineage accounting can become authoritative.

- classifies residuals without assigning unavailable history speculatively
- scopes identities and exceptional discovery to one normalized Codex home
- contains malformed, incomplete, unresolved, and ambiguous families
- bounds reconciliation to one family and streams families through two-pass discovery
- rejects files mutated during discovery
- isolates caches by accounting mode and keeps legacy as the default
- requires an opaque authorization produced only by a successful promotion evaluation before `.lineage` can become authoritative
- preserves a whole-scan rollback path

This remains reversible and does not remove legacy accounting.

## Validation

- 48 focused engine, discovery, ledger, promotion, and selector tests passed after the review fixes
- explicit no-authorization fallback coverage
- `make check`

The prior Linux x64 failure was in unrelated Claude platform-gating tests; the same lineage head passed Linux arm64 and macOS, and the corresponding x64 jobs passed on the other three PRs.


steipete:

PR #2158, opened 2026-07-14
Feature/Fix: Validate and optimize lineage token accounting

iam-brain:
## Stack dependency

This is PR 4 of 4. It follows [PR #2157](https://github.com/steipete/CodexBar/pull/2157); review and merge that PR first.

## Summary

Adds the final validation, diagnostic, and performance slice for the shadow-lineage rearchitecture.

- provides an opt-in aggregate-only replay over frozen local histories
- diagnoses reset-epoch and post-branch fingerprint theories without feeding speculative estimates into accounting
- records stage-level timing and bounded family metrics
- removes repeated copy-on-write growth and redundant timestamp parsing
- hashes JSONL during parsing to reduce mutation-check reads
- adds a bounded two-million-observation regression test
- keeps legacy removal blocked and the rollback path intact

The scanner is materially closer to OpenAI's finalized UTC totals, but it is not exact and further real-world validation remains necessary.

## Final UTC comparison

Positive percentages are overcounts versus OpenAI; negative percentages are undercounts.

| UTC day | Live app cache | vs OpenAI | Partial fix | vs OpenAI | Reviewed scanner | vs OpenAI | OpenAI official |
|---|---:|---:|---:|---:|---:|---:|---:|
| 2026-07-09 | 976,079,823 | +14.47% | 764,026,920 | -10.40% | 808,692,108 | -5.16% | 852,682,935 |
| 2026-07-10 | 7,456,347,992 | +371.86% | 1,438,632,374 | -8.96% | 1,625,777,248 | +2.88% | 1,580,199,588 |
| 2026-07-11 | 98,382,045,260 | +7,708.15% | 846,608,173 | -32.81% | 1,375,236,113 | +9.15% | 1,259,990,957 |
| **Aggregate** | **106,814,473,075** | **+2,792.45%** | **3,049,267,467** | **-17.43%** | **3,809,705,469** | **+3.16%** | **3,692,873,480** |

### Sources and interpretation

- **Live app cache:** historical totals retained by the installed CodexBar cache, not a recomputed legacy scan.
- **Partial fix:** scanner after the interleaved-cost containment fix already landed upstream.
- **Reviewed scanner:** the lineage selector after addressing review feedback with copy-stable event identity and physical-parent discovery.
- **OpenAI official:** finalized UTC daily totals used during validation.

The aggregate is much closer than the partial fix, but day-level residuals remain significant. This does not claim exact parity or support removing the legacy fallback.

## Performance

The cold retained-data replay used the same 616-document, 918,677-observation corpus for both paths.

| Calculation path | Total | Change vs legacy |
|---|---:|---:|
| Legacy scanner | 212.269 s | baseline |
| Reviewed lineage scanner | 182.249 s | -30.020 s (-14.14%) |

The lineage time breaks down as follows (minor rounding and timer-boundary overhead account for the remainder):

| Lineage phase | Time | Share of lineage total |
|---|---:|---:|
| Bounded discovery, including referenced-parent lookup | 82.903 s | 45.49% |
| Ledger document loading and hashing | 81.754 s | 44.86% |
| Family reconciliation and deduplication | 17.512 s | 9.61% |
| Family preparation, composition, and timer overhead | 0.080 s | 0.04% |

The ledger held at most one family at a time: 39 families total, 355,168 observations in the largest family, and 856,177 duplicate observations suppressed. This shows that discovery and repeated JSONL loading—not the lineage calculation itself—are the remaining performance targets.

## Validation

- opt-in frozen UTC replay with aggregate-only output
- two-million-observation regression test
- 48 focused lineage tests after review fixes
- `make check`


steipete:

PR #2161, opened 2026-07-14
ci: defer macOS tests for draft pull requests

Yuxin-Qiao:
## Summary
- defer macOS Swift shards for code-changing Draft PRs while retaining Linux lint/build checks
- report Draft deferral as a successful aggregate check with an explicit workflow summary, rather than presenting the policy as a CI failure
- re-run CI when a PR becomes ready for review or returns to Draft
- make the Kiro detached-child timing fixture return and assert parseable context output, so its 8-second bound measures PTY cleanup rather than an empty optional-context probe timing out

## Why
Independent code PR batches can exhaust the scarce hosted macOS runner queue. A Draft PR is already unmergeable, so it should receive fast portable feedback and a clear deferred-validation state, then consume macOS capacity only when promoted for review.

The Kiro timing failure was concrete: the fixture exited with empty `/context` output, while `fetch()` subsequently performs an optional context probe with an 8-second bound. That made the test's own `< 8s` assertion contradictory. This change preserves the strict bound and validates parsed context usage instead of relaxing the timeout.

## Safety
- `ready_for_review` starts a fresh workflow where code changes again require both macOS Swift shards to pass
- `converted_to_draft` starts a new run in the same concurrency group, canceling queued or running macOS work for that branch
- a code-changing Draft PR can be green only for completed portable checks; the workflow summary records that macOS validation is deferred
- non-draft PRs and pushes to `main` retain the existing macOS path gate
- no test retries, timeout relaxations, or required-gate weakening

## Validation
- `bash Scripts/test_ci_path_gate.sh`
- `bash Scripts/test_swift_test_sharding.sh`
- GitHub Actions YAML parse and shell syntax checks
- exact repaired test: `swift test --filter 'CodexBarTests.KiroStatusProbeTests/`fetch returns promptly when usage helper spawns a detached child`'`
- `git diff --check`

## Follow-up identified during validation
The broader `KiroStatusProbeTests` suite still has a separate, pre-existing failure in `accepted pipe output cannot overrun the usage deadline`: a PTY marker file remains after the deadline assertion. This PR does not hide it or add retries; it should be reproduced and fixed as its own process-cleanup regression.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches: [main]
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review, converted_to_draft]
 
 concurrency:
   group: ci-${{ github.workflow }}-${{ github.ref }}
@@ -20,6 +21,7 @@ jobs:
     timeout-minutes: 5
     outputs:
       macos-tests: ${{ steps.macos-tests.outputs.macos-tests }}
+      macos-tests-deferred: ${{ steps.macos-tests.outputs.macos-tests-deferred }}
       macos-tests-reason: ${{ steps.macos-tests.outputs.macos-tests-reason }}
       changed-path-count: ${{ steps.macos-tests.outputs.changed-path-count }}
     steps:
@@ -30,6 +32,8 @@ jobs:
       - name: Detect macOS test impact
         id: macos-tests
         shell: bash
+        env:
+          CI_PULL_REQUEST_DRAFT: ${{ github.event.pull_request.draft || false }}
         run: |
           set -euo pipefail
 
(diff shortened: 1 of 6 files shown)


steipete:
Maintainer verification complete for head `dac601090579b3a16df527d7733101e5287856ec`.

- Reviewed the full six-file diff. The `ready_for_review` and `converted_to_draft` events use the existing workflow concurrency group, draft source changes defer only the macOS shards, and the aggregate gate rejects inconsistent required/deferred/result combinations.
- `bash Scripts/test_ci_path_gate.sh`: passed.
- `bash Scripts/test_swift_test_sharding.sh`: passed.
- `swift test --filter KiroStatusProbeTests`: 55 tests passed, 0 failures, including the detached-child fixture.
- Autoreview: clean; no actionable findings.
- Exact-head CI: changes, lint, Linux x64, Linux arm64, both macOS shards, aggregate `lint-build-test`, and GitGuardian all successful. The earlier draft run correctly skipped the macOS job and passed its aggregate gate.

Merged with merge commit `18d8a051480de4bcfb5913c2df084aeeb46959dd`.

Issue #1814, opened 2026-07-01
kiro.cli (cli) available error=Kiro CLI timed out. (kiro2.10.0)


stormrise:

steipete:
Closing as fixed on current `main` by #1744 (`da693aa7`). The latest release, `v0.37.2`, predates that fix; it still used the pipe-backed Kiro path that produced the reported timeout. Kiro CLI 2.10.0's release notes do not describe a command compatibility change, and the current CLI reference still supports `kiro-cli chat --no-interactive INPUT`.

Fresh current-main proof:

- packaged `CodexBarCLI` reports CodexBar 0.37.3;
- a faithful fixture advertising `kiro-cli 2.10.0` deliberately produced no output over ordinary pipes and hit a 1-second external deadline;
- the packaged command below completed in 1.37 seconds through CodexBar's PTY path, reported provider version 2.10.0, and parsed the expected usage;
- exact child argv was `whoami`, `chat --no-interactive /usage`, and `chat --no-interactive /context`;
- no fixture process remained afterward;
- focused `KiroStatusProbeTests`: 43 passed; final commit autoreview: no actionable findings.

This machine does not have a real authenticated Kiro installation, so the fresh proof is fixture-based; #1744 also contains the existing redacted real `kiro-cli` 2.8.1 and packaged-app proof.

If this persists on a CodexBar release containing `da693aa7`, please reopen with the CodexBar version and redacted output/timing from:

```text
kiro-cli --version
kiro-cli whoami
time kiro-cli chat --no-interactive "/usage"
CodexBar.app/Contents/Helpers/CodexBarCLI usage --provider kiro --source cli --format json --json-only --no-color
CodexBar.app/Contents/Helpers/CodexBarCLI diagnose --provider kiro --format json --pretty
```

PR #2164, opened 2026-07-14
fix(kiro): Add direct API fetch for enterprise/IdC accounts

stormrise:
## Problem

Kiro usage fetching fails for enterprise/IdC accounts because  passes  to the legacy  REST API, causing a  error.

## Root Cause

Enterprise credentials have a  that gets resolved via . When this ARN is passed to , the Kiro 0.9.2 version of the API rejects it as malformed. The same issue was fixed in [kiro.rs v0.6.11](https://github.com/ZyphrZero/kiro.rs).

## Solution

Added a new  that:
- Reads credentials directly from kiro-cli's SQLite database ()
- Makes direct HTTP calls to AWS Q  API
- Uses fixed  version in User-Agent (the version that doesn't require )
- **Does NOT pass ** - this is the key fix
- Falls back to CLI strategy if API fails

## Changes

- **New file**:  - Direct API fetch strategy
- **Modified**:  - Added API strategy as primary, CLI as fallback

## Testing

- [x] Code compiles successfully
- [ ] Test with enterprise/IdC account (needs testing by maintainer)

## Reference

- kiro.rs fix: https://github.com/ZyphrZero/kiro.rs (v0.6.11)
- Key insight: Usage REST APIs (, ) should NOT include


stormrise:
before:
image

after fix:
image

to fix this issue
https://github.com/steipete/CodexBar/issues/1814

stormrise:
@clawsweeper re-review

stormrise:
@chatgpt-codex-connector re-review

steipete:
Thanks for finding the upstream `profileArn` incompatibility and for pointing to kiro.rs v0.6.11. I verified that the AWS Q endpoint and the no-`profileArn` URL shape are legitimate. This implementation is not safe to merge yet, though:

1. External-IdP enterprise auth is incomplete. `queryCredentials` retains only `access_token`, `region`, and `start_url`; the request then explicitly does nothing with `start_url`. The referenced kiro.rs implementation sends `tokentype: EXTERNAL_IDP` for external-IdP credentials and says that header is required. This branch cannot identify that auth method or send the header, so part of the target enterprise population can still fail.
2. Token lifecycle is missing. The strategy sends the cached access token without reading expiry or refresh material. After expiry, automatic mode falls back to the same CLI path this change is intended to bypass. Please define and test fresh, expiring, expired, refresh-success, and refresh-failure behavior for IdC, social, and external-IdP credentials.
3. The new automatic path is too broad. It prefers the API for social credentials too, replacing the existing CLI behavior for ordinary users while dropping CLI-only context usage and login-method data. Restrict the direct path to the affected credential classes, or prove intentional parity for every account type.
4. Credential/privacy boundaries need repair. `docs/provider.md` forbids provider strategies from directly using `FileManager`; this strategy opens the CLI SQLite store itself and hard-wires `URLSession.shared`, which also prevents isolated tests. More seriously, `generateMachineId()` reads the raw Mac platform UUID and sends it to AWS. The referenced implementation uses a Kiro credential/config machine ID or a derived value, not the raw hardware UUID. Please use injectable credential/HTTP seams and never transmit the hardware UUID.
5. Error and reliability handling are incomplete. The raw upstream response body is placed in the user-facing error, while the referenced v0.6.11 code explicitly hides sensitive upstream bodies. The new API-first network path also lacks focused timeout/retry/redaction tests.
6. There are no tests for any new behavior. The existing `KiroStatusProbeTests` exercise only the CLI path. Required coverage: credential-key/auth-method decoding, source selection, URL without `profileArn`, `tokentype`, region fallback, expiry/refresh, response mapping, fallback policy, transport timeout, and error redaction.

Local verification on exact head `0af8d2adf945b49d689eefa8173e0d7ad27491f2`:

- `swift build`: passed, with unused-value warnings in the new parser.
- `swift test --filter KiroStatusProbeTests`: 55 passed; none cover the new API strategy.
- `swiftformat --lint ...`: failed; 29 formatting findings.
- `swiftlint lint --strict ...`: failed; 5 errors.
- Structured autoreview could not run because its secret scanner correctly refused to bundle the credential-bearing diff; no bypass was used.

The screenshots show timeout-before and usage-after, but do not establish account type, selected source, token-refresh behavior, or the external-IdP header. Live enterprise/IdC verification is unavailable in this maintainer environment. Please add redacted proof for those facts after the source/test blockers above are fixed. Leaving the PR open for revision.

stormrise:
@clawsweeper @chatgpt-codex-connector  re-review

steipete:
Closing after maintainer review — for the access pattern, not the goal. The enterprise/IdC gap is real and your endpoint research is genuinely valuable, but this implementation reads OAuth tokens and client secrets directly out of Kiro's SQLite store and performs token refresh against IdC/external-IdP endpoints itself. CodexBar just completed a months-long migration (#2380/#2585) to the opposite principle: the credential owner performs credentialed work, and we never hold or refresh another app's secrets — that pattern was the root cause of our worst recurring defect, and we won't reintroduce it for a new provider.

The demand and your research now live in #2590, which defines the two acceptable designs (CLI-mediated enterprise usage, or a Kiro-provided seam). If you'd like to carry this forward, documenting what current `kiro-cli` exposes for an IdC account would be the perfect first contribution there. Thanks @stormrise.

PR #2167, opened 2026-07-14
Fix duplicate menu bar items under Icon + Percentage style

### LINKED ISSUE #2162: Menu bar items multiply when usage updates occur with "Icon + Percentage" style and "Percent" display mode
Environment:
- Menu bar style: Icon + Percentage
- Display mode: Percentage
- Merged icons mode (single provider shown individually)

Steps to reproduce:
1. Set menu bar style to "Icon and Percentage"
2. Set display mode to "Percentage"
3. Use Codex (or any provider that produces frequent usage changes)
4. Observe the menu bar while usage is updating

Expected behavior:
The menu bar icon and percentage text should update in place without creating duplicate entries.

Actual behavior:
When usage data changes frequently (e.g., during active Codex use), multiple duplicate menu bar items appear to be created. The menu bar icon area seems to be repeatedly rebuilt rather than updated in place, sometimes resulting in multiple copies of the same status item appearing.

Image

ss251:
Fixes #2162.

## Problem

With **Menu bar style = Icon + Percentage** and **Display mode = Percentage**, frequent usage changes could produce duplicate provider menu-bar items instead of updating one item in place.

The creation path checked `statusItems[provider]` before vending, but recorded the new item only after all AppKit setup returned. Any synchronous re-entry during that setup therefore saw no registered item and could vend another one.

## Fix

- Register the raw `NSStatusItem` immediately after creation, before autosave/accessibility setup or callbacks can re-enter vending.
- Keep status-item vending in a small dedicated extension.
- Add a deterministic regression test that injects a usage update during the vulnerable window and verifies the same item is reused.
- Add an Unreleased changelog entry with contributor credit.

## Before / after

| Before | After repeated live refreshes |
| --- | --- |
| Duplicate provider menu-bar items | One stable provider item after repeated refreshes |

The contributor captured six consecutive refresh cycles from a fresh debug bundle, with usage changing from 48% to 68% while one debug status item remained. The data came from a local mock endpoint; the app's real refresh timer drove each update.

## Verification

- `swift test --filter StatusItemReuseRegressionTests` — 1 passed
- `swift test --filter 'MenuBarVisibilityWatcherTests|StatusItemControllerSplitLifecycleTests'` — 61 passed
- `make check` — clean
- `make test` — 643 selections, 54/54 groups, zero failures or retries
- Autoreview — clean, confidence 0.93

Caveat: the pre-fix startup race did not reproduce reliably in three live launches. The deterministic test proves the re-entry window and exact duplicate identity failure; the live after-fix series proves stable identity across real timed refreshes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.43.1 — Unreleased
 
 ### Fixed
+- Menu bar: prevent duplicate provider items when usage updates re-enter initial status-item setup (#2162). Thanks @ss251!
 - Ollama: explain that API-key verification cannot show Cloud quota limits and direct users to browser-cookie mode (#2159). Thanks @kiranmagic7!
 - Claude: suppress duplicate all-model scoped quota rows that could appear as “All models only” beside Weekly.
 
--- a/Sources/CodexBar/StatusItemController+StatusItemVending.swift
+++ b/Sources/CodexBar/StatusItemController+StatusItemVending.swift
@@ -1,0 +1,41 @@
+import AppKit
+import CodexBarCore
+
+extension StatusItemController {
+    /// Lazily retrieves or creates a status item for the given provider.
+    func lazyStatusItem(for provider: UsageProvider) -> NSStatusItem {
+        self.vendStatusItem(for: provider)
+    }
+
+    private func vendStatusItem(
+        for provider: UsageProvider,
+        onCreated: ((NSStatusItem) -> Void)? = nil)
+        -> NSStatusItem
+    {
+        if let existing = self.statusItems[provider] {
+            return existing
+        }
+        return Self.makeStatusItem(
+            statusBar: self.statusBar,
+            identity: .provider(provider),
+            defaults: self.settings.userDefaults,
+            legacyDefaultItemIndex: self.legacyDefaultItemIndex(forNewProvider: provider),
+            onCreated: { item in
+                // Register before invoking the caller/setup callbacks: button configuration and
+                // icon-observation can synchronously re-enter vending for this provider, and an
+                // unregistered item there vends a duplicate (issue #2162).
+                self.statusItems[provider] = item
+                onCreated?(item)
+            })
+    }
+
+    #if DEBUG
+    func _test_vendStatusItem(
+        for provider: UsageProvider,
+        onCreated: @escaping (NSStatusItem) -> Void)
+        -> NSStatusItem
+    {
+        self.vendStatusItem(for: provider, onCreated: onCreated)
+    }
+    #endif
+}
--- a/Sources/CodexBar/StatusItemController.swift
+++ b/Sources/CodexBar/StatusItemController.swift
@@ -299,11 +299,12 @@ final class StatusItemController: NSObject, NSMenuDelegate, StatusItemControllin
         set { self.settings.selectedMenuProvider = newValue }
     }
 
-    private static func makeStatusItem(
+    static func makeStatusItem(
         statusBar: NSStatusBar,
         identity: StatusItemIdentity,
         defaults: UserDefaults,
-        legacyDefaultItemIndex: Int?)
+        legacyDefaultItemIndex: Int?,
+        onCreated: ((NSStatusItem) -> Void)? = nil)
         -> NSStatusItem
     {
         MenuBarStatusItemPlacementPreflight.prepare(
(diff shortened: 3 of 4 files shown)


ss251:
**Real-behavior proof** — fresh build @ `a07201c8` (this PR's head), `./Scripts/package_app.sh debug`, adhoc-signed, isolated debug domain (`com.steipete.codexbar.debug`), macOS 26.2.

Setup: Menu bar style **Icon + Percentage**, display mode **Percentage**, 60s refresh, single provider (sub2api) pointed at a local mock usage endpoint that increments usage on every poll — each poll timestamped in a hit log, so every frame below corresponds to a verified real refresh cycle, not a repaint.

Six consecutive update cycles (~60s apart), tight menu-bar crops — the percentage climbs 48% → 68% while **exactly one** status item persists (the `D`-marked debug instance; the unmarked items visible in wider frames are the separately-running production app and the screen-sharing indicator):

| t0 | t1 | t2 | t3 | t4 | t5 |
|---|---|---|---|---|---|
| t0 | t1 | t2 | t3 | t4 | t5 |

Wider context frame: context

The PR's `StatusItemReuseRegressionTests` — the deterministic mid-vend re-entrancy repro — passes on this head (`1 test passed, 0.5s`).

Honesty notes: the usage *data* comes from a constructed local mock (the *updates* are the app's own real refresh timer hitting it); and the pre-fix duplicate did **not** reproduce in 3 live launch attempts of the parent commit `c852c135` — it is a startup race (an update landing during first vend), for which the regression test remains the deterministic repro.

@clawsweeper re-review

devYRPauli (commented):
The diff itself is clean and the change is a sound one - registering the item in `statusItems[provider]` immediately after `NSStatusBar` hands it back, before the autosave/button setup, is a reasonable check-then-create hardening, and since it's the shared vend path it applies to all non-merged styles without changing steady-state behavior.

One thing I wanted to pin down before reading it as "the fix for #2162": I went looking for a production path that actually hits the check-then-create window, and couldn't find one.

- Usage refreshes arrive on a fresh main-actor turn - `observeStoreIconChanges` dispatches through `Task { @MainActor }` (`StatusItemController.swift:535`), so an update can't land synchronously inside a vend that's already running.
- The per-refresh render path doesn't vend at all: `applyIcon(for:phase:)` reads `self.statusItems[provider]?.button` directly and bails if it's absent (`StatusItemController+Animation.swift:429`) - it never calls `lazyStatusItem`.
- `makeStatusItem` (`StatusItemController.swift:302-325`) is fully synchronous - preflight, `statusBar.statusItem(withLength:)`, autosave + button config - with no await and no call back into `lazyStatusItem`, so I don't see what would re-enter the guard mid-vend.

That lines up with your note that the pre-fix duplicate didn't reproduce live and the deterministic repro is the DEBUG-only re-entrancy hook. So this looks worth taking as defensive hardening regardless.

If the goal is to actually nail the reported duplication under active use, one untouched path that does tear down and rebuild items is `recreateStatusItemsForVisibilityRecovery()` (`StatusItemController.swift:756-774`), invoked from `MenuBarVisibilityWatcher` (`:277`, `:356`, `:407`) on visibility/screen changes. That rebuild runs outside the vend guard this PR tightens, so it might be worth ruling out as the real source of a duplicate before treating #2162 as closed by this change.

steipete:
Landed as `d96d687e9ce19696095429bc87dc5b7e61ec45a9`.

Proof on exact head `3342a73dd38497f6c89e339c091cc491ca79cbb3`:

- `swift test --filter StatusItemReuseRegressionTests`: 1/1 passed.
- `swift test --filter 'MenuBarVisibilityWatcherTests|StatusItemControllerSplitLifecycleTests'`: 61/61 passed.
- `make check`: passed with zero violations.
- `make test`: 643 selections across 54/54 groups; zero failures, retries, or timeouts.
- Autoreview: clean, 0.93 confidence.
- Exact-head CI: all jobs green. One unrelated Linux x64 Claude CLI availability test failed transiently on attempt 1; the same SHA passed Linux ARM64 and Linux x64 rerun attempt 2.
- UI proof: [reported duplicate items](https://github.com/user-attachments/assets/33313e0a-a210-432d-8ac1-17957c5db068) and [stable single-item refresh result](https://raw.githubusercontent.com/ss251/CodexBar/proof-2167-assets/t5.png).

Caveat: the startup re-entry race did not reproduce reliably in three pre-fix live launches. The deterministic regression test injects the vulnerable re-entry window and proves the same provider reuses its first status item. Issue #2162 closed by the merge.

PR #2168, opened 2026-07-14
fix: retain incomplete JSONL tails

ShiroKSH:
## Summary

- retain incomplete JSONL tails, including oversized records that are still being written
- rebuild cost-usage cache artifacts on upgrade so old offsets are not reused
- preserve terminal records clipped by the configured size limit
- cover completed appends, valid final lines without a newline, complete and incomplete truncated EOF records, and cache upgrades

## Checks

- `git diff --check`


devYRPauli (approved):
LGTM. I checked this out and verified both the bug and the fix locally.

Running the PR's new test against the pre-fix source (`CostUsageJsonl.swift` on main, PR test file overlaid) reproduces a real data-corruption bug: `jsonl scanner retries an incomplete final record after append` fails with the still-forming final record partially consumed on the first pass (`firstPass` = `["{\"type\":\"message\",\"id\":\"partial"]`), `resumeOffset` advancing to 31 instead of 0, and the record's later completion mis-parsed as a bogus fragment `"\"}"` on the second pass. On this branch all four tests pass.

So the real-world effect being fixed is: an in-progress final JSONL line (a session record still being written) could be partially read as if complete, its committed offset moved past it, and its eventual completion then decoded as a spurious separate line - silent cost/usage miscount on incremental tailing. The `committedOffset` (only advanced on a newline-terminated line) plus `hasCompleteJSONTail()` (only flush a no-newline tail if it already parses as valid, non-truncated JSON) is the right, narrow fix.

Edge cases I looked at and am satisfied by:
- Empty file / file ending exactly on `\n`: `hasCompleteJSONTail()` no-ops, so no behavior change.
- Truncated tail: the `!truncated && lineBytes == current.count` guard refuses to validate a partial buffer that exceeded `prefixBytes`, which is the conservative choice.
- No double-emit: when the tail is committed early and the lone trailing `\n` arrives on a later scan, `appendSegment`/`flushLine` no-op on the zero-length segment, so the record isn't emitted twice.

Two minor, non-blocking notes:
- The second new test (`accepts a complete final record without newline`) passes on the pre-fix source too, so it reads as a non-regression/documentation test rather than proof of the fix - the `retries...` test is the one carrying the regression coverage.
- `hasCompleteJSONTail()` parses the tail once to validate, and the consumer parses it again when the line is finally accepted - a small redundant parse, not worth changing.

Nice catch on a subtle one.

steipete:
Merged as `423cf7a5c6f0709512eebc5d7f69454dc0b2d08b` after verifying exact PR head `838e64ac66cb5bac953d360ec36bef2688d9108d`.

Verification:

- Audited the full diff and the incremental scanner state machine. Offsets now advance only at newline boundaries or at EOF when the JSON tail is structurally complete. String escapes, nested containers, literals, numbers/exponents, CRLF, trailing whitespace, and records larger than the 256 KiB read buffer are covered. State remains O(1); an incomplete large record is re-read from disk instead of accumulated in memory.
- Merged current `origin/main` into the contributor branch without force-pushing. Cache artifact versions are Codex 10, Claude/Vertex AI 5, and Pi 6. Parallel PR #2228 remained open and did not collide with this scanner change.
- `swift test --filter CostUsageJsonl`: 19 tests in 3 suites passed, including adversarial truncated escape, nested string delimiter, CRLF, over-buffer, literal, number, exponent, and trailing-whitespace cases. Both scanner benchmarks passed.
- `make check`: passed; parser hash current (`577571954a2d36bc`), SwiftFormat 0/1465 changes, SwiftLint 0/1464 violations.
- `make test`: all 659 selections across 55 groups passed after two transient full-group retries; zero final failures or timeouts. The unrelated bounded-child proof was rerun directly with `swift test --filter BoundedChildProcessProofTests`: 3/3 passed.
- Final autoreview: clean; no accepted or actionable findings.
- Exact-head CI: all eight checks green, including both macOS shards, both Linux CLI builds, lint, changes, lint-build-test, and GitGuardian. GitHub reported `MERGEABLE` / `CLEAN` immediately before merge.

PR #2170, opened 2026-07-14
Add Grok weekly usage projections

Zihao-Qi:
## Summary

- Enable the existing reset-window pace projection for Grok weekly quota windows.
- Classify reset-based Grok windows using the menu model deterministic timestamp.
- Cover CLI and web weekly windows, plus monthly and unclassified negative cases.

## Why

Grok was excluded from the generic reset-window pace support used by other providers. Web-derived Grok windows also relied on the wall clock during label classification instead of the model input timestamp, which prevented deterministic weekly classification in the menu model.

The shared `UsagePace.weekly` helper remains responsible for validating reset timing and window duration.

## User impact

Weekly Grok usage rows with a valid weekly reset now show the existing deficit/reserve text, projected run-out estimate, and pace marker. Monthly and unclassified windows remain unchanged.

## Validation

- `swift test --filter GrokMenuCardModelTests` — 4 tests passed
- `swift test --skip-build --filter Grok` — 48 tests passed
- `make check` — SwiftFormat and SwiftLint passed
- `git diff --check` — passed
- Full sharded suite — discovered 643 selections; stopped after an unrelated `CLIOpenAIDashboardCacheTests` group reported 3 cache-restoration failures
- `swift test --skip-build --filter CLIOpenAIDashboardCacheTests` — all 10 tests passed in isolation

## Screenshots

Not included. The behavior is covered through the stable menu model seam without running a live Grok provider probe.

Fixes #2165

PR #2178, opened 2026-07-14
Stabilize shell cleanup timeout test

Zihao-Qi:
## Summary

- Give the Python fixture 2 seconds to reach the escaped-helper state before timeout cleanup.
- Keep the timeout behavior bounded with a 4-second elapsed assertion.
- Always remove partial PID files and kill any fixture helpers that started, even when fixture setup fails.

## Why

The previous 0.2-second timeout also acted as the fixture startup budget. Under the exact CI grouping, Python sometimes had not written the stdout helper PID before the timeout fired, so the test failed while reading the fixture file instead of testing process cleanup.

This exact missing-.stdout failure occurred six times across #2155 and #2170.

## Validation

- `swift test --filter ShellCommandLocatorProcessTests`
- 20 repeated focused runs: 20/20 passed; cleanup test 2.438–2.469 seconds
- 10 repeated runs with SettingsStoreTests, SettingsWindowAppearanceTests, ShellCommandForegroundTests, and ShellCommandLocatorProcessTests: 10/10 passed; cleanup test 2.444–2.464 seconds; full group maximum 3.924 seconds
- `make check`

`make test` was also run. Its default local group size of 12 stopped in the unrelated ClaudeOAuthCredentialsStoreCLIStorageOwnershipTests suite with existing in-memory credential-store failures; the suite reproduces alone without this patch. Upstream CI on the exact base commit c61e01e7 passes both macOS shards with the configured group size of 4.

### DIFF
--- a/Tests/CodexBarTests/ShellCommandLocatorProcessTests.swift
+++ b/Tests/CodexBarTests/ShellCommandLocatorProcessTests.swift
@@ -30,6 +30,18 @@ struct ShellCommandLocatorProcessTests {
             .path
         let stdoutPIDFile = "\(pidFile).stdout"
         let stderrPIDFile = "\(pidFile).stderr"
+        let pidFiles = [stdoutPIDFile, stderrPIDFile]
+        defer {
+            for file in pidFiles {
+                if let pidText = try? String(contentsOfFile: file, encoding: .utf8)
+                    .trimmingCharacters(in: .whitespacesAndNewlines),
+                    let pid = pid_t(pidText)
+                {
+                    kill(pid, SIGKILL)
+                }
+                try? FileManager.default.removeItem(atPath: file)
+            }
+        }
         let script = """
         import os
         import signal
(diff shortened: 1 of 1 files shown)


steipete:
Landed unchanged as `2e01544503004aed685a4a20dcee716d29ff2cdd`.

Proof:

- `swift test --filter ShellCommandLocatorProcessTests` — 2/2 passed; timeout-cleanup case completed in 2.457s
- `make check` — locale/package/script/docs checks passed; SwiftFormat clean; SwiftLint zero violations
- Codex autoreview (`--mode commit --commit HEAD`) — clean; no accepted/actionable findings
- Exact-head CI for `246e1d7002388731b01c3035988094a4c9548cef` — all checks green: lint, both Linux builds, both macOS test shards, aggregate gate, GitGuardian

No live/UI/provider proof applicable: test-only reliability change; no runtime product behavior changed.

PR #2179, opened 2026-07-15
Respect work days for Codex pace

### LINKED ISSUE #2176: Weekly pace calculation ignores "Work days" setting for the Codex provider
With the recent (possibly temporary?) removal of [5-hour rate limits from OpenAI Codex](https://x.com/thsottiaux/status/2076365965915467978), it's more important than ever for users to make sure that they're on pace with regard to their weekly limits to avoid running out.

There is a "Work days" option in the advanced settings, but it seems like this is not factored into the calculations for weekly usage limit pace, at least with the Codex provider. I'm roughly 23 hours into my 7 day limit, but I don't plan to use my work Codex account on the weekend. I have 73% quota remaining, but CodexBar seems to expect me to have 87% remaining, which seems equivalent to 13% usage per day = 1/7 — i.e. it's using a plain calendar-day model instead of my configured 5 work days.

(I also double-checked: I'm on the latest version of CodexBar, my workdays are set to 5, and historical tracking is enabled for my Codex provider)

Image
Image

### Root cause

I dug into this a bit. It seems like the bug is that `workDays` never reaches the code path that's actually computing pace here.

`UsageStore.weeklyPace(provider:window:now:)` (`Sources/CodexBar/UsageStore+HistoricalPace.swift:9-39`) reads `settings.weeklyProgressWorkDays` on line 12, but for the Codex provider, once historical tracking has a usable dataset, it takes the historical branch instead of ever using it:

```swift
// UsageStore+HistoricalPace.swift
if provider == .codex, self.settings.historicalTrackingEnabled {
    ...
    if let historical = CodexHistoricalPaceEvaluator.evaluate(
        window: window, now: now, dataset: self.codexHistoricalDataset)
    {
        resolved = historical   // <- workDays is never passed in here
    } else {
        resolved = UsagePace.weekly(..., workDays: workDays)   // only reached as a fallback
    }
}
```

`CodexHistoricalPaceEvaluator.evaluate(window:now:dataset:)` (`Sources/CodexBar/HistoricalUsagePace.swift:767`) has no `workDays` parameter at all, and its expected-usage curve is built from a plain calendar-elapsed fraction:

```swift
let uNow = Self.clamp(elapsed / duration, lower: 0, upper: 1)   // pure calendar time
...
let linearBaseline = 100 * u   // still the blend floor even with historical data present
```

There's no workday concept anywhere in this function. Even the historical medians it blends in (`CodexHistoricalDataset`) are built from raw calendar-position curves.

This lines up with the numbers in the screenshot: 23h elapsed of a 168h week = 13.7% ≈ the "14% in deficit" / "87% expected remaining" shown is the plain calendar fraction, not a workday-adjusted one. The `≈95% run-out risk` figure is also a tell that this is the historical path: `runOutProbability` is only ever populated by `UsagePace.historical(...)`; the plain `UsagePace.weekly(...)` path (which *does* respect `workDays`) always returns `nil` for it.

So the setting works correctly for every other provider, and for Codex when historical tracking is off or hasn't collected `minimumCompleteWeeksForHistorical` (3) weeks yet, but is silently a no-op for Codex the moment the historical evaluator kicks in, which is presumably the common case for anyone who's had the app running a while.

Zihao-Qi:
## Summary

- treat **Automatic** as the only mode that may use Codex historical pace predictions
- make explicit 4, 5, and 7-day schedules override historical predictions while continuing to collect history
- clarify the Work days and historical-tracking UI copy and document the behavior
- cover Automatic fallback/history and every explicit schedule with focused tests

## Why

CodexBar read the Work days preference, but once enough Codex history existed the historical evaluator took over and ignored an explicit schedule. That made a selected 4/5/7-day plan behave differently over time and contradicted the setting's promise.

The explicit schedule is now authoritative. Historical tracking remains useful in Automatic mode and continues recording in the background so switching back to Automatic works immediately.

Fixes #2176

## Validation

- `make test` — passed all 54 groups / 642 selections
- focused historical pace, preferences, and Codex provider settings tests — passed 78 tests
- `make check` — passed formatting, lint, and repository checks

### DIFF
--- a/Sources/CodexBar/PreferencesMenuPicker.swift
+++ b/Sources/CodexBar/PreferencesMenuPicker.swift
@@ -78,7 +78,7 @@ enum MenuSettingsMenuOptions {
 
     static func weeklyProgressWorkDaysLabel(_ workDays: Int?) -> String {
         switch workDays {
-        case nil: L("Off")
+        case nil: L("Automatic")
         case 4: L("4 days")
         case 5: L("5 days")
         case 7: L("7 days")
--- a/Sources/CodexBar/Providers/Codex/CodexProviderImplementation.swift
+++ b/Sources/CodexBar/Providers/Codex/CodexProviderImplementation.swift
@@ -70,12 +70,16 @@ struct CodexProviderImplementation: ProviderImplementation {
                 }
             })
         let batterySaverBinding = context.boolBinding(\.openAIWebBatterySaverEnabled)
+        let historicalTrackingSubtitle = [
+            L("Stores local Codex usage history (8 weeks) to personalize Pace predictions."),
+            "(\(L("weekly_progress_work_days_title")): \(L("Automatic")))",
+        ].joined(separator: " ")
 
         return [
             ProviderSettingsToggleDescriptor(
                 id: "codex-historical-tracking",
                 title: "Historical tracking",
-                subtitle: "Stores local Codex usage history (8 weeks) to personalize Pace predictions.",
+                subtitle: historicalTrackingSubtitle,
                 binding: context.boolBinding(\.historicalTrackingEnabled),
                 statusText: nil,
                 actions: [],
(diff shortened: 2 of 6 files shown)


steipete:
Merged as `bb142354d8d929c23335a42a28a528092b1aaf17` after exact-head verification at `e5addbba47fd4f96346e6eb98141e9803d1c4819`.

Verification:
- `git diff --check origin/main...HEAD` passed.
- `swift test --filter HistoricalUsagePaceTests` passed: 49 tests, including explicit 4/5/7-day schedule coverage and Automatic historical pacing.
- `swift test --filter PreferencesPaneSmokeTests` passed: 23 tests, including the Automatic label.
- Autoreview reported no accepted/actionable findings.
- All 8 PR checks passed: lint, both Linux builds, both macOS test shards, changes, lint-build-test, and GitGuardian.

PR #2191, opened 2026-07-15
Fix Claude background Keychain prompts

Yuxin-Qiao:
## Summary

- prevent background Auto refresh from launching either opaque Claude CLI Keychain path unless the user explicitly selected the `Always allow Keychain prompts` policy:
  - `claude auth status --json` preflight
  - delegated OAuth-expiry `/status` refresh
- prevent planning an expired Claude CLI OAuth credential for background Auto refresh when that delegated refresh is disallowed
- honor the global Disable Keychain Access setting before invoking the Claude CLI, including its child `/usr/bin/security` access
- stop background MCP-only Claude credential discovery from touching Keychain under the default user-action-only policy
- add stubbed regressions plus an opt-in live macOS proof for background and explicit-user behavior

## Root cause

CodexBar had two background Claude CLI paths: the `claude auth status --json` preflight and the delegated OAuth-expiry `/status` refresh. Both can invoke `/usr/bin/security` outside CodexBar's in-process Keychain prompt policy. That allowed recurring `Claude Code-credentials` dialogs during background refreshes even when prompts were disabled. Separately, a Security.framework MCP-only discovery path could read the same item from background work despite the user-action-only policy.

This patch treats child-process Keychain behavior as opaque: background Auto refresh does not start it unless the user explicitly opts into background Keychain access. Explicit refresh and explicit CLI use remain available.

Fixes #2115.

## Live affected-Mac proof (redacted)

Tested on a real macOS account with Claude Code 2.1.207 installed, a live OAuth login, and the default restrictive prompt policy (`onlyOnUserAction`). The opt-in proof exercises production code, not a process stub:

```console
$ LIVE_CLAUDE_KEYCHAIN_PROOF=1 swift test --disable-sandbox \
    --scratch-path /private/tmp/codexbar-pr2191-proof \
    --filter ClaudeKeychainLiveProofTests

Suite ClaudeKeychainLiveProofTests started.
Test "live background Auto skips the opaque Claude Keychain boundary" started.
Claude OAuth delegated refresh skipped by Keychain prompt policy
Test "live background Auto skips the opaque Claude Keychain boundary" passed after 0.004 seconds.
Test "live explicit user auth probe reports Claude login" started.
Test "live explicit user auth probe reports Claude login" passed after 0.195 seconds.
Suite ClaudeKeychainLiveProofTests passed after 0.200 seconds.
Test run with 2 tests in 1 suite passed after 0.200 seconds.
```

[...]

PR #2192, opened 2026-07-15
Clarify Claude OAuth recovery action

Yuxin-Qiao:
## Summary

- clarify that a suppressed Claude OAuth background repair requires clicking **Refresh** in the CodexBar menu
- remove the incorrect claim that merely opening the menu retries the repair
- cover the user-facing recovery direction in both macOS and Linux delegated-refresh regression tests

## Why

With the Keychain prompt policy set to **Only on user action**, background OAuth repair intentionally fails closed so CodexBar cannot trigger Claude, a Keychain prompt, or a browser launch. Automatic refresh after opening the menu still runs in the background, while the explicit Refresh action is user-initiated. The previous error text incorrectly promised that opening the menu would retry the repair.

## Validation

- `swift test --filter 'oauth delegated retry only on user action background suppresses delegation|appOAuthBackgroundRespectsPlatformKeychainPromptPolicy'`
- `make check`


Yuxin-Qiao:
Folded this recovery-copy correction into #2191 so the Claude Keychain/OAuth behavior has one canonical merge path.

PR #2194, opened 2026-07-15
Remove obsolete provider toggle store

Yuxin-Qiao:
## Summary

- remove the unused legacy `ProviderToggleStore` and its obsolete tests
- retain legacy `providerToggles` migration in `CodexBarConfigMigrator`
- add regression coverage proving legacy toggle values are preserved in the config-backed settings path
- make arm64 CI package setup use HTTPS for the Ubuntu ports mirror and retry transient apt downloads

## Why

Provider enablement is now config-backed. The legacy store has no production callers, while `CodexBarConfigMigrator` remains responsible for importing existing `providerToggles` data.

The arm64 CI runner repeatedly timed out reaching the HTTP-only Ubuntu ports mirror before it could install Swift's system dependency. The workflow now prefers HTTPS on that runner and retries apt downloads.

## Test

- `swift test --filter CodexBarConfigMigratorTests`
- `make check`

## Real app migration proof

I built a fresh, isolated debug `CodexBar.app` from this branch and launched its actual app executable with a temporary config path. A test-only in-process injector supplied the legacy, non-sensitive defaults before app startup; keychain access was disabled. The app then wrote the migrated config file:

```console
fresh_debug_bundle=CodexBar.app
legacy_providerToggles={codex:false, claude:true}
migrated_config_enabled={"claude": true, "codex": false}
```

This demonstrates the real app startup path preserves legacy provider enablement in the config-backed settings store, without reading an account, cookie, token, or existing user config.


Yuxin-Qiao:
@clawsweeper re-review

Added real app startup proof to the PR body. A freshly built, isolated debug `CodexBar.app` migrated legacy `providerToggles` (`codex: false`, `claude: true`) into the temporary config with the same enabled values. No account, cookie, token, or existing user config was read.

Yuxin-Qiao:
Closing to keep the open PR set focused on higher-impact user-facing fixes. The dead-code cleanup is valid but low-priority, and its shared arm64 CI workaround should not ride on this feature branch.

PR #2198, opened 2026-07-15
Hide unlimited Copilot Chat quota bars

Zihao-Qi:
## Summary

- omit Copilot quota snapshots explicitly marked `unlimited` from metered usage windows
- preserve finite Premium and Chat quota reporting
- keep the Copilot plan and identity available when an unlimited marker is the only quota in the response
- add regression coverage for unlimited-only, finite Premium plus unlimited Chat, and finite Chat payloads

## Why

Copilot can return `chat_messages` with `unlimited: true`, zero entitlement, and zero remaining. The fetcher previously converted that marker into a 0%-used secondary window with no reset date, so CodexBar rendered a permanently empty Chat bar that looked actionable but could never change.

Filtering unlimited markers at the fetcher boundary keeps real metered quotas intact while preventing non-metered metadata from becoming a usage bar.

## Validation

- `swift test --filter CopilotUsage` — 34 tests passed
- `make check` — SwiftFormat and SwiftLint passed
- `make test` — 642 selections across 54 groups passed with zero retries or failures

## Screenshot
before
Screenshot 2026-07-15 at 11 25 21

after
Screenshot 2026-07-15 at 11 29 04 AM


steipete:
Maintainer closeout at `f76caf67f7233e6b7ee166a9cf61fa6c619d9e90`:

- Exact-head CI is green: https://github.com/steipete/CodexBar/actions/runs/29495951811
- `swift test --filter CopilotUsage` passed.
- The viewport-refresh synchronization regression passed 10/10 focused repetitions, then passed both exact-head macOS CI shards.
- `make check` passed with zero SwiftFormat or SwiftLint violations.
- Autoreview is clean for the Copilot fix and the CI flake stabilization; both actionable GitHub review conversations are resolved.

The implementation preserves finite monthly fallback quotas, keeps unlimited markers in the decoded model for identity/plan behavior, and suppresses only unlimited rate-window bars.

PR #2202, opened 2026-07-15
Coalesce Claude Keychain prompts

farzanariel:
## Summary

- coalesce Claude OAuth Keychain reads within one provider refresh request
- replay the exact completed record or failure to queued and late-scheduled callers
- prevent expired credentials, cancellation, or authentication failures from producing a queue of native Keychain dialogs
- start a new coalescing epoch for a later user refresh and for the intentional post-Claude-CLI repair retry
- add deterministic coverage for expired, valid, denied, late, and post-delegation paths

## Root cause

Claude Keychain reads were protected by a mutex, but only serialized. When a read returned an expired credential, waiting callers rejected the expired memory/cache entry and each performed another interactive read. Work scheduled later in the same refresh could do the same after the first caller released the lock.

The fix gives each provider refresh a request ID and records a generation-tagged interactive-read outcome. Calls in the same request, plus callers already waiting on the prompt mutex, reuse that exact record or failure. A separate refresh receives a new request ID and can retry immediately. The deliberate post-delegation credential reload also receives a fresh epoch so an earlier expired result cannot mask Claude CLI repair.

## User impact

A single menu-open or Refresh action can require one Keychain authorization, but it no longer fans out into many duplicate dialogs when Claude credentials are expired or the first authorization is cancelled. A later user action is still allowed to retry.

## Validation

- rebased onto current `main` at `bad4bebe`; GitHub reports the PR mergeable
- 13 independent review passes total; the final three focused on lock ordering, behavior parity, and test/platform hygiene
- 90 Claude-focused tests pass on the rebased head, including:
  - concurrent expired result: one interactive read, exact replay to concurrent and late callers
  - concurrent valid result: one interactive read, exact `.claudeKeychain` replay rather than a reconstructed cache result
  - cancellation/denial with cooldown: exact failure replay within one request, then a fresh user request performs a second read
  - post-Claude-CLI repair: the retry receives a distinct request ID
  - Claude credential store, access gate, usage, and fetch-strategy suites
- `make check` passes: all portable checks, SwiftFormat, and SwiftLint (0 violations)
- signed local app bundle built from `b73d2be1` and verified with `codesign --verify --deep --strict`
- `git diff --check` passes

`make test` was also run locally. It discovered 654 Swift Testing selections and stopped on the current-main `AdaptiveRefreshTimerTests` case `manual mode performs the initial refresh but no recurring ticks` (`observed 2`, `expected 1`). The failure reproduces unchanged when the PR's only `UsageStore` edit is temporarily removed, so it is not caused by this patch. The five legacy XCTest-only files cannot run under standalone Command Line Tools because Apple ships `XCTest.framework` only with full Xcode; GitHub's Xcode runner remains authoritative for those.

## Live proof

The exact signed build is prepared for a redacted menu/Refresh proof run. This section will be updated with the observed dialog count and independent-retry result before requesting the maintainer bot's re-review.


steipete:
Verified and merged.

Current `main` still had the user-initiated refresh fan-out: the prompt mutex serialized interactive reads, but expired/error outcomes were not shared. I merged current `main` into the contributor branch, retained request-scoped exact-result replay, and hardened the boundaries so prompt-policy/read-strategy changes plus cache or credential invalidation cannot reuse stale outcomes.

Local verification (deterministic test stores and read overrides; no real Keychain prompt):

- `swift test --filter ClaudeOAuthPromptCoalescingTests`: 5/5 passed, covering concurrent reuse, a late caller after completion, new-request retry, policy change, and credential invalidation.
- `swift test --filter ClaudeOAuthDelegatedRefreshEpochTests`: 1/1 passed.
- `swift test --filter ClaudeOAuthKeychainAccessGateTests`: 7/7 passed.
- `swift test --filter ClaudeOAuthCredentialsStorePromptPolicyTests`: 12/12 passed.
- `swift test --filter KeychainPromptSafetyAuditTests`: 10/10 passed.
- `make check`: passed; SwiftFormat clean and SwiftLint reported 0 violations across 1,473 files.
- `git diff --check`: passed.
- Required autoreview: clean; no accepted/actionable findings.

Exact-head CI for `77f41bc0885760a4b59bc1b4b423d309704ba21d` was fully green in run https://github.com/steipete/CodexBar/actions/runs/29546394343: changes, lint, both Linux builds, both macOS test shards, aggregate, and GitGuardian all passed.

Merged with merge commit `03c4ec5c174df6bc8d80f4699e06616de2914d0a`.

Issue #2204, opened 2026-07-15
Finish migrating test-override seams off unsynchronized process-global statics to task-local scopes


anagnorisis2peripeteia:
The provider/keychain override seams that tests use exist in two forms: a per-test-isolated
`@TaskLocal` store exposed via a `with…OverrideForTesting { }` scope, and a **legacy process-global**
`nonisolated(unsafe) static var …Override` with a `set…OverrideForTesting(_:)` setter. The getter
checks the task-local first, then the process-global.

The codebase is already ~90% on the task-local form — **624** `with…OverrideForTesting` call sites vs
**60** `set…OverrideForTesting`. The remaining process-global statics are `nonisolated(unsafe)` and
written by parallel tests — a data race under `swift test --parallel`, the same class as #2201 / #2203.
It's flaky (fires only when one test's write overlaps another's read), which is why plain CI doesn't
catch it.

**No production impact** — these statics are never written outside `…ForTesting` seams, so the app
only ever reads their default `nil` and runs the real path. But it's a latent hazard: any future
regression test that writes one under parallel execution will flake, exactly as the stub-handler
races did.

**Fix (finishes the migration already underway):** convert the `set…OverrideForTesting(v) …
set…(nil)` test sites to the existing `with…OverrideForTesting(v) { … }` task-local scope, then remove
the process-global static + its setter. Task-locals are per-task-isolated, so this **eliminates** the
race rather than locking it, and matches where the getter and the other 624 call sites already point.

### Status (verified against `main` on 2026-07-16)

The migration is **not yet complete** — a current-`main` inventory finds **9** process-global
`nonisolated(unsafe)` test-override statics still present. Merged work so far covers subsets:

- ✅ Cookie importers (`importSessionOverride…` across Alibaba / MiMo / Perplexity / Manus / Windsurf / Devin) — migrated task-local, #2212
- ➖ Claude keychain overrides — partial, #2207 (several statics still remain, listed below)
- ➖ `CookieHeaderCache` display-intervals — **locked** with `NSLock` as a stopgap, not yet task-local, #2205

### Status — COMPLETE (all genuine unsynchronized test-override statics migrated)

- [x] **PR1 — #2239:** remove 3 already-dead `private` statics — `KeychainCacheStore.globalServiceOverride`, `KeychainAccessPreflight.checkGenericPasswordOverride`, `ClaudeOAuthCredentials.credentialsURLOverride` (pure dead-seam removal).
[...]

PR #2205, opened 2026-07-15
Lock CookieHeaderCache display-interval test overrides

anagnorisis2peripeteia:
## Summary

`CookieHeaderCache.displayStalenessIntervalOverride` and `displayUnavailableRetryIntervalOverride` are mutable static test overrides. Production display code reads them while tests can update them from another thread, creating a data race under parallel Swift Testing.

This uses the existing `NSLock` pattern already used for `legacyBaseURLOverride`: both override reads and writes now use one dedicated lock. Production defaults and behavior stay unchanged.

The regression test was strengthened during maintainer review. Its ThreadSanitizer stress run now exercises the actual `CookieHeaderCache` setters and display-interval readers instead of a duplicate probe.

Exact-head CI also exposed an existing shell cleanup test timeout on a loaded macOS runner before the child could publish its PID-readiness marker. The setup budget is now 5 seconds instead of 2 while preserving child-side readiness semantics and the strict cleanup assertions.

## Proof

- `CODEXBAR_TSAN_STRESS=1 swift test --sanitize=thread --filter DisplayIntervalOverrideConcurrencyTests`: 1/1 passed; no ThreadSanitizer race.
- `swift test --filter ShellCommandLocatorProcessTests`: 2/2 passed.
- `make check`: passed; zero violations.
- `make test`: 644 selections across 54/54 groups; zero failures, retries, or timeouts.
- Autoreview: clean, 0.91 confidence.

No changelog entry: internal test synchronization only.

Part of #2204.

### DIFF
--- a/Sources/CodexBarCore/CookieHeaderCache.swift
+++ b/Sources/CodexBarCore/CookieHeaderCache.swift
@@ -296,6 +296,10 @@ public enum CookieHeaderCache {
         self.displayIntervalOverrideLock.withLock { self.displayUnavailableRetryIntervalOverride = interval }
     }
 
+    static func displayIntervalsForTesting() -> (staleness: TimeInterval, unavailableRetry: TimeInterval) {
+        (self.currentDisplayStalenessInterval, self.currentDisplayUnavailableRetryInterval)
+    }
+
     static func resetDisplayCacheForTesting() {
         self.displayCacheLock.lock()
         self.displayCache.removeAll()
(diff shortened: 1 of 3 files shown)


steipete:
Landed as `6ad100452f81d9a00e972ce66feea93c239bb6a5` after maintainer review and exact-head CI.

Proof on `ba363186c2c7d5a7c7a84b3548e24c6a7bdb8a23`:

- Production-backed ThreadSanitizer stress test: 1/1 passed; no race.
- Shell cleanup focused tests: 2/2 passed with child-side PID readiness preserved.
- `make check`: zero violations.
- `make test`: 644 selections, 54/54 groups, zero retries or timeouts.
- Autoreview: clean, 0.91 confidence.
- Exact-head CI: all eight checks green.

No user-visible behavior change. Thanks @anagnorisis2peripeteia.

PR #2207, opened 2026-07-16
Remove Claude keychain global test overrides

kiranmagic7:
## Summary
- remove the legacy process-global Claude keychain data/fingerprint test overrides
- keep the Claude keychain test path on the existing task-local override scope
- drop now-redundant nil cleanup calls from the affected tests
- add a barrier-backed concurrency regression proving independent task-local values and parent restoration
- stabilize the Linux planner fallback test by scoping its synthetic CLI resolver override

Addresses part of #2204.

## Testing
- `swift test --filter 'keychain overrides stay isolated across concurrent tasks'` (1 test passed)
- `make check`
- `make test` (646 selections, 54 groups, zero failures, zero retries)
- `autoreview --mode branch --base origin/main --parallel-tests 'make test'` (clean, 0.93)
- final `autoreview --mode branch --base origin/main` after Linux CI repair (clean, 0.95)

No changelog entry: internal test synchronization only; no user-visible behavior change.

### DIFF
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeOAuth/ClaudeOAuthCredentials+TestingOverrides.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeOAuth/ClaudeOAuthCredentials+TestingOverrides.swift
@@ -2,8 +2,6 @@ import Foundation
 
 #if DEBUG
 extension ClaudeOAuthCredentialsStore {
-    nonisolated(unsafe) static var claudeKeychainDataOverride: Data?
-    nonisolated(unsafe) static var claudeKeychainFingerprintOverride: ClaudeKeychainFingerprint?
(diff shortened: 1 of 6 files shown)


steipete:
Landed as `b62c9020f283fefa090caf1b5c5b82029ff28cbd`. Thanks @kiranmagic7!

Maintainer proof on final head `13807e7ad60d09dd9033d8b34e02f020e16dd809`:

- focused task-local isolation regression: 1/1 passed
- `make check`: clean
- `make test`: 646 selections, 54/54 groups, zero failures, zero retries
- final branch autoreview: clean (0.95)
- exact-head CI: 8/8 green, including Linux x64/ARM64 and both macOS shards

CI also exposed a pre-existing ARM-only Linux planner-test flake. The maintainer follow-up scopes the synthetic Claude CLI resolver override to the fetch under test; the repaired ARM job passed in 6m39s.

Issue #2193, opened 2026-07-15
Bug: Local token usage & estimated cost are under-reported due to ignoring subagent rollout session logs


qiuruiyu:
#### **Summary**
CodexBar's local log parser under-reports the total daily/monthly token usage and estimated cost when users run complex tasks. This happens because CodexBar's local scanning logic relies solely on top-level session IDs registered in `~/.codex/session_index.jsonl` and fails to scan rollout files generated by spawned **subagents (child threads)**.

#### **How to Reproduce / Technical Details**
1. Run a complex coding task in Codex CLI or Claude Code via the Codex MCP plugin that triggers agentic loops (where the model defines and spawns subagents).
2. Codex CLI handles these subagents by creating separate execution threads. In the filesystem, they write their own session rollout logs under:
   `~/.codex/sessions/YYYY/MM/DD/rollout-<subagent-session-id>.jsonl`
3. Because these subagents are internal child threads, their session IDs are **not** registered in the top-level user session index `~/.codex/session_index.jsonl`.
4. CodexBar's local parser reads `session_index.jsonl` to locate files to parse. As a result, it completely bypasses the subagent rollout logs, leading to a massive gap in reported vs. actual token usage.

#### **Real-world Example from Logs**
On **July 14, 2026** (local timezone UTC+8), I ran a task that spawned multiple subagents (`Gauss`, `Hubble`, `Schrodinger`, etc.).
* **Actual Usage (Direct Log Sum of all files in directory)**: **331M tokens** with an estimated cost of **$269.96**.
* **CodexBar's Chart Output**: **83M tokens** with an estimated cost of **$65.71**.

##### **The math behind the discrepancy:**
* The parent session (`019f6137-a3e2-7c72-ace4-126c54369a4a`) cost **$60.07**. Combined with minor VSCode / Exec sessions, it totaled exactly **$65.71** (what CodexBar displayed).
* However, the subagent sessions (`Gauss` costing $52.37, `Hubble` costing $52.04, `Schrodinger` costing $51.23) were completely ignored by CodexBar's scanner because their session IDs were missing from `session_index.jsonl`.
* Note: Although the cloud-based "Weekly/Monthly Limit" percentage (fetched via browser cookies/API) correctly counted the subagent requests, the "Estimated Cost" charts, project list, and total token usage meters were under-reported.

#### **Expected Behavior**
[...]

PR #2208, opened 2026-07-16
Fix restarted Codex subagent usage accounting

harjothkhara:
### Summary

- recognize explicit Codex `source.subagent` provenance in both fast and fallback metadata parsers
- treat spawned subagent counters as independent instead of subtracting the parent's unrelated cumulative baseline
- key each rollout by its per-thread `id`, not the multi-agent tree's shared `session_id`
- preserve inherited-baseline subtraction for ordinary forks and invalidate stale parser caches

Refs #2193.

### Why

The reporter's follow-up showed that CodexBar found and scanned all rollouts. The undercount happened later: `forked_from_id` made CodexBar subtract the parent's cumulative total from a subagent whose counter restarted at zero.

Codex protocol metadata marks spawned children with `source.subagent`. Those counters are independent. A bare `parent_thread_id` remains an ordinary fork because it can carry a continuing parent counter. Current multi-agent metadata also uses one shared `session_id` for the tree and a unique `id` for each thread, so rollout identity must prefer `id` across both supported metadata locations.

### Verification

- `swift test --filter CostUsageScannerBreakdownTests` — 87 tests passed
- `swift test --filter CostUsageScannerTests` — passed
- `make check` — 0 SwiftLint violations; parser hash, formatting, shell, docs, and repository checks passed
- autoreview — clean after fixing shared-tree identity and mixed metadata layout findings
- regression uses parent and child records with the same `session_id` but distinct root-level `id` values

### Behavior proof

An isolated production-style parent fixture reports 60,000,000 input / 48,000,000 cached / 3,000,000 output tokens. Its restarted subagent ends at 62,200 / 51,000 / 3,200. The combined report now returns 60,062,200 / 48,051,000 / 3,003,200, including the child exactly once.

The negative control remains intact: an ordinary continuing fork at 1,001 / 900 / 101 against a 1,000 / 900 / 100 parent baseline reports only 1 / 0 / 1.

### AI assistance

Codex assisted with investigation, implementation, tests, review, and proof preparation.


steipete:
Maintainer closeout on `59a18fcc4e3729560f2e02ee701ef5202b1fd631`:

- Confirmed the fix uses explicit Codex `source.subagent` provenance and preserves parent-baseline subtraction for ordinary forks.
- `make check`: passed, including current parser hash, SwiftFormat, and SwiftLint with 0 violations.
- `swift test --filter CostUsageScannerBreakdownTests`: 87/87 passed.
- Synthetic black-box `CodexBarCLI cost` proof: a restarted subagent contributed its full +50 input / +10 cache-read / +5 output; an ordinary continuing fork contributed only that same post-fork delta; a warm-cache repeat was identical. No credentials, browser data, or Keychain access used.
- Hardened session identity selection so multi-agent rollouts use their distinct thread `id`, not their shared tree `session_id`; both fast and fallback parsers remain aligned.
- Structured autoreview: original branch and the final identity-hardening commit both clean, no accepted/actionable findings.
- Added the missing contributor credit for @harjothkhara to the changelog entry.
- Exact-head CI: all required Linux, macOS, lint, aggregate, and security checks passed.

Merged as `8489602f84ca278a4cb6944a97cca009b4412a87`. Thanks @harjothkhara and @qiuruiyu!

hhh2210:
This merge regresses the copied-parent-prefix `source.subagent` layout already documented in #2108.

I froze 2,010 local JSONL files and ran the same CodexBarCLI cost scan with empty isolated caches. Only the scanner revision changed:

| Revision | Jul 16 tokens | Unknown |
| --- | ---: | ---: |
| pre-#2208 `e419a88d` | 289,150,407 | 4,100,938 |
| #2208 `8489602f` | 4,059,888,540 | 886,306,282 |
| after #2118 `2acf34cd` | 4,059,888,540 | 886,306,282 |

A warm repeat on the #2208 cache was identical. #2118 cannot repair this because #2208's subagent branch bypasses the parent resolver.

One representative child starts with a leaf `session_meta` marked `source.subagent`, followed by the parent `session_meta` and 1,845 copied `token_count` rows stamped at the spawn second. That prefix grows from 32,584 to 298,905,426 tokens. Only 20 later rows belong to the child and grow the counter to 300,571,631.

- pre-#2208 counted 1,666,205 tokens
- #2208 counted 300,571,631 tokens
- the overcount is exactly the 298,905,426-token copied prefix

A conservative same-spawn-second check across 38 Jul 16 subagent rollouts finds 18 non-empty copied prefixes totaling 3,527,023,205 tokens. Their later growth totals 387,607,038 tokens. The Unknown spike has the same cause: copied parent rows arrive before the leaf `turn_context` supplies its model.

The unconditional branch in `CostUsageScanner.swift` that treats every `source.subagent` file as an independent counter is the problem. `source.subagent` is provenance, not proof of counter semantics. The synthetic fixture in #2208 covers reset-only children, while #2108 already documents the other real protocol shape.

The follow-up needs phase-aware accounting:

- resolve the parent baseline for every fork
- suppress the copied prefix through the fork boundary
- switch to an independent epoch only after an actual counter reset is observed, preserving the #2193 case
- cover reset-only, copied-prefix, nested depth-2, mixed-family, cold-cache, and warm-cache cases

Please do not ship the current scanner state. Unless a patch is already in flight, I will take the follow-up and provide a real-shape fixture plus frozen-corpus proof.

Refs #2108 and #2037.

hhh2210:
Follow-up implemented in #2228.

The patch keeps #2208's reset-only behavior, but moves the decision to a rollout-shape classifier. A subagent with only its leaf metadata owns an independent counter; a subagent with embedded ancestor metadata is a copied-prefix rollout. For the copied shape, the scanner either starts at the verified local leaf marker or retains the parent-baseline fallback.

That removes the unconditional `source.subagent` assumption without reverting the valid #2193 fix. On the frozen 2,012-file corpus, Jul 16 falls from 4,086,288,236 tokens / 886,306,282 Unknown on current main to 231,475,556 / 0 Unknown. Cold, warm, and forced scans are identical.

The PR also keeps #1042's producer-hash contract: parser semantics rotate the generated hash; `codex-vN` remains reserved for incompatible cache schema/layout changes.

PR #2135, opened 2026-07-13
Improve DeepSeek detailed usage

Zihao-Qi:
## Summary

- enrich the existing DeepSeek balance with today/current-month cost and token totals, request counts, model/category breakdowns, and a 30-day token chart
- discover DeepSeek Platform sessions from Chrome local storage, validate every discovered profile, and offer only valid profiles in Settings
- keep the balance and profile selector stable while detailed usage refreshes or a profile changes
- scope each profile choice to the active API credential and saved-account slot to prevent cross-account data mixing
- bound optional Platform work so a slow or uncooperative browser/session probe cannot delay the required balance

## Why

DeepSeek's documented API exposes the account balance but not the detailed usage visible at `platform.deepseek.com/usage`. The Platform dashboard uses private endpoints for that data.

This change keeps the documented balance fetch authoritative and required. Platform usage is optional enrichment: failures, invalid sessions, and timeouts preserve the balance instead of failing the provider.

## Compared with #1910

This builds on the same general idea of reading DeepSeek Platform usage, but uses a narrower and more defensive session path:

- reads the Platform `userToken` from live Chrome local storage instead of relying on cookies or persisting a manual browser token
- discovers all Chrome profiles and validates each session before showing it
- lists only valid profiles and requires an explicit choice for a new or changed API credential
- fingerprints the API credential/account scope without persisting the credential
- keeps imported Platform tokens in memory and logs only opaque profile identifiers
- adds hard timeout behavior, profile-catalog preservation, and stable transition presentation

Reference: #1910

## Privacy and reliability

- DeepSeek Platform tokens are never written to CodexBar config
- legacy DeepSeek `cookieHeader` values are removed during config normalization
- private Platform endpoints may change; parser failures degrade to balance-only data
- optional session resolution is cancellation-aware and hard-bounded

## Validation

- `swift test --filter DeepSeek` — 91 tests passed
- `swift test --filter ProviderSettingsDescriptorTests` — 34 tests passed
- `make check` — passed with 0 lint violations
- `git diff --check` — passed

## UI verification

Validated locally with multiple Chrome profiles:

- balance remains visible during profile changes
- the profile selector remains visible and shows `Refreshing` while disabled
- invalid/expired sessions are omitted
- detailed usage repopulates after the selected profile refresh completes

## Screenshots
Screenshot 2026-07-13 at 14 22 24
Screenshot 2026-07-13 at 14 22 51

PR #2216, opened 2026-07-16
Improve DeepSeek detailed usage

joeVenner:
## Summary
- Adds DeepSeek Platform detailed usage support from #2135.
- Supports browser-only DeepSeek usage and keeps detailed usage separate from API-key balance fetching.

## Verification
- swift test --filter DeepSeekUsageFetcherTests


steipete:
Thanks @joeVenner for carrying this work forward and validating it. PR #2135 contains Zihao-Qi’s original DeepSeek commits and is being merged as the canonical change so first-author credit stays intact. The additional UsageFetcher+Testing.swift split is useful; please send that focused refactor as its own small PR.

Issue #2222, opened 2026-07-16
Claude: spurious session+weekly reset confetti from transient near-zero CLI-probe reading (shared claude:claude detector key, no weekly boundary guard)


you3fen:
### Summary

Full-screen **session + weekly reset confetti fired for Claude when no reset had occurred**. Persisted detector state shows the trigger was a transient near-zero usage reading from the Claude CLI probe, recorded under the shared fallback detector key `claude:claude`. This is the Claude-side sibling of #2054 / #1986 (which were fixed for Codex ownership); the Claude weekly path currently has no boundary guard at all, so a single bogus low sample is enough to celebrate.

### Environment

- CodexBar 0.43.0 (104), auto-update on
- macOS 27 beta (Darwin 27.0.0)
- Refresh frequency: 2 minutes
- `debugDisableKeychainAccess = 1` (set to avoid Keychain prompts), `claudeOAuthKeychainPromptMode = onlyOnUserAction`
- Because Keychain access is disabled, Claude usage is served by the CLI probe (PTY scrape of `claude /usage`) — snapshot `resetDescription` strings are raw CLI panel text (e.g. `Resets9:50pm(Asia/…)`), and file log confirms `Keychain access disabled; skipping migration` followed by `claude` subprocess probes.

### What happened

On 2026-07-16 at ~12:44 UTC, full-screen confetti appeared. Nothing had reset:

- Claude session window: 3% used, resets 13:50 UTC (in the future at the time)
- Claude weekly window: 2% used, resets 2026-07-22 19:00 UTC
- Codex weekly: 34% used, resets 2026-07-23 04:22 UTC

This has recurred several times over the past days.

### Forensic evidence (persisted detector state vs. reality)

Decoded `sessionLimitResetDetectorStates` / `weeklyLimitResetDetectorStates` from defaults right after the event:

| | `claude:claude` session | `claude:claude` weekly |
|---|---|---|
| `lastObservedAt` | 2026-07-16 12:44:52 UTC | 2026-07-16 12:44:52 UTC |
| `wasAboveThreshold` | **false** (≤1% used) | **false** (≤1% used) |
| `resetBoundary` | 13:50 UTC (correct) | Jul 22 19:00 UTC (correct) |

But the widget snapshot written **90 seconds later** (12:46:30 UTC) shows `primary.usedPercent = 3`, `secondary.usedPercent = 2` — both above the 1% threshold.

[...]

PR #2223, opened 2026-07-16
Fix spurious Claude weekly reset confetti from transient near-zero samples

OfficialAbhinavSingh:
**Partially addresses #2222** (the Claude weekly boundary-guard path; see "Scope" below — this intentionally does not fully close the issue).

## Problem
Claude weekly limit-reset celebrations fire on a single bogus near-zero usage reading (e.g. a CLI usage-probe misparse). The `resetBoundaryAllowsPost` gate in `postLimitResetCelebrationIfNeeded` has a boundary-advance guard for the **session** lane and for **Codex weekly** (added for #2054), but Claude weekly fell through to the unconditional `else { true }` — so one transient zero was enough to celebrate, even when the reset boundary had not moved.

Reported forensics on the issue show the transient sample carried the *correct, unchanged* reset boundary, yet a whole-panel misread dropped both session and weekly to ~0 and both fired.

## Fix
Route Claude weekly crossings through the same boundary-advance guard as Codex, via a shared `weeklyResetBoundaryAllowsCelebration(provider:previous:current:)` helper: a below-threshold crossing only celebrates if the reset boundary actually advanced.

Claude uses a more permissive variant than Codex, because Claude OAuth snapshots can legitimately have **no** boundary:
- **Unchanged** boundary (present on both sides, within the 2-minute equivalence tolerance) → suppressed (the #2222 transient-zero case).
- Boundary **missing** on either side → still allowed (a genuine boundary-less OAuth reset must still celebrate).
- Boundary **advanced** → celebrates.

Codex behavior is byte-for-byte unchanged (it requires both a prior and current boundary).

## Scope
This fixes the Claude **weekly boundary-guard** path only. #2222 also reports a stale-boundary **session** path and an identity-less shared `claude:claude` detector-key concern. Those are larger, separable changes and are intentionally **out of scope** here — left as follow-ups rather than folded into this narrow guard. Hence "Partially addresses" rather than "Closes".

## Tests
- `claude weekly celebration ignores transient zero when reset boundary is unchanged` — transient zero (same boundary) suppressed; a subsequent genuine week-advance celebrates exactly once.
- `claude weekly celebration posts when a genuine reset drops the boundary` — prior-boundary → nil-boundary genuine reset still celebrates.
- `weekly same-hour drop with unchanged boundary coalesces history without celebrating` — same-hour history coalescing preserved, no false celebration.
Existing nil-boundary Claude reset coverage (`weekly quota celebration posts when weekly usage resets to zero`, repeated-low, first-seen) is unaffected.

## Review feedback addressed (clawsweeper — commits `b02e3ac`, `048f444`)
Both first-round P1 findings were correct and are fixed:
1. **Preserve boundary-less Claude resets** — added a `requiresCurrentBoundary` flag to `limitResetBoundaryAdvanced` (default `true`, so session/Codex are unchanged); Claude passes `false`, so a stored boundary followed by a boundary-less genuine reset (`B → nil`) is no longer wrongly suppressed. Added focused coverage.
2. **Same-hour history fixture** — an earlier `+7d` edit exceeded the 2-minute tolerance and started a second history segment. Restored an in-tolerance boundary; the same-boundary drop is now correctly a suspect sample, so the test asserts coalescing **without** a celebration.
Also fixed a sibling-test compile issue (made `WeeklyLimitResetEventRecorder` internal so the split test file resolves it).

## On "real behavior proof"
This is a pure-logic change fully covered by deterministic Swift Testing cases (transient-zero suppressed, genuine week-advance celebrates, boundary-dropped reset celebrates, same-hour coalescing without celebration). I don't own a Mac, so I can't attach a native-app run/recording — the CI `swift-test-macos` job and a maintainer's signed run are the runtime confirmation, same situation as #2102. Happy to adjust anything.

### DIFF
--- a/Sources/CodexBar/UsageStore+LimitResetBoundary.swift
+++ b/Sources/CodexBar/UsageStore+LimitResetBoundary.swift
@@ -1,13 +1,37 @@
+import CodexBarCore
 import Foundation
 
 extension UsageStore {
     nonisolated static func limitResetBoundaryAdvanced(
         previous: Date?,
         current: Date?,
-        requiresPreviousBoundary: Bool = false) -> Bool
+        requiresPreviousBoundary: Bool = false,
+        requiresCurrentBoundary: Bool = true) -> Bool
     {
         guard let previous else { return !requiresPreviousBoundary }
-        guard let current else { return false }
+        // A missing current boundary cannot confirm an advance. Callers that trust
+        // boundary-less snapshots (e.g. Claude OAuth, which may omit resetsAt on a genuine
+        // reset) pass requiresCurrentBoundary: false so the crossing is still allowed.
+        guard let current else { return !requiresCurrentBoundary }
         return !self.areEquivalentPlanUtilizationResetBoundaries(previous, current) && current > previous
+    }
+
+    /// Whether a weekly below-threshold crossing should post a limit-reset celebration.
+    ///
+    /// Codex boundaries are always present, so a real advance requires both a prior and a
+    /// current boundary (#2054). Claude OAuth snapshots may legitimately omit the boundary, so
+    /// neither is required: only a confirmed *unchanged* boundary is suppressed as a suspect
+    /// near-zero sample (#2222); a missing boundary on either side still allows a genuine reset.
+    nonisolated static func weeklyResetBoundaryAllowsCelebration(
+        provider: UsageProvider,
+        previous: Date?,
+        current: Date?) -> Bool
+    {
+        let isCodex = provider == .codex
+        return self.limitResetBoundaryAdvanced(
+            previous: previous,
+            current: current,
+            requiresPreviousBoundary: isCodex,
+            requiresCurrentBoundary: isCodex)
     }
 }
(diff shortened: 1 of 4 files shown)


OfficialAbhinavSingh:
Pausing this (converting to draft). #2231 merged an overlapping Claude-weekly-confetti fix whose model treats a genuine reset as usage → 0 with the **reset boundary unchanged**, which directly conflicts with this PR's boundary-advance guard (it would suppress that case, breaking the new dedup tests).

That's a design question, not a mechanical rebase, so I've raised it on #2222 rather than force a conflicting change. Will revisit once the direction is clear — happy to rework toward the dedup-style "confirm the drop" approach or close this if the merged fix is sufficient.

steipete:
Thanks @OfficialAbhinavSingh for the careful diagnosis and coverage here.

I’m closing this approach because the field evidence in #2230 shows Claude can refill weekly usage without advancing the reported reset boundary. An unchanged-boundary guard would therefore suppress a genuine reset celebration. This PR also changes the existing “weekly quota celebration posts when reset lands mid hour without history split” behavior from celebrate to suppress, which makes that risk concrete.

#2231 and #2224 cover the two observed bug classes without deciding the contested boundary semantics: duplicate celebration after a real refill, and false low-confidence identity-less CLI samples.

If false weekly celebrations continue on identified accounts, please come back with verbose logs that show whether a genuine Claude weekly reset advances the boundary. That evidence would let us revisit boundary handling safely. Thank you again for pushing this investigation forward.

PR #2224, opened 2026-07-16
fix: confirm identity-less Claude reset samples

### LINKED ISSUE #2222: Claude: spurious session+weekly reset confetti from transient near-zero CLI-probe reading (shared claude:claude detector key, no weekly boundary guard)

Yuxin-Qiao:
## Summary

- require a second low-usage sample before celebrating a Claude reset when the observation has no account identity and therefore uses the shared `claude:claude` detector key
- preserve compatibility with previously persisted detector state
- add coverage for a transient simultaneous session/weekly low sample followed by recovery, plus a real reset confirmed by the next sample

## Root cause

Identity-less Claude CLI probe samples share one persisted detector key. A transient partial/near-zero CLI read can therefore look like a reset for both the session and weekly windows. The detector previously posted immediately on the first threshold crossing.

## Behavior

Only the low-confidence shared Claude path waits for one subsequent low sample (one refresh interval). Identified accounts retain the existing immediate behavior.

## Test

- `swift test --filter "UsageStorePlanUtilizationTests"` (72 tests passed)

Fixes #2222.


steipete:
Merged after exact-head verification.

Verification on head `9232b5ed022e42da2fcefff6505098c71fce1666`:

- `swift test --filter UsageStorePlanUtilizationTests`: 78/78 tests passed.
- `make check`: SwiftFormat clean; SwiftLint reported 0 violations across 1,462 files.
- Branch autoreview against `origin/main`: no accepted/actionable findings.
- Exact-head CI: changes, lint, both Linux builds, both macOS test shards, and GitGuardian all passed.

Merge commit: `9fa8e918c88b69732bcf5299431efcc8b0654b4d`

Issue #2214, opened 2026-07-16
Proposal: unify credential access policy across providers


kreitter:
## Summary

CodexBar currently has several overlapping ways to control credential access:

- the global **Disable Keychain access** setting;
- provider-specific prompt controls;
- Security.framework no-UI reads;
- browser Safe Storage and cached credentials;
- provider-specific denial and timeout cooldowns;
- CodexBar-launched provider CLIs whose descendants may access Keychain outside CodexBar's process.

These controls do not currently form one user-understandable policy.

This is a product and architecture proposal. I will not open an implementation PR until maintainers indicate whether this direction is desirable.

## Problem

The same user preference can produce different behavior depending on the provider or credential path:

- CodexBar can directly access Security.framework.
- Browser credential import may access a browser's Safe Storage item.
- Provider CLIs can access credentials or launch descendant tools outside CodexBar's direct control.
- Providers have accumulated separate settings, fallback rules, and cooldown behavior.
- A setting that appears to prohibit prompts cannot truthfully guarantee that behavior if an ungoverned child process still runs.

Fixing individual call sites does not resolve the underlying policy fragmentation.

Related reports demonstrate that this is broader than one provider or one Keychain item: #2115, #2213, #1681, #1056, and #2024.

## Proposed contract

Adopt one provider-neutral credential-access policy with three user-facing modes:

1. **Standard**
   - Automatic credential sources remain available.
   - A denial, cancellation, or indeterminate credential-capable child pauses automatic retries.
   - Explicit recovery remains available.

2. **No prompts**
   - Operations that may display credential UI, or whose behavior is unknown, do not run.
   - No-UI reads and safe CodexBar-owned cache operations may continue.

3. **Disabled**
   - CodexBar does not access credentials for the selected scope.

A provider override may make the global policy stricter, but providers should not define separate prompt-policy vocabulary.

## Enforcement boundaries

The same policy would be enforced through three shared gateways:

1. Security.framework and CodexBar-owned credential storage.
2. Browser credential and Safe Storage access.
3. External processes launched by the CodexBar app.

Provider implementations would declare the operation and its credential risk. They would not independently decide whether it is allowed.

Standalone CLI commands deliberately run by the user would remain outside app-background restrictions.

## Failure and recovery

Denial, cancellation, timeout, and indeterminate child behavior would eventually use one provider-and-operation-scoped failure model.

[...]

PR #2225, opened 2026-07-16
Block background Chromium Keychain access

### LINKED ISSUE #2213: why will codexbar request my keychain permission?
Image

Yuxin-Qiao:
## Summary

- block Chromium browser-cookie imports during background provider refreshes
- preserve browser-cookie access for user-initiated refreshes and the existing Keychain pre-alert flow
- update gate tests to prove background paths do not even perform the Keychain preflight

## Root cause

Chromium cookie decryption reads `Chrome Safe Storage` from Keychain. The previous background path could proceed into the browser importer whenever an attribute-only preflight appeared allowed, even though the later password-data read could require user interaction.

## Impact

Automatic refreshes no longer initiate the Chromium Keychain path. A user-initiated refresh remains the consent boundary for browser-cookie authentication.

## Validation

- targeted test changes use `KeychainAccessPreflight` test overrides and do not touch a real Keychain item
- `git diff --check`
- the full local SwiftPM test executable was blocked by unrelated auxiliary-target timestamp churn after the modified core module compiled; this is recorded rather than treated as a passing test

## Follow-up

SweetCookieKit is being hardened separately so no host can implicitly promote a no-UI Safe Storage read to an interactive one. Once released, CodexBar should bump that dependency.

Fixes #2213
Related: #2214

### DIFF
--- a/Sources/CodexBarCore/BrowserCookieAccessGate.swift
+++ b/Sources/CodexBarCore/BrowserCookieAccessGate.swift
@@ -82,6 +82,12 @@ public enum BrowserCookieAccessGate {
     public static func shouldAttempt(_ browser: Browser, now: Date = Date()) -> Bool {
         guard browser.usesKeychainForCookieDecryption else { return true }
         guard !KeychainAccessGate.isDisabled else { return false }
+        guard ProviderInteractionContext.current == .userInitiated else {
+            self.log.info(
+                "Skipping background Chromium cookie import to avoid a Keychain prompt",
+                metadata: ["browser": browser.displayName])
+            return false
+        }
         if self.deniedBrowsersForTesting?.contains(browser) == true {
             return self.isExplicitRetryAllowed(for: browser)
         }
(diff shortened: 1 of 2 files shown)


steipete:
Merged as bad4bebebc39a7f69ce90036ac97dc9d569d80c8.

Verification:
- Audited the complete two-file diff. The guard only blocks background access for browsers whose cookie decryption uses Keychain; Safari, Firefox, and Zen remain allowed, and user-initiated Chromium access remains allowed.
- Ran swift test --filter BrowserDetectionTests: 28 tests passed, 0 failures.
- Autoreview completed with no actionable findings.
- Exact head cc661d7ecce5b82791055bff133bed928b7b0722 passed GitGuardian, lint, changes, both Linux CLI jobs, lint-build-test, and both macOS shards. macOS shard durations: 27m09s and 27m17s.
- GitHub reported MERGEABLE/CLEAN immediately before merge.

PR #2227, opened 2026-07-16
Token-gated dashboard snapshots for codexbar serve (restart of #1026)

jethac:
This is the narrower, token-only restart of #1026, taking the first option from the close comment: token-only auth with explicit acceptance of the trusted-network/plain-HTTP threat model. The pairing flow is gone entirely — that was the blocking bug (a one-time code sent as a plaintext GET query parameter while the docs claimed it never touched the network). A pairing redesign, if wanted, can be its own PR later.

## What this adds

`GET /dashboard/v1/snapshot` on `codexbar serve`, gated by a static bearer token:

- `--dashboard-token <token>`, or preferably the `CODEXBAR_DASHBOARD_TOKEN` env var (a flag value leaks via `ps`; the env var wins over the flag). Empty/whitespace tokens are startup errors.
- `--host <ip>` (default `127.0.0.1`; `localhost` normalizes to `127.0.0.1`). A non-loopback host refuses to start without a token, and refuses to start without `--allow-plain-http` — passing that flag is the explicit, operational acceptance that the bearer token crosses the network in cleartext on every request.
- On a **non-loopback** bind, the token gates **all data routes** — `/usage`, `/cost`, and the snapshot — so account data is never exposed to the network unauthenticated. `/health` stays open. On the default loopback bind, `/usage` and `/cost` are unchanged and unauthenticated; the snapshot route fails closed (`401`) when no token is configured.
- The token is only ever read from `Authorization: Bearer <token>`. A `?token=` query string is never accepted, and there's a wire-level test proving it.
- Token comparison is constant-time (fixed-length SHA-256 digest comparison, no short-circuit).
- The snapshot handler rides the existing serve machinery — `CLIServeResponseCache` + `CLIServeOperationCoordinator`, config-fingerprint keying, `--refresh-interval` TTL, `--request-timeout` deadline — with auth checked before any cache access so unauthenticated requests can neither warm nor read the cache.

## Close-comment checklist

Each bullet from the #1026 close, and where it's addressed:

- *"conflicted and has no current CI proof"* → fresh branch off current `main`; four logical commits, each building and passing its tests independently. Full local run: `./Scripts/test.sh` — 6281 tests, 0 failures; `./Scripts/lint.sh lint` clean.
- *"sends it in a plaintext HTTP GET query and returns the bearer credential over HTTP"* → pairing flow deleted; nothing credential-shaped ever appears in a URL; query-string tokens are rejected with a test asserting it on raw response bytes.
- *"docs … documents transport truthfully"* → `docs/dashboard-api.md` states plainly that transport is plain HTTP, the token crosses the network in cleartext on every request, and anyone on the path can capture and replay it. Deployment guidance covers loopback (default), a TLS-terminating reverse proxy (Caddy example), and explicit `--allow-plain-http` acceptance for trusted segments. No sentence anywhere claims a secret "never traverses the network."
- *"missing `Cache-Control: no-store`"* → snapshot responses, all dashboard errors, and every `401` carry `no-store` (asserted at the wire level). The last feature commit extends `no-store` to `/usage` and `/cost`, which also carry account data — it's deliberately separate and droppable if you'd rather not change those routes' headers.
- *"tests bypass the raw HTTP path"* → `CLIServeRawHTTPTests.swift` boots the real `CLILocalHTTPServer` on an ephemeral port, connects with a plain TCP socket, writes raw HTTP/1.1 bytes, and asserts on raw response bytes — driving the production request handler end to end (route → auth → cache → builder). Cases: no/wrong/query-string token → `401` + `WWW-Authenticate: Bearer` + `no-store`; correct token → `200` + decodable snapshot JSON; duplicate `Authorization` → `400`; non-GET → `405`; non-loopback gating of `/usage`/`/cost`; `/health` open.
- *"preserve current serve/cache/session semantics"* → no parallel bespoke cache; the snapshot shares the per-provider operation coordinators with `/usage`/`/cost`, so concurrent requests dedupe provider fetches exactly as today.

Along the way, live verification caught that `--allow-plain-http` would have been silently dropped (Commander keys `ParsedValues` by property name, not kebab-case); there's now a regression test that parses real argv through Commander's actual parser.

## Decisions I'd like your read on (none block review)

1. **Fail-closed on loopback:** with no token configured, the snapshot route answers `401` rather than being open like `/usage`/`/cost`. Consistent everywhere, but happy to open it on loopback if you prefer parity with the existing routes.
2. **`no-store` on `/usage`/`/cost`:** implemented as its own commit so it's droppable.
3. **Wildcard binds accept any `Host` header:** a literal reading ("bind host in the allowed set") would reject every real request to `--host 0.0.0.0`. Since all data routes require the bearer token off-loopback, a DNS-rebinding page can't attach the header, so `.any` is safe; specific non-loopback hosts still get loopback + that name only.

Notes: the socket layer is IPv4-only (pre-existing), so `--host ::` fails at bind with a startup error rather than being supported. The snapshot payload keeps #1026's shape (identity redacted, one row per provider), re-fitted to current `main`'s provider/payload code.


steipete:
Maintainer verification:

- Reviewed the complete 14-file change against the token-only restart specification from #1026. Kept loopback snapshot fail-closed behavior, `no-store` on `/usage` and `/cost`, and the token-mitigated wildcard-bind Host policy. Confirmed default loopback behavior remains unchanged, token precedence is environment over flag, comparison is constant-time, query-string tokens are rejected, and non-loopback startup requires both a token and `--allow-plain-http`.
- Updated contributor head to `583114d7c670345e77ce924f03c018a77ba3b842`, including current `main` and maintainer hardening/tests.
- `swift test --filter 'CLIServe(Auth|Router|RawHTTP)Tests|DashboardSnapshotBuilderTests'`: 82 tests in 4 suites passed.
- Final follow-up proof: `swift test --filter 'CLIServeRawHTTPTests|DashboardSnapshotBuilderTests'`: 28 tests in 2 suites passed.
- `make test`: 668 selections across 56 groups passed; one transient group recovered on the harness retry, with zero terminal failures.
- `make check`: passed; SwiftFormat reported 0 files requiring formatting and SwiftLint reported 0 violations.
- Manual smoke with `CODEXBAR_DASHBOARD_TOKEN` set: snapshot without token returned 401; `Authorization: Bearer ...` returned 200; query-string-only token returned 401.
- CI run https://github.com/steipete/CodexBar/actions/runs/29548421640 passed: lint, both Linux builds/tests, both macOS Swift-test shards, aggregate check, and GitGuardian all green.
- Merged as `547b2789e4c52a4b1643f6a30de20d07c19ce9a4`.

PR #2231, opened 2026-07-16
Prevent duplicate Claude weekly reset confetti

### LINKED ISSUE #2230: Claude weekly reset confetti can fire twice after a stale usage rebound
### Summary

After a genuine Claude weekly quota refill, CodexBar can show weekly reset confetti twice for the same reset when the usage source briefly returns a stale pre-refill value.

### Evidence

Local persisted history and event logs for one Claude account showed:

| Observation | Value |
|---|---:|
| Usage sequence around the refill | `73% → 0% → 73% → 0%` |
| Weekly reset boundary across all samples | `2026-07-20 00:00 UTC` (unchanged) |
| Weekly reset celebrations | 2 |
| Time between celebrations | 14.5 seconds |
| Later usage sample | 6% at 04:56, with the same boundary |

The first drop to 0% was the real refill. The following 73% was a stale rebound that re-armed the detector, so the next correct 0% sample produced duplicate confetti.

Claude can refill the weekly token allowance without changing the displayed weekly counting boundary, so requiring a boundary change would suppress valid celebrations.

### Expected behavior

- One Claude weekly refill produces at most one celebration, even if usage briefly rebounds to a stale value.
- Reset-detector recovery is tracked independently per Claude account and survives an app restart.
- A later genuine refill may celebrate after post-reset usage recovery has been confirmed.

### Relation to #2222

Related to #2222, but distinct:

- #2222 covers a **false first celebration** caused by a bogus low sample such as `3% → 0% → 3%`.
- This issue covers a **duplicate celebration after a real refill**, such as `73% → 0% → 73% → 0%`.

Zihao-Qi:
## Summary

- prevent a stale Claude weekly usage rebound from rearming duplicate reset confetti
- require two consecutive above-threshold observations before rearming, persisted separately per account
- migrate legacy Claude weekly low states into the new recovery-confirmation state
- preserve Claude's existing behavior where a genuine refill may occur without changing the weekly boundary

## Root cause

After a genuine refill, a stale pre-refill sample could immediately rearm the threshold detector:

`73% → 0% (event posted) → 73% → 0% (duplicate event posted)`

The detector now remains disarmed until two consecutive readings confirm that weekly usage has genuinely resumed. A low reading interrupts that confirmation.

## Tests

- `swift test --filter UsageStorePlanUtilizationTests` — 76 passed
- `make check` — passed with 0 violations
- `make test` — the full suite remains blocked by eight failures in the untouched `ClaudeOAuthCredentialsStoreCLIStorageOwnershipTests` suite; the same failures reproduce when that suite is run independently

Regression coverage includes the exact observed sequence, a stale rebound delayed by 30 minutes, restart persistence, multiple-account isolation, a genuine later reset, and legacy-state migration.

Fixes #2230.

Related to #2222, but does not address its false first celebration.

### DIFF
--- a/Sources/CodexBar/UsageStore+LimitResetCelebration.swift
+++ b/Sources/CodexBar/UsageStore+LimitResetCelebration.swift
@@ -1,0 +1,187 @@
+import CodexBarCore
+import Foundation
+
+extension UsageStore {
+    private nonisolated static let limitResetThreshold = 1.0
+    private nonisolated static let claudeWeeklyRecoveryObservationCount = 2
+
+    struct LimitResetDetectorState: Codable, Equatable {
+        let wasAboveThreshold: Bool
+        let lastObservedAt: Date
+        let sourceRawValue: String?
+        var resetBoundary: Date?
+        var recoveryAboveThresholdCount: Int?
+
+        init(
+            wasAboveThreshold: Bool,
+            lastObservedAt: Date,
+            sourceRawValue: String?,
+            resetBoundary: Date? = nil,
+            recoveryAboveThresholdCount: Int? = nil)
+        {
+            self.wasAboveThreshold = wasAboveThreshold
+            self.lastObservedAt = lastObservedAt
+            self.sourceRawValue = sourceRawValue
+            self.resetBoundary = resetBoundary
+            self.recoveryAboveThresholdCount = recoveryAboveThresholdCount
+        }
+    }
+
+    struct LimitResetDetectionContext {
+        let provider: UsageProvider
+        let account: ProviderTokenAccount?
+        let snapshot: UsageSnapshot
+        let accountKey: String?
+        let capturedAt: Date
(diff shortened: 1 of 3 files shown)


steipete:
Merged after exact-head verification.

Verification on head `3dff14e7b4a40a968186eb906226fa9ec6821b08`:

- `swift test --filter UsageStorePlanUtilizationTests`: 76 tests passed, 0 failed, including the exact 73 → 0 → 73 → 0 sequence, restart persistence, legacy-state migration, and multi-account coverage.
- `make check`: SwiftFormat clean; SwiftLint reported 0 violations across 1,453 files.
- Branch autoreview against `origin/main`: no accepted/actionable findings.
- Exact-head CI: changes, lint, both Linux builds, both macOS test shards, and GitGuardian all passed.

Merge commit: `6fb8c012296e817264525e70e4feb00b7b733ce0`

PR #2233, opened 2026-07-16
site: add CrossModel to the provider grid

hujuncheng:
Adds a CrossModel card to the landing page provider grid, placed next to
the other API-key aggregators.

The grid renders each logo as a CSS mask (`.provider-logo::before`), so the
asset has to be a single-colour silhouette that gets tinted with `--brand`
on hover. `docs/logos/crossmodel.svg` is therefore the bare CrossModel mark
— solid upper arc, deck, dashed reflection — rather than the white-on-
gradient badge, and `--brand` is #7C3AED, the leading stop of the brand
gradient.

Provider count is unchanged (`crossmodel` is already in `UsageProvider`),
so no count or locale strings needed updating.

Verified:
- `node Scripts/check-site-locales.mjs`
- `node Scripts/check-documentation-links.mjs`
- local render of `docs/index.html`, default and hover states

Happy to drop this if the grid is meant to stay a curated subset.


hujuncheng:
Closing this — with #2254 removing the CrossModel integration on hosted-relay eligibility grounds, adding the card to the provider grid doesn't make sense right now. Happy to revisit if we publish an imprint and clarify upstream authorization. Thanks for the review.

PR #2247, opened 2026-07-17
Preserve Claude snapshots across CLI parse failures

kiranmagic7:
## Summary
- Preserve the existing Claude snapshot when repeated terminal CLI usage parse failures reach the failure gate, so the card can surface the failure without falling back to `Not fetched yet`.
- Gate parse-failure preservation on the terminal fetch attempt kind, so Web/OAuth parse failures still clear stale prior data once the failure gate surfaces the error.
- Keep a prior Max/Pro/Team/Ultra subscription snapshot visible when a later CLI probe returns a subscription-only response without quota data; Education/no-prior paths still use `Limits not available`.
- Add regression coverage for repeated CLI parse failures, subscription-only responses with prior subscription limits, and Web parse failures clearing stale snapshots.

Refs #2241.

## Tests
- `make check`
- `git diff --check upstream/main...HEAD`
- `swiftc -parse Sources/CodexBar/UsageStore+Refresh.swift`
- `swiftc -parse Tests/CodexBarTests/ClaudeResilienceTests.swift`
- `swiftc -parse Tests/CodexBarTests/ClaudeEducationAvailabilityTests.swift`
- `swiftc -parse Tests/CodexBarTests/ClaudeWebRefreshResilienceTests.swift`

Attempted but not completed locally:
- `swift test --filter ClaudeResilienceTests`
- `swift test --filter ClaudeEducationAvailabilityTests`
- `swift test --skip-update --filter ClaudeResilienceTests`
- `swift test --skip-update --filter ClaudeWebRefreshResilienceTests`

Those SwiftPM runs stalled while resolving/downloading the Sparkle 2.9.3 binary artifact before tests executed. A direct `curl -L --max-time 60` download of the same artifact completed successfully earlier, so this looks like a local SwiftPM artifact resolver stall rather than a source-level test failure.

I did not run live Claude provider probes because the repository guidance says live provider probes, browser-cookie imports, and real account paths should not be run unless explicitly requested. The behavior here is covered with stubbed provider strategies and the hosted CI test shards.

## Compatibility Risk
Low. The change only affects Claude refresh failure publication when a prior Claude snapshot already exists, and the follow-up gate keeps stale Web/OAuth parse failures from being preserved as CLI data.

Issue #2234, opened 2026-07-16
CodexBarCLI serve (Linux) crashes SIGILL every ~2-4h: pipe fd leak → EMFILE → Pipe() yields fd=-1 → precondition in FileHandle.readabilityHandler


psl75011:
## Summary

`CodexBarCLI serve` on Linux crashes with **SIGILL every ~2–4 hours of uptime**. Root cause (established from two symbolized core dumps + live fd measurements): the daemon **leaks pipe file descriptors** on every provider-version-detection subprocess spawn. When the process hits `RLIMIT_NOFILE` (soft 1024), swift-corelibs-foundation's `Pipe()` init takes its EMFILE path and returns `FileHandle`s with **fd = -1** instead of failing. `ProcessPipeCapture.start()` then sets `readabilityHandler` on that handle, which trips `precondition(_fd >= 0, "Bad file descriptor")` in `FileHandle._checkFileHandle()` — Swift preconditions compile to `ud2`, hence SIGILL at a byte-identical instruction address on every crash.

## Environment

- CodexBar **v0.43.0** (`CodexBarCLI-v0.43.0-linux-musl-x86_64.tar.gz`), also observed on v0.41.0
- Ubuntu 26.04 x86_64, systemd service: `CodexBarCLI serve --port 9120 --refresh-interval 120 --request-timeout 30`
- Providers enabled: codex, claude, grok, zai, antigravity, deepseek

## Evidence

**5 SIGILL crashes at the byte-identical address** (kernel log):

```
traps: CodexBarCLI[3927817] trap invalid opcode ip:1ded552 sp:... error:0 in CodexBarCLI[1bec552,9b1000+1e29000]
traps: CodexBarCLI[38225]   trap invalid opcode ip:1ded552 sp:... error:0 in CodexBarCLI[1bec552,9b1000+1e29000]
```

**Symbolized backtrace** (`coredumpctl debug`, identical `#0`–`#4` on both cores; only the caller differs):

```
#0  Swift runtime failure: precondition failure () at .../swift-corelibs-foundation/Sources/Foundation/FileHandle.swift:89
#1  FileHandle._checkFileHandle()                       at FileHandle.swift:89
#2  FileHandle.monitor(forReading:resumed:handler:)     at FileHandle.swift:129
#3  FileHandle.readabilityHandler.setter                 at FileHandle.swift:187
[...]

PR #2252, opened 2026-07-17
fix(claude): preserve last-good snapshot across CLI probe failures (fixes #2241)

### LINKED ISSUE #2241: Claude: card shows 'Not fetched yet' instead of last-good data — CLI probe failures (8.6% success rate) silently drop the cached snapshot
### Summary

The Claude menu card frequently shows **"Not fetched yet"** (no data, no error) instead of the last-good snapshot, until a manual refresh. Verbose file logs over ~10 hours show the CLI probe succeeding only **23 of 268 attempts (8.6%)**, with failure streaks up to **110 minutes**; on non-preservable failures the Claude card **silently drops the cached snapshot** instead of keeping stale data the way the Codex card does.

Same setup as #2222 (you asked for verbose logs there — this is the second symptom from the same machine).

### Environment

- CodexBar 0.43.0 (104), refresh: 2 minutes
- Claude Code CLI 2.1.207; heavy concurrent Claude Code usage on the machine
- Keychain access disabled (`debugDisableKeychainAccess = 1`), so Claude usage runs entirely on the CLI-probe path (PTY scrape of `claude /usage`, then direct no-TTY fallback)
- macOS 27 beta (Darwin 27.0.0) — honest caveat: this may amplify the probe failure *rate*; it does not explain the data-dropping behavior below

### Probe reliability (verbose file log, ~10 h)

Per-hour PTY probe launches vs. successful scrapes:

| Hour (UTC) | launches | scrape ok |
|---|---|---|
| 13–14 | 10 | 2 |
| 14–15 | 28 | 2 |
| 15–16 | 30 | 3 |
| 16–17 | 28 | 3 |
| 17–18 | 24 | 1 |
| 18–19 | 26 | 3 |
| 19–20 | 30 | 2 |
| 20–21 | 28 | 2 |
| 21–22 | 30 | 2 |
| 22–23 | 24 | 0 |

- Failures have a uniform signature: `Claude CLI session started` → `session stopping` after **~4.5 s**, no scrape result. Successes take ~17 s (or ~30 s with the logged `looked like startup output; retrying once` retry). The rate is equally bad while the machine idles overnight, so background load/scripts don't correlate.
- Every failing cycle then runs the fallback `claude /usage` as a plain subprocess (`claude-direct-usage`, no TTY). It exits `status=0` in ~2.5 s but never yields a parseable usage panel — with CLI 2.1.207 the panel doesn't render without a TTY, so this fallback is structurally a no-op that still consumes the cycle.
- `claude-auth-status` succeeds every cycle (~0.7 s, exit 0), so binary resolution, PATH, and login state are all healthy — the failure is specific to panel scraping.

### The actual bug: failures erase last-good data with no error

While Codex shows either "Refreshing" or a stale snapshot after failures, the Claude card ends up with **snapshot = nil and error = nil**, which renders as "Not fetched yet". In `UsageStore+Refresh.swift` (v0.43.0, `handleProviderFetchFailure`), two paths produce exactly that state:

1. **Quota-unavailable misclassification** (~line 971): if a failed scrape's error matches `isSubscriptionQuotaUnavailableDescription`, the handler deliberately removes the snapshot and sets `errors[provider] = nil` (built for Education accounts with no quota panel). Once entered, the state is sticky: subsequent failures keep `knownLimitsAvailability = .unavailable` with no error until the next successful scrape republishes. My account is a regular Max subscription with a quota panel, so any entry into this branch is a misread.
2. **Surfaced non-preservable failure** (~line 1026): parse failures are not in the preserve list (only timeouts/network/rate-limit/cancellation are), so once the failure gate opens, the snapshot is removed alongside the surfaced error; later cycles can clear the error again, landing in the same nil/nil state.

The verbose log records no refresh-outcome line for Claude fetch failures (only subprocess start/exit), so I can't tell from logs alone which of the two branches fired — happy to run any added logging or a test build to pin it down (per your offer in #2222).

### Suggestions

- Preserve the last-good Claude snapshot across CLI probe/parse failures, matching the Codex behavior and the direction of the 0.43.1 OAuth work ("preserve each account's last-good usage") — ideally with an "updated Xm ago" stale marker instead of silent data loss.
- Require corroboration (e.g., two consecutive readings, or an explicit subscription notice string) before entering the sticky quota-unavailable state.
- Treat an exit-0-but-unparseable `claude-direct-usage` result as "strategy unavailable" rather than a normal failure that counts toward gates/cleanup.
- Log Claude refresh outcomes (strategy + error class) at verbose level so field reports can distinguish these branches.

Yuxin-Qiao:
## Summary

Fixes #2241 by preserving the last-good Claude usage snapshot across two failure paths that currently clear it:

1. **Subscription-only probe responses** on accounts that already have a Max/Pro/Team/Ultra snapshot are no longer misclassified as quota-unavailable, which used to erase the snapshot and set `knownLimitsAvailability` to `.unavailable`.
2. **Repeated CLI parse failures** are now treated as preservable failures when prior data exists, so the card keeps showing stale data while the failure gate still surfaces the error after the gate threshold.

Education / no-prior-data behavior is unchanged: those cases still drop the snapshot and show "Limits not available".

## Test plan

- `swift test --filter Issue2241ReproTests` — added new regression tests that fail on current main and pass with this change.
- `swift test --filter ClaudeResilienceTests` — includes new repeated CLI parse failure case.
- `swift test --filter ClaudeEducationAvailabilityTests` — includes new subscription-only response with prior subscription limits case.

## Compatibility

Low. The change only affects Claude refresh failure publication when a prior snapshot already exists. It does not alter other providers or the Education/no-prior-data path.

### DIFF
--- a/.github/workflows/repro-2234.yml
+++ b/.github/workflows/repro-2234.yml
@@ -1,0 +1,48 @@
+name: Repro 2234 SIGILL
+
+on:
+  push:
+    branches: [issue2241-repro]
+  workflow_dispatch:
+
+jobs:
+  repro:
+    runs-on: ubuntu-24.04
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Swift toolchain version
+        run: swift --version
+
+      - name: Run fd leak repro
+        shell: bash
+        run: |
+          set -euo pipefail
+          cd repro/2234
+          swift run Repro leak 2>&1 | tee leak.log
+
+      - name: Run EMFILE/SIGILL repro
+        shell: bash
+        run: |
+          set -euo pipefail
+          cd repro/2234
+          # Tight fd limit forces the EMFILE branch quickly.
+          ulimit -n 64
+          set +e
+          swift run Repro emfile 2>&1 | tee emfile.log
+          exit_code=$?
+          set -e
+          echo "emfile exit code: $exit_code"
+          if [[ $exit_code -ne 0 ]]; then
+            echo "Process crashed as expected (SIGILL or other signal -> exit $exit_code)"
+          fi
+
+      - name: Upload repro logs
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: repro-2234-logs
+          path: |
+            repro/2234/leak.log
+            repro/2234/emfile.log
--- a/.github/workflows/repro-2234.yml
+++ b/.github/workflows/repro-2234.yml
@@ -8,21 +8,27 @@ on:
 jobs:
   repro:
     runs-on: ubuntu-24.04
-    timeout-minutes: 15
+    timeout-minutes: 20
     steps:
       - uses: actions/checkout@v4
 
       - name: Swift toolchain version
         run: swift --version
 
-      - name: Run fd leak repro
+      - name: Run fd leak repro (unfixed baseline)
         shell: bash
         run: |
           set -euo pipefail
           cd repro/2234
           swift run Repro leak 2>&1 | tee leak.log
 
-      - name: Run EMFILE/SIGILL repro
+      - name: Run real ProcessPipeCapture regression tests
+        shell: bash
+        run: |
+          set -euo pipefail
+          swift test --filter ProcessPipeCaptureLinuxTests 2>&1 | tee test.log
+
+      - name: Run EMFILE/SIGILL repro (unfixed baseline)
         shell: bash
         run: |
           set -euo pipefail
@@ -46,3 +52,4 @@ jobs:
           path: |
             repro/2234/leak.log
             repro/2234/emfile.log
+            test.log
--- a/Sources/CodexBarCore/Host/Process/ProcessPipeCapture.swift
+++ b/Sources/CodexBarCore/Host/Process/ProcessPipeCapture.swift
@@ -1,4 +1,7 @@
 import Foundation
+#if os(Linux)
+import Glibc
+#endif
 
 package final class ProcessPipeCapture: @unchecked Sendable {
     package static let defaultMaxBytes = 1 * 1024 * 1024
@@ -13,6 +16,7 @@ package final class ProcessPipeCapture: @unchecked Sendable {
     private var didReachEOF = false
     private var isStopping = false
     private var continuation: CheckedContinuation<Void, Never>?
+    private var usesReadableHandler = true
 
     package init(
         pipe: Pipe,
@@ -25,6 +29,22 @@ package final class ProcessPipeCapture: @unchecked Sendable {
     }
 
     package func start() {
+        #if os(Linux)
+        // swift-corelibs-foundation's readabilityHandler setter duplicates the
+        // underlying fd to create a dispatch source. If the process is already at
+        // EMFILE, that dup fails and the setter traps with precondition(_fd >= 0),
+        // producing the SIGILL reported in issue #2234. Defensively probe whether
+        // we can duplicate the fd ourselves; if not, fall back to synchronous
+        // reading in stopAndSnapshot() instead of installing the handler.
+        let fd = self.handle.fileDescriptor
+        let probe = Glibc.dup(fd)
+        guard probe != -1 else {
+            self.usesReadableHandler = false
+            return
+        }
+        Glibc.close(probe)
+        #endif
+
         self.handle.readabilityHandler = { [weak self] handle in
             self?.handleReadableData(from: handle)
         }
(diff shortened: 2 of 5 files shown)


Yuxin-Qiao:
@clawsweeper re-review

This PR now contains the focused #2241 fix and regression tests. The branch was replaced with the correct implementation after the earlier review identified the unrelated #2234 reproducer.

## Evidence (non-screenshot / non-recording)

### 1. Before the fix: both regression tests fail on current main

```
✘ Test \"subscription only response erases prior Claude subscription snapshot\" recorded an issue at Issue2241ReproTests.swift:74:21: Expectation failed: (store.snapshot(for: .claude)?.updatedAt → nil) == (prior.updatedAt → 2027-01-15 08:00:00 +0000)
✘ Test \"subscription only response erases prior Claude subscription snapshot\" recorded an issue at Issue2241ReproTests.swift:76:21: Expectation failed: (store.knownLimitsAvailability(for: .claude) → .unavailable) == .available
✘ Test \"repeated CLI parse failures keep prior Claude snapshot while surfacing error\" recorded an issue at Issue2241ReproTests.swift:172:17: Expectation failed: (secondResult.updatedAt → nil) == (prior.updatedAt → 2027-01-15 08:00:00 +0000)
✘ Test run with 2 tests in 1 suite failed after 0.502 seconds with 3 issues.
```

### 2. After the fix: the same tests pass

```
✔ Test \"subscription only response erases prior Claude subscription snapshot\" passed after 0.226 seconds.
✔ Test \"repeated CLI parse failures keep prior Claude snapshot while surfacing error\" passed after 0.234 seconds.
✔ Suite Issue2241ReproTests passed after 0.234 seconds.
✔ Test run with 2 tests in 1 suite passed after 0.234 seconds.
```

### 3. Why the change is bounded and correct

The diff only touches `UsageStore+Refresh.swift` failure publication for Claude:

- **Subscription-only misclassification:** when the probe returns `subscriptionQuotaUnavailableDescription`, we now check whether a prior snapshot exists and is a real subscription (`Max`/`Pro`/`Team`/`Ultra`). If so, we keep the snapshot and mark availability as `.available`. Education accounts and accounts without prior data still fall through to the original path that drops the snapshot and shows `Limits not available`.

- **CLI parse failures:** the `shouldPreservePriorSnapshot` policy is extended for Claude to include `ClaudeStatusProbeError.parseFailed` and `ClaudeUsageError.parseFailed`, unless the parse failure message itself is the subscription-quota-unavailable marker. This means repeated parse failures keep the last-good snapshot while still allowing the failure gate to surface the error once the suppression threshold is crossed.

### 4. Test commands run locally

```bash
swift test --filter Issue2241ReproTests
swift test --filter ClaudeResilienceTests
swift test --filter ClaudeEducationAvailabilityTests
```

The PR branch now contains:
- `Sources/CodexBar/UsageStore+Refresh.swift` — the actual fix
- `Tests/CodexBarTests/Issue2241ReproTests.swift` — new source-level reproducer tests
- `Tests/CodexBarTests/ClaudeResilienceTests.swift` — additional CLI parse-failure regression test
- `Tests/CodexBarTests/ClaudeEducationAvailabilityTests.swift` — additional subscription-only with prior subscription limits test

The unrelated #2234 reproducer and workflow are no longer in this branch.

Yuxin-Qiao:
@clawsweeper re-review

Pushed review feedback addressing the auth-failure preservation concern:

1. `UsageStore+Refresh.isClaudeCLIUsageParseFailure` now excludes parse failures that indicate authentication issues (`token expired`, `authentication error`, or the `Run \`claude login\`` hint). Those follow the existing non-preservable path, so the snapshot is dropped once the failure gate surfaces the error.

2. Added a regression test in `Issue2241ReproTests`: `Claude authentication parse failure clears prior snapshot when surfaced` verifies that after two consecutive auth parse failures the prior snapshot is removed and the error is published.

Local result:
```
✔ Test "subscription only response erases prior Claude subscription snapshot" passed
✔ Test "Claude authentication parse failure clears prior snapshot when surfaced" passed
✔ Test "repeated CLI parse failures keep prior Claude snapshot while surfacing error" passed
✔ Suite Issue2241ReproTests passed
```

The unrelated `KiroStatusProbeTests` timing failure in the previous CI run appears to be a flaky timing test (elapsed 8.66s vs 8.0s threshold) and is not touched by this change.

Yuxin-Qiao:
Fixed the lint failure (SwiftFormat wrap/multiline brace issues in `UsageStore+Refresh.swift`) and force-pushed. Local `lint-macos` now passes cleanly.

Yuxin-Qiao:
Closing this PR: the branch accidentally carries the #2234 ProcessPipeCapture repro/fix (force-pushed here during local branch confusion), which now lives cleanly in #2258. The #2241 fix described in the title/body is already covered by #2247. The standalone reproducer for #2234 remains available in this PR's history if ever needed.

PR #2258, opened 2026-07-17
fix(ProcessPipeCapture): close read handle and defend against EMFILE on Linux (#2234)

### LINKED ISSUE #2234: CodexBarCLI serve (Linux) crashes SIGILL every ~2-4h: pipe fd leak → EMFILE → Pipe() yields fd=-1 → precondition in FileHandle.readabilityHandler

Yuxin-Qiao:
## Summary

- Fixes the file descriptor leak in `ProcessPipeCapture` on Linux that leads to `RLIMIT_NOFILE` exhaustion and the subsequent SIGILL crash described in #2234.
- In `stopAndSnapshot()`, explicitly `close()` the read-end `FileHandle` after capturing output. On swift-corelibs-foundation, clearing `readabilityHandler` alone does not release the internal dup'd monitor fd, so the pipe read end leaks over time.
- In `start()`, defensively probe whether another fd can be duplicated before installing `readabilityHandler`. If the process is already at EMFILE, the Foundation setter traps with `precondition(_fd >= 0)` and SIGILL. When the dup probe fails, fall back to synchronous draining in `stopAndSnapshot()` instead.
- Adds a Linux regression test `ProcessPipeCaptureLinuxTests` that spawns 100 child processes and asserts the process does not leak pipe read-end fds.

## Test plan

- [x] Reproduced SIGILL with the standalone reproducer on `issue2241-repro`.
- [x] Verified the fd-leak regression test passes on Ubuntu 24.04 (`swift test --filter ProcessPipeCaptureLinuxTests`).
- [x] The repro workflow now reports the (unfixed) baseline `Repro emfile` still crashes as expected, while the real `ProcessPipeCapture` test passes.

Closes #2234

### DIFF
--- a/Sources/CodexBarCore/Host/Process/ProcessPipeCapture.swift
+++ b/Sources/CodexBarCore/Host/Process/ProcessPipeCapture.swift
@@ -1,4 +1,7 @@
 import Foundation
+#if os(Linux)
+import Glibc
+#endif
 
 package final class ProcessPipeCapture: @unchecked Sendable {
     package static let defaultMaxBytes = 1 * 1024 * 1024
@@ -13,6 +16,7 @@ package final class ProcessPipeCapture: @unchecked Sendable {
     private var didReachEOF = false
     private var isStopping = false
     private var continuation: CheckedContinuation<Void, Never>?
+    private var usesReadableHandler = true
 
     package init(
         pipe: Pipe,
@@ -25,6 +29,22 @@ package final class ProcessPipeCapture: @unchecked Sendable {
     }
 
     package func start() {
+        #if os(Linux)
+        // swift-corelibs-foundation's readabilityHandler setter duplicates the
+        // underlying fd to create a dispatch source. If the process is already at
+        // EMFILE, that dup fails and the setter traps with precondition(_fd >= 0),
+        // producing the SIGILL reported in issue #2234. Defensively probe whether
+        // we can duplicate the fd ourselves; if not, fall back to synchronous
+        // reading in stopAndSnapshot() instead of installing the handler.
+        let fd = self.handle.fileDescriptor
+        let probe = Glibc.dup(fd)
+        guard probe != -1 else {
+            self.usesReadableHandler = false
+            return
+        }
+        Glibc.close(probe)
+        #endif
+
         self.handle.readabilityHandler = { [weak self] handle in
             self?.handleReadableData(from: handle)
         }
@@ -115,6 +135,27 @@ package final class ProcessPipeCapture: @unchecked Sendable {
         continuation?.resume()
     }
 
+    private func drainSynchronously() {
+        while !self.isStopping {
+            let chunk = self.handle.availableData
+            self.condition.lock()
+            if chunk.isEmpty {
+                self.isFinished = true
+                self.didReachEOF = true
+                self.condition.broadcast()
+                self.condition.unlock()
+                return
+            }
+            let remainingBytes = max(0, self.maxBytes - self.data.count)
+            if remainingBytes > 0 {
+                self.data.append(chunk.prefix(remainingBytes))
+            }
+            self.onData?()
+            self.condition.broadcast()
+            self.condition.unlock()
+        }
+    }
+
     private func waitUntilFinished() async {
         await withCheckedContinuation { continuation in
             self.condition.lock()
@@ -131,6 +172,12 @@ package final class ProcessPipeCapture: @unchecked Sendable {
     private func stopAndSnapshot() -> Data {
         self.handle.readabilityHandler = nil
 
+        // If we could not install the readability handler (e.g. EMFILE on Linux),
+        // read whatever data is available synchronously before closing the fd.
+        if !self.usesReadableHandler {
+            self.drainSynchronously()
+        }
+
         let continuation: CheckedContinuation<Void, Never>?
         let snapshot: Data
         self.condition.lock()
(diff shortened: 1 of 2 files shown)


steipete:
Merged as `7f8d33f8451c55691ca570f88fb06f4ffa2c6526` after verification on exact head `46730121be2f814f56f352876906990a609a77b4`.

Verification:

- `swift test --filter SubprocessRunnerTests` on macOS: 12/12 passed, including large output, cancellation, hung-process, detached-child, and timeout stress coverage.
- `swift test --filter ProcessPipeCaptureLinuxTests` in a Swift 6.3.3 Noble Docker environment with `libsqlite3-dev`: 5/5 outer tests passed; the isolated EMFILE child also reported 5/5. Coverage includes real `RLIMIT_NOFILE` exhaustion, setup-failure cleanup, blocked callbacks, continuous output, and 100-process descriptor reuse/leak stress.
- `make check`: SwiftFormat checked 1,421 files with no changes required; SwiftLint reported 0 violations across 1,420 files; repository checks passed.
- Maintainer autoreview of the full `origin/main...46730121` diff: no actionable findings.
- Exact-head CI run [29605817833](https://github.com/steipete/CodexBar/actions/runs/29605817833): Linux x64 and arm64 builds/tests green; macOS shards green (28m02s and 28m51s); lint, changes, aggregate `lint-build-test`, and GitGuardian green.

The Linux capture path now uses a nonblocking `DispatchSourceRead`, closes the read handle through source cancellation, bounds shutdown even under continuous writers or blocked callbacks, and avoids allocating a preflight descriptor under EMFILE. Contributor diagnosis and original fix retained in the merge history. Thanks @Yuxin-Qiao.

Issue #2253, opened 2026-07-17
Settings… menu item silently does nothing after the keepalive window is recreated as an unconfigured shell (relaunch required)


you3fen:
### Summary

After the app has been running for a while, clicking **Settings…** in the menu silently does nothing — no window, no error, no log line. Quitting and relaunching CodexBar fixes it until the next occurrence. This has recurred many times over several days.

Window-server forensics captured the failure live: the settings-open path relies on a notification relay through the invisible `CodexBarLifecycleKeepalive` window, and at failure time that window had been **recreated as an unconfigured shell** (untitled, default 500×500, normal layer) — so the relay's subscriber is gone and the open-settings notification lands in a void.

Third report from the same machine as #2222 / #2241.

### Environment

- CodexBar 0.43.0 (104)
- macOS 27 beta (Darwin 27.0.0), single built-in display, **3 Spaces** (possible trigger for window recreation; see below)
- Periodic `System memory pressure level=warning` entries in the verbose file log during the affected uptime

### Evidence: window-server state, broken vs. healthy

`CGWindowListCopyWindowInfo` for the CodexBar process, captured **while the bug was live** and again **right after relaunch**:

| | broken (before restart) | healthy (after restart) |
|---|---|---|
| title | *(empty)* | `CodexBarLifecycleKeepalive` |
| layer | 0 (normal) | 3 (floating, as configured) |
| bounds | 500×500 @ (0,482) | 20×20 @ (755,507) |
| onscreen | false | true |

The broken-state window looks exactly like a freshly created SwiftUI `WindowGroup` shell that never went through `HiddenWindowView.onAppear`: default size, no title, normal level. No Settings window existed on any Space at failure time (also confirmed via AX), ruling out the "opened on another Space" theory.

### Why that kills the Settings menu item

The menu action path is:

1. `StatusItemController.openSettings(pane:)` posts `.codexbarOpenSettings` (`StatusItemController+Actions.swift`)
2. `HiddenWindowView.onReceive(.codexbarOpenSettings)` calls `@Environment(\.openSettings)` (`HiddenWindowView.swift`)

Both links live inside the keepalive window's view hierarchy, and `onAppear` locates the window **by title** (`$0.title == "CodexBarLifecycleKeepalive"`). Once the window is recreated as an untitled shell:

- the title lookup can never re-apply the configuration, and
- the `onReceive` subscriber from the original view is gone, so the posted notification is simply dropped.

[...]

PR #2259, opened 2026-07-17
Fix Settings menu silently failing after keepalive window recreation

devYRPauli:
Addresses #2253. The keepalive lifecycle window was configured in `HiddenWindowView.onAppear` by locating it with a title string match (`$0.title == "CodexBarLifecycleKeepalive"`). When the window scene is recreated, that one-shot configuration does not re-run, leaving an unconfigured shell, and both settings-open paths then fail silently: the notification relay has no subscriber and the AppKit `sendAction` return value was discarded.

This change reconfigures the window through an `NSViewRepresentable` whose `viewDidMoveToWindow()` re-fires on scene recreation, and assigns a stable `NSWindow.identifier`. Both open-settings paths now detect an unhandled relay (via a synchronous ack) and fall back to the alternate path with diagnostics instead of failing silently.

Verified with `make build`, `make test` (0 failures), and `make check` (0 violations). Note: I could not reproduce the original macOS 27 recreation trigger on macOS 26.5, so this is defensive hardening of the identified paths rather than a fix confirmed against the live trigger.


steipete:
Merged as `ac96913543bdba752b3290583856fedd9e3d376b`.

Verification:

- Reviewed the complete external diff and traced the keepalive recreation path. The recreated shell lacked the SwiftUI notification relay/configuration; the merged change configures every newly attached keepalive window and gives both Settings entry points an acknowledged primary route plus the alternate fallback.
- Added regression coverage for the recreated-shell lifecycle callback and missing-relay fallback. `swift test --filter SettingsWindowOpeningTests`: 1 test passed, 0 failures.
- `make check`: passed; SwiftFormat reported 0 files requiring formatting and SwiftLint reported 0 violations.
- Maintainer autoreview after the final test hardening: clean, no accepted/actionable findings.
- Exact head `0a4c708208c1f6c12488c6f358fddb50be8e15c5`: changes, lint, both Linux builds, both macOS test shards, aggregate lint-build-test, and GitGuardian all passed. macOS shards completed in 30m50s and 21m40s.

Fresh-bundle live UI automation verification follows after merge. This comment records source review, regression, and CI proof; it does not claim that downstream live check yet.

PR #2263, opened 2026-07-17
Fix Claude probe session pollution

### LINKED ISSUE #2251: Claude CLI usage probe mints a new empty account session per probe, polluting the claude.ai session list (dozens of <host>-<word>-<word> entries)
## Summary

CodexBar's Claude CLI usage probe (`ClaudeCLISession`, working dir `~/Library/Application Support/CodexBar/ClaudeProbe`) launches a fresh headless `claude` CLI session on every probe. Each launch registers a new, permanently-empty session on the user's Anthropic account. Over weeks this pollutes the claude.ai / Claude desktop session list with dozens of entries auto-named `<hostname>-<adjective>-<noun>` (e.g. `macosvm-snug-moore`, `macosvm-typed-lovelace`, `macosvm-eager-conway`) — 36+ on my account so far, accumulating at roughly one per day since the ClaudeProbe workdir was created (Jun 7).

The user has no way to tell these come from CodexBar — they look like a rogue agent spawning sessions, which is how I initially investigated it.

## Environment

- CodexBar 0.43.0 (104), installed in `/Applications`, launched as Login Item
- Claude Code CLI 2.1.212, `~/.local/bin/claude`
- macOS 15.7.3 (arm64), hostname `macOSvm`
- Claude source: auto (OAuth path failing → CLI fallback active; `claudeOAuthDelegatedRefreshLastAttemptAtV1` = Jul 8)

## Evidence

- `~/.claude/projects/-Users-elpinguino-Library-Application-Support-CodexBar-ClaudeProbe/` is created when the probe runs — but contains **zero transcripts** (probes send no messages), so every probe leaves a nameless server-side session.
- Cross-referencing `~/.claude/session-env/` (one entry per locally-started session) against `~/.claude/projects/` transcripts: 11 transcript-less "phantom" sessions Jul 8–17, ~1/day, the last at 10:50 today.
- Prefs contain `claudeCLIUsageRateLimitBlockedUntilV1` — the probe even got rate-limited by Anthropic today, so the probe frequency is also hitting API limits.
- All words in the generated session names match the adjective/noun word lists embedded in the `claude` CLI binary (30/30 from my session list screenshot), confirming these are CLI-minted untitled sessions, not another device.

## Expected

Usage probing should not create durable account-side sessions, or at minimum should reuse one persistent session. Options that would fix it:

1. Reuse a single session (`--resume`/fixed `--session-id`) for all probes instead of minting a new one each time.
2. If the CLI/API gains a way to mark a session ephemeral/unlisted, use it.
3. Document the behavior and surface it in the UI ("Claude CLI probing creates sessions named <host>-…") so users don't chase ghosts.

## Actual

Every CLI probe = one new permanent `<host>-<random>-<random>` session in the account list; no cleanup, no attribution.

steipete:
## Summary

- persist one installation-local Claude probe UUID in the existing `ClaudeProbe` working directory
- pass that UUID through `--session-id` for both direct and watchdog-backed Claude CLI launches
- remove stale probe-owned transcripts before launch and keep the identifier file permission-restricted

## Root cause

The CLI usage source starts an interactive Claude session so it can invoke `/usage` and `/status`. It previously launched without a session identity, causing every cold probe launch to register another empty account-side session.

Claude CLI 2.1.212 still has no non-interactive quota command. Its `--no-session-persistence` flag is print-mode-only, so it cannot drive the interactive `/usage` panel. Reusing an explicit session ID preserves every field the existing probe collects without creating a new identity per refresh.

Background delegated refresh is already policy-gated after #2191, so this currently affects user-initiated CLI refreshes rather than scheduled background probes. Repeated refreshes and cold launches still polluted the account session list.

## Verification

- `swift test --filter ClaudeCLISessionTests` — 3 passed
- `swift test --filter ClaudeDirectUsageFallbackTests` — 3 passed
- `swift test --filter ClaudeProbeWorkingDirectoryTests` — 6 passed
- `make check` — SwiftFormat clean; SwiftLint 0 violations; repository checks passed
- autoreview — clean, no accepted/actionable findings

Live account-side counting was unavailable: the working local Claude CLI 2.1.206 is logged out, while the global 2.1.212 wrapper is missing its native optional dependency. No credentialed probe was attempted.

Closes #2251

### DIFF
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeCLISession.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeCLISession.swift
@@ -10,6 +10,8 @@ import Foundation
 actor ClaudeCLISession {
     static let shared = ClaudeCLISession()
     private static let log = CodexBarLog.logger(LogCategories.claudeCLI)
+    private static let probeSessionIDFilename = ".codexbar-session-id"
+    private static let fallbackProbeSessionID = UUID()
     #if DEBUG
     @TaskLocal private static var sessionOverrideForTesting: ClaudeCLISession?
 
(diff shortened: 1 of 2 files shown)


steipete:
Merged as `cfb7b8036087cc1186b963ada7bec2ce9efa71e2` from exact head `2d60bbb8266505652f5d9bb2542e8f910070140d`.

Verification:

- `swift test --filter ClaudeCLISessionTests`: 3 tests passed.
- `swift test --filter ClaudeDirectUsageFallbackTests`: 3 tests passed.
- `swift test --filter ClaudeProbeWorkingDirectoryTests`: 6 tests passed.
- `make check`: passed; SwiftFormat 0/1535 files need formatting, SwiftLint 0 violations across 1534 files, and all repository checks passed.
- `git diff --check`: clean.
- Autoreview: clean; no accepted/actionable findings.
- CI on exact head: changes, lint, Linux x64, Linux arm64, macOS shards 0/2 and 1/2, aggregate lint-build-test, and GitGuardian all passed. macOS shards completed in 36m42s and 30m06s.
- Merge state before merge: CLEAN and MERGEABLE. No approval was required by repository policy.

Live claude.ai session-list counting could not be run from this worktree: the usable installed Claude CLI reports `loggedIn: false`, while the newer global wrapper is missing its native optional package. This limitation was documented before merge. Source inspection and tests verify every direct/watchdog probe launch now receives the same persisted UUID through `--session-id`, with a stable process-local fallback when persistence is unavailable.

PR #2280, opened 2026-07-18
Allow single-account claude-swap cards

possibilities:
## Summary

- add `claudeSwapShowSingleAccount`, an opt-in Claude provider setting that prefers claude-swap cards when exactly one account is available
- apply the same account-count precedence to the menu-bar app and `codexbar cards`, while retaining the existing 2+ account default
- preserve zero-account fallback and explicit CLI account/source overrides; document and test config compatibility on macOS and Linux

## Why

The claude-swap integration currently requires at least two rows before it replaces the ambient Claude presentation. Users who intentionally manage a single account through claude-swap therefore cannot use its valid quota snapshot when the ambient Claude account is unavailable or different.

Changing the existing threshold globally would risk replacing ambient presentation for users with stale or sentinel-only single rows. This keeps that behavior backward-compatible and makes the alternate precedence explicit.

Follow-up to #2188 and the claude-swap precedence introduced in #1909.

## Testing

- `swift build --product CodexBarCLI`
- [redacted live proof](https://github.com/steipete/CodexBar/pull/2280#issuecomment-5009539899): a real one-account setup renders `[claude-swap]`, the active account marker, and both quota windows
- synthetic `cswap --list --json` fixture renders `@ synthetic@example.com [active]` from a one-account config without invoking ambient Claude
- focused macOS/Linux tests added for disabled/enabled single-account behavior, sentinel rows, explicit bypass, and config Codable compatibility
- `make check` portable, documentation, locale, and SwiftFormat checks pass locally; local SwiftLint cannot load SourceKit under the Command Line Tools-only environment
- the Command Line Tools-only installation also lacks Swift `Testing` and the SwiftUI `PreviewsMacros` plugin, so focused tests and the app target are delegated to CI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,6 @@
 
 ## 0.44.1 — Unreleased
 
-### Added
-- Claude: add an opt-in setting to show claude-swap cards when only one account is available, while preserving the
-  ambient Claude presentation by default.
-
 ### Fixed
 
 ## 0.44.0 — 2026-07-17
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 0.44.1 — Unreleased
 
+### Added
+- Claude: add an opt-in setting to show claude-swap cards when only one account is available, while preserving the
+  ambient Claude presentation by default.
+
 ### Fixed
 
 ## 0.44.0 — 2026-07-17
--- a/Sources/CodexBar/Providers/Claude/ClaudeProviderImplementation.swift
+++ b/Sources/CodexBar/Providers/Claude/ClaudeProviderImplementation.swift
@@ -26,6 +26,7 @@ struct ClaudeProviderImplementation: ProviderImplementation {
         _ = settings.claudeOAuthKeychainReadStrategy
         _ = settings.claudeWebExtrasEnabled
         _ = settings.claudeSwapEnabled
+        _ = settings.claudeSwapShowSingleAccount
         _ = settings.claudeSwapExecutablePath
     }
 
@@ -86,6 +87,9 @@ struct ClaudeProviderImplementation: ProviderImplementation {
         let claudeSwapBinding = Binding(
             get: { context.settings.claudeSwapEnabled },
             set: { context.settings.claudeSwapEnabled = $0 })
+        let claudeSwapShowSingleAccountBinding = Binding(
+            get: { context.settings.claudeSwapShowSingleAccount },
+            set: { context.settings.claudeSwapShowSingleAccount = $0 })
 
         return [
             ProviderSettingsToggleDescriptor(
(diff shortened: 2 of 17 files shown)


possibilities:
Addressed the concrete review finding: removed the release-owned `CHANGELOG.md` entry in 531200b2.

The PR body now includes synthetic one-account runtime proof. I am waiting for explicit operator approval before running the real Keychain/account-backed `codexbar cards --provider claude` probe needed for redacted live proof; this repository’s safety policy forbids live provider probes without that approval.

possibilities:
Real one-account proof from the freshly installed Keeper-managed build (`fork_sha=531200b2`, rebased mode), with the account identifier redacted:

```console
$ codexbar cards --provider claude --no-color
╭─────────────────────────────────────╮
│ Claude [claude-swap]                │
│ @ <redacted-account> [active]       │
│ ─────────────────────────────────── │
│ Session                    79% left │
│ [ ━━━━━━━━━━━━━━━━━━━━━━━━        ] │
│ Reset in 1h 22m                     │
│                                     │
│ Weekly                     22% left │
│ [ ━━━━━━                          ] │
│ Reset in 19h 22m                    │
╰─────────────────────────────────────╯
```

The resolved personal config has `claudeSwapShowSingleAccount: true`, and claude-swap reported exactly one active account. The output demonstrates that the real account now selects the `[claude-swap]` card and preserves its quota windows instead of falling through to ambient Claude.

The zero-account and explicit account/source bypass paths remain covered by the injectable macOS/Linux tests; I did not force either path against the live account because that would intentionally invoke the ambient provider flow.

@clawsweeper re-review

possibilities:
I decided to not use this. I'd close it but it looks like it's ready to go in so I'll go hands off, but I won't be a user of the feature so possibly makes most sense to close.

steipete:
Merged as `83f5a7e27d37b99c3a55666df6438a0bb2e533f9`.

Verification:

- Audited the full diff and confirmed the adapter remains opt-in. App refresh and CLI card replacement still require `claudeSwapEnabled`; zero-account lists retain ambient Claude output; one-account cards require the new explicit setting.
- Added maintainer docs commit `b30c1457b9035afaaca10a06271ab691bf177acf` to document the single-account opt-in exception.
- Ran a synthetic source-blind CLI behavior fixture: one sentinel account rendered one `Claude [claude-swap]` card with its problem text, and the fake adapter recorded exactly one `--list --json` invocation.
- `swift test --filter 'CLICardsClaudeSwapTests|ClaudeSwapMenuPrecedenceTests|ProviderSettingsDescriptorTests'`: 66 tests passed across 4 suites.
- `swift test --filter ShellCommandLocatorProcessTests`: 2 tests passed; this confirmed the first CI attempt's unrelated missing-temp-file failure was flaky.
- `make check`: documentation links, SwiftFormat, and SwiftLint passed with 0 violations.
- Maintainer autoreview: clean, no actionable findings.
- Exact head `b30c1457b9035afaaca10a06271ab691bf177acf`: all 9 CI checks passed after rerunning the single unrelated flaky macOS shard. Final state was CLEAN and MERGEABLE.

PR #2282, opened 2026-07-18
Fix widget token refresh cadence

### LINKED ISSUE #2103: Widget token/cost data should follow the global refresh frequency instead of a fixed hourly TTL

steipete:
## Summary

- derive token/cost snapshot freshness and its timer from the global refresh frequency
- apply a five-minute minimum to protect WidgetKit reload budgets; use the adaptive policy nominal interval; disable automatic token refresh in Manual mode
- restart the token timer when settings change, update widget docs/preferences copy, and cover cadence plus cache hydration behavior

## Verification

- `swift test --filter UsageStoreTokenRefreshCadenceTests` — 3 tests / 8 argument cases passed
- `swift test --filter UsageStoreCachedTokenHydrationTests` — 5 tests passed
- `swift test --filter 'UsageStore(ManualTokenRefresh|TokenRetryPolicy|WidgetSnapshot)Tests'` — 19 tests passed
- `make test` — 714 selections across 60 groups passed; one unrelated Claude fixture race recovered on the harness retry
- `make check` — locale checks, SwiftFormat, SwiftLint, docs, packaging, and repository checks passed
- autoreview — no accepted or actionable findings

Closes #2103


steipete:
Merged as `5d1d22630513c335619ffa8f2b4ed07a8c46c92a`.

Verification:

- `swift test --filter UsageStoreTokenRefreshCadenceTests`: 3 tests / 8 argument cases passed.
- `swift test --filter UsageStoreCachedTokenHydrationTests`: 5 tests passed.
- `swift test --filter 'UsageStore(ManualTokenRefresh|TokenRetryPolicy|WidgetSnapshot)Tests'`: 19 tests passed.
- `make test`: all 714 selections in 60 groups passed; one unrelated Claude temporary-fixture race recovered on the harness retry.
- `make check`: locale, SwiftFormat, SwiftLint, docs, packaging, and repository checks passed; 0 lint violations.
- Autoreview before commit and again before merge: no accepted/actionable findings; final confidence 0.94.
- CI run `29627447634`, attempt 2, exact head `f15a66d06eaf2bfe92918b1b9bac8b125649c84d`: all jobs green. Attempt 1 macOS shard 0/2 hit an unrelated missing SwiftPM cache directory before test discovery; the failed-job rerun passed.

Issue #2103 auto-closed by this merge.

PR #2285, opened 2026-07-18
Fix compact subagent copied prefixes

### LINKED ISSUE #2108: Codex real fork corpus: 5,427 replayed parent events across rewritten timestamps

hhh2210:
## Summary

CodexBar 0.44.0 includes #2228, but compact single-leaf subagent rollouts can still copy the parent counters before their first turn without embedding ancestor metadata. The scanner classifies that shape as independent, so the copied prefix appears as `Unknown model`.

This patch keeps `source.subagent` as lineage evidence and adds a narrowly gated second proof path:

- propose a local suffix only when an explicit parent exists, nonzero totals precede the first `turn_context`, and `inter_agent_communication(trigger_turn: true)` is physically adjacent;
- promote that candidate only when its input, cached-input, and output totals exactly match the parent's snapshot at the fork boundary;
- keep mismatch and unresolved-parent cases independent;
- make every candidate outcome depend on the parent snapshot, so a parent change invalidates the cached child classification.

The copied prefix is suppressed rather than relabeled. In the affected live and frozen corpora,
attributed model totals remain unchanged.

## Evidence

The released 0.44.0 build showed the residual on Jul 17:

An isolated scan of the same local files produced identical cold, warm, and forced results:

| Jul 17 | CodexBar 0.44.0 | Patched scanner |
|---|---:|---:|
| Total tokens | 549,936,863 | 500,408,358 |
| `gpt-5.6-sol` | 485,647,099 | 485,647,099 |
| `codex-auto-review` | 14,761,259 | 14,761,259 |
| Unknown | 49,528,505 | 0 |

The patch removes exactly the Unknown copied-prefix bucket.

I also reran the frozen 2,012-file corpus used to validate #2228. Its canonical manifest SHA-256 is
`7ec548cf8a49821c9abba5d0901e1d130d4546316e40f92f2cbb0decb1240368`:

- cold, warm, and forced scans are identical;
- Jul 16 remains 231,475,556 tokens, $186.318118, with zero Unknown and zero strict-matcher residual;
- the previous 30-day Unknown bucket falls from 307,441,975 to zero, and total tokens fall by exactly the same amount;
- only Jul 13 and Jul 14 change, by exactly their previous Unknown buckets;
- ten compact children that previously had lineage-only cache entries gain parent snapshot dependencies;
  the existing dependency is preserved, with no missing or unexpected dependency modes.

The proof used parser hash `48ac20dad61e9a7f`. It was run on `origin/main@1f422d579437abc4301a5b4d3cdb57e491445d5a` plus this patch. The branch is now rebased onto `9e9f739b6a96e2a786467adb59f1ce6b35f1d6ac`; the intervening commits do not change `Vendored/CostUsage` or the Codex subagent scanner paths.

## Maintenance invariant

An embedded ancestor can prove a copied prefix by itself. A compact rollout cannot: it needs both a first-turn boundary and an exact parent snapshot match. Missing or mismatched evidence must remain independent, never be guessed or relabeled. Future scanner changes should preserve both the positive and negative cases in these tests, including parent-dependent cache invalidation.

## Tests

- `make check`
- `CodexSubagentRolloutShapeTests` (17 tests)
- `CodexCompactSubagentAccountingTests` (3 tests)
- `CodexSubagentAccountingIntegrationTests` (11 tests)
- `CostUsageCacheTests` (9 tests)
- `CostUsageScannerBreakdownTests` (92 tests)

The focused suites passed with `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1` and
`DYLD_FRAMEWORK_PATH=.build/out/Products/Debug`; the latter supplies Sparkle's local build path to Xcode beta's test runner.

The full `make test` runner stops in `ClaudeOAuthCredentialsStoreTemporaryKeychainCacheTests`
(four assertions, also reproducible when that suite is run alone). This diff does not touch Claude credential storage.

This extends the rollout-shape coverage added in #2228. It does not introduce or replace the broader Phase 2 lineage ledger.

Fixes #2108
Refs #2037


hhh2210:
CI note: the only failing check is the Linux ARM64 run. It fails in the untouched
`TestsLinux/PlatformGatingTests.swift:128`, where both parameterized cases compare a temporary
Claude CLI invocation log. This PR does not change that test or the Claude provider.

The same base SHA (`9e9f739b`) shows the reciprocal flake on `main`: ARM64 passed while Linux x64
failed in a different temporary-file test. On this PR, Linux x64 passed and ARM64 failed. Both macOS
shards, lint, and the aggregate `lint-build-test` gate passed.

- PR failure: https://github.com/steipete/CodexBar/actions/runs/29633363184/job/88051196896
- base `main` run: https://github.com/steipete/CodexBar/actions/runs/29632108080

I tried to rerun the failed job, but GitHub restricts upstream reruns to repository admins. I am
keeping the scanner patch unchanged rather than mixing an unrelated Linux test stabilization into
this PR.

steipete:
Verified before merge:

- Reviewed the full scanner/parser/test diff and the #2228 cache contract. The compact shape is only promoted after an explicit parent, first-turn adjacent trigger, nonzero pre-boundary totals, and an exact parent snapshot match. Mismatch or unresolved parents remain independently counted.
- Confirmed cache safety: candidate classifications always retain the parent dependency key; parent changes invalidate the child. `Scripts/regenerate-codex-parser-hash.sh --check` reported `48ac20dad61e9a7f`. The producer hash rotated; no cache schema/layout changed, so no artifact-version bump was needed.
- Ran `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --filter 'Codex(Subagent|CompactSubagent)|CostUsageScannerBreakdown|CostUsageCacheTests'`: 129 tests in 5 suites passed.
- Ran `make check`: SwiftFormat clean; SwiftLint found 0 violations in 1,574 files; all repository checks passed.
- Ran maintainer autoreview twice, including immediately before merge: no accepted/actionable findings.
- Exact PR head `acf512d6ac408cca623b54addd242254b3d5bda9` passed all 9 checks in CI run 29657884872: lint, Linux x64, Linux arm64, Linux musl, both macOS shards, aggregate, changes, and GitGuardian.

Merged as `dbcba17b28b2841d38017a30c52ded5fb330a9e0`.

PR #2265, opened 2026-07-17
fix: clamp Abacus credit usage percent to 100%

OfficialAbhinavSingh:
## Summary

`AbacusUsageSnapshot.toUsageSnapshot()` computes the credit percent as `(used / total) * 100` **without** the `min(100, max(0, …))` clamp that every sibling credit provider applies — `OpenAIAPICreditBalanceFetcher`, `CommandCodeUsageSnapshot`, `CodebuffUsageSnapshot`. On an overage / shrunk-grant state (`creditsUsed > creditsTotal`) it flows e.g. `15000/10000 → 150` straight into `RateWindow.usedPercent`, which does not clamp — so the headline bar reads past full.

Same class as the recently merged #2255 (Cursor plan-percent clamp).

## Fix

```diff
-            (used / total) * 100.0
+            min(100, max(0, (used / total) * 100.0))
```

One line, matching the sibling idiom. No behavior change for in-range plans.

## Test

`TestsLinux/AbacusUsageSnapshotLinuxTests.swift`:
- overage `creditsUsed 15000 / creditsTotal 10000 → 100` (was `150`)
- in-range guard `250 / 1000 → 25`

Red before the source change (`150 ≠ 100`), green after.

## Verification

Built and tested on Linux (`swift:6.3.3`) — `AbacusUsageSnapshot` is in cross-platform `CodexBarCore`, so the test runs in the Linux shard. Build clean, `swiftlint --strict` + `swiftformat --lint` clean.

## Proof (real run)

The `AbacusUsageSnapshotLinuxTests` suite is executed on CI by the **`build-linux-cli`** job's "Swift Test (Linux only)" step (`swift test --parallel`) — both `linux-x64` and `linux-arm64` passed on this PR.

Below is a local red→green run in `swift:6.3.3` (`x86_64-unknown-linux-gnu`) with the exact same input (`creditsUsed: 15000, creditsTotal: 10000`); the one-line clamp is the only difference between the two runs.

**Before the fix** — clamp removed, overage flows through as `150`:
```
>>> line 25:  (used / total) * 100.0
✘ Test "credit overage clamps used percent to 100" recorded an issue at AbacusUsageSnapshotLinuxTests.swift:21:9:
  Expectation failed: (usage.primary?.usedPercent → 150.0) == (100 → 100.0)
✘ Test run with 2 tests in 1 suite failed with 1 issue.
```

**After the fix** — clamp applied (this PR, `c70597b`), overage caps at `100`:
```
[...]

PR #2255, opened 2026-07-17
fix: clamp Cursor plan usage percent to 100%

OfficialAbhinavSingh:
## Summary

On a usage-based Cursor plan in overage (`used > limit`), the headline plan-usage percent could exceed 100%. In `CursorStatusProbe`, the plan-ratio branch computed `(planUsedRaw / planLimitRaw) * 100` **without** the `normalizeTotalPercent` clamp that every sibling branch in the same function applies — so e.g. 15000/10000 flowed through as `150`, into `RateWindow.usedPercent`.

Overage is already surfaced separately (on-demand USD), so the headline percent is meant to cap at 100.

## Fix

Wrap that one branch in `normalizeTotalPercent` (`max(0, min(100, v))`), identical to the other branches:

```diff
-            (planUsedRaw / planLimitRaw) * 100
+            normalizeTotalPercent((planUsedRaw / planLimitRaw) * 100)
```

One line. No behavior change for in-range plans.

## Test

Added `plan ratio caps at 100 percent when usage exceeds the limit` — `used: 15000, limit: 10000` now yields `planPercentUsed == 100` (was `150`). Red before the source change, green after.

## Verification

- Source compiles clean on Linux (`swift:6.3.3`, `swift build --product CodexBarCLI` — `CursorStatusProbe.swift` is in `CodexBarCore`).
- The new test lives in the macOS `CodexBarTests` target, so the `swift-test-macos` CI job validates it.

Small and self-contained — happy to adjust or close if the clamp-on-overage behavior isn't wanted.

PR #2293, opened 2026-07-18
fix: clamp ElevenLabs usage percent to 100%

OfficialAbhinavSingh:
## Summary

`ElevenLabsUsageSnapshot` computes three `usedPercent` values **without** the `min(100, max(0, …))` clamp that every sibling credit provider applies — `OpenAIAPICreditBalanceFetcher`, `CommandCodeUsageSnapshot`, `CodebuffUsageSnapshot`, and the recently merged `AbacusUsageSnapshot` (#2265):

- `usedPercent` (character usage) applies a `max(0, …)` floor but **no ceiling** (`ElevenLabsUsageFetcher.swift:78`)
- both voice-slot windows apply **neither** bound (`:130`, `:140`)

ElevenLabs models overage explicitly via its `current_overage` field, so `characterCount > characterLimit` is a real, documented state. It flows e.g. `150000/100000 → 150` (and voice `12/10 → 120`) straight into `RateWindow.usedPercent`, which does not clamp — so the bar reads past full.

Same class as the merged #2255 (Cursor) and #2265 (Abacus) percent clamps.

## Fix

```diff
- return max(0, Double(self.characterCount) / Double(self.characterLimit) * 100)
+ return min(100, max(0, Double(self.characterCount) / Double(self.characterLimit) * 100))

- usedPercent: Double(used) / Double(limit) * 100,      // voice-slots  (x2)
+ usedPercent: min(100, max(0, Double(used) / Double(limit) * 100)),
```

Matches the sibling idiom. No behavior change for in-range values. The `characterLimit > 0` / `limit > 0` guards are untouched.

## Test

`TestsLinux/ElevenLabsUsageSnapshotLinuxTests.swift`:
- character overage `150000/100000 → 100` (was `150`)
- voice-slot overage `12/10 → 100` (was `120`)
- in-range guard `25000/100000 → 25`

## Proof (real run)

The suite runs on CI in the **`build-linux-cli`** job's "Swift Test (Linux only)" step (`swift test --parallel`). Local red→green run in `swift:6.3.3` (`x86_64-unknown-linux-gnu`), same inputs, clamp is the only difference:

**Before the fix** — clamps removed, overage flows through as `150` / `120`:
```
>>> line 78:  return max(0, Double(self.characterCount) / Double(self.characterLimit) * 100)
>>> line 130/140:  usedPercent: Double(used) / Double(limit) * 100,
✘ "character overage clamps used percent to 100": Expectation failed: (usage.primary?.usedPercent → 150.0) == 100
✘ "voice slot overage clamps used percent to 100": Expectation failed: (voice?.window.usedPercent → 120.0) == 100
✘ Test run with 3 tests in 1 suite failed with 2 issues.
```

**After the fix** — clamps applied (this PR):
```
✔ "voice slot overage clamps used percent to 100" passed
✔ "character overage clamps used percent to 100" passed
✔ "in-range character usage maps to its percent" passed
✔ Test run with 3 tests in 1 suite passed
```

Small and self-contained — happy to adjust or close if the clamp-on-overage behavior isn't wanted.

### DIFF
--- a/Sources/CodexBarCore/Providers/ElevenLabs/ElevenLabsUsageFetcher.swift
+++ b/Sources/CodexBarCore/Providers/ElevenLabs/ElevenLabsUsageFetcher.swift
@@ -75,7 +75,7 @@ public struct ElevenLabsUsageSnapshot: Codable, Sendable, Equatable {
 
     public var usedPercent: Double {
         guard self.characterLimit > 0 else { return 0 }
-        return max(0, Double(self.characterCount) / Double(self.characterLimit) * 100)
+        return min(100, max(0, Double(self.characterCount) / Double(self.characterLimit) * 100))
     }
 
     public var remainingCharacters: Int {
@@ -127,7 +127,7 @@ public struct ElevenLabsUsageSnapshot: Codable, Sendable, Equatable {
                 id: "voice-slots",
                 title: "Voice slots",
                 window: RateWindow(
-                    usedPercent: Double(used) / Double(limit) * 100,
+                    usedPercent: min(100, max(0, Double(used) / Double(limit) * 100)),
                     windowMinutes: nil,
                     resetsAt: nil,
                     resetDescription: "\(used) / \(limit)")))
@@ -137,7 +137,7 @@ public struct ElevenLabsUsageSnapshot: Codable, Sendable, Equatable {
                 id: "professional-voices",
                 title: "Professional voices",
                 window: RateWindow(
-                    usedPercent: Double(used) / Double(limit) * 100,
+                    usedPercent: min(100, max(0, Double(used) / Double(limit) * 100)),
                     windowMinutes: nil,
                     resetsAt: nil,
                     resetDescription: "\(used) / \(limit)")))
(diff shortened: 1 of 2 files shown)


steipete:
Maintainer verification:

- Reviewed the full two-file diff and adjacent ElevenLabs mapping. The clamp remains at the provider snapshot boundary, consistent with #2255 and #2265; raw overage counts remain visible in the summaries.
- `git diff --check origin/main...HEAD`: passed.
- `swift build --target CodexBarCore`: passed. One pre-existing StepFun interpolation warning was emitted; no ElevenLabs errors.
- `/Users/steipete/Projects/agent-skills/skills/autoreview/scripts/autoreview --mode branch --base origin/main --parallel-tests "swift build --target CodexBarCore"`: clean, no accepted/actionable findings.
- Exact head `fa595b11c6163a8e4f66f25c4489f197c379fa93`: CI 8/8 passed, including both Linux builds/tests, both macOS test shards, lint, and GitGuardian.
- Merged with merge commit `d3c3fdf004804011be93fb682c9c2b1fc52ff487`.

PR #2274, opened 2026-07-17
Add weekly session-equivalent headroom

steipete:
## Summary

- estimate remaining weekly quota as full 5-hour windows from the median burn across the latest seven completed active session windows
- show the estimate verdict first, then the numeric headroom line, only after at least three reliable samples
- support Codex, Claude, Antigravity/Gemini, and generic providers with unambiguous session/weekly pairs
- respect weekly work-day settings, isolate provider/account/window history, and memoize the existing history scan
- localize the new pluralized output and accessibility labels across all 23 catalogs

Addresses #2261. The issue remains open for reporter confirmation.

## Verification

- `swift test --filter SessionEquivalentForecastTests` — 23 tests passed
- `swift test --filter 'SessionEquivalentForecastTests|UsageStorePlanUtilizationTests'` — 101 tests passed
- source-blind behavior contract — 19 feature cases passed
- `make check` — SwiftFormat, SwiftLint, localization, docs, scripts, and repository gates passed
- all 23 stringsdict catalogs passed `plutil -lint`
- full sharded suite on the feature head before the final isolation hardening — 700 selections passed; exact-head CI is the final full-suite gate
- repeated branch autoreview; all accepted findings fixed, including named-window isolation and preservation of generic weekly reset/history behavior

## Presentation

Verdict leads; numeric estimate follows. The reporter has not answered the ordering question yet, so this uses the owner-approved default.

Issue #2261, opened 2026-07-17
The information I really need from this app


hdsheena:
<img width="713" height="303" alt="Image" src="https://github.com/user-attachments/assets/cd7ba983-ec0a-4b93-a45c-cee53cc4b897" />
when I won't have enough 5 hour intervals to use up my weekly, or how many intervals it will take to use up my weekly, cause I hate leaving weekly unused because I've exceeded my 5 hours :'(

steipete:
This is a good framing. The gauges tell you where you are, not what you can do. The real question is "will my weekly go unused because the 5-hour window keeps gating me?", and today CodexBar makes you do that arithmetic yourself.

A fair bit is already there in 0.44.0:

- The weekly pace line on the provider card: "X% in reserve / in deficit", "Runs out in ..." or "Lasts until reset", plus a run-out risk percentage. A big reserve is the current signal for "you're going to strand weekly quota". This works for Antigravity/Gemini weekly rows too.
- Session pace ("Projected empty in ...") for the 5-hour window.
- Settings > Menu > "Weekly progress work days", so projections don't assume you code weekends.
- "Show reset time when quota runs out" flips an exhausted gauge into a countdown.
- Opt-in predictive pace alerts before a window runs dry.

What you're actually asking for is different: "remaining weekly is about N full 5-hour windows, and only M windows are left before reset". That doesn't exist yet, and it's a genuinely good headline metric, so we're going to build it on top of the pacing engine. One honest caveat: the dashboard only gives us two percentages and reset times. 5-hour % and weekly % aren't the same units, so the count has to be derived from your own recent burn per window and will be clearly labeled an estimate.

Question back: at decision time, which helps more, the number ("~4 full windows of weekly left, 9 windows until reset") or the verdict ("you can't exhaust your weekly even running every window flat out")? We'll likely ship both, but your answer decides what leads.

steipete:
Shipped in #2274 (merge `1f422d579437abc4301a5b4d3cdb57e491445d5a`). Thanks for the sharp framing, @hdsheena.

The provider card now shows both estimate lines when a compatible 5-hour and weekly window has enough history:

1. The verdict leads: whether weekly can run out before reset at the recent pace, or approximately how many windows early it can run out.
2. The number follows: approximately how many full 5-hour windows of weekly quota remain, alongside how many windows remain until reset.

[...]

PR #2301, opened 2026-07-18
Normalize session equivalent forecasts

Zihao-Qi:
## Summary

- Measure weekly burn and session consumption over the same stored observation interval.
- Normalize aligned partial-session samples to a full five-hour allowance.
- Reject partial histories whose denominator cannot be established safely.
- Keep the existing provider/account/history isolation and median aggregation unchanged.

## Why

The merged estimator treated raw weekly burn from a partially used five-hour window as the cost of a full allowance. The first patch normalized by the window's maximum session usage, but that could cover a different time interval from the weekly delta.

The revised estimator pairs session and weekly observations captured by the same refresh, calculates both deltas over that shared interval, and then scales the result to a full allowance. Exact 0% and 100% boundary anchors remain supported; ambiguous partial samples are ignored.

The regression fixture passes 20%, 40%, and 100% partial-session histories through `UsageStore`'s hourly storage canonicalization. The previous implementation reported 12.5% weekly burn per full allowance; the interval-aligned implementation reports the expected 10% for all three samples.

## Validation

- `swift test --filter SessionEquivalentForecastTests` — 37 tests passed.
- `make check` — passed.
- `make test` — reached unrelated failures in untouched Claude OAuth/keychain ownership suites; the estimator suite passes independently.

Related: #2261
Follow-up to: #2274

### DIFF
--- a/Sources/CodexBar/SessionEquivalentForecast.swift
+++ b/Sources/CodexBar/SessionEquivalentForecast.swift
@@ -151,12 +151,18 @@ struct SessionEquivalentForecast: Equatable, Sendable {
 enum SessionEquivalentBurnEstimator {
     static let defaultSampleLimit = 7
     static let minimumSampleCount = 3
-    private static let boundaryTolerance: TimeInterval = 75 * 60
+    private static let observationAlignmentTolerance: TimeInterval = 0
     private static let resetEquivalenceTolerance = SessionEquivalentForecast.resetTolerance
 
     private struct SessionGroup {
         let resetsAt: Date
+        var entries: [PlanUtilizationHistoryEntry]
         var maximumUsedPercent: Double
+    }
+
+    private struct BurnObservation {
+        let sessionUsedPercent: Double
+        let weeklyEntry: PlanUtilizationHistoryEntry
     }
 
     static func estimate(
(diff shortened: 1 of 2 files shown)


Zihao-Qi:
@clawsweeper re-review

Updated in f54cdee6: normalization now uses session and weekly observations captured by the same refresh, and the regression fixture passes through UsageStore history canonicalization. The focused suite passes 37 tests and make check passes.

steipete:
Verified and merged.

- Reviewed the complete two-file diff and adjacent estimator/history-storage paths. No dependency, network, workflow, credential, or unrelated scope changes.
- Checked the #2274 contract: full-window display boundaries remain floored, fractional capacity still drives the verdict, and fewer than three accepted samples still returns nil.
- Added d4116d9475fd33649fe5491435c3ff03bde529f6 so boundary lookup chooses the nearest eligible sample at or before the boundary; added a straddling-boundary regression.
- `swift test --filter SessionEquivalentForecastTests`: 38 tests passed, 0 failures.
- `make check`: passed; SwiftFormat clean and SwiftLint reported 0 violations.
- `/Users/steipete/Projects/agent-skills/skills/autoreview/scripts/autoreview --mode branch --base origin/main`: clean, no accepted/actionable findings.
- Exact-head CI run 29655684953 for d4116d9475fd33649fe5491435c3ff03bde529f6: all checks passed, including both macOS shards, Linux x64/arm64, Linux musl, lint, aggregate, and GitGuardian.

Merge commit: 167db4dca93efc692acf9374e849d7dd30c72f66

PR #2184, opened 2026-07-15
fix: show StepFun credit-plan usage instead of false 0%

douxy1994:
## Summary

StepFun credit-based subscription plans (e.g. Mini, Pro — `plan_family: 2`) report usage through `plan_credit_rate_limit.subscription_credit_left_rate`, not the 5-hour/weekly rate windows. For these plans the API returns `five_hour_usage_left_rate: 0` and `weekly_usage_left_rate: 0` (meaning "no rate-limit window configured"), but CodexBar interpreted the `0` as "0% remaining → 100% used", showing a false **0% available** even when 96%+ of the credit remained.

This PR also fixes a secondary issue: the `Oasis-Webid` was hardcoded, so browser-imported (Manual mode) tokens — which carry a different `device_id` — caused `"auth failed: oasis-token is embezzled"` errors.

## Root cause

**`StepFunUsageFetcher.swift`** — `StepFunRateLimitResponse` decoded only the rate-window fields (`five_hour_usage_left_rate`, `weekly_usage_left_rate`) and completely ignored the `plan_credit_rate_limit` / `plan_family` fields. Credit-plan responses were mapped to `(1.0 - 0) × 100 = 100% used`.

Additionally, `webID` was a hardcoded constant. The StepFun Dashboard validates that the `Oasis-Webid` header matches the token JWT's `device_id` claim; a mismatch yields HTTP 401 `"oasis-token is embezzled"`.

### `StepFunUsageFetcher.swift`
1. **New response types**: `StepFunPlanCreditRateLimit` and `StepFunPlanCreditBucket` decode the `plan_credit_rate_limit` object (including `subscription_credit_left_rate`, `topup_credit_left_rate`, and `credit_buckets`).
2. **`StepFunFlexibleNumber`** now also decodes JSON-string-encoded numbers (the API returns `credit_total`/`credit_residual` as strings like `"400000000"`).
3. **Credit-plan detection**: `isCreditPlan` checks `plan_family == 2`, with a fallback heuristic (rate fields 0 + reset_time 0 + credit data present).
4. **`toUsageSnapshot()`**: for credit plans, the credit balance becomes the primary window and the meaningless 0%-left rate windows are dropped.
5. **Dynamic `webID(forToken:)`**: extracts `device_id` from the refresh-token JWT payload, falling back to the hardcoded default only during the initial device-registration/login flow.

### Tests (`StepFunUsageFetcherTests.swift`)
4 new test cases:
- `parses credit-plan response and maps credit as primary window`
[...]

PR #2312, opened 2026-07-18
Fix StepFun login web ID

Zihao-Qi:
## Summary

- derive the StepFun login web ID from the anonymous token before building the Cookie header
- add a regression assertion for the complete password-login Cookie header

## Root cause

The login request interpolated `self.webID`, which is a function reference after `webID(forToken:)` was introduced. Swift emitted a warning and the resulting Cookie header could contain a function debug description instead of the required web ID.

## Impact

Password-based StepFun fallback login now sends a valid `Oasis-Webid` value and the production build no longer emits the function-interpolation warning.

## Validation

- `swift test --filter StepFunTokenRefreshTests` — 8 tests passed
- `swift build -c release` — passed without the StepFun warning
- `make check` — passed


steipete:
Merged as `ac7b2b5e3606ee42b3e564d5afb8c67877dd6754` from exact head `aa9b486cc360e46f3c1a066226391b839e4e2d02`.

Verification:

- `swift test --filter StepFunTokenRefreshTests`: 9 tests passed, including separate fallback-ID and JWT `device_id` password-login cases.
- `make check`: passed; SwiftFormat reported 0 files requiring changes and SwiftLint reported 0 violations across 1,576 files.
- Structured autoreview against `origin/main`: clean, no accepted/actionable findings; overall result “patch is correct” (0.95 confidence).
- Exact-head CI: all 9 checks green — changes, lint, Linux x64, Linux arm64, Linux musl, both macOS shards, aggregate `lint-build-test`, and GitGuardian.
- Source review confirmed the anonymous token’s derived web ID is reused for both the `oasis-webid` header and the `Oasis-Webid` cookie value, avoiding the function-reference interpolation introduced by #2184.

Live StepFun password-login verification was unavailable because no live StepFun account was available in this environment. The request contract was verified with the focused URLProtocol regression test. Thank you @Zihao-Qi for the focused follow-up!

Issue #2117, opened 2026-07-13
Slow Claude usage refresh


dstier-git:
# Summary

The length of time it takes Claude limits to update since a refresh has triggered has ballooned recently. Observed using the "Auto" source for Claude usage.

Claude's refresh time was always slower than Codex and other major providers, but recently it has gotten far slower.

## Empirical evidence

Here is a comparison of the amount of time it took today for the most current rate to be fetched from 3 common providers:

|  | Codex | Cursor | Claude |
|--------|--------|--------|--------|
| Time since app is opened | 2.31s | 1.98s | **14.35s** |
| Time since "Refreshing" shows | 0.5s | Instant | **13.1s** | 

## Environment 
- Macbook Pro 2024, 24GB RAM, M4. 
- CodexBar 0.42.1. 
- Claude source="auto"

I cannot say with certainty when this issue began, although it was certainly before CodexBar 0.42.1 (current) and likely less than a full week ago.

steipete:
The likely main latency source is now removed from the normal background Auto path by #2191: CodexBar no longer waits for the 5-second `claude auth status --json` preflight during a default-policy background refresh. That opaque CLI probe was running on every Auto refresh and could consume the full timeout before the fetch plan continued.

#2191 is merged on `main` as `5f136af0899f9867165bae5b23b899065b2793f0`. Please re-test Claude refresh timing after the next CodexBar release and report the observed duration here. I am keeping this issue open until that release-build timing is confirmed.

jeroenev:
similar issue with OpenCode GO, takes about 23 seconds to get an update 😅 
```
time codexbar usage --provider opencodego && time codexbar usage --provider codex && time codexbar usage --provider claude
== OpenCode Go (local) ==
5-hour: 100% left [===========-]
Resets in 1h 36m
Weekly: 77% left [=========---]
Resets in 1d 9h
Monthly: 65% left [=======-----]
Resets in 2d 20h

real    0m19.736s
user    0m3.129s
sys     0m1.886s
== Codex 0.144.5 (oauth) ==
Weekly: 99% left [===========-]
Resets in 6d 23h
Credits: 0 left
Limit Reset Credits: 0 available
Account: xxxx@xxx.xxx
Plan: Plus

real    0m2.643s
user    0m0.823s
sys     0m0.662s
== Claude 2.1.211 (claude) ==
Session: 18% left [==----------]
Pace: 37% in deficit | Expected 45% used | Projected empty in 30m
[...]

PR #2316, opened 2026-07-19
Prefer OpenCode Go local usage in Auto mode

kiranmagic7:
## Summary
- prefer OpenCode Go local history before the web fetch only for unscoped Auto requests
- keep selected-token-account, manual-cookie, and workspace-scoped Auto requests web-first so scoped web resolution runs before local fallback
- keep explicit Web source web-only, and keep web fallback when local history is unavailable
- update strategy tests for unscoped local-first, scoped web-first, explicit Web, and fallback behavior

## Why
The OpenCode Go timing reported in #2117 shows `codexbar usage --provider opencodego` spending about 20 seconds. Auto previously ran the web strategy first, so a slow or missing web session could delay the fast local SQLite reader even when local history could satisfy an unscoped usage display.

The follow-up keeps that local-first optimization only for unscoped Auto requests. If the fetch context carries a selected token account, a manual cookie, or a workspace override from settings or `CODEXBAR_OPENCODEGO_WORKSPACE_ID`, Auto keeps the prior web-first order so the selected scope is applied before any local fallback.

## Tests
- `swiftc -parse Sources/CodexBarCore/Providers/OpenCodeGo/OpenCodeGoProviderDescriptor.swift`
- `swiftc -parse Tests/CodexBarTests/OpenCodeGoProviderStrategyTests.swift`
- `git diff --check`
- `swift test --filter OpenCodeGoProviderStrategyTests`
- `make check`

Not run: live OpenCode Go account/cookie probes. `AGENTS.md` says live provider probes, browser-cookie imports, `codexbar usage` against real accounts, and real SecItem reads require an explicit request; this PR uses focused strategy tests instead.

## Compatibility risk
Medium-low. This still changes OpenCode Go Auto precedence for unscoped requests, but selected token accounts, manual cookies, and workspace-scoped requests preserve the prior web-first behavior. `--source web` remains web-only, and Auto still falls through to web when local history is missing or unreadable.

### DIFF
--- a/Sources/CodexBarCore/Providers/OpenCodeGo/OpenCodeGoProviderDescriptor.swift
+++ b/Sources/CodexBarCore/Providers/OpenCodeGo/OpenCodeGoProviderDescriptor.swift
@@ -50,8 +50,8 @@ public enum OpenCodeGoProviderDescriptor {
             return [OpenCodeGoUsageFetchStrategy()]
         }
         return [
-            OpenCodeGoUsageFetchStrategy(),
             OpenCodeGoLocalUsageFetchStrategy(),
+            OpenCodeGoUsageFetchStrategy(),
         ]
     }
 }
--- a/Sources/CodexBarCore/Providers/OpenCodeGo/OpenCodeGoProviderDescriptor.swift
+++ b/Sources/CodexBarCore/Providers/OpenCodeGo/OpenCodeGoProviderDescriptor.swift
@@ -49,10 +49,33 @@ public enum OpenCodeGoProviderDescriptor {
         if context.sourceMode == .web {
             return [OpenCodeGoUsageFetchStrategy()]
         }
+        if self.requiresScopedWebStrategy(context: context) {
+            return [
+                OpenCodeGoUsageFetchStrategy(),
+                OpenCodeGoLocalUsageFetchStrategy(),
+            ]
+        }
         return [
             OpenCodeGoLocalUsageFetchStrategy(),
             OpenCodeGoUsageFetchStrategy(),
         ]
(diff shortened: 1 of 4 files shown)


steipete:
Merged as `9f2cf4809a0d9a4053c850a0b3b47bd27e82fc08`.

Verification:

- Reviewed the full four-file diff and traced `ProviderFetchPipeline` ordering and fallback behavior. Unscoped Auto now prefers device-local SQLite history, while selected token accounts, manual cookies, and workspace overrides remain web-first; explicit Web remains web-only.
- Added coverage for blank settings/environment workspace overrides so whitespace-only values remain unscoped.
- `swift test --filter OpenCodeGoProviderStrategyTests`: 9 tests passed in 1 suite.
- `make check`: passed, including repository checks, SwiftFormat (0/1578 files requiring formatting), and SwiftLint (0 violations in 1577 files).
- Final branch autoreview against `origin/main`: no actionable findings; patch judged correct (0.94 confidence).
- Exact head `1912bb5642ed18b7f2b9ae570a69a980888dc6a2`, CI run `29690930584`: all required checks green. The first Linux arm64 attempt hit the unrelated `CLICardsClaudeSwapTests` four-issue flake; `OpenCodeGoLinuxTests` and `OpenCodeGoPercentUnitLinuxTests` passed in that attempt, and the failed-only arm64 rerun passed.
- Confirmed the exact head is contained in `origin/main` after merge.

No live provider/account probe was run: repository policy requires explicit authorization for real provider and Keychain-backed probes, and this strategy-order behavior is fully exercised through the provider seam without touching credentials.

PR #2311, opened 2026-07-18
Serialize flaky Claude CLI platform-gating cases

Chipagosfinest:
## Summary

- serialize only the two parameterized Claude CLI platform-gating cases
- preserve the per-mode assertion that logged-out `.auto` and `.cli` runs invoke only `auth status --json`
- keep the containment test-only; no production concurrency behavior changes

Fixes #2309.

This flake surfaced in #2292's unrelated Linux x64 shard. Both parameterized cases started together and failed the same invocation assertion even though each already uses UUID-scoped binary and log paths. The deeper shared runtime boundary is not proven, so this PR takes the narrowest safe containment.

## Validation

- `swift test --filter PlatformGatingTests` — 10/10 passed
- `make check` — passed; 0 SwiftFormat changes and 0 SwiftLint violations across 1,572 files
- `make test` — passed all 716 selections in 60/60 groups, no retries or timeouts
- `git diff --check` — passed

Linux x64 CI is the authoritative regression gate for the originally observed flake.

## Exact-head Linux proof

- Linux x64 passed on commit `27230574533f2ee2566dd8f0ed3e1771b304bbef`: https://github.com/steipete/CodexBar/actions/runs/29659074330/job/88118445749
- Linux arm64 passed on the same commit: https://github.com/steipete/CodexBar/actions/runs/29659074330/job/88118445750

Both jobs retained the full Linux test step and CLI smoke test. The remaining macOS jobs are queued and unchanged by this `TestsLinux`-only diff.

PR #2320, opened 2026-07-19
Isolate Claude CLI gating tests

steipete:
## Summary

- isolate `TTYCommandRunnerActiveProcessRegistry` behind a task-local state override so shutdown-fence tests cannot reject unrelated concurrent Claude login launches
- serialize the complete Linux `PlatformGatingTests` suite, completing the partial Claude CLI gating serialization from #2311
- make Claude auth-status timeout overrides task-local and widen only test fixture deadlines that proved load-sensitive; production defaults and behavior are unchanged
- harden the named Kiro pipe-deadline assertion without changing the probe implementation

## Root cause

`TTYCommandRunnerEnvTests` is serialized only within its own suite. Its shutdown-fence cases mutate the process-global `TTYCommandRunnerActiveProcessRegistry.isShuttingDown` flag, while `ClaudeLoginRunnerTests` concurrently call `TTYCommandRunner.run`. When the fence wins, `beginLaunch()` rejects the Claude fixture launch and all three login tests receive `launchFailed`/empty output instead of their expected auth link or exit result.

The exact concurrent reproducer was:

```
swift test --filter 'ClaudeLoginRunnerTests|TTYCommandRunnerEnvTests'
```

It failed before the fix and now passes all 31 tests. The isolated registry state is task-local; the one test that crosses to a detached `Thread` explicitly captures its scoped state.

The Linux gating failure is a neighboring isolation class: #2311 serialized only the parameterized logged-out test's argument cases, not that test against the rest of `PlatformGatingTests`. Suite-level serialization closes that gap, while task-local auth timeout control keeps slow arm64 fake-CLI fixtures inside a test-only budget.

## Verification

- `swift test --filter ClaudeLoginRunnerTests`
- `swift test --filter 'ClaudeLoginRunnerTests|TTYCommandRunnerEnvTests'` — 31 tests passed
- arm64 Docker `PlatformGatingTests` — 10 consecutive focused runs passed
- `swift build -c release` — passed in 218.64s
- `make check` — SwiftFormat clean; SwiftLint 0 violations in 1,577 files
- three consecutive `make test` runs on the final debug implementation:
  - 60/60 first-pass groups, 719 selections, 0 retries/timeouts, 692.8s
  - 60/60 first-pass groups, 719 selections, 0 retries/timeouts, 575.6s
  - 60/60 first-pass groups, 719 selections, 0 retries/timeouts, 477.9s
- autoreview: no accepted/actionable findings

`swift test -c release` was also attempted after making the new seams configuration-independent. It compiled past this patch and then hit existing unrelated DEBUG-only OAuth test seams in `ClaudeOAuthDelegatedRefreshLinuxTests`; the production release build is green.

No changelog edit here; this is an internal test-infrastructure fix.

### DIFF
--- a/Sources/CodexBarCore/Host/PTY/TTYCommandRunner.swift
+++ b/Sources/CodexBarCore/Host/PTY/TTYCommandRunner.swift
@@ -8,101 +8,159 @@ import Musl
 import Foundation
 
 private enum TTYCommandRunnerActiveProcessRegistry {
-    private static let condition = NSCondition()
-    private nonisolated(unsafe) static var processes: [pid_t: ProcessInfo] = [:]
-    private nonisolated(unsafe) static var isShuttingDown = false
-    private nonisolated(unsafe) static var launchesInProgress = 0
-
     private struct ProcessInfo {
         let binary: String
         var processGroup: pid_t?
     }
 
-    @discardableResult
+    private final class State: @unchecked Sendable {
+        private let condition = NSCondition()
+        private var processes: [pid_t: ProcessInfo] = [:]
+        private var isShuttingDown = false
+        private var launchesInProgress = 0
+
+        @discardableResult
+        func register(pid: pid_t, binary: String) -> Bool {
+            guard pid > 0 else { return false }
+            self.condition.lock()
+            defer { self.condition.unlock() }
+            guard !self.isShuttingDown else { return false }
+            self.processes[pid] = ProcessInfo(binary: binary, processGroup: nil)
+            return true
+        }
+
+        func beginLaunch() -> Bool {
+            self.condition.lock()
+            defer { self.condition.unlock() }
+            guard !self.isShuttingDown else { return false }
+            self.launchesInProgress += 1
+            return true
+        }
+
+        func endLaunch() {
+            self.condition.lock()
+            self.launchesInProgress = max(0, self.launchesInProgress - 1)
+            if self.launchesInProgress == 0 {
+                self.condition.broadcast()
+            }
+            self.condition.unlock()
+        }
+
+        func updateProcessGroup(pid: pid_t, processGroup: pid_t?) {
+            guard pid > 0 else { return }
+            self.condition.lock()
+            guard var existing = self.processes[pid] else {
+                self.condition.unlock()
+                return
+            }
+            existing.processGroup = processGroup
+            self.processes[pid] = existing
+            self.condition.unlock()
+        }
+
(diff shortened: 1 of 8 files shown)


steipete:
Merged as `663099e8f9fb2680dabeaa811e844dd6090ea9aa` after exact-head CI completed green.

Root cause: `TTYCommandRunnerActiveProcessRegistry.isShuttingDown` was process-global. The shutdown-fence cases in `TTYCommandRunnerEnvTests` set that flag while `ClaudeLoginRunnerTests` concurrently entered `TTYCommandRunner.run`; `beginLaunch()` then rejected the unrelated Claude fixture launch, producing `launchFailed`/empty output and a missing auth link. Suite-level `.serialized` never isolated those two separate suites.

Fix: the active-process registry now resolves through a task-local isolated `State` in tests, while its default remains the shared production state. The detached-thread shutdown case explicitly captures its scoped registry state because task locals do not propagate to a raw `Thread`. `PlatformGatingTests` is serialized at suite scope, completing #2311's partial parameter-case serialization. Claude auth-status timing uses a task-local test override; the remaining timing changes are fixture-only, including the named Kiro deadline assertion. Production defaults are unchanged.

Verification:

- `swift test --filter ClaudeLoginRunnerTests` passed.
- `swift test --filter 'ClaudeLoginRunnerTests|TTYCommandRunnerEnvTests'`: 31 tests passed concurrently. This command failed before the registry isolation and passed after it.
- arm64 Docker focused `PlatformGatingTests`: 10 consecutive runs passed.
- `swift build -c release`: passed in 218.64s.
- `make check`: SwiftFormat clean; SwiftLint reported 0 violations in 1,577 files.
- Three consecutive `make test` runs on the final debug implementation each passed 60/60 groups and all 719 selections with 0 first-pass failures, 0 retries, and 0 timeouts: 692.8s, 575.6s, and 477.9s.
- Final local autoreview and branch autoreview both reported no accepted/actionable findings.
- Exact-head CI run `29673366301` on `a3cac151f9c05b9a33822e9116a7363b6b63dd36`: 9 passed, 0 pending, 0 failed, 0 cancelled. This includes `build-linux-cli (linux-arm64, ubuntu-24.04-arm)`, Linux x64, Linux musl, both macOS test shards, lint, and security checks.

`swift test -c release` was additionally attempted. It compiled past this patch's configuration-independent test seams, then stopped on pre-existing unrelated DEBUG-only OAuth seams in `ClaudeOAuthDelegatedRefreshLinuxTests`; the production release build above is green.

PR #2327, opened 2026-07-19
Prefer weekly windows in switcher progress

### LINKED ISSUE #2324: Switcher rows “Weekly progress” uses the primary/session quota
## Summary

In the merged menu, `Switcher rows: Weekly progress` does not show the weekly quota for most providers. It shows each provider's primary quota instead, which is commonly the 5-hour/session window.

This is easiest to see with Claude when session and weekly usage differ substantially: the small line under Claude follows the session value. The same row can therefore compare different time horizons across providers.

## Why Codex can look correct while Claude does not

A sanitized live check on the official v0.45.1 build showed the exact asymmetry:

- Codex had no active `primary` window and did have a 7-day `secondary` window, so `primary ?? secondary` fell back to weekly.
- Claude had both a 5-hour `primary` window and a 7-day `secondary` window, so the same expression selected the 5-hour session.

No usage percentages or account data are included here. The important part is that Codex only looks weekly when its primary value is absent; this is fallback behavior, not the `Weekly progress` setting selecting weekly.

## Steps to reproduce

1. Enable the combined menu-bar icon.
2. Set `Settings > Menu Bar > Combined icon > Switcher rows` to `Weekly progress`.
3. Set `Settings > Menu > Usage bars fill` to `As remaining`.
4. Open the merged menu with Claude, Codex, or another provider that exposes distinct session and weekly quotas.
5. Compare the switcher line with the session and weekly rows in the provider card.

## Expected

`Weekly progress` should use a real 7-day quota window whenever the provider exposes one. If no weekly window exists, falling back to the provider's available primary quota is reasonable.

If primary quota is the intended behavior, the option should be renamed to something like `Usage progress` and the switcher should offer a per-provider metric setting, just as the menu-bar item does.

## Actual

The setting is labelled `Weekly progress`, but `switcherWeeklyWindow` returns `primary ?? secondary` for the default provider path. For Claude and OpenCode Go, that selects the session window rather than weekly remaining.

The mismatch is visible in v0.45.1:

- [`Weekly progress` setting label](https://github.com/steipete/CodexBar/blob/v0.45.1/Sources/CodexBar/Resources/en.lproj/Localizable.strings#L569-L574)
- [`switcherWeeklyWindow` default selection](https://github.com/steipete/CodexBar/blob/v0.45.1/Sources/CodexBarCore/UsageFetcher.swift#L469-L496)

Commit [`11f4cc8d`](https://github.com/steipete/CodexBar/commit/11f4cc8d13419f9b8f2147bd81a00decaa7ec439) deliberately changed switcher bars from weekly remaining to primary quotas, but the user-facing setting and function name still say weekly.

## Environment

- CodexBar 0.45.1 (108), official notarized release
- macOS

Anneo22:
## Summary

- prefer an actual 7-day quota for the combined switcher's `Weekly progress` line
- choose the most constrained weekly lane when a provider publishes several model-specific limits
- preserve the existing provider-specific fallback when no weekly quota exists

## Why

The current selector generally uses `primary ?? secondary`. Codex can therefore appear correct only when its primary window is absent, while Claude selects its 5-hour primary window even when a 7-day secondary window is available. That makes the same switcher row compare different time horizons across providers.

This is a draft because commit `11f4cc8d` deliberately changed the switcher toward primary quotas. The patch implements the behavior promised by the current `Weekly progress` label, but the maintainer should confirm that product contract before merge.

## Tests

- `swift test --filter StatusItemControllerMenuTests`: 12 tests passed
- `make check`: passed, including SwiftFormat and SwiftLint with 0 violations
- `make test`: ran all 60 groups; the only failure was the unrelated locale-sensitive `ZenMuxProviderTests` date assertion (`12 Apr 2026` on this UK-locale machine versus expected `Apr 12, 2026`)

Fixes #2324

### DIFF
--- a/Sources/CodexBarCore/UsageFetcher.swift
+++ b/Sources/CodexBarCore/UsageFetcher.swift
@@ -467,6 +467,13 @@ public struct UsageSnapshot: Codable, Sendable {
     }
 
     public func switcherWeeklyWindow(for provider: UsageProvider, showUsed: Bool) -> RateWindow? {
+        // This surface is labelled "Weekly progress", so prefer a real 7-day lane when one is
+        // available. Some providers publish model-specific weekly lanes in extraRateWindows.
+        if let weekly = self.mostConstrainedSwitcherWeeklyWindow() {
+            return weekly
+        }
+
+        // Keep the existing provider-specific fallback for providers without a weekly allowance.
         switch provider {
         case .factory:
             // Factory prefers secondary window
@@ -494,6 +501,16 @@ public struct UsageSnapshot: Codable, Sendable {
         default:
             return self.primary ?? self.secondary
         }
+    }
+
+    private func mostConstrainedSwitcherWeeklyWindow() -> RateWindow? {
+        let standardWindows = [self.primary, self.secondary, self.tertiary].compactMap(\.self)
+        let namedWindows = self.extraRateWindows?
+            .filter(\.usageKnown)
+            .map(\.window) ?? []
+        return (standardWindows + namedWindows)
+            .filter { $0.windowMinutes == 7 * 24 * 60 }
+            .max { $0.usedPercent < $1.usedPercent }
     }
 
     public func accountEmail(for provider: UsageProvider) -> String? {
--- a/Tests/CodexBarTests/StatusItemControllerMenuTests.swift
+++ b/Tests/CodexBarTests/StatusItemControllerMenuTests.swift
@@ -29,6 +29,7 @@ struct StatusItemControllerMenuTests {
         primary: RateWindow?,
         secondary: RateWindow?,
         tertiary: RateWindow? = nil,
+        extraRateWindows: [NamedRateWindow]? = nil,
         providerCost: ProviderCostSnapshot? = nil)
         -> UsageSnapshot
     {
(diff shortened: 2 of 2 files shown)


steipete:
Maintainer verification:

- `swift test --filter StatusItemControllerMenuTests`: 12 tests passed, including weekly-over-session selection, most-constrained named weekly selection, and the no-weekly provider fallback.
- `make check`: passed; SwiftFormat found 0 files needing changes and SwiftLint found 0 violations across 1,577 files.
- `make test`: all 60 groups passed. One unrelated synthetic Grok child-process timing test in `BoundedChildProcessProofTests` timed out once; the repository harness retry passed all 66 tests in that group.
- Structured branch autoreview against `origin/main`: clean, with no actionable findings.
- Exact-head GitHub CI [run 29689779554](https://github.com/steipete/CodexBar/actions/runs/29689779554): all 9 checks passed for `7c84ad105b2b8989ad96dd1d4fbc965236436fbf`.

No live-account UI screenshot was taken because repository policy prohibits real provider/Keychain probes without an explicit request; the switcher selection behavior was verified at its stable model seam.

Merged with merge commit `7bd390c42d358b476e9e03abe62c7088ce42bed7`.

PR #2336, opened 2026-07-19
Keep full-session estimate visible while idle

Zihao-Qi:
## Summary

- Keep a learned full-session forecast available when the current five-hour window is idle and has no reset timestamp.
- Continue validating a current reset when one is present.
- Continue excluding historical session windows whose reset is still in the future.
- Refresh the idle forecast cache on minute boundaries so newly completed history becomes eligible.
- Cover idle forecast persistence and cache invalidation with focused tests.

## Why

An idle Claude five-hour window can have no `resetsAt` value. The forecast pipeline treated that missing reset as a hard failure even when enough completed history already existed to calculate the weekly full-session estimate. As a result, a learned estimate disappeared while idle and returned as soon as the user consumed any session quota.

The current session reset is not needed for the estimate itself. Weekly remaining quota and the learned weekly burn per completed full session are sufficient, so this change makes the reset optional without weakening the weekly-window or historical-evidence guards.

Because historical eligibility depends on the current time, idle cache keys also include a minute bucket. This avoids indefinitely reusing a stale `nil` or older estimate after a recorded session reset passes.

## User impact

Once CodexBar has learned a full-session estimate, it remains visible during idle periods and picks up newly completed historical sessions within the menu's minute-level refresh cadence.

## Validation

- `swift test --filter SessionEquivalentForecastTests` — 40 tests passed.
- `make check` — passed with no formatting or lint violations.
- Independent review found no actionable issues.
- `make test` was attempted earlier in development; it stopped in unrelated Claude OAuth temporary-keychain-cache tests (`ClaudeOAuthCredentialsStoreTemporaryKeychainCacheTests`).

### DIFF
--- a/Sources/CodexBar/SessionEquivalentForecast.swift
+++ b/Sources/CodexBar/SessionEquivalentForecast.swift
@@ -51,7 +51,6 @@ struct SessionEquivalentForecast: Equatable, Sendable {
               == self.sessionWindowMinutes,
               weeklyWindow.windowMinutes.map({ PlanUtilizationSeriesName.weekly.canonicalWindowMinutes($0) })
               == self.weeklyWindowMinutes,
-              let sessionResetsAt = sessionWindow.resetsAt,
               let weeklyResetsAt = weeklyWindow.resetsAt,
               weeklyWindow.usedPercent.isFinite,
               (0...100).contains(weeklyWindow.usedPercent),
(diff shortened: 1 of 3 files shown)


steipete:
Verified and merged. The final exact-head review of `66dee1672` and the required Codex autoreview were clean.

The earlier Linux x64 failure was isolated to a transient launch failure in the unchanged `CLICardsClaudeSwapTests` temporary executable fixture; the first rerun then stopped earlier on an apt repository mirror-sync mismatch. On the justified final rerun, the x64 release build, full Linux test suite (including that fixture), and CLI smoke test all passed. The remaining macOS, arm64 Linux, musl, lint, aggregate, and security checks were also green.

CI: https://github.com/steipete/CodexBar/actions/runs/30398256241/attempts/3

Thanks for the contribution, @Zihao-Qi!

Issue #2107, opened 2026-07-13
Allow more than 3 providers in the Overview tab (compact density mode)


elijah7x:
**Is your feature request related to a problem?**

CodexBar's value prop is keeping *many* AI coding providers visible at once (57+ supported), but in practice the popover's Overview tab only fits **3 providers on screen**. For someone tracking several providers in parallel — in my case z.ai, Claude, Codex, and more I'd like to add — that means scrolling instead of a single-glance overview, which is exactly the moment a menu bar app is supposed to serve.

This isn't only about tile height. `docs/ui.md` documents that the Overview tab intentionally "renders **up to 3 provider rows**" and the **Display → "Overview tab providers"** setting is capped at 3. So the limit is part intentional cap, part density: even with the cap raised, today's tile layout wouldn't fit many more.

**Current behavior (v0.42.0)**

Each provider renders as a tall tile, and a multi-window provider makes it acute. z.ai renders three windows (Tokens / MCP / 5-hour), each with its own bar + percentage + reset line, plus a 24h usage bar chart and a "most used model" line — one z.ai tile consumes roughly a third of the popover on its own. Conversely, a provider with only a status line (Claude "Refreshing…") still reserves a full tile whose content-to-padding ratio is ~20–30%. Net result: 3 providers fill the viewport. (Screenshot attached.)

**Why it matters**

The target user — a developer running multiple AI tools — is precisely the one who wants to see *all* active providers' remaining quota at a glance, not three at a time. The current density works against the product's own pitch.

**Proposed**

Two independent levers (either helps; the density mode is the real fix):

1. **Compact / dense display mode** — an opt-in toggle under Settings → Display, alongside the existing "Overview tab providers", "Menu bar metric", and icon/label/bar/reset-time controls. **Default unchanged** so existing users are unaffected. In compact mode:
   - Collapse each provider to ~1–2 rows: icon + name + primary-window bar + percentage + reset on a single dense line.
   - Multi-window providers (Tokens/MCP/5-hour, session/weekly) show the primary window by default; secondary windows reveal on hover/click or via a small expander.
   - Hide the 24h usage chart and peak/model-count extras in the compact list (keep them in the per-provider detail card).
   - Reduce inter-tile and section padding.

[...]

PR #2314, opened 2026-07-18
Allow up to six Overview providers

BobbyWang0120:
## Summary

- raise the shared merged Overview provider limit from three to six
- keep selection persistence, reconciliation, chooser copy, and menu rendering aligned with the shared cap
- update the UI documentation to describe the six-provider limit

## Why

The existing three-provider cap prevents users who monitor several AI providers from including them in the merged Overview. Six providers offers a more useful upper bound while preserving the current provider-row presentation and ordering behavior.

This addresses the provider-cap portion of #2107. The compact-density mode and secondary-window disclosure proposed there remain out of scope for this PR.

## User impact

Users can select and persist up to six providers in **Settings → Menu Bar → Overview providers**. When six are selected, additional providers remain unavailable until one of the selected providers is deselected.

## Validation

- `make check` — passed; documentation links, SwiftFormat, SwiftLint, locales, packaging, and repository checks are clean
- `make test` — all 716 selections passed in 60/60 groups with no failures, retries, or timeouts
- focused settings and menu suites — passed
- manual packaged-app verification — the requester confirmed the chooser displays a six-provider maximum and works correctly with the local build

## Visual verification

The updated chooser and six-provider selection boundary were manually verified in the packaged macOS app. No compact-layout changes are included.

PR #2339, opened 2026-07-19
Make merged-Overview provider limit user-configurable

Vit129:
## Summary
- The merged-icon Overview tab's provider cap was hardcoded (`docs/ui.md`, issue #2107). This makes it user-configurable via a Settings stepper (1-12), backed by `UserDefaults`.
- Rebased on current `main`, which already raised the hardcoded default from 3 to 6 — kept that as the new configurable default so existing users see no behavior change.

## Test plan
- [x] `swift build --product CodexBar -c debug` succeeds
- [ ] Manual: Settings > Menu Bar > "Overview tab limit" stepper changes how many providers render in the merged Overview icon

### DIFF
--- a/Sources/CodexBar/PreferencesMenuBarPane.swift
+++ b/Sources/CodexBar/PreferencesMenuBarPane.swift
@@ -3,7 +3,9 @@ import SwiftUI
 
 @MainActor
 struct MenuBarPane: View {
-    private static let maxOverviewProviders = SettingsStore.mergedOverviewProviderLimit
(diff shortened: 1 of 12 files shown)


steipete:
Closing: main already raised the merged-Overview cap to six (#2314), and the remaining ask in #2107 is a compact density mode rather than a configurable cap — this PR doesn't address that half, and the Overview redesign discussion in #2578 supersedes the direction. Thanks @Vit129.

Issue #1855, opened 2026-07-03
z.ai (BigModel CN) usage bars are wrong: MCP shows 100%, token windows show 0%


jackyngtf:
CodexBar 0.37.5, Windows, API region = BigModel CN (`open.bigmodel.cn`)

## What's wrong

All three usage bars display incorrect values for the BigModel China region:
- **MCP / tools → 100%** (actually 14%)
- **Tokens (5h) → 0%** (actually 8%)
- **Tokens (weekly) → 0%** (actually 7%)

## Raw API response

`GET https://open.bigmodel.cn/api/monitor/usage/quota/limit` (auth via Bearer token):

```json
{"code":200,"data":{"limits":[
  {"type":"TIME_LIMIT","unit":5,"number":1,"usage":1000,"currentValue":147,"remaining":853,"percentage":14,"nextResetTime":1784706344993,"usageDetails":[{"modelCode":"search-prime","usage":84},{"modelCode":"web-reader","usage":41},{"modelCode":"zread","usage":8}]},
  {"type":"TOKENS_LIMIT","unit":3,"number":5,"percentage":8,"nextResetTime":1783049703178},
  {"type":"TOKENS_LIMIT","unit":6,"number":1,"percentage":7,"nextResetTime":1783496744998}
],"level":"pro"},"success":true}
```

## Likely parsing bugs

1. **`TIME_LIMIT`**: the `usage` field (value `1000`) appears to be read as "used" instead of total, so it computes 100%. The correct `percentage` field (`14`) is ignored. (`currentValue`=147, `remaining`=853 → real usage is 14%.)
2. **`TOKENS_LIMIT` windows**: the `percentage` field (`8` and `7`) is not being read, so they show 0%. These entries have no `usage`/`currentValue`/`remaining` — only `percentage`.

`codexbar diagnose --provider zai` confirms the same wrong values (`secondary 100%`, `primary 0%`, `model_specific 0%`).

Auth config is correct (`auth.configured: true`, token read via `ZAI_API_TOKEN`). It's purely a response-parsing issue for the BigModel CN format.

PR #2342, opened 2026-07-19
fix: clamp z.ai raw-percentage usedPercent fallback to 0...100

OfficialAbhinavSingh:
## Summary

`ZaiLimitEntry.usedPercent` returns `computedUsedPercent` (which already clamps `min(100, max(0, …))`) when z.ai reports quota fields. But when those fields are absent it falls back to `return self.percentage` — the **raw API percentage, unclamped**:

```swift
public var usedPercent: Double {
    if let computed = self.computedUsedPercent { return computed }
    return self.percentage            // raw, can be < 0 or > 100
}
```

z.ai is known to omit/misreport quota fields (see #1855), so an out-of-range `percentage` flows straight into `RateWindow.usedPercent` and the menu bar — every other provider, and this struct's own computed branch, clamp to `0…100`.

## Fix

```diff
- return self.percentage
+ return min(100, max(0, self.percentage))
```

Matches `computedUsedPercent` (line 150) and sibling providers. No change for in-range values.

## Test

`TestsLinux/ZaiUsedPercentLinuxTests.swift` (usage `== nil` forces the fallback path):
- `150 → 100`, `-5 → 0` (clamped)
- `42 → 42` (in-range unchanged)
- computed-path precedence: `usage 100 / used 25 / percentage 999 → 25` (raw 999 does not leak)

## Proof (real run)

Runs on CI in the **`build-linux-cli`** "Swift Test (Linux only)" step. Local red→green in `swift:6.3.3`:

**Before the fix** — fallback returns the raw value:
```
✘ "raw percentage fallback clamps above 100":  fallbackUsedPercent(150) → 150.0 == 100   FAILED
✘ "raw percentage fallback clamps below 0":     fallbackUsedPercent(-5)  → -5.0  == 0     FAILED
✔ in-range + computed-path tests pass
✘ Test run with 4 tests in 1 suite failed with 2 issues.
```

**After the fix** — all pass:
```
✔ Suite ZaiUsedPercentLinuxTests passed
✔ Test run with 4 tests in 1 suite passed
```

Small and self-contained — happy to adjust or close if not wanted.

### DIFF
--- a/Sources/CodexBarCore/Providers/Zai/ZaiUsageStats.swift
+++ b/Sources/CodexBarCore/Providers/Zai/ZaiUsageStats.swift
@@ -84,7 +84,9 @@ extension ZaiLimitEntry {
         if let computed = self.computedUsedPercent {
             return computed
         }
-        return self.percentage
+        // The raw API percentage can fall outside 0...100 (z.ai omits/misreports quota fields);
+        // clamp it like computedUsedPercent and every sibling provider instead of surfacing it raw.
+        return min(100, max(0, self.percentage))
     }
 
     public var windowMinutes: Int? {
--- a/TestsLinux/ZaiUsedPercentLinuxTests.swift
+++ b/TestsLinux/ZaiUsedPercentLinuxTests.swift
@@ -1,0 +1,52 @@
+#if os(Linux)
+import CodexBarCore
+import Foundation
+import Testing
+
+struct ZaiUsedPercentLinuxTests {
+    /// usage == nil forces computedUsedPercent to return nil, exercising the raw-percentage fallback.
+    private func fallbackUsedPercent(_ percentage: Double) -> Double {
+        ZaiLimitEntry(
+            type: .tokensLimit,
+            unit: .hours,
+            number: 5,
+            usage: nil,
+            currentValue: nil,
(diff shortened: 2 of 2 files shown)


steipete:
Final maintainer review and isolated Codex autoreview were clean on exact head `39d86dae22e2522d65154ed65c483428a85be573`. The fallback clamp is bounded to raw z.ai percentages, preserves the computed-path precedence, and adds upper-bound, lower-bound, in-range, and precedence coverage.

Exact-head hosted CI was fully green, including lint, both macOS test shards, Linux x64/arm64 builds, musl, and the aggregate gate: https://github.com/steipete/CodexBar/actions/runs/29695150250

The focused local `swift test --filter ZaiProviderTests` build completed, but this machine's Xcode-beta SwiftPM test host could not load its generated Sparkle framework, so no local assertions ran; the hosted matrix above provides the exact-head execution proof, including the Linux target containing the new tests.

Related parser context: https://github.com/steipete/CodexBar/issues/1855 (reference only; no issue state change made here).

Merged as `fb32b844a4986af2ac4e5d8379b627561f111841`. Thank you for the careful fallback hardening and focused coverage, @OfficialAbhinavSingh!

Issue #2328, opened 2026-07-19
Qwen Cloud Token Plan (Personal/Solo) uses a different API than the Team-edition provider, and Linux gating blocks it


Yach0:
# Qwen Cloud "Token Plan (Personal/Solo)" — separate API from the existing Alibaba Token Plan provider, and Linux gating blocks it

The existing `alibabatokenplan` provider targets the **Team Edition** on
`modelstudio.console.alibabacloud.com` / `bailian.console.aliyun.com` via
`GetSubscriptionSummary` (`product=BssOpenAPI-V3`).

Qwen Cloud's **personal/solo** Token Plan (`home.qwencloud.com`) uses a
completely different console, gateway, and payload — and, unlike the Team
Edition's credit-pool model, it reports **rolling 5-hour + weekly usage
windows**, i.e. the same shape as Claude/Codex.

Verified end-to-end on Linux with plain `curl` (cookies only, no WebKit, no
`bx-ua`/`bx-umidtoken` anti-bot headers required).

## Gateway

```
POST https://cs-data.qwencloud.com/data/api.json
  ?product=sfm_bailian
  &action=IntlBroadScopeAspnGateway
  &api=zeldaHttp.apikeyMgr.%2Ftokenplan%2Fpersonal%2Fapi%2Fv2%2F<endpoint>

Content-Type: application/x-www-form-urlencoded
Origin:  https://home.qwencloud.com
Referer: https://home.qwencloud.com/billing/subscription/token-plan-individual
Cookie:  login_qwencloud_ticket, login_aliyunid_pk, cna, isg
```

Body (form-encoded):

```
product=sfm_bailian
action=IntlBroadScopeAspnGateway
sec_token=<secToken>
region=ap-southeast-1
params={"Api":"zeldaHttp.apikeyMgr./tokenplan/personal/api/v2/<endpoint>",
        "Data":{"cornerstoneParam":{"domain":"home.qwencloud.com",
                                    "consoleSite":"QWENCLOUD",
                                    "console":"ONE_CONSOLE",
                                    "xsp_lang":"en-US",
                                    "protocol":"V2",
                                    "productCode":"p_efm"}},
        "V":"1.0"}
```

`sec_token` is obtainable with cookies alone — no DevTools paste needed:

```
GET https://home.qwencloud.com/tool/user/info.json  ->  .data.secToken
```

This is the same `/tool/user/info.json` path the existing
[...]

PR #2356, opened 2026-07-20
fix: allow Alibaba/Qwen Token Plan usage on Linux with a manual cookie

OfficialAbhinavSingh:
## Summary

`sourceModeRequiresWebSupport` blanket-rejects web-source providers off macOS, but it already carries per-provider escape hatches for `opencodego`, `commandcode`, `cursor`, and `qoder` when their cookie source is **manual** — those fetches are plain `URLSession` + cookies, and only browser cookie *auto-import* actually needs macOS.

`alibabatokenplan` (Alibaba / Qwen Cloud Token Plan) was missing this exemption, so on Linux:

```
$ codexbar usage --provider alibaba-token-plan
Error: selected source requires web support and is only supported on macOS.
```

…even with `cookieSource: "manual"` + a configured cookie header. `AlibabaTokenPlanProviderSettings` already conforms to `ProviderCookieSettings`, so the same exemption applies cleanly.

This is the "Blocker on Linux" from #2328 (the full Qwen Cloud Personal API variant is a separate, larger follow-up).

## Fix

```swift
if provider == .alibabatokenplan,
   settings?.alibabaTokenPlan?.cookieSource == .manual
{
    return false
}
```

Placed alongside the existing `commandcode` / `qoder` exemptions (not `#if os(Linux)`-gated, since a manual cookie works on any platform; only `cursor`'s app-auth path is Linux-specific). No effect on macOS (already supports web). Only `.manual` is exempted — `.auto` / `.off` still require web support.

## Test

`TestsLinux/AlibabaTokenPlanLinuxTests.swift`:
- `cookieSource == .manual` → does **not** require web support (reachable on Linux)
- `cookieSource == .auto` → still requires web support (unchanged)

## Proof (real run)

Runs on CI in the **`build-linux-cli`** "Swift Test (Linux only)" step. Local red→green in `swift:6.3.3`:

**Before the fix** (exemption removed) — manual cookie is still gated:
```
✘ "manual cookie source does not require macOS web support"  FAILED  (sourceModeRequiresWebSupport → true)
✔ "auto cookie source still requires web support off macOS"  passed
✘ Test run with 2 tests failed with 1 issue.
```

**After the fix** — both pass:
```
✔ "manual cookie source does not require macOS web support" passed
✔ "auto cookie source still requires web support off macOS" passed
✔ Test run with 2 tests in 1 suite passed
```

Small and self-contained — happy to adjust or close if not wanted.

### DIFF
--- a/Sources/CodexBarCLI/CLIUsageCommand.swift
+++ b/Sources/CodexBarCLI/CLIUsageCommand.swift
@@ -694,6 +694,13 @@ extension CodexBarCLI {
         {
             return false
         }
+        if provider == .alibabatokenplan,
+           settings?.alibabaTokenPlan?.cookieSource == .manual
+        {
+            // The Alibaba/Qwen Token Plan fetch is plain URLSession + cookies; only browser
+            // cookie auto-import needs macOS, so a manual cookie header works off macOS too.
+            return false
+        }
         #if os(Linux)
         if provider == .cursor,
            settings?.cursor?.cookieSource != .off
--- a/TestsLinux/AlibabaTokenPlanLinuxTests.swift
+++ b/TestsLinux/AlibabaTokenPlanLinuxTests.swift
@@ -1,0 +1,30 @@
+#if os(Linux)
+import Foundation
+import Testing
+@testable import CodexBarCLI
+@testable import CodexBarCore
+
+struct AlibabaTokenPlanLinuxTests {
+    @Test
+    func `manual cookie source does not require macOS web support`() {
+        // The Alibaba/Qwen Token Plan fetch is plain URLSession + cookies, so a manually
+        // configured cookie header must be usable off macOS (matches qoder/commandcode).
+        #expect(!CodexBarCLI.sourceModeRequiresWebSupport(
+            .web,
+            provider: .alibabatokenplan,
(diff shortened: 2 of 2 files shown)


steipete:
Reviewed the exact head `5f2194445d654d33dd4755b09b977e024946649a` and the adjacent CLI gate and provider-strategy code. The diff is narrowly limited to allowing explicit Alibaba Token Plan manual-cookie settings through the non-macOS web-support gate; automatic browser import remains gated, and there are no provider parsing changes.

Verification:

- Exact-head CI: [run 29744620970](https://github.com/steipete/CodexBar/actions/runs/29744620970)
- All nine checks passed, including Linux x64/arm64 builds and Linux-only tests, the Linux musl build, both macOS test shards, lint, the aggregate gate, and GitGuardian.
- Isolated Codex autoreview completed cleanly with no actionable findings.
- A local focused `swift test --filter CLIEntryTests/test_sourceModeRequiresWebSupportIsProviderAware` was attempted, but the compiler stopped before test execution because the shared runner ran out of disk; no source or test failure was observed. Hosted CI provides the exact-head platform proof.

Thank you @OfficialAbhinavSingh for the focused fix and Linux regression coverage.

PR #2368, opened 2026-07-21
Show local Codex credit fallback as remaining percent

hengjiqi-oss:
## Summary

- In the local recovery bundle only, render a positive Codex credits fallback as a whole remaining percentage when a valid credit limit is available.
- Preserve raw credit values for the upstream bundle and for missing, zero, or otherwise invalid limits.
- Add characterization coverage for the local-only and fallback paths.

## Why

When Codex returns only a credits balance, the menu bar previously showed a raw number (for example, `42675.6`) while other quota sources use percentages. The local recovery build can safely calculate and display the remaining percentage only when its limit is known.

## Validation

- `swift test --filter CodexConsumerProjectionCharacterizationTests` (7 tests passed)

## Scope

The changed rendering behavior is deliberately gated to `com.steipete.codexbar.localrecovery`; the upstream bundle retains its existing behavior.

### DIFF
--- a/Sources/CodexBar/StatusItemController+Animation.swift
+++ b/Sources/CodexBar/StatusItemController+Animation.swift
@@ -986,9 +986,11 @@ extension StatusItemController {
     {
         guard let remaining = credits?.remaining, remaining > 0 else { return nil }
         if isLocalRecoveryBuild,
-           let remainingPercent = credits?.codexCreditLimit?.remainingPercent
+           let creditLimit = credits?.codexCreditLimit,
+           creditLimit.limit > 0,
+           creditLimit.remainingPercent > 0
         {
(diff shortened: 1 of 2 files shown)


steipete:
Thanks for this draft. It's been sitting since late July with the review still asking for real-behavior proof of the credit-fallback percentage display. If you're still pursuing it, could you rebase/refresh it and add that proof within the next two weeks? Otherwise we'll fold it into the next stale-queue cleanup — a fresh PR later is always welcome if the idea still applies.

steipete:
Closing after the requested refresh and proof deadline expired: this draft remains unrefreshed, and its behavior is gated exclusively to a local-recovery bundle identifier absent from shipped CodexBar. If you have a current user-facing credit fallback with real-account proof, please open a focused replacement against main.

PR #2371, opened 2026-07-21
Prevent WidgetKit reload loop in BurnDownRefreshSchedule (fix #2369)

### LINKED ISSUE #2369: [macOS 26.5.2] Sustained chronod disk writes while CodexBar widget reload state is active
## Environment

- CodexBar 0.45.2
- macOS 26.5.2 (25F84)
- Apple Silicon (`arm64`)

## Observation

On one Mac, with CodexBar and its widget left running, the `chronod` process continued accumulating disk writes. This is a local observation; CodexBar is not being asserted as the proven root cause.

Two read-only samples of the `chronod` process resource counters were taken 36 seconds apart:

- 2026-07-21 08:37:28 UTC: cumulative bytes written `148,540,096,512`
- 2026-07-21 08:38:04 UTC: cumulative bytes written `148,561,920,000`
- Delta: `21,823,488` bytes / 36 seconds = about `606,208 B/s` (`0.606 MB/s`, or about `52.4 GB/day` if sustained)

The counter is the process-level `ri_diskio_byteswritten` value from macOS `proc_pid_rusage`; it covers `chronod` I/O generally, not only one database file.

During the same paired reads, `ReloadState` contained 48,620 rows overall and about 40,232 rows whose bundle identifier contains `com.steipete.codexbar.widget`. The matching row count stayed bounded at about 40,232, while the latest `DateReloadRequested` advanced by about 35 seconds between the two reads, indicating that existing reload state continued to be updated/rotated rather than simply growing without bound. Earlier local inspection of this installation observed CodexBar widget reload entries at roughly 6-second intervals.

The final on-disk SQLite sizes observed were:

- Main `chrono.sql`: 8,257,536 bytes
- `chrono.sql-wal`: 32,768 bytes

The main database file is far smaller than the process's cumulative writes (about 17,991x smaller). This is consistent with SQLite/WAL write amplification, although the process counter alone does not prove that every written byte came from this database.

## Request

Could you please advise whether this reload/write pattern is expected on macOS 26.5.2, and how to distinguish normal WidgetKit/`chronod` behavior from an app-induced reload loop? If CodexBar can reduce or coalesce the reload activity, guidance on the appropriate cadence would be appreciated.

No account information, credentials, usernames, absolute local paths, or private widget data are included.

Yuxin-Qiao:
This PR prevents a WidgetKit reload loop where the widget daemon `chronod` sustained high disk writes when a provider's rate limit reset time was calculated very close to `now` (e.g. within a few seconds).

To fix this, we've enforced a minimum 5-minute reload interval safety threshold (per Apple's recommendation) on all dynamic timeline refreshes.

### Changes
- Enforced `minimumInterval = 5 * 60` in `BurnDownRefreshSchedule.nextRefresh`.
- Added test coverage for clamping behavior in `CodexBarWidgetProviderTests.swift`.

### Verification
- Focused and sharded unit tests passed successfully.

### DIFF
--- a/Sources/CodexBarWidget/BurnDownWidgetProvider.swift
+++ b/Sources/CodexBarWidget/BurnDownWidgetProvider.swift
@@ -147,6 +147,7 @@ struct BurnDownState {
 }
 
 enum BurnDownRefreshSchedule {
+    private static let minimumInterval: TimeInterval = 5 * 60
     private static let maximumInterval: TimeInterval = 30 * 60
 
     static func nextRefresh(
@@ -161,7 +162,12 @@ enum BurnDownRefreshSchedule {
             .filter { $0 > now }
             .min()?
             .addingTimeInterval(1)
-        return min(fallback, nextReset ?? fallback)
+        if let nextReset {
+            let target = min(fallback, nextReset)
+            let minimumDate = now.addingTimeInterval(self.minimumInterval)
+            return max(minimumDate, target)
+        }
+        return fallback
     }
 }
 
--- a/Tests/CodexBarTests/CodexBarWidgetProviderTests.swift
+++ b/Tests/CodexBarTests/CodexBarWidgetProviderTests.swift
@@ -826,11 +826,25 @@ struct CodexBarWidgetProviderTests {
             provider: .codex,
             primaryUsed: 20,
             secondaryUsed: 30,
+            primaryReset: now.addingTimeInterval(360),
+            secondaryReset: now.addingTimeInterval(420))
+
+        #expect(BurnDownRefreshSchedule.nextRefresh(snapshot: snapshot, provider: .codex, now: now)
+            == now.addingTimeInterval(361))
+    }
+
+    @Test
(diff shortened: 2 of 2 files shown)


steipete:
Merged after a final exact-head review of `83b6b91f05370857a891f1bc229b4063253443b2` found no blocking defects. The scheduler change is narrow: reset-driven WidgetKit timeline refreshes now have a five-minute minimum while retaining the existing 30-minute fallback and the reset+1s behavior outside that minimum window.

Verification:

- All nine hosted checks passed on the exact head, including both macOS Swift test shards, lint, Linux builds, and security checks: https://github.com/steipete/CodexBar/actions/runs/29818995941
- The focused scheduling coverage exercises both the clamped near-reset case and an unclamped reset beyond five minutes.
- Structured autoreview reported no accepted/actionable findings.
- A local focused test build was also attempted, but did not reach test execution because the linker hit `ENOSPC`; the exact-head hosted macOS shards are the completed test proof.

Thank you, @Yuxin-Qiao, for the focused fix and regression coverage.

PR #2372, opened 2026-07-21
Fix Claude scoped weekly pace

smkwray:
## Summary

- enable the existing pace projection for Claude extra rate windows
- cover a model-scoped seven-day `Fable only` window in the Claude account-card regression test

## Why

Claude model-scoped weekly limits are emitted as extra rate windows with a seven-day duration and reset date, so they contain everything the standard pace calculation needs. The extra-window pace helper was restricted to Codex and Antigravity, which left Claude's `Fable only` card showing usage and reset information but no reserve/deficit or run-out projection.

## Impact

Claude scoped weekly cards now display the same pace detail as the main weekly card. Other providers and non-weekly extra windows remain unchanged.

## Validation

- `git diff --check`
- release app target built successfully and the patched bundle was exercised with a live Claude `Fable only` card
- portable repository checks and SwiftFormat passed
- the focused Swift test could not run locally because this machine's standalone Command Line Tools installation lacks XCTest; upstream CI should exercise the added assertion
- SwiftLint could not start locally because the standalone Command Line Tools installation lacks `sourcekitdInProc.framework`


steipete:
Merged after full verification. Claude model-scoped seven-day windows (e.g. "Fable only") now show reserve/deficit pace detail.

Narrowed during review: the original guard enabled pace for any Claude extra window, which would have silently applied to any future 300-minute Claude window too — broader than the weekly fix intended. It's now restricted to `windowMinutes == 10080`, with negative coverage proving a non-weekly Claude extra window gets no pace detail.

Note on CI: the aggregate gate had been failing across reruns because `swift-test-macos` was *skipped* — the branch was stale enough that the path-change filter saw no macOS-relevant changes. Merging main in regenerated that comparison and the macOS shards then ran green. Thanks @smkwray — nice first contribution!

PR #2389, opened 2026-07-22
Touch Bar overlay, multi-provider widget, and reliability fixes

Vit129:
## Summary
- Add multi-provider Overview desktop widget
- Add persistent Touch Bar overlay with real provider logos, tap-to-show usage graph
- Fix duplicate Touch Bar registration, taps using Button instead of onTapGesture
- Fix widget reset times, Antigravity per-pool cadence, Kiro picker, wake refresh
- Log widget snapshot read/decode failures instead of silently swallowing them
- Fix Touch Bar taps, per-provider window labels, and IDE-only visibility

## Test plan
- [x] `swift build` passes
- [ ] Manual check: Touch Bar overlay shows provider cards, tap opens usage graph
- [ ] Manual check: Overview desktop widget renders multiple providers


steipete:
Closing after maintainer review: the Touch Bar overlay depends on dynamically loaded private DFRFoundation system-modal APIs, and the last Touch Bar Mac was discontinued in 2023 — that's a permanent private-API and hardware-test burden we can't take on. The desktop Overview widget half is inseparable from that surface in this branch. If there's demand for the widget alone, a clean WidgetKit-only PR would be reviewed on its merits. Thanks @Vit129 for the substantial work here.

PR #2391, opened 2026-07-22
Let users pick which providers show on the Overview widget and Touch Bar

Vit129:
## Summary
Both the Overview desktop widget and the Touch Bar overlay (#2389) previously took the first N enabled providers with no way to choose which ones.

- Overview widget: `AppIntentConfiguration` with 4 provider-picker slots, editable from the native "Edit Widget" popover
- Touch Bar overlay: new Settings > Menu Bar section with a checkbox picker (max 3 providers)
- Falls back to the previous first-N behavior when nothing is explicitly selected

**Depends on #2389** — this branch is stacked on top of it, so this diff only makes sense once that one lands. Rebasing onto `main` once #2389 merges will shrink this PR to just the picker commit.

## Test plan
- [x] `swift build` passes
- [x] New `SettingsStoreTests` cases: persist/normalize round-trip, default first-N fallback, add/remove respecting the cap
- [ ] Manual check: Edit Widget popover shows 4 provider slots and applies selection
- [ ] Manual check: Settings > Menu Bar > Touch Bar providers picker applies live


steipete:
Closing together with #2389: this branch contains all of #2389 (including the private Touch Bar runtime) plus the picker layer, so the same decision applies. A picker for a future WidgetKit-only Overview widget would be welcome on top of such a PR. Thanks @Vit129.

Issue #2159, opened 2026-07-14
Ollama Usages seems not showing now.


jkryanchou:
Currently, I have re-added Ollama APIKey into Codexbar, while it seems not showing the actual credit/token usages now. 
Doesn't it support it now?

<img width="321" height="288" alt="Image" src="https://github.com/user-attachments/assets/c47f7335-1928-40c4-99fb-16138afb6caf" />

devYRPauli:
I dug into this and there's a specific code-level reason the usage stays blank in `.auto` mode with only an API key configured.

The fetch pipeline runs strategies in order (cookie/web first, API key as fallback) and returns on the first success without merging - `ProviderFetchPipeline.fetch` in `Sources/CodexBarCore/Providers/ProviderFetchPlan.swift:248-256`. When there's no usable Ollama browser session, the cookie strategy throws `noSessionCookie`, `shouldFallback` allows it, and the API-key strategy runs instead.

The problem is that the API-key strategy "succeeds" but carries no usage data. `OllamaAPIUsageSnapshot.toUsageSnapshot()` in `Sources/CodexBarCore/Providers/Ollama/OllamaUsageFetcher.swift:701-713` returns:

    UsageSnapshot(primary: nil, secondary: nil, tertiary: nil, providerCost: nil, ...)

so the pipeline reports success with an all-nil snapshot and no error. `MenuCardView` only appends a usage metric when `snapshot.primary != nil` (`Sources/CodexBar/MenuCardView.swift:1152`), so the card ends up empty and silent.

This is by design at the API level - Ollama's token API doesn't expose the Cloud quota windows (noted in `docs/ollama.md`), so the API key path was only ever meant to validate access, not report usage. The actual bug is that when the pipeline falls back to the API-only path it renders a blank card instead of telling you the quota bars need browser-cookie auth.

To confirm your case specifically: is CodexBar signed into ollama.com in the browser you have set as the cookie source? If you can grab a verbose log (or a diagnose export), it'll show whether the cookie strategy is throwing `noSessionCookie` and falling back here, versus a genuine cookie-parsing regression - those are two different fixes.

Smallest fix I can see: when the pipeline falls back to the API-only strategy, surface a "sign in to Ollama for quota" state instead of a silently-empty snapshot - either attach that hint in `OllamaAPIFetchStrategy`, or have the card detect an all-nil Ollama snapshot and show the sign-in prompt. Happy to send a small PR for the messaging path if that direction sounds right.

[...]

Issue #2072, opened 2026-07-11
No Ollama session cookie found. Please sign in at https://ollama.com/signin in your browser.


fishcharlie:
I keep getting the error `No Ollama session cookie found. Please sign in at https://ollama.com/signin in your browser.`. I have already signed into Ollama.com on both Safari & Brave.

It doesn't really give much detail about how to fix this or even troubleshoot/debug it.

OS Version: macOS 26.5.2 (25F84)
CodexBar Version: 0.42.0

steipete:
@fishcharlie, the targeted cookie-access hints are now merged in #2249. Could you please retest Ollama usage with Safari and Brave?

Safari should now identify Full Disk Access as the blocker when its cookie database cannot be read. Brave should identify a declined Keychain cookie-decryption prompt and direct you to retry with a manual refresh; a disabled Chromium Keychain setting now gets its own enablement hint.

Keeping this issue open until you confirm the behavior on your setup.

fishcharlie:
@steipete Thanks! It is improved. But I'm still slightly confused about what "manual refresh" means. Does that mean just clicking the "Refresh" button in CodexBar under Ollama? Or does it mean something else? Not exactly sure what the "manual refresh" refers to.

The error I'm getting now is: "Chrome cookie decryption was declined in Keychain; retry with a manual refresh."

fishcharlie:
Oh maybe manual refresh means setting the Cookie manually in the CodexBar settings. If so, I took the `__Secure-session` Cookie from Brave, put it into those settings, but I'm still getting an error.

I've tried with both the prefix `Cookie: <BRAVE_COOKIE>` and just `<BRAVE_COOKIE>` since the placeholder doesn't really make it clear which one it should be. Both have the same result.

<img width="992" height="732" alt="Image" src="https://github.com/user-attachments/assets/df229ce3-9bae-4e08-8728-d416161fc95b" />

SuperSalsa20:
I can reproduce a closely related failure on CodexBar 0.45.2 with Ollama Cloud.

Environment/conditions:

- Ollama is enabled with Cookie source set to Auto.
- The user is signed in to Ollama in Chrome.
- Chrome contains a current `ollama.com` `__Secure-session` cookie (the value was not inspected or included here).
- Full Disk Access is granted to CodexBar.
- No Keychain prompt appears during a normal refresh.

Observed behavior:

1. A normal/background refresh reports:
[...]

PR #2404, opened 2026-07-23
Cache validated Ollama browser sessions

hxy91819:
## Summary

- reuse the last validated Ollama browser session instead of scanning every browser on each refresh
- keep the cached session on offline, transport, and general parse failures; only invalidate it on authentication failures
- preserve multi-browser fallback when a cached account is valid but does not expose quota data
- skip optional Safari Full Disk Access errors during automatic fallback while retaining explicit Safari recovery guidance
- add a staged Refresh action for automatic Cookie mode and hide it when Ollama is explicitly API-only

## Root cause

Ollama automatic refresh imported browser cookies on every poll. When the network was unavailable, no candidate could be validated and the final browser-import outcome could replace the real connectivity failure with a Safari permission or generic sign-in error.

## User impact

After one successful browser-cookie refresh, temporary network loss now reports the network failure without rescanning browsers or discarding the validated session. Authentication failures still trigger re-import, and users can explicitly refresh browser cookies from Settings.

## Validation

- `swift test --filter 'OllamaUsageFetcherTests|OllamaUsageFetcherRetryMappingTests|ProviderSettingsDescriptorTests'` (99 tests passed)\n- `make check`\n- structured Codex autoreview: clean, no accepted/actionable findings\n- release bundle built, ad-hoc signed, and deep-signature verified\n- `/Applications/CodexBar.app` launched with `CodexGitCommit=667beb1b`\n\n`make test` consistently stops in the unrelated existing `AdaptiveRefreshTimerTests` group: two timer cases time out after 30 seconds with `CancellationError`, including on the automatic retry. The Ollama-focused suites pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.45.3 — Unreleased
 
 ### Fixed
+- Ollama: skip inaccessible Safari cookies when another browser cookie store can be read.
 
 ## 0.45.2 — 2026-07-19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 ## 0.45.3 — Unreleased
 
 ### Fixed
-- Ollama: skip inaccessible Safari cookies when another browser cookie store can be read.
+- Ollama: skip inaccessible Safari cookies during automatic browser fallback while preserving explicit Safari
+  permission guidance.
 
 ## 0.45.2 — 2026-07-19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,8 @@
 ## 0.45.3 — Unreleased
 
 ### Fixed
-- Ollama: skip inaccessible Safari cookies during automatic browser fallback while preserving explicit Safari
-  permission guidance.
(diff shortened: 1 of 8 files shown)


steipete:
Merged after a final maintainer review of exact head `4b908586b003a5100306bf06bf0f9cd34fe60cb7`.

The full diff and current Ollama/cache boundaries were re-read with no blocking correctness defect found. The cache-first path invalidates only on authentication failures, retains the validated session on transport failures, preserves the cached parse error when browser fallback also fails, and limits Safari permission suppression to automatic fallback. The mixed fallback case also preserves a later non-Safari access error, with regression coverage for Safari denial followed by Brave denial.

Exact-head CI is fully green: https://github.com/steipete/CodexBar/actions/runs/30332987241. This includes both macOS test shards, lint, Linux builds, the aggregate gate, and security checks. An independent structured review also returned no blocking findings. A local focused test attempt built successfully but could not launch XCTest because the local bundle loader could not locate Sparkle.framework, so that attempt is not counted as test proof; hosted CI is the execution proof.

Known limitation: live offline and authentication-expiry behavior was not obtainable in this environment because real provider authentication and browser-cookie access were prohibited. Please report any regressions on #2072 or #2159.

Thank you, @hxy91819, for the focused implementation and regression coverage.

PR #2416, opened 2026-07-23
Fix duplicate menu bar items from leftover Codex hooks

uclort:
## Summary

- Treat leftover `--hook-event` launches as a no-op before SwiftUI/AppKit initializes.
- Keep the normal application launch path unchanged.
- Add focused launch-mode regression coverage and a changelog entry.

Follow-up to #2167 and continues the fix for #2162.

The previous revision attributed the duplication to status-item visibility recovery. The new process-level evidence below identified the actual root cause, so that visibility-recovery change has been fully reverted.

## Reproduction

1. Install [bob-zebedy/CodexBar](https://github.com/bob-zebedy/CodexBar) and enable its Codex Hook option.
2. The fork registers ten Codex lifecycle events in `~/.codex/hooks.json`, each using this handler with a five-second timeout:

   `'/Applications/CodexBar.app/Contents/MacOS/CodexBar' --hook-event`

3. Replace that app with upstream CodexBar at the same `/Applications/CodexBar.app` path without first removing the user-level hook registrations.
4. Enable Codex and DeepSeek as separate menu bar items, then use Codex normally.
5. Prompts, tool calls, permission requests, compaction, and session/subagent events temporarily add another complete set of provider icons. The extra set disappears after roughly five seconds.

The configured one-minute usage refresh is not the trigger. Codex activity is the trigger because it fires the leftover lifecycle hooks.

## Root cause

The other CodexBar implementation uses `--hook-event` as a lightweight child-process mode: it reads one event from stdin, records it, and exits before creating its menu bar UI. Its hook registration targets the generic installed app path rather than a fork-specific helper.

After the bundle at that path is replaced with upstream CodexBar, the registration still resolves successfully. Upstream CodexBar did not recognize `--hook-event`, so every lifecycle event started the full SwiftUI/AppKit application. Each transient process created its own merged, Codex, and DeepSeek `NSStatusItem` set. Codex then terminated the handler at its five-second timeout, explaining why the duplicates eventually disappeared.

Unified logging showed distinct CodexBar PIDs creating distinct autosave-named status items. A 500 ms Core Graphics window trace captured the exact sequence: the two normal provider windows became four when a hook process started, then returned to two when that process exited.

#2167 fixed reentrant status-item vending within one process. It could not address this case because these complete icon sets came from separate application processes.

## Why handle this in upstream

Both applications use the same bundle name and default installation path, so switching between them can leave a valid user-level integration pointing at whichever bundle currently occupies that path. Manually deleting the hooks is a workaround, but an unknown compatibility launch argument should not initialize a full menu bar UI ten times per Codex interaction.

Handling the established hook-mode argument at the process boundary is deliberately narrow: it prevents visible duplication and unnecessary app initialization without editing the user's hooks, adopting the fork's workflow-recording behavior, or touching provider and refresh business logic.

## Fix

Introduce a small explicit application entry point that resolves `CommandLine.arguments` before calling `CodexBarApp.main()`:

- normal launches continue into the existing app unchanged;
- launches containing the exact `--hook-event` argument return success before SwiftUI/AppKit initializes;
- similar arguments such as `--hook-events` remain normal launches.

The patch is limited to the app entry point, launch-mode tests, and the changelog. The earlier menu visibility changes are no longer part of the PR.

## Verification

- `swift test --filter CodexBarLaunchModeTests` — 4 tests passed.
- `make check` — SwiftFormat clean, strict SwiftLint reported 0 violations, and all repository checks passed.
- Release bundle built, installed at `/Applications/CodexBar.app`, and launched with Codex + DeepSeek enabled.
- Before the fix: each hook launch changed the provider window count from 2 to 4 with new window IDs, then back to 2 after the timeout.
- After the fix: 50 consecutive invocations of the installed `--hook-event` command kept the maximum provider window count at 2; the original two window IDs remained unchanged and all 50 processes exited successfully.

`make test` was also run. The affected launch-mode tests passed; the aggregate checkout remains non-green because of pre-existing assertions in `SessionEquivalentForecastTests`, `SpendDashboardControllerTests`, and `SpendDashboardTokenProvenanceTests`, none of which are touched by this patch.

### Inspectable runtime evidence

- [Redacted installed-bundle terminal transcript](https://github.com/user-attachments/files/30381703/codexbar-hook-event-runtime-proof.txt)
- Captured against reviewed head `4e1c2468400dd7b20246f7eec7ebe3981735304c` with Codex and DeepSeek visible as two separate status items.
- The monitor sampled only the two provider status-item window names every 1 ms via `CGWindowListCopyWindowInfo`; it did not read account data, provider configuration, tokens, endpoints, or hook payloads.
- The attached transcript records the before-fix peak of four status items and the after-fix result of 50 successful zero-exit invocations with a peak of two and unchanged window IDs.

### Compatibility policy for maintainer decision

This PR intentionally reserves only the exact `--hook-event` argument as a successful pre-UI compatibility no-op. It does not recognize similar arguments, edit or preserve hook configuration, consume hook payloads, or adopt the third-party workflow protocol. The purpose is limited to preventing stale registrations at the shared default app path from launching a visible second application. Whether upstream wants to keep that narrow compatibility contract is explicitly left to the maintainers.


uclort:
@clawsweeper re-review

Added inspectable, redacted installed-bundle runtime evidence to the PR body: [terminal transcript](https://github.com/user-attachments/files/30381703/codexbar-hook-event-runtime-proof.txt).

It captures the before-fix 2 → 4 → 2 status-item sequence and the after-fix result: 50 consecutive zero-exit hook invocations, no timeouts, a peak count of 2, and unchanged window IDs. The PR body also now makes the narrow compatibility policy explicit for maintainer decision.

steipete:
Merged after full verification. Per the owner decision, the exact `--hook-event` argument is now a permanent no-UI compatibility contract: it's classified before `CodexBarApp.main()` runs, so stale hooks from another CodexBar installation exit successfully instead of spawning duplicate menu-bar processes. Final diff is a tight 2-file/+48 change (the contributor changelog hunk was dropped per repo policy — maintainer adds it on release). Exceptional evidence in the PR body: the 2→4→2 status-item reproduction and 50 clean post-fix invocations. Exact-head CI fully green (https://github.com/steipete/CodexBar/actions/runs/30424725501). Thanks @uclort!

PR #2421, opened 2026-07-23
Fix localized session equivalents

Yuxin-Qiao:
## Summary
- format session-equivalent plural strings with the locale of the resource bundle selected by L
- keep application-language fallbacks and plural rules aligned
- add regression coverage for resource-locale fallback

## Why
When the system locale and the selected or fallback application language differ, the previous formatter could use plural rules from one language with strings from another. For example, English could render 1 windows instead of 1 window.

## Validation
- focused Swift test filter for SessionEquivalentForecastTests and LocalizationBundleCacheTests: 43 tests passed
- make check: passed
- full make test: the prior SessionEquivalentForecastTests failure is fixed; the run now reaches group 49 and stops on unrelated Spend Dashboard test isolation failures. Existing #2418 addresses the Controller subset; remaining Token Provenance coverage will be kept separate.

## UI proof
Not applicable: localized text behavior is covered by deterministic unit tests.

## Provider / privacy
No provider access, credentials, account data, or network behavior changed.

### DIFF
--- a/Sources/CodexBar/Localization.swift
+++ b/Sources/CodexBar/Localization.swift
@@ -214,7 +214,22 @@ func L(_ key: String, language: String) -> String {
 }
 
 func codexBarLocalizedLocale() -> Locale {
-    let language = resolvedAppLanguage()
+    codexBarLocale(forLanguage: resolvedAppLanguage())
+}
+
+/// Returns the locale of the resource bundle currently selected by `L`.
+///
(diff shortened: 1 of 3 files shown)


Yuxin-Qiao:
Superseded by #2418, which now contains this localized plural fix together with the complete macOS test-baseline restoration and full validation.

PR #2430, opened 2026-07-23
fix: overlay authoritative OpenCode Go web windows onto local usage

aseelye:
**Summary**
In unscoped Auto mode, OpenCode Go usage windows come from local `opencode.db` cost sums against the plan limits ($12/$30/$60). The monthly window is anchored at the earliest local usage row instead of the real billing cycle, so for users with history predating their cycle anchor it sums too much — exceeding the $60 limit and clamping to 100% used ("0% left") — and shows a fabricated reset date. Reproduced with a real account: dashboard showed 64% used / resets Aug 10, while CodexBar showed 0% left / reset in 3d.

**Fix**
When a cached or manual `opencode.ai` session cookie is available, the local strategy now overlays the server-reported rolling/weekly/monthly percentages and reset countdowns (plus Zen balance and renewal date) onto the local snapshot, keeping the local daily cost history. Falls back to the previous local-only behavior when web usage is unavailable. `URLError.cancelled` from URLSession is normalized to `CancellationError` so cancelled refreshes can't complete with stale estimates.

No strategy-order changes, no fresh browser imports on this path, scoped web-first behavior untouched.

**Verification**
- `swift test --filter OpenCodeGo` — 87 tests pass (10 new in `OpenCodeGoWebOverlayTests`)
- `make check` — 0 violations
- Real account before/after: Monthly 0% left / reset 3d 17h → 36% left / reset ~17d, matching the opencode.ai dashboard

### DIFF
--- a/Sources/CodexBarCore/Providers/OpenCodeGo/OpenCodeGoProviderDescriptor.swift
+++ b/Sources/CodexBarCore/Providers/OpenCodeGo/OpenCodeGoProviderDescriptor.swift
@@ -84,33 +84,66 @@ struct OpenCodeGoLocalUsageFetchStrategy: ProviderFetchStrategy {
     let id: String = "opencodego.local"
     let kind: ProviderFetchKind = .localProbe
 
+    typealias LocalSnapshotLoader = @Sendable (ProviderFetchContext) throws -> OpenCodeGoUsageSnapshot
+    typealias WebUsageOverlayFetcher = @Sendable (ProviderFetchContext, String) async throws
+        -> OpenCodeGoUsageSnapshot?
+
+    private let localSnapshotLoader: LocalSnapshotLoader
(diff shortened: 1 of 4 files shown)


steipete:
Merged after full verification. OpenCode Go now overlays authoritative web quota windows onto local usage when a cached session exists, fixing monthly usage that was anchored to the oldest local row instead of the billing cycle — while preserving local daily history, propagating cancellation, and keeping local output intact when the web call fails.

Fixed during review: an invalid cached cookie was silently retried forever, so a stale inaccurate estimate persisted until manual re-import. Invalid credentials now evict on authentication failure while being retained on transport failure — matching the Grok cookie-cache semantics from #2458, so a flaky network does not log you out. Thanks @aseelye — great first contribution!

PR #2434, opened 2026-07-24
Fix duplicate weekly row from garbled Claude all-models capture

guhyun9454:
## Summary

On a CLI-auth Claude account the menu can show the weekly (all models) limit **twice**: once as the real **Weekly** row and again as a bogus model-scoped row titled **"all modls only"** (note the dropped character), sitting right next to it with a slightly different percentage/reset.

## Root cause

`claude /usage` is a redrawing TUI, and CodexBar scrapes it through a PTY. A half-painted capture frame can drop a character from `Current week (all models)`, so the same window is captured twice in one scrape — once clean (`all models`) and once garbled (`all modls`).

`extractScopedWeeklyUsages` filtered the all-models bucket out of the model-scoped windows with an **exact** string compare:

```swift
guard normalizedModel != "allmodels", !normalizedModel.isEmpty else { continue }
```

The clean copy is dropped (and used for the Weekly row), but the garbled copy normalizes to `allmodls != allmodels`, slips through, and is surfaced as an extra `NamedRateWindow` → the duplicate `"all modls only"` row. This is the same class of duplicate the 0.44.0 fix ("suppress duplicate all-model scoped quota rows that could appear as 'All models only' beside Weekly") targeted, but the exact-string guard is defeated by a single-character capture corruption.

## Fix

Match the all-models bucket with a small edit-distance tolerance (Levenshtein ≤ 2) instead of exact equality, so minor TUI capture corruption no longer escapes the filter. Genuine model-scoped names (Opus, Sonnet, Fable, …) are far outside that window, so they are unaffected.

## Test plan

Added `parse claude status ignores garbled all-models weekly duplicate` to `StatusProbeTests` (feeds a clean + garbled all-models frame plus a Fable window; asserts Weekly is captured and the only scoped row left is `Fable only`).

Because `@testable import CodexBar` links the full app target (whose SwiftUI `#Preview` macro needs full Xcode) I couldn't run the whole `swift test` suite locally on a Command-Line-Tools-only box, so I also verified the change end-to-end by linking the real compiled `CodexBarCore` and calling the public `ClaudeStatusProbe.parse` directly:

**Before the fix** (garbled input):
```
weeklyPercentLeft: Optional(34)
extraRateWindows (2):
  - title=all modls only  used=67.0  reset=Resets Jul 24 at 1:59pm (Asia/Seoul)
  - title=Fable only      used=71.0  reset=Resets Jul 24 at 2pm (Asia/Seoul)
RESULT: BUG — duplicate all-models row(s): ["all modls only"]
```

**After the fix:**
```
Case 1: garbled all-models duplicate
        weeklyPercentLeft=Optional(34)  extraRateWindows=["Fable only"]
  PASS  weekly (all models) still captured
  PASS  no duplicate all-models row; only Fable survives
Case 2: clean panel (regression guard)
        weeklyPercentLeft=Optional(34)  extraRateWindows=["Fable only"]
  PASS  weekly parsed
  PASS  clean panel unchanged: only Fable
Case 3: real scoped model preserved
        opusPercentLeft=Optional(90)  extraRateWindows=[]
  PASS  weekly parsed
  PASS  Sonnet-only scoped limit still recognized

ALL CHECKS PASSED
```

Logs are from synthetic fixtures — no account identity/PII. `swiftformat` was run on the changed files; `swiftlint` couldn't load `sourcekitdInProc` on this CLT-only box (CI covers lint).

### DIFF
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeStatusProbe.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeStatusProbe.swift
@@ -784,7 +784,14 @@ extension ClaudeStatusProbe {
             return normalizedLine.contains(normalizedLabel)
         }
         guard let actualModel = self.weeklyModelName(from: line) else { return false }
-        return self.normalizedForLabelSearch(actualModel) == self.normalizedForLabelSearch(expectedModel)
+        let expectedNormalized = self.normalizedForLabelSearch(expectedModel)
+        let actualNormalized = self.normalizedForLabelSearch(actualModel)
+        // When looking up the all-models weekly bucket, tolerate garbled TUI captures ("all modls")
+        // so the Weekly percent/reset are still recovered even if the clean copy never survived.
+        if self.isAllModelsWeeklyModel(expectedNormalized) {
+            return self.isAllModelsWeeklyModel(actualNormalized)
+        }
+        return actualNormalized == expectedNormalized
     }
 
     private static func crossesLabelBoundary(
@@ -798,7 +805,12 @@ extension ClaudeStatusProbe {
             return !normalizedLine.contains(normalizedLabel)
         }
         guard let actualModel = self.weeklyModelName(from: line) else { return true }
-        return self.normalizedForLabelSearch(actualModel) != self.normalizedForLabelSearch(expectedModel)
+        let expectedNormalized = self.normalizedForLabelSearch(expectedModel)
+        let actualNormalized = self.normalizedForLabelSearch(actualModel)
+        if self.isAllModelsWeeklyModel(expectedNormalized) {
+            return !self.isAllModelsWeeklyModel(actualNormalized)
+        }
+        return actualNormalized != expectedNormalized
     }
 
     private static func weeklyModelName(from line: String) -> String? {
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeStatusProbe.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeStatusProbe.swift
@@ -341,6 +341,17 @@ extension ClaudeStatusProbe {
     }
 
     private static func extractPercent(labelSubstring: String, context: LabelSearchContext) -> Int? {
+        // Prefer an exact label match; only fall back to a fuzzy (garbled-capture) match when no exact
+        // copy yields a value, so a clean row always wins over a corrupted duplicate regardless of order.
+        self.extractPercent(labelSubstring: labelSubstring, context: context, allowFuzzy: false)
+            ?? self.extractPercent(labelSubstring: labelSubstring, context: context, allowFuzzy: true)
+    }
+
+    private static func extractPercent(
+        labelSubstring: String,
+        context: LabelSearchContext,
+        allowFuzzy: Bool) -> Int?
+    {
         let lines = context.lines
         let label = self.normalizedForLabelSearch(labelSubstring)
         for (idx, line) in lines.enumerated() {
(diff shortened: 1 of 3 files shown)


guhyun9454:
Pushed `aef55fd` addressing the review:

- **Codex (P2 — garbled-only capture loses Weekly):** good catch. The tolerant all-models match is now applied in the Weekly percent/reset extraction path too (`matchesLabel` / `crossesLabelBoundary`), so a capture where only the corrupted `all modls` label survived still populates `weeklyPercentLeft` / `secondaryResetDescription` instead of dropping the quota — and it's still kept out of the scoped rows. Added a dedicated test for that shape.
- **`lint` (`type_body_length`):** moved the new cases out of `StatusProbeTests` into their own `ClaudeAllModelsWeeklyDuplicateTests` type so the struct stays under the 800-line limit.
- **`swift-test-macos (0, 2)`:** the failure there is `SpendDashboardControllerTests` (`totalCost → nil`), unrelated to this change — the other shard `(1, 2)` and this parser's tests pass. Looks like pre-existing flakiness in that suite.

Verified by linking the built `CodexBarCore` and calling `ClaudeStatusProbe.parse` directly (full `swift test` needs Xcode; this box only has Command Line Tools):

```
Case 1: garbled all-models duplicate   -> weekly=34, extra=["Fable only"]        PASS
Case 2: clean panel (regression guard) -> weekly=34, extra=["Fable only"]        PASS
Case 3: real scoped model preserved    -> opusPercentLeft=90                     PASS
Case 4: garbled-only all-models        -> weekly=34, reset recovered, no bogus   PASS
```

guhyun9454:
Pushed `2f2b8d7` after an additional deep review of the tolerant-matching logic:

- **(High) Ordering regression:** with fuzzy matching enabled, an earlier *garbled* `all modls` row could have shadowed a later exact `all models` row when selecting the Weekly percent/reset. Fixed by making `extractPercent`/`extractReset` try an **exact match first and only fall back to the fuzzy match** — a clean row now always wins regardless of order. Fuzzy matching is opt-in (`allowFuzzy`) on `matchesLabel`/`crossesLabelBoundary`.
- **Test coverage:** expanded to pin the behavior — exact-wins-over-earlier-garbled (distinct 66% vs 67% / different resets), scoped-only input leaves `weeklyPercentLeft == nil`, single-character insertion (`all modelss`) is treated as all-models, and a distance-3 corruption (`almdls`) intentionally stays a **visible** scoped row rather than being silently absorbed (documents the edit-distance boundary).

Re-verified all six shapes against the built `CodexBarCore`:

```
1 duplicate            weekly=34  extra=[Fable only]          PASS
2 garbled-before-exact weekly=34  reset=2pm  extra=[]         PASS  (exact wins)
3 garbled-only         weekly=34  reset=2pm  extra=[Fable]    PASS  (recovered)
4 insertion            weekly=34  extra=[Fable only]          PASS
5 scoped-only          weekly=nil extra=[Haiku only]          PASS
6 dist-3 corruption    weekly=nil extra=[almdls only]         PASS  (stays visible)
```

Note on CI: the red `swift-test-macos (0, 2)` is an unrelated pre-existing flake in the SpendDashboard suites (`totalCost → nil`) — it fails a *different* suite each run and the sibling shard `(1, 2)`, which runs this change's tests, passes.

steipete:
Merged after full verification. The fix restricts fuzzy label matching to the aggregate "all models" capture with bounded edit distance, prefers exact rows before fuzzy fallback, recovers weekly/reset values from garbled-only captures, and preserves genuine model-scoped rows — six fixtures pin the behavior. Exact-head CI: all nine checks green (https://github.com/steipete/CodexBar/actions/runs/30329306328, macOS shard 0 completed green after the earlier report). Review + autoreview clean. Thanks @guhyun9454!

PR #2402, opened 2026-07-23
Fix Chutes quota detail rendering

kiranmagic7:
Fixes #2399.

## Summary

- treat Chutes quota counts as detail instead of reset descriptions across menu cards, native menus, and CLI output
- preserve real reset timestamps when present
- add regression coverage for detail-with-reset and detail-without-reset paths

## Root cause

Chutes stores quota counts such as `40/100 requests` in `RateWindow.resetDescription`. The shared renderers either treated that field as reset text or dropped it when `resetsAt` existed.

## Verification

- `swift test --skip-build --no-parallel --filter 'Chutes(Provider|Presentation)Tests'` — 14 tests passed
- `make check` — passed (SwiftFormat: 0 files; SwiftLint: 0 violations)
- `git diff --check` — passed
- `make test` — stopped on two unrelated `AdaptiveRefreshTimerTests` timeouts; both failures reproduce unchanged on base commit `cc8da27c`

Compatibility risk: low. This only changes presentation for Chutes quota windows; real reset timestamps remain rendered separately.

Review focus: please sanity-check the detail-backed provider handling across the three renderers.

PR #2436, opened 2026-07-24
Preserve hooks in config persistence

### LINKED ISSUE #2432: Hooks silently disabled since 54cfd1a3: custom CodexBarConfig CodingKeys drops `hooks`
Hooks configured in Settings > Hooks or in the config file are silently dropped, so no hook can ever fire.

**Corrected root cause: this is a semantic merge conflict between two PRs merged three minutes apart.** The original hooks PR was fine.

- `2026-07-17T06:45:57Z` #2001 merged. `CodexBarConfig` still used synthesized `Codable`, so the new `hooks` field round-tripped correctly. The manual verification in that PR did work.
- `2026-07-17T06:48:58Z` #2254 merged (commit `54cfd1a3`, "chore: retire unverified provider relays"). It added a custom `private enum CodingKeys { case version, case providers }` plus a custom `init(from:)` to skip unknown provider ids. Its patch context shows no `hooks` field, so it branched from a pre-hooks `main`. Git merged both cleanly because the hunks do not overlap.

A custom `CodingKeys` also drives the synthesized encoder, so from that commit on `hooks` is neither decoded nor encoded. Every config read drops the block and every write erases it. Hooks were announced in 0.44.0, which shipped at `16:57Z` the same day, so no released build has ever run a hook.

Repro on 0.45.2, same code on `main`:

```
$ codexbar hooks enable
Hooks: enabled
$ codexbar hooks list
Hooks: disabled
No rules configured.
```

Not the running app racing the file. Same result against an isolated config:

```
$ python3 -c "import json; p='/tmp/cfg.json'; d=json.load(open(p)); d['hooks']={'enabled':True,'events':[{'id':'r','enabled':True,'event':'quota_low','provider':'codex','threshold':0.9,'executable':'/usr/local/bin/x','arguments':[],'timeoutSeconds':60}]}; json.dump(d,open(p,'w'))"
$ CODEXBAR_CONFIG=/tmp/cfg.json codexbar hooks list
Hooks: disabled
No rules configured.
$ CODEXBAR_CONFIG=/tmp/cfg.json codexbar hooks enable
Hooks: enabled
$ python3 -c "import json; print(json.load(open('/tmp/cfg.json')).get('hooks'))"
None
```

So the documented `hooks` example in `docs/configuration.md` cannot take effect, and `hooks enable` erases a hand-written block instead of reading it. Via the app the symptom is the same: rules added in Settings > Hooks never reach `~/.config/codexbar/config.json`, which keeps `"hooks": null`.

Two-line fix in `Sources/CodexBarCore/Config/CodexBarConfig.swift`:

```diff
     private enum CodingKeys: String, CodingKey {
         case version
         case providers
+        case hooks
     }
```

```diff
         self.providers = providers
+        self.hooks = try container.decodeIfPresent(HooksConfig.self, forKey: .hooks)
     }
```

Why CI missed it: both PRs were green on their own heads, and neither ran against the other's change. `HookRule` and `HooksConfig` decode correctly in isolation, and the hook suites build configs in memory, so nothing round-trips a config file with hooks through `CodexBarConfig`. A regression test that encodes a config containing hooks, decodes it, and asserts the rules survive would cover this, and would also catch the next field added to `CodexBarConfig` without a matching `CodingKeys` case.

kiranmagic7:
Fixes #2432.

## Summary

- include top-level `hooks` in the custom `CodexBarConfig` coding keys
- restore persisted hook settings during config decoding
- add regression coverage for a full config round trip

## Root cause

`CodexBarConfig`'s custom `CodingKeys` listed only `version` and `providers`. That excluded `hooks` from both synthesized encoding and the custom decoder, so config load/save cycles silently discarded hook settings.

## Verification

- `swift test --filter CodexBarConfigHooksTests` — the regression failed before the fix with `decoded.hooks == nil`, then passed
- `swift test --skip-build --no-parallel --filter '(CodexBarConfigHooksTests|CodexBarConfigUnknownProviderTests|CodexActiveSourceConfigTests|CLIHooksTests|ConfigValidationTests)'` — 48 tests passed
- `make check` — passed (SwiftFormat: 0/1583 files; SwiftLint: 0 violations in 1582 files)
- `git diff --check` — passed
- `make test` — stopped after its automatic retry on two unchanged `AdaptiveRefreshTimerTests` that each threw `CancellationError()` after 30 seconds:
  - `menu open advances a long idle timer during refresh without postponing an earlier tick`
  - `coding activity advances a long idle timer without postponing an earlier tick`

Both timer failures are pre-existing on base `cc8da27c` and outside this two-file config/test diff.

Compatibility risk: low. Existing configs without `hooks` continue to decode that field as `nil`; this does not change hook execution, validation, or provider behavior.

Review focus: please sanity-check the custom top-level `Codable` boundary and round-trip coverage.

### DIFF
--- a/Sources/CodexBarCore/Config/CodexBarConfig.swift
+++ b/Sources/CodexBarCore/Config/CodexBarConfig.swift
@@ -8,6 +8,7 @@ public struct CodexBarConfig: Codable, Sendable {
     private enum CodingKeys: String, CodingKey {
         case version
         case providers
+        case hooks
     }
 
     private enum ProviderCodingKeys: String, CodingKey {
@@ -46,6 +47,7 @@ public struct CodexBarConfig: Codable, Sendable {
             try providers.append(ProviderConfig(from: providerDecoder))
         }
         self.providers = providers
+        self.hooks = try container.decodeIfPresent(HooksConfig.self, forKey: .hooks)
     }
 
     public static func makeDefault(
--- a/Tests/CodexBarTests/CodexBarConfigHooksTests.swift
+++ b/Tests/CodexBarTests/CodexBarConfigHooksTests.swift
@@ -1,0 +1,28 @@
+import CodexBarCore
+import Foundation
+import Testing
+
+struct CodexBarConfigHooksTests {
+    @Test
+    func `hooks survive config round trip`() throws {
+        let hooks = HooksConfig(
+            enabled: true,
+            events: [
+                HookRule(
+                    id: "quota-low",
+                    event: .quotaLow,
+                    provider: "codex",
+                    threshold: 0.9,
+                    executable: "/usr/bin/true",
+                    timeoutSeconds: 30),
+            ])
+        let config = CodexBarConfig(
+            providers: [ProviderConfig(id: .codex, enabled: true)],
+            hooks: hooks)
+
+        let data = try JSONEncoder().encode(config)
+        let decoded = try JSONDecoder().decode(CodexBarConfig.self, from: data)
+
+        #expect(decoded.hooks == hooks)
+    }
+}


kiranmagic7:
CI failure attribution:

- The failed macOS shard is isolated to untouched `SpendDashboardTokenProvenanceTests`: `cached token account activation does not prove a forced refresh` and `forced successful empty publication removes prior spend without warning` produced three missing-cost assertions. Failed job: https://github.com/steipete/CodexBar/actions/runs/30070614342/job/89410541690
- The exact four-suite CI selection passes 25/25 on head `856bbd268baac903b3171c338bc4bdf7a7d6f65e`:
  `swift test --skip-build --no-parallel --filter '(^CodexBarTests\.SpendDashboardRequestTimeTests/|^CodexBarTests\.SpendDashboardSourceConcurrencyTests/|^CodexBarTests\.SpendDashboardTokenProvenanceTests/|^CodexBarTests\.StatusItemAnimationCodexCreditsTests/)'`
- The same hosted-only failure was previously isolated against untouched base `cc8da27cec92029a6435bfee4a703a719290234e`: https://github.com/steipete/CodexBar/pull/2402#issuecomment-5054911721
- This PR only changes `CodexBarConfig` hooks persistence and adds its round-trip regression test; the other macOS shard and all Linux/lint/security jobs passed.

I attempted the failed-job rerun, but GitHub requires repository admin rights. Could a maintainer rerun the failed shard?

geofffranks:
@steipete Can we get some help on this? It fixes some fundamental issues that currently render hooks completely non-functional, and results in some commands completely overwriting/deleting all hook configuration from the config file, even when they are valid.

fcakyon:
hooks don't work in codexbar, can you please take a look @steipete ?

kiranmagic7:
Blocker update: I reproduced the shared macOS dashboard failure again today. Applying only the fixture `UserDefaults` isolation already proposed in #2418 made the previously failing four-suite selection pass 25/25; the separate fixture-clock patch was unnecessary and has been closed.

#2418 is currently mergeable with all nine visible checks green. Once it lands, I will rebase this hooks-only branch and rerun its targeted checks. No hooks-path change is warranted for the unrelated dashboard failure.

steipete:
Reviewed the full exact-head diff at `6199a8ec986af0aec417d8268ee8fabb645cefb3` and the adjacent config encoding, normalization, and persistence paths. The change is narrowly correct: `hooks` now participates in `CodingKeys` and custom decoding, and the new round-trip regression test covers the failure mode.

Hosted CI passed on the exact head: [run 30329308078](https://github.com/steipete/CodexBar/actions/runs/30329308078). Both macOS test shards, Linux builds, lint, the aggregate test gate, and GitGuardian completed successfully. Independent structured review reported no actionable findings.

Thanks @kiranmagic7 for the focused fix and regression coverage.

PR #2442, opened 2026-07-24
Price proxy models in Claude usage logs

EClinick:
## Summary
- price OpenAI models recorded in Claude Code logs through CodexBar’s existing bundled and models.dev pricing paths
- fall back to unambiguous models.dev pricing for other proxy backends while leaving ambiguous model IDs unpriced
- retry Claude cost scans after a cross-provider catalog refresh

## Why
Anthropic-compatible proxies such as CLIProxyAPI can route Claude Code to non-Anthropic models. Claude Code records those model IDs in its JSONL history, but CodexBar previously searched only the Anthropic pricing slice. Models such as `gpt-5.6-sol` therefore showed token counts without a dollar cost and were omitted from totals.

## Tests
- `swift test --filter CostUsagePricingTests` (48 passed)
- `swift test --filter CostUsageFetcherUnknownModelPricingTests` (6 passed)
- `make check` (passed)
- `make test` reached 48 successful groups, then failed three unrelated assertions in `SpendDashboardTokenProvenanceTests`; the same three failures reproduce on untouched `origin/main` at `cc8da27c`


EClinick:
## Live verification

Built and launched the local PR bundle (`0.45.2 (109)`) from `task/ec/claude-cross-provider-pricing`. Claude’s 30-day Usage & Spend view now prices `gpt-5.6-sol` at **$255.44** and includes it in the Claude total.

### 30-day overview

codexbar-pr2442-overview

### Model breakdown

codexbar-pr2442-models

EClinick:
I have a better solution and will create a new PR soon.

PR #2446, opened 2026-07-25
Handle claude-swap relogin status

### LINKED ISSUE #2444: Claude-swap: unrecognized usageStatus "relogin_required" from cswap 0.23.0
## Summary

`ClaudeSwapUsageStatus` does not recognize the `relogin_required` status emitted by claude-swap (cswap) 0.23.0. Accounts with a dead refresh token report `"usageStatus": "relogin_required"` in `cswap list --json`, which falls into the `.unknown` case and displays:

> Unrecognized claude-swap status: relogin_required

## Expected behavior

`relogin_required` should be mapped to a known case (similar to `token_expired`) with a user-friendly message like "Re-login required. Switch to this account in Claude Code and run `cswap add` to refresh credentials."

## Actual behavior

The account row shows the raw "Unrecognized claude-swap status: relogin_required" error string.

## Reproduction

1. Install cswap 0.23.0 and register two Claude accounts
2. Let one account's refresh token expire (or log out from Claude Code for that account)
3. Run `cswap list --json` — the expired account shows `"usageStatus": "relogin_required"`
4. CodexBar 0.45.2 displays the unrecognized status error on that account row

## Environment

- CodexBar 0.45.2 (latest release, 2026-07-19)
- claude-swap (cswap) 0.23.0 (latest)
- macOS 15.x

## Relevant code

- `Sources/CodexBarCore/Providers/Claude/ClaudeSwap/ClaudeSwapAccountList.swift` — `ClaudeSwapUsageStatus.init(rawValue:)` only handles: `ok`, `token_expired`, `api_key`, `keychain_unavailable`, `no_credentials`, `unavailable`
- `Sources/CodexBarCore/Providers/Claude/ClaudeSwap/ClaudeSwapAccountProjection.swift` — `errorText(for:)` produces the "Unrecognized" string for `.unknown` cases

## Suggested fix

Add `case "relogin_required": self = .reloginRequired` to `ClaudeSwapUsageStatus` and a corresponding user-friendly message in `errorText(for:)`.

kiranmagic7:
Fixes #2444.

## Problem

cswap 0.23.0 can report `"usageStatus": "relogin_required"` for an inactive account. CodexBar currently treats that value as unknown, so the account card shows the raw status instead of recovery guidance.

## Change

- Map `relogin_required` to a dedicated Claude swap usage status.
- Show re-login guidance and keep the affected account non-activatable until its credentials are refreshed in claude-swap.
- Cover the reported schema-v1 payload in parser, projection, and CLI card tests.

This does not change credential reading or account switching behavior.

## Verification

- `swift test --filter ClaudeSwapListParserTests`: the new regression failed before the implementation, then all 13 tests passed.
- `swift test --no-parallel --filter ClaudeSwap`: 69 tests passed.
- `make check`: passed (SwiftFormat: 0/1582 files; SwiftLint: 0 violations in 1581 files).
- `git diff --check`: passed.
- `make test`: stopped after its automatic retry because two unchanged `AdaptiveRefreshTimerTests` timed out with `CancellationError()` at 30 seconds:
  - `menu open advances a long idle timer during refresh without postponing an earlier tick`
  - `coding activity advances a long idle timer without postponing an earlier tick`

Both timer failures reproduce when that untouched suite runs alone.

Compatibility risk is low: this only changes handling for the newly recognized status string. Unknown future values still use the existing fallback.

Review focus: whether the recovery text and non-activatable state match the expected cswap behavior.

### DIFF
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeSwap/ClaudeSwapAccountList.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeSwap/ClaudeSwapAccountList.swift
@@ -75,6 +75,7 @@ public struct ClaudeSwapUsageWindow: Equatable, Sendable {
 public enum ClaudeSwapUsageStatus: Equatable, Sendable {
     case ok
     case tokenExpired
+    case reloginRequired
     case apiKey
     case keychainUnavailable
     case noCredentials
@@ -85,6 +86,7 @@ public enum ClaudeSwapUsageStatus: Equatable, Sendable {
         switch rawValue {
         case "ok": self = .ok
         case "token_expired": self = .tokenExpired
+        case "relogin_required": self = .reloginRequired
         case "api_key": self = .apiKey
         case "keychain_unavailable": self = .keychainUnavailable
         case "no_credentials": self = .noCredentials
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeSwap/ClaudeSwapAccountProjection.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeSwap/ClaudeSwapAccountProjection.swift
@@ -101,7 +101,7 @@ public enum ClaudeSwapAccountProjection {
         case .tokenExpired:
             "Token expired. Switch to this account in claude-swap to refresh it."
         case .reloginRequired:
-            "Re-login required. Switch to this account in claude-swap to refresh it."
+            "Re-login required. Re-authenticate this account in claude-swap."
         case .apiKey:
             "API-key account; subscription usage is unavailable."
         case .keychainUnavailable:
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeSwap/ClaudeSwapAccountProjection.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeSwap/ClaudeSwapAccountProjection.swift
@@ -100,6 +100,8 @@ public enum ClaudeSwapAccountProjection {
                 : nil
         case .tokenExpired:
             "Token expired. Switch to this account in claude-swap to refresh it."
+        case .reloginRequired:
+            "Re-login required. Switch to this account in claude-swap to refresh it."
         case .apiKey:
             "API-key account; subscription usage is unavailable."
         case .keychainUnavailable:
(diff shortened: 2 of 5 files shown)


kiranmagic7:
The failed macOS shard is outside this patch.

It reports 10 missing-`totalCost` assertions in the untouched `SpendDashboardControllerTests`; the aggregate gate is red only because that shard failed. I ran the workflow's exact four-suite selection at both revisions:

- PR head `7b4932e2bb2ae6d4d3458ba149f04700d56377a1`: reproduced the same 10 issues.
- Untouched base `cc8da27cec92029a6435bfee4a703a719290234e`: reproduced the same 10 issues.

The branch diff remains limited to the five Claude Swap parser/projection/test files. The focused Claude Swap tests, `make check`, and `git diff --check` remain green.

GitHub rejected a failed-job rerun because the fork account does not have repository admin rights. Could a maintainer rerun the failed jobs?

The separate ClawSweeper request for redacted proof from a real `relogin_required` claude-swap setup remains open; fixture-backed tests are not being presented as live-account proof.

kiranmagic7:
Refreshed this branch onto current `main` now that #2418 landed. New head: `0813d366ef4673f8590ff5de8fc19ab55dbbf9d5`. The effective PR diff remains the same five Claude Swap files (32 additions, 1 deletion).

Verification on the refreshed tree:

- `swift test --no-parallel --filter ClaudeSwap` — 69 tests in 8 suites passed
- shared spend-dashboard/status regression selection — 25 tests in 4 suites passed
- `make check`
- `git diff --check origin/main...HEAD`

Hosted checks are now running on the new head.

The separate request for redacted real `relogin_required` behavior proof remains open. I did not run a live claude-swap account, Keychain, or provider probe; repository guidance calls for fixture/stub testing unless that probe is explicitly requested.

steipete:
Merged after full verification. claude-swap's `relogin_required` status is now recognized instead of surfacing as an unknown-status error, the account stays non-activatable while it needs re-auth, and recovery guidance is shown.

Adjusted during review: the original guidance named a specific command ("Switch to this account in claude-swap") that the issue's evidence didn't actually establish — the report pointed at a `cswap add` / Claude Code flow instead. Rather than risk sending users down a wrong path, the text is now generic ("re-authenticate this account in claude-swap"). Exact-head CI green; 69 focused tests. Thanks @kiranmagic7!

Issue #2353, opened 2026-07-20
Feature Request - Hide Daily Routines bar in Claude


tavlean:
Pease add an option to hide "Daily Routines" bar from Claude

steipete:
[#2450](https://github.com/steipete/CodexBar/pull/2450) has landed. It removes the fabricated “Daily Routines 0%” row when Anthropic returns a null routines payload, which may reduce the need for a toggle for affected accounts.

It does not add a user-controlled hide setting, so this toggle request remains open for the owner’s product decision.

steipete:
Done via two landed pieces: #2450 removed the fabricated "Daily Routines 0%" row when Anthropic returns a null payload, and #2358 (just merged) adds the setting to hide the row entirely even when real routines data exists — default-on, so nothing changes unless you flip it. Thanks @tavlean for the request and @Zihao-Qi for the implementation. Ships in the next release.

PR #2450, opened 2026-07-25
Hide Daily Routines bar when the routines payload is null

urda:
## Summary

- treat a routines key that is present but `null` as absent instead of synthesizing a 0% placeholder window
- update the two regression tests to assert the window is omitted

## Why

Accounts without routines data receive `null` for these aliases, so the placeholder rendered a permanent phantom "Daily Routines 0% used" row matching no quota. claude.ai renders from `limits`, which carries no routines lane, and shows no bar. Nothing is lost: the removed row was hardcoded to 0%. Accounts with a real routines window are unchanged.

Refs #2353, though this adds no toggle.

## Validation

- `make check` — 0 violations across 1,581 files
- `swift test --filter 'ClaudeOAuthTests|ClaudeWebUsageExtraWindowTests|ClaudeExtraWindowQuotaWarningTests|MenuCardModelTests'` — 127 tests passed
- `make test` — 60 of 61 groups passed; `SpendDashboardTokenProvenanceTests` failures are pre-existing
- local debug build on a live account whose payload returns `seven_day_cowork: null`: the card fetches successfully and renders Session, Weekly and Fable only, with no Daily Routines row

Claude usage card with no Daily Routines row

Claude usage card with no Daily Routines row (menu)

The inverse case (a real routines payload still rendering) is covered by the regression tests rather than a capture: this account returns `null`, so there is no local way to produce one.

### DIFF
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeOAuth/ClaudeOAuthUsageFetcher.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeOAuth/ClaudeOAuthUsageFetcher.swift
@@ -208,18 +208,14 @@ struct OAuthUsageResponse: Decodable {
         in container: KeyedDecodingContainer<DynamicCodingKey>,
         keys: [String]) -> (window: OAuthUsageWindow?, sourceKey: String?)
     {
-        var firstNullKey: String?
         for keyName in keys {
             guard let key = DynamicCodingKey(stringValue: keyName) else { continue }
             guard container.contains(key) else { continue }
             if let value = try? container.decodeIfPresent(OAuthUsageWindow.self, forKey: key) {
                 return (value, keyName)
             }
-            if firstNullKey == nil {
-                firstNullKey = keyName
-            }
         }
-        return (nil, firstNullKey)
+        return (nil, nil)
     }
 
     private static func decodeValue<T: Decodable>(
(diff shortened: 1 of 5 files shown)


urda:
@clawsweeper re-review

Body now includes a runtime capture: the Claude card on an account whose payload returns `seven_day_cowork: null`, fetching successfully with no Daily Routines row.

urda:
@clawsweeper re-review

The two prior runs (`30176252501`, `30177607322`) both completed without publishing a verdict, so the durable comment still shows the 09:48 UTC review. The PR body has carried the requested runtime capture since 22:41 UTC.

urda:
@clawsweeper re-review + status?

steipete:
Maintainer verification complete on exact head `c8958358cabd024beefa1df46efa2061f3b0a887`.

I reread the full diff and the adjacent OAuth/web parser paths, including the null-payload regressions and the existing populated-alias coverage. AutoReview found no blocking defect. Full hosted CI passed across lint, Linux x64/arm64, Linux musl, both macOS test shards, and the aggregate gate: [CI run 30327078930](https://github.com/steipete/CodexBar/actions/runs/30327078930).

Merged as [`cae98731410431479f093a9570ab0aeb155e3162`](https://github.com/steipete/CodexBar/commit/cae98731410431479f093a9570ab0aeb155e3162). Thanks for the focused fix, @urda!

urda:
Let's go 🥳

PR #2454, opened 2026-07-25
Fix CLI version path resolution

geofffranks:
## Problem ##

When the CLI is invoked through a relative path or symlink, version lookup used `argv[0]` relative to the current working directory. This could prevent CodexBar from finding the adjacent `VERSION` file or app-bundle metadata and result in a missing or incorrect displayed version.

## Summary ##

• Resolve CLI version metadata from the actual executable path instead of relative `argv[0]`.
• Preserve symlink-aware `VERSION` and app-bundle `Info.plist` lookup.
• Add regression coverage for bundle executable resolution.

### DIFF
--- a/Sources/CodexBarCLI/CLIIO.swift
+++ b/Sources/CodexBarCLI/CLIIO.swift
@@ -55,14 +55,37 @@ extension CodexBarCLI {
 
     static func currentVersion(
         bundle: Bundle = .main,
-        executablePath: String? = CommandLine.arguments.first) -> String?
+        executablePath: String? = nil) -> String?
     {
+        let executablePath = executablePath ?? Self.runningExecutablePath(bundle: bundle)
         if let version = self.currentVersion(bundleVersion: nil, executablePath: executablePath) {
             return version
         }
         return self.currentVersion(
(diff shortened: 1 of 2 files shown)


geofffranks:
Proof:
```
$ alias | grep codexbar
$ codexbar -V
CodexBar
$ codexbar --version
CodexBar
$ which codexbar
/opt/homebrew/bin/codexbar
$ /opt/homebrew/bin/codexbar -V
CodexBar 0.45.2
$ cd /opt/homebrew/bin/
$ ./codexbar -V
CodexBar 0.45.2
$ codexbar -V
CodexBar 0.45.2
$ cd ~
$ codexbar -V
CodexBar
```

steipete:
Merged after full verification. CLI version metadata now resolves from the running executable (`_NSGetExecutablePath` on Darwin, `/proc/self/exe` on Linux) with symlink resolution and normalized bundle fallback, so `codexbar --version` reports correctly regardless of how it was invoked.

Added during review: a real regression exercising the reported failure — invocation via a stale relative `argv[0]` from a different working directory, and through a symlink — since the original test only covered `Bundle.executableURL` and would have passed even with the bug present. Also switched a test helper to the failable UTF-8 conversion the repo lints for. Thanks @geofffranks — nice first contribution!

PR #2462, opened 2026-07-26
Support LongCat Firefox cookie imports

### LINKED ISSUE #2463: LongCat automatic cookie import ignores Firefox sessions
## Summary

LongCat usage is not detected when the user is signed in to `https://longcat.chat/platform/product` in Firefox and the provider's cookie source is set to Automatic.

## Steps to reproduce

1. Sign in to LongCat at `https://longcat.chat/platform/product` using Firefox.
2. Enable the LongCat provider in CodexBar.
3. Leave **Cookie source** set to **Automatic**.
4. Trigger a user-initiated refresh.

## Actual result

CodexBar does not discover the valid Firefox session, so LongCat usage cannot be fetched automatically.

## Expected result

LongCat Automatic cookie import should check Firefox after Chrome and use the valid `longcat.chat` session without requiring a manually pasted Cookie header.

## Root cause

LongCat's provider metadata hard-codes its automatic browser import order to `[.chrome]`. The shared browser-cookie importer already supports Firefox, and Firefox does not add a Chromium Safe Storage Keychain prompt, but LongCat never includes Firefox in its candidate list.

## Proposed fix

Keep Chrome first for existing users and add Firefox as the fallback, with a regression test locking the provider import order.

Draft fix: #2462

No cookies, tokens, or other account credentials are included in this report.

akshayprabhu200:
## Summary

- add Firefox as a Chrome-first fallback for LongCat automatic cookie imports
- preserve the existing Chrome precedence without adding Safari or unrelated Chromium browser probes
- update the dedicated browser-order regression test to lock the provider metadata and Firefox fallback

## Root cause

LongCat's automatic cookie import order was hard-coded to `[.chrome]`. Although the shared cookie importer already supports Firefox and other providers use it successfully, Firefox was never included in LongCat's candidate list, so a valid `longcat.chat` Firefox session could not be discovered.

## Real behavior proof

Ran a credential-safe local proof on a machine signed in to LongCat only in Firefox. The proof used the exact SweetCookieKit `0.4.1` revision pinned by this checkout and the same Firefox cookie query and LongCat endpoints used by the production importer.

```text
[proof] configured LongCat order: Chrome -> Firefox
[proof] Chrome LongCat cookie rows: 0
[proof] Firefox candidate profiles with LongCat cookies: 1
[proof] Firefox profile #1: authenticated LongCat session discovered
[proof] user-current: HTTP 200, envelope 0
[proof] tokenUsage: HTTP 200, envelope 0, totalToken present
[proof] cookie values and account fields: redacted by harness
```

The Chrome check was count-only and did not decrypt or print cookie data. The Firefox proof imported the live cookie store, sent authenticated GET requests only to `longcat.chat`, and printed only status/schema evidence. No cookie names, values, account identity, quota values, or response bodies were retained.

Firefox uses the shared Gecko SQLite/WAL reader and does not use Chromium Safe Storage, so this fallback does not introduce a browser Keychain prompt. It does expand LongCat's file-based browser probing from Chrome to Firefox when Automatic import runs; Chrome remains first.

## Verification

- real Firefox-only LongCat discovery proof above — authenticated `user-current` and `tokenUsage` requests succeeded
- `make check` — all portable repository checks and SwiftFormat passed; the initial SwiftLint invocation could not load SourceKit from the active Command Line Tools path
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer .build/lint-tools/bin/swiftlint --strict` — 0 violations across 1,581 files
- `swift test --filter BrowserCookieOrderStatusStringTests` — could not start because this checkout requires Swift tools 6.2 while the installed Xcode/Swift toolchain is 5.10
- `make test` — same local Swift 6.2 toolchain prerequisite prevented test discovery; upstream CI remains the compilation and test gate

## Changelog

LongCat: detect Firefox web sessions during Automatic cookie import.

No screenshots needed; this changes provider cookie-source selection only.

Fixes #2463

### DIFF
--- a/Sources/CodexBarCore/Providers/Providers.swift
+++ b/Sources/CodexBarCore/Providers/Providers.swift
@@ -291,10 +291,11 @@ public enum ProviderBrowserCookieDefaults {
         #endif
     }
 
-    /// LongCat Auto imports only from Chrome by default to avoid prompting unrelated browser keychains.
+    /// LongCat Auto keeps Chrome first for existing users, then checks Firefox without adding
+    /// an unrelated browser Keychain prompt.
     public static var longcatCookieImportOrder: BrowserCookieImportOrder? {
         #if os(macOS)
-        [.chrome]
+        [.chrome, .firefox]
         #else
         nil
         #endif
--- a/Tests/CodexBarTests/BrowserCookieOrderLabelTests.swift
+++ b/Tests/CodexBarTests/BrowserCookieOrderLabelTests.swift
@@ -102,9 +102,13 @@ struct BrowserCookieOrderStatusStringTests {
     }
 
     @Test
-    func `longcat cookie imports default to chrome only`() {
-        #expect(ProviderDefaults.metadata[.longcat]?.browserCookieOrder == [.chrome])
-        #expect(ProviderBrowserCookieDefaults.longcatCookieImportOrder == [.chrome])
+    func `longcat cookie import order supports chrome and firefox`() {
+        let order = ProviderDefaults.metadata[.longcat]?.browserCookieOrder ?? Browser.defaultImportOrder
+        #expect(order == ProviderBrowserCookieDefaults.longcatCookieImportOrder)
+        #expect(order == [.chrome, .firefox])
+        #expect(order.first == .chrome)
+        #expect(order.contains(.firefox))
+        #expect(!order.contains(.safari))
     }
     #endif
 }
--- a/Tests/CodexBarTests/BrowserCookieOrderLabelTests.swift
+++ b/Tests/CodexBarTests/BrowserCookieOrderLabelTests.swift
@@ -103,12 +103,14 @@ struct BrowserCookieOrderStatusStringTests {
 
     @Test
     func `longcat cookie import order supports chrome and firefox`() {
-        let order = ProviderDefaults.metadata[.longcat]?.browserCookieOrder ?? Browser.defaultImportOrder
-        #expect(order == ProviderBrowserCookieDefaults.longcatCookieImportOrder)
-        #expect(order == [.chrome, .firefox])
-        #expect(order.first == .chrome)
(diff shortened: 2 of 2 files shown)


akshayprabhu200:
@clawsweeper re-review

akshayprabhu200:
I believe this bounded LongCat override follows the `AGENTS.md` guidance (“default Chrome-only when possible … override via browser list when needed”): the redacted live proof shows Chrome-only does not work for a Firefox-only LongCat session, and Firefox adds no Chromium Keychain prompt.

The closest merged precedent is #1985, which intentionally expanded Mistral Automatic from Chrome-only to Chrome → Firefox → Safari so Firefox users are detected without Manual mode. LongCat currently exposes only Automatic / Manual cookie header / Off; I could not find an existing per-provider browser picker, so making Firefox explicitly selectable would require a broader settings/config feature rather than a focused compatibility fix.

Would you accept the LongCat-specific Chrome → Firefox exception, or would you prefer that broader opt-in browser-selection UX first?

steipete:
Merged after final review of the exact head and the surrounding LongCat cookie-import path.

Verification:

- Confirmed Automatic import remains Chrome first, falls back to Firefox, and does not add Safari.
- Confirmed Firefox uses the existing non-Keychain browser-cookie path.
- Structured autoreview reported no accepted/actionable findings.
- Exact-head CI passed: https://github.com/steipete/CodexBar/actions/runs/30326268798
  - lint, GitGuardian, Linux x64/arm64, Linux musl, both macOS test shards, and the aggregate gate all passed.
  - One macOS shard was rerun after SwiftPM's shared cache lost its `swift-log` repository during dependency resolution; the rerun completed successfully.
- The PR documents successful live Firefox-only LongCat discovery and authenticated endpoint proof with credentials and account data redacted.

Thank you for the focused fix and the live Firefox validation, @akshayprabhu200!

PR #2390, opened 2026-07-22
Stabilize dashboard fixture clock

ProspectOre:
## Summary

Stabilize Spend Dashboard controller tests by injecting the same fixed clock used by their fixture snapshots.

## Root cause

Several tests used a July 15 fixture while `SpendDashboardSource.makeRequest` used the real current date. Once the calendar moved beyond the fixture window, valid fixture spend was filtered out and the assertions saw `$0`.

## Change

- add one fixture-clock controller helper
- route the eight affected controller constructions through it
- preserve the production request clock and all dashboard behavior

## Verification

- `xcodebuildmcp swift-package test --package-path /Users/alec/Dev/codexbar-dashboard-ci --filter SpendDashboardControllerTests --parallel false --output text` — 24 tests across 3 suites passed
- `swiftformat --lint Tests/CodexBarTests/SpendDashboardControllerTests.swift` — clean
- `swiftlint --strict` — 0 violations
- `make check` completes relevant validation and stops only at the repository's pre-existing Node 25 CommonJS/ESM mismatch in `docs/site.js`

PR #2464, opened 2026-07-27
Stabilize token provenance fixture clock

kiranmagic7:
## Summary

Stabilize the spend dashboard token-provenance tests by injecting the same fixed clock used by their July 16 fixture snapshots.

## Root cause

Three controller constructions called `SpendDashboardSource.makeRequest` with the real current date. Once the calendar moved beyond the fixtures' selected dashboard window, valid fixture spend was filtered out and the cost assertions received `nil`.

The same three failures appear in recent macOS test-shard runs for #2402 and #2436.

## Change

- add one fixture-clock controller helper
- route the three affected controller constructions through it
- preserve the production request clock and all dashboard behavior

This complements #2390, which fixes the same failure mode in `SpendDashboardControllerTests`.

## Verification

- before the change, the affected 25-test selection failed with three issues
- after the change, the same selection passed: 25 tests across 4 suites
- `make check` — clean, including SwiftFormat and SwiftLint (0 violations in 1581 files)
- `make test` — stopped in unrelated `AdaptiveRefreshTimerTests`: two 30-second `CancellationError()` failures reproduced on the automatic retry and in an isolated run

## Risk and review focus

Test-only change; no production or public API impact. Please focus on whether the fixed fixture clock should remain scoped to these three dashboard controller constructions.

### DIFF
--- a/Tests/CodexBarTests/SpendDashboardTokenProvenanceTests.swift
+++ b/Tests/CodexBarTests/SpendDashboardTokenProvenanceTests.swift
@@ -124,9 +124,7 @@ struct SpendDashboardTokenProvenanceTests {
         store.activateCachedTokenAccountSnapshot(provider: .mistral, accountID: account.id)
         #expect(store.tokenSnapshotPublicationRevision(for: .mistral) == baselineRevision)
         store._test_providerRefreshOverride = { _ in }
-        let controller = SpendDashboardController(requestBuilder: { mode in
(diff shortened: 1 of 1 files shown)


kiranmagic7:
Closing as superseded by #2418. I retested today’s failing 25-test selection with only #2418’s isolated `UserDefaults` change, and it passes 25/25. The fixed-clock addition here is therefore unnecessary duplicate coverage while that root repair is active.

PR #2466, opened 2026-07-27
Fix Command Code rolling and monthly usage windows

derekszen:
## Summary
- read Command Code API-key credentials from the CLI auth file or environment
- fetch live 5-hour, weekly, and monthly billing windows through the alpha billing API
- preserve browser-cookie fallback while allowing Linux auto mode to use API-key auth
- expose reset timestamps and correct quota labels in provider output
- add cross-platform and Linux coverage for auth, headers, percentages, and reset dates

## Verification
- `make test`
- `./Scripts/lint.sh lint-linux`
- SwiftFormat lint on changed files
- live `codexbar usage --provider commandcode --format json --pretty` against a Go account

### DIFF
--- a/Sources/CodexBarCLI/CLIUsageCommand.swift
+++ b/Sources/CodexBarCLI/CLIUsageCommand.swift
@@ -689,10 +689,15 @@ extension CodexBarCLI {
                 return false
             }
         }
-        if provider == .commandcode,
-           settings?.commandcode?.cookieSource == .manual
-        {
-            return false
+        if provider == .commandcode {
+            if sourceMode == .auto,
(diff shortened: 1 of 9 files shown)


steipete:
Landed in two parts: the rolling/monthly window parsing from this PR is on main via #2630 (your commit carried as co-author; changelog credits you), and the direct-API source idea now has a proper design home in #2629 — the research note there documents what your key-discovery approach does, where it stands against the owner-mediated credential policy (the same boundary that shaped the Claude and Kiro decisions), and what an acceptable shape would look like. If you want to pursue the API source, #2629 is the place. Closing this PR in favor of the split. Thanks @derekszen!

PR #2475, opened 2026-07-28
fix: allow MiniMax usage on Linux with a configured API key

### LINKED ISSUE #2474: minimax provider fails on Linux — API-key path doesn't need WebKit but platform gate blocks it before pipeline (same root cause as #1596)
CodexBar version: 0.45.2 (musl Linux x86_64, tarball from v0.45.2 release) OS: Debian 12
Reproduction
bash

$ codexbar config enable --provider minimax
$ printf '%s' "$MINIMAX_API_KEY" | codexbar config set-api-key --provider minimax --stdin
$ codexbar --provider minimax --format json --pretty
[
  {
    "error" : {
      "code" : 1,
      "kind" : "runtime",
      "message" : "Error: selected source requires web support and is only supported on macOS."
    },
    "provider" : "minimax",
    "source" : "auto"
  }
]

Same result with explicit --source api. The error fires before the API-key fetch path is even reached.
Root cause

Identical to #1596 (closed): a platform-level gate in CLIUsageCommand.swift short-circuits the provider when the source mode requires "web support":
swift

#if !os(macOS)
if Self.sourceModeRequiresWebSupport(effectiveSourceMode, provider: provider, ...) {
    return Self.webSourceUnsupportedOutput(...)
}
#endif

For minimax the descriptor advertises [.auto, .api, .web] (and .web always triggers the gate under auto). Even when the user has supplied an explicit API key, the gate fires first.
Why this is fixable

The minimax API endpoint is a plain HTTPS GET with Bearer auth — no WebKit, no browser cookie import, no JS evaluation. Verified working on the same Linux host with curl:
bash

$ curl -sS -X GET "https://api.minimax.io/v1/token_plan/remains" \
    -H "Authorization: Bearer $MINIMAX_API_KEY" \
    -H "Content-Type: application/json"
# (response omitted — contains my real quota)

With a bogus key to show the auth handshake works end-to-end on Linux:
json

{"base_resp":{"status_code":1004,"status_msg":"login fail: Please carry the API secret key in the 'Authorization' field of the request header"}}

So the network round-trip and auth parsing work fine on Linux; only the CLI gate is blocking it. The CN endpoint (https://api.minimaxi.com/v1/token_plan/remains) is the same shape.
Suggested fix

Same approach as #1596 should apply: when the user has supplied an explicit credential (API key via set-api-key, or --source api/--source oauth), skip the platform gate. The web/cookie paths can stay macOS-gated since they do need WebKit, but the .api and .oauth paths are pure HTTP.

Concretely: in sourceModeRequiresWebSupport, return false whenever the resolved source mode is .api or .oauth and the corresponding credential is present in the resolved token account / config. The macOS-only restriction should only apply to the actual code paths that touch WebKit.
Same pattern in other providers

This is now the third instance of the same gate being too aggressive. Filed for visibility:

    #1596 (closed) — opencodego with manual cookie
    #2328 (open) — qwencloud token plan via cookies-only API
    This issue — minimax via API key

A single fix to the platform-gate logic would resolve all three (and any future API-key-only providers added to the descriptor set).
Environment
text

$ codexbar --version
CodexBar 0.45.2
$ uname -a
Linux ... 5.10.134 ... x86_64 GNU/Linux
$ ./CodexBarCLI --version
CodexBar 0.45.2

OfficialAbhinavSingh:
Closes #2474.

## Problem

On Linux, `minimax` fails even with an API key configured:

```
$ codexbar config enable --provider minimax
$ printf '%s' "$MINIMAX_API_KEY" | codexbar config set-api-key --provider minimax --stdin
$ codexbar --provider minimax --format json
"message" : "Error: selected source requires web support and is only supported on macOS."
```

`MiniMaxProviderDescriptor` advertises `sourceModes: [.auto, .web, .api]`, so under Auto the CLI's
`sourceModeRequiresWebSupport` gate returns `true` purely because `.web` is in the list — the API
strategy never gets a chance to run. But `minimax.api` is `kind: .apiToken`, a plain HTTPS request
with `Authorization: Bearer …`: no WebKit, no cookie import, no JS.

## Fix

Exempt Auto from the gate when a MiniMax API key actually resolves, mirroring the existing
credential exemptions in the same function (`.factory` reads `FactorySettingsReader.apiKey`,
`.kimi` reads `ProviderTokenResolver.kimiAPIToken`):

```swift
if provider == .minimax,
   sourceMode == .auto,
   environment.map({ MiniMaxAPISettingsReader.apiToken(environment: $0) != nil }) == true
{
    // The MiniMax API fetch is plain HTTPS + Bearer auth; only its web/cookie path
    // needs macOS, so a configured API key works off macOS too.
    return false
}
```

`MiniMaxAPISettingsReader.apiToken` covers both `MINIMAX_API_KEY` and `MINIMAX_CODING_API_KEY`, and
`ProviderConfigEnvironment.directAPIKeyEnvironmentKey` maps `.minimax` to that same key — so keys
stored via `config set-api-key` are seen here too, not just environment variables.

Scope is deliberately narrow: explicit `--source web` and the no-credential case still require web
support, so nothing that genuinely needs macOS is bypassed.

## Scope note

The issue also reports `--source api` failing. I could not reproduce that from source: the gate's
final switch already returns `false` for `case .cli, .oauth, .api`, so the explicit-API path should
not hit it. This PR therefore fixes only the Auto path, which is reproducible and covered below.

## Validation

Built and tested on Linux in `swift:6.3.3`; macOS CI validates the rest.

**Green — with the fix (4/4):**

```
$ swift test --filter MiniMaxLinuxTests
◇ Suite MiniMaxLinuxTests started.
✔ Test "configured API key does not require macOS web support" passed after 0.001 seconds.
✔ Test "coding plan API key also skips the web-support gate" passed after 0.001 seconds.
✔ Test "explicit web source still requires web support even with an API key" passed after 0.001 seconds.
✔ Test "auto without an API key still requires web support off macOS" passed after 0.001 seconds.
✔ Test run with 4 tests in 1 suite passed after 0.003 seconds.
```

**Red — same tests with the fix reverted (simulating current `main`):**

```
$ swift test --filter MiniMaxLinuxTests
✘ Test "configured API key does not require macOS web support" recorded an issue at MiniMaxLinuxTests.swift:12:9:
  Expectation failed: !(CodexBarCLI.sourceModeRequiresWebSupport(.auto, provider: .minimax,
  environment: (… → ["MINIMAX_API_KEY": "sk-api-test"])) → true)
✘ Test "coding plan API key also skips the web-support gate" recorded an issue at MiniMaxLinuxTests.swift:22:9:
  Expectation failed: !(CodexBarCLI.sourceModeRequiresWebSupport(.auto, provider: .minimax,
  environment: (… → ["MINIMAX_CODING_API_KEY": "sk-cp-test"])) → true)
✔ Test "explicit web source still requires web support even with an API key" passed after 0.001 seconds.
✔ Test "auto without an API key still requires web support off macOS" passed after 0.003 seconds.
✘ Test run with 4 tests in 1 suite failed after 0.007 seconds with 2 issues.
```

Note the two negative controls pass in **both** states — they confirm the gate still fires for
explicit `.web` and for the no-credential case, so the exemption is scoped rather than a blanket
bypass.

**Lint:**

```
$ swiftlint --strict
Done linting! Found 0 violations, 0 serious in 1582 files.
```

**Not tested:** a live MiniMax account round-trip — I don't have MiniMax credentials. The issue
reporter already verified the endpoint answers on Linux via `curl`, and this change only affects
the CLI platform gate, not the request itself.

Competing open PR scan: none — no other open PR references #2474 or touches
`sourceModeRequiresWebSupport`.

### DIFF
--- a/Sources/CodexBarCLI/CLIUsageCommand.swift
+++ b/Sources/CodexBarCLI/CLIUsageCommand.swift
@@ -784,10 +784,12 @@ extension CodexBarCLI {
             guard sourceMode == .auto || sourceMode == .cli else { return false }
             return environment.map { FactorySettingsReader.apiKey(environment: $0) != nil } == true
         case .minimax:
-            // The MiniMax API fetch is plain HTTPS + Bearer auth; only its web/cookie path
-            // needs macOS, so a configured API key works off macOS too.
-            guard sourceMode == .auto else { return false }
-            return environment.map { MiniMaxAPISettingsReader.apiToken(environment: $0) != nil } == true
+            // The MiniMax API fetch is plain HTTPS + Bearer auth, so a configured key works off
+            // macOS. Standard `sk-api-` keys are the exception: Auto resolves them to the Coding
+            // Plan web strategy, which still needs the macOS-only web path.
+            guard sourceMode == .auto, let environment else { return false }
+            guard MiniMaxAPISettingsReader.apiToken(environment: environment) != nil else { return false }
+            return MiniMaxAPISettingsReader.apiKeyKind(environment: environment) != .standard
         case .mimo:
             guard sourceMode == .auto, let environment else { return false }
             return MiMoLocalUsageFallback.cacheExists(environment: environment)
--- a/Sources/CodexBarCLI/CLIUsageCommand.swift
+++ b/Sources/CodexBarCLI/CLIUsageCommand.swift
@@ -254,7 +254,9 @@ extension CodexBarCLI {
     }
 
     private static func accountSelections(from accounts: [ProviderTokenAccount]) -> [ProviderTokenAccount?] {
-        if accounts.isEmpty { return [nil] }
+        if accounts.isEmpty {
+            return [nil]
+        }
         return accounts.map { Optional($0) }
     }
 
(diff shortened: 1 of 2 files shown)


OfficialAbhinavSingh:
@clawsweeper re-review

steipete:
Maintainer review — thanks @OfficialAbhinavSingh, this needed a correction before it could land, and I've pushed it rather than bouncing it back to you.

**The defect:** the exemption was applied to *any* configured MiniMax key, but `MiniMaxAPIFetchStrategy.isAvailable` explicitly refuses standard `sk-api-` keys. So under Auto, a standard key resolves to `MiniMaxCodingPlanFetchStrategy` — the macOS-only web path. On Linux the gate would open and the fetch would then fail with `noAvailableStrategy`, i.e. the same failure in a less obvious place. The exemption only actually holds for coding-plan (`sk-cp-`) and unknown-prefix keys, which do resolve to the plain HTTPS + Bearer API strategy.

**What I changed** (`4783dafd`): the exemption now additionally requires `apiKeyKind != .standard`, with a comment explaining why. Your first Linux test asserted the old behavior, so I inverted it to assert that a standard key still requires web support, and added a pipeline-level test that an exempted `sk-cp-` key really does resolve a Linux-capable strategy (`minimax.api`) — gate agreement alone wasn't proving the path works end to end.

I also pushed `9e8a2a85` earlier: your change tipped `sourceModeRequiresWebSupport` past the SwiftLint cyclomatic-complexity limit (it had grown one hand-written branch per provider), so I grouped the exemptions into `webSupportExempt` / `cookieSourceExempt` / `credentialExempt` helpers. Behavior-preserving — I compared every provider case against the pre-refactor version, and 119 focused tests across 17 suites pass.

Driving CI now; will land once green.

OfficialAbhinavSingh:
Thanks for pushing the correction rather than bouncing it back — it's the right fix.

I did read `isAvailable` gating on `apiKeyKind`, but never traced what that condition gated, so I never verified a strategy was actually reachable for the key class I was exempting, and my test then locked in the wrong behavior. Taking forward: assert the resolved pipeline contains the expected strategy id, not just the gate's boolean.

The complexity refactor makes sense too — that function was already close to the limit before I added another branch.

steipete:
Merged after full verification. MiniMax now works on Linux when a usable API credential is configured — scoped, after review, to keys that actually resolve a Linux-capable strategy: standard `sk-api-` keys still require the macOS web path because `MiniMaxAPIFetchStrategy` refuses them and Auto falls back to the Coding Plan page. Exempting those would only have moved the failure to `noAvailableStrategy`.

Includes a maintainer refactor grouping the CLI web-support exemptions into `webSupportExempt` / `cookieSourceExempt` / `credentialExempt` (your change tipped the function past the SwiftLint complexity limit), verified behavior-identical per provider, plus a pipeline-level regression proving an exempted key resolves `minimax.api`. The earlier macOS shard red was a 50-minute runner timeout, not a test failure. Thanks @OfficialAbhinavSingh!

PR #2379, opened 2026-07-21
Add Charm Hyper provider

JavaGT:
Closes #2318

## Summary
- add Charm Hyper as an API-key provider using `GET /v1/credits`
- display the remaining Hypercredit balance without inventing a usage percentage or reset time
- add settings, token-account, CLI, diagnostics, icon, documentation, and provider-site integration
- add request/parsing/registration tests

## Verification
- `swift build --target CodexBarCore`
- `swift build --target CodexBarCLI`
- `make format`
- `make check` passes repository, documentation, locale, generated-hash, and SwiftFormat checks; SwiftLint cannot run on this host because its Command Line Tools installation cannot load `sourcekitdInProc.framework`
- focused Swift tests could not execute because this host lacks the Swift Testing module

No live credentials or dashboard cookies are stored in the change.

Issue #2318, opened 2026-07-19
Feature Request: Add Charm Hyper (hyper.charm.land) as a provider


JavaGT:
### Summary

Would love to see CodexBar support [Charm Hyper](https://hyper.charm.land) — Charm's coding-optimized inference provider (official provider for the Crush agent). It uses a Hypercredit-based billing model (1 Hypercredit = 5¢) with free monthly grants, daily-refreshing subscriptions, and prepaid bundles that never expire.

The web dashboard shows a reset panel ("Plan / Shred / Next Hypercredit refresh in 13 hours"), which lines up with CodexBar's per-provider reset-countdown feature.

### Investigation findings

Sharing what I turned up while looking into feasibility. Not prescribing any particular implementation — just recording the evidence so a maintainer can pick scope.

**Endpoints probed (authenticated with `HYPER_API_KEY`):**

| Path | Status | Notes |
| --- | --- | --- |
| `GET /v1/models` | 200 (public) | OpenAI-shaped list with per-model `pricing.input`/`pricing.output`/`cache_create`/`cache_hit` |
| `GET /v1/credits` | 200 | Returns `{"balance": <int>}` only — 21 bytes, no `reset_at`, no `plan`, no `limit` |
| `GET /v1/me`, `/v1/usage`, `/v1/account`, `/v1/subscription`, `/v1/plan`, `/v1/key` | 404 | No account/usage/plan endpoints on the API-key side |
| `GET /api/*` | 404 | No separate dashboard API surface |

**Chat completion response (`POST /v1/chat/completions`) carries metadata in two places:**

- Body: `usage.remaining.hypercredits` (float) and `usage.cost.usd` / `usage.cost.hypercredits` per response
- Headers: `x-ratelimit-limit-day`, `x-ratelimit-limit-hour`, `x-ratelimit-remaining-day`, `x-ratelimit-remaining-hour` (request windows, not Hypercredit refresh)

**Dashboard "Plan / Shred / Next Hypercredit refresh in 13h" panel** is server-rendered HTML behind cookie auth, not exposed via any API-key endpoint I could find.

**Auth conventions:**
- `HYPER_API_KEY` env var — matches Crush and `@charmland/pi-hyper-provider` conventions
- OAuth subscription flow also exists (used by Crush), but API key is the simpler path

[...]

PR #2476, opened 2026-07-28
Add Charm Hyper balance provider

akshayprabhu200:
## Summary

- add Charm Hyper as an opt-in API-key provider using `HYPER_API_KEY`, saved provider keys, and token accounts
- fetch the documented `GET https://hyper.charm.land/v1/credits` endpoint and show the result as a Hypercredit balance in the menu bar, menu card, and CLI
- register Hyper across provider configuration, diagnostics, widgets, icons, documentation, and generated provider counts
- cover request construction, response validation, credential handling, menu-bar/card presentation, CLI output, and icon availability

## Context

Relates to #2318 and credits the original direction in #2379.

This draft differs from #2379 in two reviewed behavior details:

- it uses Charm's maintained `https://hyper.charm.land/v1/credits` contract instead of the incorrect `api.hyper.charm.land` host
- it renders `Balance: <amount> HC` instead of treating a balance as `Cost: <amount> / 0`

The API contract and `HYPER_API_KEY` convention are corroborated by Charm's maintained [pi Hyper provider](https://github.com/charmbracelet/pi-hyper-provider/blob/main/src/credits.ts) and [Crush provider documentation](https://github.com/charmbracelet/crush).

The implementation deliberately does not inspect browser cookies or reuse OAuth refresh tokens. It only sends an explicitly configured API key to Charm Hyper.

### Hypercredit unit semantics

Charm [currently documents](https://hyper.charm.land/) **1 Hypercredit = 5¢ USD**. This provider keeps the API's native HC balance authoritative, matching how CodexBar preserves other provider-specific credit units. It does not relabel HC as USD or hard-code a conversion into the usage snapshot, because the published rate is product pricing and may change independently of the API contract. A future presentation-only USD equivalent could be shown as approximate without replacing the HC value.

## Validation

- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer make check`
- `git diff --check`
- `make test` was invoked locally, but test discovery cannot start because the installed Xcode 15.3 provides Swift 5.10 while this package requires Swift 6.2; the draft PR's Xcode 26 macOS CI is the authoritative full-suite run

### Redacted live behavior proof

After the implementation was complete, a temporary API key was supplied only through hidden process input and used for one real request against the exact endpoint configured by this provider. The credential, authorization header, team identifier, and account identity were not logged or retained.

```text
LIVE_PROOF provider=hyper endpoint=/v1/credits http=200
Balance: 90 HC
```

This establishes that the production endpoint accepts the documented bearer credential, returns the numeric `balance` contract parsed by `HyperUsageFetcher`, and produces the HC balance presentation covered by the focused CLI/menu regression tests. No Keychain or browser-cookie import was used.

## UI evidence

No runtime screenshot is included because the installed local Xcode cannot build the Swift 6.2 app. The redacted live request above verifies the production API contract, while focused model tests assert the exact menu-bar, menu-card, and CLI balance presentation.

## Decision requested

Please confirm that a balance-only API-key v1 is acceptable for #2318. This remains a draft while that product decision and the full CI result are pending.


akshayprabhu200:
@clawsweeper re-review

steipete:
Status from maintainer triage: your implementation (API-key-only, native HC balance) and #2379 (`GET /v1/credits`) both target issue #2318, and your redacted live proof is appreciated. We're getting a Charm Hyper account so this can be verified properly before landing either.

Direction we're leaning: support **both** auth paths rather than picking one — API key when present, and a signed-in session when available (preferring the richer source) — consistent with how most CodexBar providers behave. We'll converge the two PRs into that shape once testing is possible, crediting both of you.

Heads up: this branch is now conflicted against main (Qwen Cloud, ZoomMate, and Alibaba Personal providers landed, which moved provider counts, registries, docs, and the generated parser hash). No need to churn on rebases until the direction is settled. Thanks @akshayprabhu200!

steipete:
Update on Charm Hyper: Peter is on the Charm Hyper **waitlist**, so we can't verify either implementation against a live account just yet.

Rather than leave two competing PRs sitting stale, we're combining them into a single new PR that implements both auth paths:

- **API key present** → infer usage/balance from it (the #2476 shape).
- **Signed-in session** → prefer it when available, since it exposes more (the #2379 `/v1/credits` shape).

Both of you will be credited as co-authors on the combined PR. It'll be prepared, tested, and CI-green so it can merge the moment account access comes through and the live behavior checks out. We'll link it here when it's up — thanks for the work and the patience, @JavaGT and @akshayprabhu200.

steipete:
The combined Charm Hyper PR is up: #2502 — "Add Charm Hyper provider (API key + session)".

It implements both auth paths as planned: a signed-in `hyper.charm.land` session is preferred in Auto mode (imported from Chrome or supplied as a manual Cookie header), falling back to an explicit API key from configuration, `HYPER_API_KEY`, or a token account. Balance data stays honest — no inferred plan limit, usage percentage, quota, or reset timestamp — and malformed successful responses fail closed rather than hiding an upstream contract change.

Both of you are recorded as co-authors on the combined commit, and the PR credits #2379 and #2476 explicitly.

It is **not merging yet**: Peter is on the Charm Hyper waitlist, so nobody has confirmed the behavior against a real account. Once access lands we'll verify session auth, the session-over-key preference and fallback, API/session balance parity, and HC rendering — then it goes in. CI is green apart from one macOS shard still queued.

Keeping #2379 and #2476 open for now so you can review the combined result and flag anything we got wrong about your implementations.

steipete:
Closing in favor of #2502, which incorporates the correct /v1/credits host and native balance presentation from this PR (you're credited as co-author there). The live-proof discipline from this branch is being repeated on #2502's exact head before it lands. Thanks @akshayprabhu200.

PR #2482, opened 2026-07-28
Prevent premature Codex weekly reset confetti

### LINKED ISSUE #2481: Codex weekly reset confetti can fire before the prior reset boundary
### Summary

CodexBar can publish a confirmed 0% Codex weekly snapshot and play reset confetti even when the prior reset boundary is still days away.

### Observed

- 2026-07-28 03:09:20Z: 100% used, resets 2026-08-02 10:17:56Z
- 2026-07-28 03:59:23Z: 0% used, resets 2026-08-04 03:59:21Z
- A repeated low sample advertised an equivalent rolling boundary within the existing two-minute tolerance.
- Usage later returned above zero.

The August 2 boundary had not elapsed, so this was not a genuine reset.

### Cause

#2056 requires the boundary to advance, and #2064 requires two matching low samples, but neither requires the previous boundary to be due. A degraded or fallback source that reports a rolling `now + 7 days` boundary can therefore pass both checks.

### Expected behavior

Confirmed weekly lows and reset confetti should only be accepted once the previous reset boundary has elapsed, with a small clock-skew tolerance. Until then, CodexBar should preserve the prior public snapshot, history, widgets, and celebration detector baseline.

### Related

#1986, #2054, #2056, #2064

gmkbenjamin:
## Summary
- Prevent premature Codex weekly-reset confetti by keeping confirmed early weekly lows private until the **previously published** reset boundary is due.
- The prior-boundary gate is now **unconditional** for both rolling and stable early-boundary patterns (aligns with #2481).
- Close/indeterminate observations still request a delayed confirmation; if still early after the delay, preserve previous.

Fixes #2481

## Contract
Confirmed weekly lows and reset celebration publish only after the prior reset boundary elapses. There is no stable-boundary early-publish exception.

## Real behavior proof
Focused decision-path proof on current head (cannot safely force a live weekly rollover):

- [`docs/screenshots/codex-false-reset-decision-paths-2026-07-31.txt`](https://github.com/gmkbenjamin/CodexBar/blob/codex/fix-false-reset-confetti/docs/screenshots/codex-false-reset-decision-paths-2026-07-31.txt)
- [`docs/screenshots/codex-false-reset-runtime-2026-07-31.txt`](https://github.com/gmkbenjamin/CodexBar/blob/codex/fix-false-reset-confetti/docs/screenshots/codex-false-reset-runtime-2026-07-31.txt)

Covers rolling preserve, stable preserve before prior boundary, and delayed confirmation.

## Test plan
- [x] Rolling early low before prior boundary preserves previous
- [x] Stable early low before prior boundary preserves previous
- [x] Indeterminate close observation requests delayed confirmation
- [x] Publication surfaces stay private for pre-boundary confirmed lows
- [x] `make check`
- [x] `swift test --filter 'CodexWeeklyResetConfirmationTests|CodexWeeklyResetPublicationTests'`

## Commands run
- `make check` (pass)
- focused weekly-reset confirmation/publication tests (pass)


gmkbenjamin:
@clawsweeper re-review — added packaged integration-runtime evidence while keeping the temporal rollover claim bounded to deterministic regression tests.

harjothkhara:
Reviewed at 8faae03c. The fix looks right: I traced #2481's timestamps through both implementations and base returns `.publishConfirmation` where this branch returns `.preservePrevious` via the `.rolling` path. The publication test asserting `.codexbarWeeklyLimitReset` never fires is the right proof shape.

Two things before this is ready:

1. The proof block is pinned to `025dd6d1`, but head is `8faae03c` and `c10db738` "Handle intentional Codex resets" landed in between — that commit added the `.stable` publish path and the 2s retry. Worth re-running and re-pinning so the runtime proof covers the behavior that's actually shipping.

2. The CHANGELOG entry is missing the usual `(#2482, fixes #2481). Thanks @gmkbenjamin!` suffix the surrounding entries carry.

Smaller notes:

- The summary says the guard was applied to reset celebration detection, but `UsageStore+LimitResetCelebration.swift` isn't in the diff — the confetti is suppressed indirectly, because the gate returns nil before `recordPlanUtilization` runs. Might be worth saying that explicitly.
- #2481 asks for lows to be accepted only once the prior boundary has elapsed, but `.stable` still publishes early when both samples agree on a fixed boundary. That seems intentional for genuine plan-change resets — could the trade-off be named in the body? It's the part the compatibility risk label is pointing at.
- `Task.sleep(for: .seconds(2))` matches `minimumBoundaryObservationSeconds = 2` only by coincidence; deriving one from the other would keep the retry correct if the constant changes.

gmkbenjamin:
Addressed @harjothkhara review + ClawSweeper proof staleness on `8c7d3aba`:

1. **Re-pinned proof** to current head (packaged launch + focused decision-path output covering rolling preserve / stable intentional publish / delayed confirmation).
2. **CHANGELOG** now uses `(#2482, fixes #2481). Thanks @gmkbenjamin!` under `0.46.1`, and names the stable-vs-rolling trade-off.
3. **Summary clarified:** confetti suppressed indirectly via publication gate; no `UsageStore+LimitResetCelebration` diff.
4. **Retry sleep** derives from `CodexWeeklyResetConfirmation.minimumBoundaryObservationSeconds`.
5. Merged current `main`.

@clawsweeper re-review

gmkbenjamin:
Rebased onto current `main` and took ClawSweeper’s recommended contract:

1. **Unconditional prior-boundary gate** — `.stable` and `.rolling` early lows both `preservePrevious` until the previously published boundary is due.
2. Updated confirmation + publication tests; refreshed decision-path proof on current head.
3. No release-owned `CHANGELOG.md` edit on this tip.

`make check` + focused weekly-reset tests pass.

@clawsweeper re-review

steipete:
Thanks @gmkbenjamin — the fix for #2481 is now on main via #2568, which keeps your root-cause analysis and regression tests (your original commit is cherry-picked in the branch, and the changelog credits you).

The landed shape simplifies the mechanism: instead of the EarlyBoundaryPattern classifier with delayed re-confirmation, `confirmationDecision` gains one prior-boundary guard (`confirmation.updatedAt < previousBoundary - 120s` → preserve previous). All three classifier branches ultimately produced the same pre-boundary outcome, so the guard delivers identical protection without the added latency or the extra provider request. Publication admission still precedes snapshot persistence, history recording, and reset detection, so the guard covers all affected surfaces.

Closing in favor of #2568 — appreciate the report and the investigation.

gmkbenjamin:
Happy to defer to #2568 here — it keeps the core fix and the test coverage, and a single prior-boundary guard is a simpler shape than the `EarlyBoundaryPattern` classifier plus the delayed third fetch. Thanks for the changelog credit.

Should I close this in favor of #2568? I've deliberately held off on resolving the `CHANGELOG.md` conflict and re-triggering CI here, since both look moot if this is superseded — happy to do either if you'd rather keep this one alive.

One note in case it's useful either way: the `swift-test-macos (1, 2)` failure on this PR isn't a test failure. Every test passed; the shard hit the 50-minute Actions timeout partway through group 91/95. Shard 1 takes ~30–38 min on healthy runs, so the margin is thin under runner load. I opened #2579 about a related wall-clock flake in `CodexModelsPerformanceTests`, which red-flagged an unrelated PR of mine the same way.

PR #2483, opened 2026-07-28
fix: validate LiteLLM and LLM Proxy base URLs before sending the API key

OfficialAbhinavSingh:
## Summary

`LiteLLMSettingsReader.baseURL` and `LLMProxySettingsReader.baseURL` built their URL with a bare
`URL(string:)` — no scheme, user-info, or host validation — and the API key is then sent to that URL
as a bearer token:

```swift
// LiteLLMSettingsReader.swift:16 (and LLMProxySettingsReader.swift:16)
guard let raw = self.cleaned(environment[self.baseURLEnvironmentKey]) else { return nil }
return URL(string: raw)
```
```swift
// LiteLLMUsageFetcher.swift:391 / LLMProxyUsageFetcher.swift:216
request.setValue("Bearer \(apiKey)", forHTTPHeaderField: "Authorization")
```

Every other provider override already goes through `ProviderEndpointOverrideValidator` — Sub2API,
ClawRouter, Wayfinder, z.ai, OpenRouter, NeuralWatt, Codebuff, AzureOpenAI. These two were the
outliers, so `http://…` and `https://user:pass@…` were accepted and used as a credential destination.

The value isn't environment-only: `ProviderConfigEnvironment.swift:65-68` maps `enterpriseHost` from
`~/.codexbar/config.json` onto `LITELLM_BASE_URL` / `LLM_PROXY_BASE_URL`, and
`CodexBarConfigValidation.swift` validates that field only for `sub2api` — so litellm/llmproxy got
neither validation nor a warning.

Scope note, stated plainly: the API key lives in that same `config.json`, so anyone who can write the
file can already read the key. This is defense-in-depth and consistency, not a privilege boundary
being crossed — which is why it's a normal PR rather than a security advisory.

## Fix

Both readers now use `ProviderEndpointOverrideValidator().validatedURLAllowingLoopbackHTTP(raw)`,
matching `WayfinderSettingsReader`: loopback HTTP still works for self-hosted proxies, any other host
must be HTTPS, and embedded credentials are rejected. A failing override is ignored, so the existing
`missingBaseURL` path reports it instead of silently using it; both messages now read "Missing or
invalid …" and state the rule.

**Behavior change worth calling out:** a non-loopback plain-HTTP base URL (e.g.
`http://10.0.0.5:4000`) is now rejected where it previously worked. That is the point of the change —
it's a cleartext bearer token to a remote host — and both docs pages only ever documented HTTPS
(`https://litellm.example.com`, `https://proxy.example.com`). Documented in `docs/litellm.md`,
`docs/llm-proxy.md`, and the changelog. Happy to relax it to allow private ranges if you'd rather not
break those setups.

## Validation

Built and tested on Linux in `swift:6.3.3`; macOS CI covers the rest.

**Red — fix reverted, tests kept (i.e. current `main`).** The failures show the vulnerable values
being accepted:

```
✘ liteLLMRejectsRemoteHTTPBaseURLBeforeSendingKey ... Expectation failed:
  (LiteLLMSettingsReader.baseURL(environment: [... : "http://attacker.test"])
     → http://attacker.test) == nil
✘ liteLLMRejectsBaseURLWithEmbeddedCredentials ... Expectation failed:
  (LiteLLMSettingsReader.baseURL(environment: [... : "https://user:pass@attacker.test"])
     → https://user:pass@attacker.test) == nil
✘ llmProxyRejectsRemoteHTTPBaseURLBeforeSendingKey ... → http://attacker.test) == nil
✘ llmProxyRejectsBaseURLWithEmbeddedCredentials ... → https://user:pass@attacker.test) == nil
✘ Test run with 15 tests in 1 suite failed after 0.006 seconds with 4 issues.
```

**Green — with the fix:**

```
✔ Suite ProviderEndpointOverrideSecurityLinuxTests passed after 0.009 seconds.
✔ Test run with 15 tests in 1 suite passed after 0.010 seconds.
```

The two acceptance tests (`https://litellm.example.com`, `http://127.0.0.1:4000`,
`http://localhost:4000`, `http://[::1]:4000`, `https://proxy.example.com`, `http://127.0.0.1:8080`)
pass in **both** states — they confirm self-hosted loopback setups keep working and that the
exemption is scoped rather than a blanket bypass. The 9 pre-existing tests in that suite also stay
green, so the sibling providers are unaffected.

```
$ swiftlint --strict
Done linting! Found 0 violations, 0 serious in 1586 files.
```

**Not tested:** a live LiteLLM/LLM Proxy round-trip — I don't have credentials for either. The change
only affects whether the configured URL is accepted, not the request itself.

Competing open PR scan: none — no open PR touches these readers or
`ProviderEndpointOverrideValidator`.

## CI note

Rebased onto current `main` (`be15ba0a`). The changelog entry that caused the conflict has since
been dropped (see below), so `CHANGELOG.md` is now identical to upstream.

Two checks are red on that merge commit, and I don't believe either is caused by this change:

- `swift-test-macos (0, 2)` — not a test failure. The job hit the 50-minute action timeout
  (`The action 'Swift Test' has timed out after 50 minutes`) with tests still passing through
  group 93/94. Shard 1 of the same run finished in 29m and passed.
- `lint-build-test` — fails in 6s purely as the downstream gate
  (`macOS test gate/result mismatch: required=true deferred=false result=failure`), so it clears
  once the shard does.

Everything else is green: both Linux builds, the musl build, lint, changes, GitGuardian.

This diff can't reach the macOS shard: it changes two Linux-buildable settings readers, two error
strings that no test asserts, `TestsLinux/*` (not run on the macOS shards), plus docs and the
changelog. Recent branches based on older `main` ran these shards at 29–40 min against the 50-minute
cap; this branch merged the latest landings, so it runs more tests and tipped over. Happy to rebase
again or split anything out if that would help — I can't re-run the job myself.

## Review follow-up (`6727ecd3`)

Both ClawSweeper findings are addressed.

**Surface rejected endpoint configuration (P2).** This was a real defect — thanks for catching it. A
rejected override made `baseURL` nil, and both availability checks treated nil as "not configured",
so the strategy was skipped and the user got generic unavailability instead of the new message.

Fixed the way the sibling providers already do it: added `invalidEndpointOverride(key)` to both error
enums (wording mirrors `WayfinderSettingsError`) and `hasBaseURLOverride(environment:)` to both
readers, which reports whether a value is configured regardless of validity. `isAvailable` now keys
off that in the Core strategies and the two macOS provider implementations, so a rejected override
reaches `fetch` and throws an error naming the setting and the accepted URL shape:

> LiteLLM base URL override LITELLM_BASE_URL is invalid. Use an HTTPS URL, or plain HTTP for loopback
> addresses only, without embedded credentials.

That also answers the P1 upgrade risk: an existing remote plain-HTTP config now says exactly which
key to fix instead of the provider quietly disappearing. The `missingBaseURL` text is back to its
original wording, since "invalid" has its own case now.

**Release-owned changelog (P3).** Dropped; `CHANGELOG.md` is byte-identical to upstream. My mistake —
I checked an open PR rather than merged ones when inferring the convention.

Two tests added: one asserting a rejected override still reads as configured (so the error can
surface), one asserting each message names its env key, HTTPS, and the loopback exemption.

```
$ swift test --filter ProviderEndpointOverrideSecurityLinuxTests
✔ Test run with 17 tests in 1 suite passed after 0.005 seconds.
$ swiftlint --strict
Done linting! Found 0 violations, 0 serious in 1645 files.
```

Still no live LiteLLM/LLM Proxy account here, so the configured-provider transcript remains the one
gap I can't close myself.


OfficialAbhinavSingh:
@clawsweeper re-review

steipete:
Maintainer verification complete on head `169714299fa7e29dd1c395e586a97fb0dc61dcd1`.

Final review found no blocking issue. LiteLLM and LLM Proxy now reject public plain-HTTP endpoints and URL user info while preserving HTTP for loopback, RFC 1918, IPv4 link-local, IPv6 unique-local/link-local, and `.local` proxy hosts. The stricter loopback-only validator used by other providers remains unchanged, and missing configuration is still surfaced separately from a rejected override.

Focused local proof: `swift test --filter ProviderEndpointOverrideSecurityLinuxTests` passed 18 tests, including range boundaries, IPv6, mDNS, public-HTTP rejection, and shared-validator isolation. Hosted CI is fully green in [run 30462181955](https://github.com/steipete/CodexBar/actions/runs/30462181955), including Linux x64/arm64, musl, both macOS test shards, lint, and the aggregate gate.

Merged as `c52b0efa13b0169064063a20ca95b87ca8c4435f`. Thank you for the careful security analysis and for following through on the invalid-configuration error path, @OfficialAbhinavSingh!

steipete:
Merged after full verification, with the owner-approved scope: endpoint overrides are validated before any bearer token is sent, and plain HTTP is accepted for **loopback and private networks** — RFC1918 (10/8, 172.16/12, 192.168/16), link-local 169.254/16, IPv6 unique-local fc00::/7 and link-local fe80::/10, plus `.local` mDNS hostnames. Public hosts over plain `http://` remain rejected, which is the point: credentials never leave the machine in cleartext to the open internet, while self-hosted LiteLLM/LLM Proxy setups on a home or office LAN keep working.

The missing-vs-rejected configuration distinction in the surfaced error is preserved. Exact-head CI: all nine checks green. Thanks @OfficialAbhinavSingh!

Issue #2449, opened 2026-07-25
Feature request: option to display cost estimates in a non-USD currency (e.g. GBP)


hedche:
**What I'd like**

A way to have CodexBar show the Claude cost/estimate figures in my own currency (GBP in my case) instead of always `$`.

**Why**

My Anthropic account is billed in GBP. The dashboard shows everything in £:

- £16.09 spent
- £15.00 monthly spend limit
- £58.91 current balance

But the estimate CodexBar shows in the menu bar is always in `$`, so I'm mentally converting every time and it doesn't line up with what I actually see on my Anthropic billing page.

**What I found in the code (for context)**

From a quick look, this seems intentional rather than a bug:

- The local-log cost estimate is hardcoded to USD — `UsageFormatter.usdString` / `currencyString(...)` pin it to `.currency(code: "USD")` with an `en_US` locale, so it's `$` regardless of region. Totally understand why: it's an estimate at API token rates, not a real bill.
- The "extra usage" / spend-limit line *does* already respect the currency the API reports (`OAuthExtraUsage.currency` → `ProviderCostSnapshot.currencyCode`), which is great.

**Possible options (whatever's easiest / makes sense)**

1. A currency picker in Settings → Display, applied to the estimate figures.
2. Or just follow the currency the Claude API already reports for the account (the same value the extra-usage line uses), so it "just works" without a setting.

Thanks for CodexBar — it's become a permanent fixture in my menu bar.

steipete:
Non-USD cost display landed via #2490. You can now pick a display currency, and CodexBar converts cost estimates using daily-updated rates.

Privacy note on the design: rate fetching is **opt-in by construction** — while you're on USD (or Auto), CodexBar never contacts the exchange-rate service at all. It only fetches once you've actually selected a non-USD currency, caches for 24 hours, and falls back to bundled rates offline.

Please confirm GBP renders correctly for you on the next release — keeping this open until then. Thanks for the request!

steipete:
Non-USD cost display shipped in #2490 (GBP included), and the currency picker keeps growing — CZK just landed via https://github.com/steipete/CodexBar/pull/2886. Closing as completed; if a currency you need is missing, a small PR adding it to the picker + CurrencyExchange is very welcome.

PR #2485, opened 2026-07-28
Add daily spend ledger

sahilaidev:


steipete:
Thanks for the contribution @sahilaidev, and sorry for the blunt outcome — closing this one, with reasons worth having on record:

The PR bundles three unrelated things: a daily spend ledger, GBP currency conversion via Coinbase polling, and fork-packaging customization. The packaging changes aren't applicable upstream, and the currency piece overlaps #2490, which takes an app-wide approach and is the one we're pursuing for #2449. There's also a behavior bug in the currency path: existing users get switched to GBP by default, and every successful rate refresh re-enables GBP even after the user opts out — a preference has to stay opted out once set.

The **daily spend ledger** on its own is the interesting idea here. If you'd like to bring it back as a focused PR — ledger only, no currency conversion, no packaging changes, with tests covering the request-aggregation and coverage semantics — that's a much easier review and we'll look at it properly.

PR #2437, opened 2026-07-24
fix(alibaba): restore Token Plan 5-hour and weekly windows

rohitsabu:
## Summary

- fetch Alibaba Token Plan rate limits from the current authenticated console RPC
- map fractional 5-hour and 1-week percentages plus millisecond reset timestamps into CodexBar rate windows
- retain the existing subscription-summary request as a fallback and optional monthly credits window
- add redacted parser and request-contract regression coverage

## Regression evidence

The live authenticated international dashboard currently calls `zeldaHttp.apikeyMgr./tokenplan/personal/api/v2/usage` and returns `per5HourPercentage`, `per1WeekPercentage`, `per5HourResetTime`, and `per1WeekResetTime` inside a nested JSON envelope. No cookie, token, account identifier, or response value is included in this change.

A fail-then-pass isolated harness using the redacted fixture verified:

- `0.0769` maps to `7.69%` over 300 minutes
- `0.0261` maps to `2.61%` over 10,080 minutes
- millisecond reset timestamps map to `Date`
- legacy subscription-summary data remains the fallback

## Verification

- `swift build --product CodexBarCLI` on current `main`: pass
- SwiftFormat for all four changed files: pass
- `git diff --check`: pass
- focused `swift test --filter AlibabaTokenPlanProviderTests`: environment-blocked locally because this Mac currently has Command Line Tools only and no `Testing` module; the test sources are included for full-Xcode CI

The change is provider-only: no mobile, signing, entitlement, credential, or private configuration files are touched.

Issue #2370, opened 2026-07-21
Alibaba Token Plan: personal (Solo) subscriptions render no quota rows — ProductCode is hardcoded to the Teams plan


LeoLin990405:
## Summary

On a **personal (Solo) Alibaba Token Plan subscription**, the Alibaba Token Plan card authenticates fine but renders no quota rows — `primary`, `secondary` and `tertiary` are all `null`. The card shows only the provider title and `Plan: Token Plan`.

The subscription-summary request is hardcoded to the **Teams** product code, and the console returns an empty payload for a Solo subscription.

## Environment

- CodexBar **0.45.0** (also reproduced on 0.41.0)
- macOS, menu-bar app + `codexbar` CLI
- Alibaba Model Studio / Bailian, region `cn-beijing`
- Subscription: **Token Plan "Pro 套餐", personal (Solo)**, auto-renew on
- Auth: cookie (web), working

## Observed

`codexbar usage --provider alibabatokenplan --format json` (redacted):

```json
{
  "usage": {
    "identity": { "providerID": "alibabatokenplan", "loginMethod": "TOKEN PLAN" },
    "primary": null,
    "secondary": null,
    "tertiary": null,
    "updatedAt": "2026-07-21T09:03:25Z"
  },
  "source": "web",
  "provider": "alibabatokenplan"
}
```

Text output is just:

```
== Alibaba Token Plan (web) ==
Plan: Token Plan
```

Auth itself is healthy — debug log shows the sec_token being resolved and a 200 response:

```
info …alibaba-token-plan: bodyBytes=225 userInfoHost=bailian.console.aliyun.com  Resolved Alibaba Token Plan sec_token from user info
info …alibaba-token-plan: bodyBytes=407 contentType=application/json;charset=UTF-8 status=200  Alibaba Token Plan HTTP response
```

So the provider reaches the API and gets a 200 — the 407-byte body simply carries no quota fields.

## Root cause (source-level)

`AlibabaTokenPlanUsageFetcher` builds the subscription-summary request with a product code that is hardcoded to the **Teams** plan:

`Sources/CodexBarCore/Providers/Alibaba/AlibabaTokenPlanAPIRegion.swift`

```swift
public var tokenPlanProductCode: String {
    switch self {
    case .international:
        "sfm_tokenplanteams_dp_intl"
    case .chinaMainland:
        "sfm_tokenplanteams_dp_cn"
    }
}
```

[...]

Issue #2349, opened 2026-07-20
Alibaba Token Plan (Personal/Solo, Singapore ap-southeast-1) — quota not parsed; auth succeeds but body parser returns null


halilertekin:
### Problem

The `alibaba-token-plan` provider in CodexBar detects the login cookie correctly and reaches `modelstudio.console.alibacloud.com` over HTTPS, but the parsed usage is always `primary=null, secondary=null, tertiary=null`. Only `loginMethod=TOKEN PLAN` is exposed, so the menubar widget shows "Plan: Token Plan" without any quota/reset windows.

This is distinct from #612 (which is about `alibaba-coding-plan` INTL users with no auth path). This issue is about `alibaba-token-plan` for **Singapore Personal/Solo subscriptions**, where auth works but quota parsing does not.

### Environment
- CodexBar: latest stable (menubar build, macOS 14+)
- Account: Alibaba Cloud **International (INTL)** Singapore workspace
- Plan: **Token Plan — Personal/Solo edition** (not Coding Plan, not Team edition)
- Region: `ap-southeast-1`
- Authentication: Chrome Default profile cookies on `modelstudio.console.alibabacloud.com` (Section "Alibaba Token Plan")
- API key (`sk-sp-H.RIXR.{signature}`): confirmed valid for inference endpoint `https://token-plan.ap-southeast-1.maas.aliyuncs.com/apps/anthropic/v1/messages` (returns 200 OK), but token-plan team endpoint (`sfm_tokenplanteams_dp_cn` ProductCode) rejects it.

### Symptoms

`codexbar usage --provider alibaba-token-plan --json` returns:
```json
[
  {
    "provider": "alibabatokenplan",
    "source": "web",
    "usage": {
      "loginMethod": "TOKEN PLAN",
      "secondary": null,
      "tertiary": null,
      "updatedAt": "2026-07-20T02:24:04Z",
      "primary": null,
      "identity": {
        "providerID": "alibabatokenplan",
        "loginMethod": "TOKEN PLAN"
      }
    }
  }
]
```

Verbose log (`--log-level trace`) confirms:
```
apiHost=modelstudio.console.alibacloud.com dashboardCookieNames=[…27 cookies including login_aliyunid, hsite…] hasCSRF=1 region=intl secTokenSource=resolved
bodyBytes=409 contentType=application/json;charset=UTF-8 status=200
```

[...]

PR #2487, opened 2026-07-28
Add Alibaba Token Plan Personal/Solo variants (mainland + international)

steipete:
## Summary

- add explicit international and China-mainland Personal/Solo variants inside the existing Alibaba Token Plan provider
- route Personal usage to the dedicated rolling-window gateways while leaving both existing Team variants on `GetSubscriptionSummary`
- rebuild browser cookies independently for the dashboard and quota hosts, and support cookie-only Personal requests without requiring `sec_token`
- parse shared 5-hour/7-day usage, reset times, subscription tier, and tier quota metadata from issue-documented fixtures
- keep weekly-only payloads visible and label Personal rows as 5-hour / 7-day across the menu, descriptors, and widget snapshots

Based on the investigation by @LeoLin990405 (#2370) and @halilertekin (#2349). The international endpoint work from @rohitsabu's closed #2437 also informed the coordinated rewrite, with contributor credit preserved in the commit.

## Verification

- `swift test --filter AlibabaTokenPlan` — 46 tests passed
- focused settings picker test — passed
- `make check` — passed (format, strict lint, locale/docs/package gates)
- isolated autoreview — clean, no accepted/actionable findings

Live provider authentication was not run here, so this intentionally relies on the reporters confirming both variants in the next release, mirroring the #2361 approach. Issues #2370 and #2349 should remain open until that confirmation.

### DIFF
--- a/Sources/CodexBar/MenuCardView+ModelHelpers.swift
+++ b/Sources/CodexBar/MenuCardView+ModelHelpers.swift
@@ -288,11 +288,15 @@ extension UsageMenuCardView.Model {
             Sub2APIProviderDescriptor.primaryLabel(details: snapshot.sub2APIUsage) ?? input.metadata.sessionLabel
         } else if input.provider == .amp {
             AmpProviderDescriptor.primaryLabel(details: snapshot.ampUsage) ?? input.metadata.sessionLabel
+        } else if input.provider == .alibabatokenplan {
+            AlibabaTokenPlanProviderDescriptor.primaryLabel(window: snapshot.primary) ?? input.metadata.sessionLabel
         } else {
             input.metadata.sessionLabel
         }
         let secondaryLabel = if input.provider == .amp {
             AmpProviderDescriptor.secondaryLabel(details: snapshot.ampUsage) ?? input.metadata.weeklyLabel
+        } else if input.provider == .alibabatokenplan {
+            AlibabaTokenPlanProviderDescriptor.secondaryLabel(window: snapshot.secondary) ?? input.metadata.weeklyLabel
         } else {
             input.metadata.weeklyLabel
         }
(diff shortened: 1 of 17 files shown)


LeoLin990405:
Reporter of #2370 here — thanks for the coordinated rewrite and the credit.

**China-mainland Personal/Solo: confirmed working against a live account.** I wired the same routing locally (a `chinaMainlandPersonal` region → `bailian-cs.console.aliyun.com` + the `.../tokenplan/personal/api/v2/{usage,subscription,quota-config}` gateway, cookie-only, no `sec_token`) and it renders real data on my live Solo account:

```
5-hour: usedPercent=0  windowMinutes=300
7-day:  usedPercent=0  windowMinutes=10080   (src=web)
```

The two things this PR calls out both hold up in practice:
- cookie-only works — no `sec_token` needed;
- the host-scoped cookie rebuild matters — a flat aliyun cookie header gets `BailianGateway.Login.NotLogined`; only cookies scoped to `bailian-cs.console.aliyun.com` (most-specific host winning on name clashes) authenticate. Good that the PR rebuilds cookies independently per host.

**+1 on the Codex review's `consoleSite` finding — it's a real typo, not a false positive.** The mainland variant is unaffected (different `consoleSite`), which is why my mainland test passes, but the International Personal path sends `"MODELSTUDIO_ALBABACLOUD"` while the existing `AlibabaCodingPlanAPIRegion.consoleSite` uses `"MODELSTUDIO_ALIBABACLOUD"` (with the `I`). Worth fixing before merge so the intl Personal gateway doesn't get an unrecognized site id — and maybe assert the intl Personal request body in the tests too, since only the mainland body is currently checked.

I can only speak to the **mainland** variant end-to-end (I'm a mainland Solo account); I can't authenticate the international gateway myself. Happy to re-run mainland against the shipped build once it lands to close out #2370.

steipete:
Merged after full verification (owner-approved architecture: Personal/Solo as explicit variants inside the Alibaba provider). Mainland Personal targets the `bailian-cs.console.aliyun.com` gateway with host-scoped cookie assembly; international Personal covers the Model Studio Singapore path; a shared rolling-window parser serves both; cookie-only auth works (no secToken requirement); weekly-only responses render; existing Team variants are untouched (regressions pin that). 46 focused tests + make check green; autoreview clean; exact-head CI fully green after a musl SDK infra rerun (https://github.com/steipete/CodexBar/actions/runs/30403595624). Based on the investigations by @LeoLin990405 (#2370) and @halilertekin (#2349) — live confirmation requested from both on the next release.

PR #2492, opened 2026-07-29
Classify StepFun Coding vs Token Plan by payload shape, not a hard-coded plan_family

### LINKED ISSUE #2491: StepFun: classify Coding Plan vs Token Plan by payload shape, not a hard-coded plan_family == 2
## Summary

StepFun's `isCreditPlan` classifies a plan as credit-based by trusting `plan_family == 2` first. Since StepFun's **2026-06-18 Step Plan upgrade** two billing models now run side by side, and both can reach this code path, so classification should key off the payload *shape* rather than a single magic family id.

## Background — two models coexist

Per StepFun's own upgrade notice ([`docs/zh/step-plan/upgrade-notice`](https://platform.stepfun.com/docs/zh/step-plan/upgrade-notice)):

- **Coding Plan** (grandfathered): 按请求次数(Prompt), **5 小时 / 周限额** — rolling 5-hour + weekly windows.
- **Token Plan** (current, since 2026-06-18): **Credit 月池** — a monthly Credit pool, "统一 Credit 额度体系,按 Token 使用量换算".

The upgrade is one-way; users with auto-renew keep the old **Coding Plan**, everyone else moves to **Token Plan** on expiry. So a released build can see either model in the wild.

`QueryStepPlanRateLimit` distinguishes them by shape:

```jsonc
// Token Plan (Credit): live rate windows are 0 with reset_time "0" = "no window"
{ "five_hour_usage_reset_time": "0", "weekly_usage_reset_time": "0",
  "plan_family": 2,
  "plan_credit_rate_limit": { "subscription_credit_left_rate": 1, "credit_buckets": [ … ] } }

// Coding Plan: live rolling windows (non-zero reset_time), no meaningful credit pool
{ "five_hour_usage_left_rate": 0.8, "five_hour_usage_reset_time": "1746000000",
  "weekly_usage_left_rate":  0.6, "weekly_usage_reset_time":  "1746500000" }
```

## The fragility

```swift
var isCreditPlan: Bool {
    if let family = self.planFamily?.value, family > 0 {
        return family == 2          // ← trusted before the payload shape
    }
    …
}
```

Because the family id wins first, a **windowed** payload that also carried `plan_family == 2` — or any future change to the family id StepFun assigns — routes a rolling-window plan into the credit renderer, which drops the real 5h/weekly windows and shows a credit balance the plan doesn't have. The 5h/weekly fields are right there in the same response, so the classifier has enough to decide from shape.

## Proposed fix

Classify by the shape the payload actually carries; use `plan_family` only to break a genuine tie:

1. a live rolling window (`*_reset_time > 0`) → Coding Plan (windows);
2. no live window + a Credit pool present → Token Plan (credit);
3. neither (e.g. a brand-new plan with no usage yet) → fall back to `plan_family == 2`.

## Verification scope

I can only exercise a live **Token Plan** account (which keeps rendering its credit balance under the shape-driven path). I don't have a grandfathered **Coding Plan** account, so that branch is covered by a fixture built from the documented shape + the fields observed on the rate-limit endpoint — mirroring the fixture-based approach used for the Alibaba Personal work in #2487.

PR incoming.

LeoLin990405:
## Summary

Closes #2491.

StepFun's `isCreditPlan` trusted `plan_family == 2` ahead of the payload shape. After the **2026-06-18 Step Plan upgrade** the grandfathered **Coding Plan** (rolling 5-hour/weekly windows) and the current **Token Plan** (monthly Credit pool) both run in the wild, so a windowed payload that ever carried `plan_family == 2` — or any future family-id change — would route a rolling-window plan into the credit renderer, dropping the real windows and showing a credit balance the plan doesn't have.

This classifies by the shape the response actually carries and keeps `plan_family` only as a tie-breaker:

- a live rolling window (`*_reset_time > 0`) → Coding Plan (windows);
- no live window + a Credit pool present → Token Plan (credit);
- neither → fall back to `plan_family == 2`.

The credit renderer already only fires when a plan has a Credit pool and no live windows, so the current Token Plan path is unchanged; this just stops the family id from overriding a live window.

## Verification

- `swift test --filter StepFun` — 39 passed, including two new cases: a live-window payload that also reports `plan_family == 2` stays on the rolling-window renderer, and the ambiguous-payload fallback to the family id.
- `Scripts/lint.sh` (strict) — 0 violations; SwiftFormat — no changes.
- **Live**: ran against a real **Token Plan** account — still classifies as credit and renders the credit balance as before (no regression).

I don't have a grandfathered **Coding Plan** account to exercise the window branch live, so that path is covered by a fixture built from the documented shape + the fields observed on `QueryStepPlanRateLimit`, mirroring the fixture-based approach used for the Alibaba Personal variants in #2487.

### DIFF
--- a/Sources/CodexBarCore/Providers/StepFun/StepFunUsageFetcher.swift
+++ b/Sources/CodexBarCore/Providers/StepFun/StepFunUsageFetcher.swift
@@ -81,28 +81,30 @@ public struct StepFunRateLimitResponse: Decodable, Sendable {
         self.status == 1
     }
 
-    /// Credit-based plans (plan_family=2) report usage via `plan_credit_rate_limit`
-    /// instead of the five-hour / weekly rate windows. Those rate fields are 0 with
-    /// reset_time "0" — meaning "no window configured", NOT "fully consumed".
+    /// StepFun runs two Step Plan billing models side by side after the 2026-06-18
+    /// upgrade (docs/zh/step-plan/upgrade-notice): the grandfathered **Coding Plan**
+    /// meters rolling **5-hour / weekly** windows, while the current **Token Plan**
+    /// meters a monthly **Credit** pool via `plan_credit_rate_limit` (its rate windows
+    /// come back as 0 with `reset_time` `"0"` — "no window configured", not "used up").
+    ///
+    /// Classify by the shape the payload actually carries rather than trusting
+    /// `plan_family` alone: a live rolling window means Coding Plan; no window plus a
+    /// Credit pool means Token Plan. `plan_family` (2 == the Credit family) is only a
+    /// tie-breaker for an ambiguous payload (e.g. a brand-new plan with neither a live
+    /// window nor credit yet), so a future family-id change can't silently flip a
+    /// windowed plan onto the credit renderer or vice versa.
     var isCreditPlan: Bool {
-        // plan_family 2 = credit-based subscription plans (e.g. Mini, Pro).
-        if let family = self.planFamily?.value, family > 0 {
-            return family == 2
+        let hasLiveWindow = (self.fiveHourUsageResetTime?.value ?? 0) > 0
+            || (self.weeklyUsageResetTime?.value ?? 0) > 0
+        if hasLiveWindow {
+            return false
         }
-        // Fallback heuristic: if both rate windows are 0 with no reset time, but
-        // credit data is present, treat as a credit plan.
-        if let credit = self.planCreditRateLimit,
-           (credit.subscriptionCreditLeftRate?.value ?? 0) > 0
-        {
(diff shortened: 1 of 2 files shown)


steipete:
Merged after full verification. StepFun plans are now classified by payload shape — live reset windows take precedence over `plan_family`, with the family-ID fallback retained.

Fixed during review: `hasCreditPool` required a *positive* credit balance or non-empty buckets, so an exhausted zero-credit pool was misclassified — exactly the state a user hits when they've run out, and worst if the family ID is missing or changed. It now detects credit-field *presence* (including an explicit zero and the top-up field), with regressions for the exhausted-pool and zero-with-top-up cases. Thanks @LeoLin990405!

PR #2221, opened 2026-07-16
Doubao: query Coding/Agent Plan usage via arkcli CLI

start3015:
## Summary

- Replace the Volcengine AK/SK signed API with `arkcli usage plan` CLI (SSO-based) for fetching Coding Plan and Agent Plan usage
- Split Doubao menu card into "Coding Plan" and "Agent Plan" sections with group headers
- Replace placeholder smiley icon with official Volcengine logo SVG

## Why

The signed Volcengine API (`GetCodingPlanUsage` / `GetAgentPlanUsage`) required users to obtain AK/SK credentials separately and did not work well. The `arkcli` CLI uses the same SSO session the user already has, making setup trivial - just install arkcli and `arkcli auth login`.

## Changes

- `DoubaoUsageFetcher`: new `fetchCodingPlanUsage(runArkcli:)` runs `arkcli usage plan` and decodes the JSON response. Old Volcengine signer code removed.
- `DoubaoAPIFetchStrategy`: try arkcli CLI first, fall back to Ark API key rate-limit probe.
- `MenuCardView`: split Doubao metrics into two groups (Coding Plan / Agent Plan) with section headers. Other providers unaffected (`guard provider == .doubao`).
- `ProviderIcon-doubao.svg`: official Volcengine logo (MIT-licensed, from thesvg.org).
- Tests updated to match new arkcli JSON format and strategy interface.

Closes #1724.

Issue #1835, opened 2026-07-02
Doubao: support Volcengine Agent Plan bearer keys via /api/plan/v3


TsezzzRay:
## Summary

CodexBar's current Doubao provider probes `https://ark.cn-beijing.volces.com/api/coding/v3/chat/completions`. That works for Coding Plan-style endpoints, but Volcengine Ark **Agent Plan** uses a different OpenAI-compatible endpoint:

```text
https://ark.cn-beijing.volces.com/api/plan/v3/chat/completions
```

An Agent Plan bearer key is valid against `/api/plan/v3`, but CodexBar sends it to `/api/coding/v3`, where it returns `401 AuthenticationError`.

## Environment

- CodexBar: 0.37.2, build 92
- Provider: Doubao
- Config path: `~/.codexbar/config.json`
- Key type: Volcengine Ark Agent Plan API key (`ark-...`, redacted)
- Region: `cn-beijing`

## What I observed

The Volcengine Ark console Agent Plan page shows these base URLs:

```text
Anthropic-compatible: https://ark.cn-beijing.volces.com/api/plan
OpenAI-compatible:   https://ark.cn-beijing.volces.com/api/plan/v3
```

The subscription page URL was:

```text
https://console.volcengine.com/ark/region:cn-beijing/subscription/agent-plan
```

## Reproduction

1. Enable Doubao in CodexBar.
2. Configure an Agent Plan API key in `~/.codexbar/config.json`:

```json
{
  "id": "doubao",
  "enabled": true,
  "apiKey": "ark-...redacted..."
}
```

3. Run:

```bash
/Applications/CodexBar.app/Contents/Helpers/CodexBarCLI usage --provider doubao --format json --pretty --log-level debug
```

4. CodexBar returns:

```text
Doubao API error (401): The API key or AK/SK in the request is missing or invalid.
```

## Direct endpoint checks

Using the same redacted Agent Plan key:

```bash
curl https://ark.cn-beijing.volces.com/api/coding/v3/chat/completions \
  -H 'Content-Type: application/json' \
  -H 'Authorization: Bearer <redacted-agent-plan-key>' \
  -d '{"model":"doubao-seed-2.0-code","messages":[{"role":"user","content":"hi"}],"max_tokens":1}'
```

returns `401 AuthenticationError`.

But:

```bash
curl https://ark.cn-beijing.volces.com/api/plan/v3/chat/completions \
  -H 'Content-Type: application/json' \
[...]

PR #2496, opened 2026-07-29
Doubao: read Agent Plan (AFP) usage on the AK/SK API source

LeoLin990405:
## Summary

The Doubao `.api` (AK/SK) source signs `GetCodingPlanUsage`, which only
covers a **Coding Plan**. A Volcengine Ark account holds a Coding Plan
*or* an **Agent Plan** — not both — so on an Agent Plan account the Coding
Plan is *reclaimed* and `GetCodingPlanUsage` returns `Status` only, with no
`QuotaUsage`:

```json
{"Result":{"Status":"Reclaimed","UpdateTimestamp":1785322689}}
```

The strict decoder treats a missing `QuotaUsage` as a hard failure, so the
account surfaces **“Failed to parse Doubao response”** and shows no usage
at all on the `.api` source.

This teaches the AK/SK path to read the Agent Plan:

1. A quota-less Coding Plan result decodes as **empty** instead of throwing.
2. When the Coding Plan carries no active quota, fall back to the sibling
   **`GetAFPUsage`** action (AFP = "Agent Flow Points"), signed with the
   same AK/SK, version, and region. Its rolling windows map onto the
   existing `agent_*` rate-window levels:

   ```json
   {"Result":{"PlanType":"medium",
     "AFPFiveHour":{"Quota":10000,"Used":0,"ResetTime":-1},
     "AFPWeekly":{"Quota":35000,"Used":8750,"ResetTime":1785686400000},
     "AFPMonthly":{"Quota":100000,"Used":25000,"ResetTime":1787846399000}}}
   ```

So the `.api` source now renders an Agent Plan account **identically to the
arkcli `.cli` path** — the 5-hour / Weekly / Monthly windows — without
requiring `arkcli` to be installed or an SSO login.

The fallback is deliberately narrow: it fires only on a `200` whose Coding
Plan has no active quota (e.g. `Status: "Reclaimed"`). Hard errors (403,
etc.) still surface as-is, and an active Coding Plan is returned as before,
with no extra request.

## Relationship to existing Doubao work

- **#2221** (merged) reads Coding/Agent Plan usage through `arkcli usage
  plan` on the `.cli` source. This PR brings **Agent Plan** parity to the
  `.api` (AK/SK) source, so users with AK/SK credentials no longer need to
  install `arkcli` or do an SSO login just to see Agent Plan usage.
- **#1835** is about Agent Plan **bearer** keys and the OpenAI-compatible
  `/api/plan/v3/chat/completions` probe — a distinct auth path. This PR
  does not touch that; it uses the AK/SK-signed `GetAFPUsage` usage action.

## Verification

- `swift test --filter Doubao` — 67 passed, including a new
  `DoubaoAgentPlanUsageTests` suite (4 cases): AFP window → `agent_*`
  mapping (incl. `AFPDaily` correctly *not* surfaced), the
  reclaimed-Coding-Plan → Agent-Plan fallback (asserts the second request
  targets `GetAFPUsage`), a reclaimed Coding Plan decoding as empty rather
  than throwing, and an active Coding Plan returning without probing AFP.
- `swiftlint --strict` — 0 violations; SwiftFormat — clean.
- **Live**: ran the built CLI against a real Agent Plan (**Medium**)
  account on the `.api` source. It now returns the 5-hour (0%), Weekly, and
  Monthly windows with correct resets — matching that account's `.cli`
  output — instead of the parse error.

`GetAFPUsage` was confirmed both from a live `200` response and by
cross-checking `arkcli`'s own usage operation set; it needs no extra
request parameters (empty signed body, `Version=2024-01-01`, service `ark`).

### DIFF
--- a/Sources/CodexBarCore/Providers/Doubao/DoubaoUsageFetcher.swift
+++ b/Sources/CodexBarCore/Providers/Doubao/DoubaoUsageFetcher.swift
@@ -266,6 +266,12 @@ public struct DoubaoUsageFetcher: Sendable {
     private static let apiURL = URL(string: "https://ark.cn-beijing.volces.com/api/coding/v3/chat/completions")!
     private static let codingPlanAPIURL = URL(
         string: "https://open.volcengineapi.com/?Action=GetCodingPlanUsage&Version=2024-01-01")!
+    /// Agent Plan usage lives behind a sibling Volcengine Top OpenAPI action (`GetAFPUsage`,
+    /// AFP = "Agent Flow Points"), signed with the same AK/SK the Coding Plan path uses. An
+    /// account holds a Coding Plan *or* an Agent Plan, so `GetCodingPlanUsage` returns no
+    /// active quota for an Agent Plan account and we fall back to this action.
+    private static let agentPlanAPIURL = URL(
+        string: "https://open.volcengineapi.com/?Action=GetAFPUsage&Version=2024-01-01")!
 
     /// Closure that runs `arkcli usage plan` and returns raw stdout.
     public typealias ArkcliRunner = @Sendable () async throws -> Data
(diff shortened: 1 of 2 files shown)


steipete:
Merged after full verification. Doubao now reads Agent Plan (AFP) usage on the AK/SK API source, complementing the Ark CLI path.

Fixed during review: a broad `try?` was swallowing AFP 403s, transport failures, decode errors, and cancellation, surfacing them as an empty success — so an auth or network problem looked identical to "no usage," which is the worst failure mode for a usage tool. Those now propagate through the normal error taxonomy, with tests for 403, malformed response, transport failure, and cancellation. Thanks @LeoLin990405!

PR #2506, opened 2026-07-29
Preserve Claude widget quota on partial refresh (#2480 by @ChenZiHong-Gavin)

steipete:
## Summary

- preserve the last real Claude Session/Weekly widget rows when a partial refresh only has fresh local token/cost history
- persist quota ownership and invalidate preserved rows after credential changes, account switches, provider cleanup, or an explicit quota-unavailable result
- scope default OAuth quota ownership to the profile identity established by #2484, and fail closed for legacy ownerless snapshots

## Ownership hardening

PR #2480 compared an optional owner key. Default Claude OAuth snapshots and legacy snapshots both had no owner, so they compared equal; after profile-scoped credential caching landed, that could show profile A's stale quota under profile B.

This replacement derives a stable non-nil OAuth owner from the existing credentials-profile identity. Preservation requires both the previous snapshot and the current profile to have the same non-nil owner. Legacy ownerless snapshots are deliberately dropped once rather than risk crossing an ownership boundary. Token-account owners continue using their existing account-scoped cache keys.

## Credit and replacement context

This recreates and hardens #2480 by @ChenZiHong-Gavin. Their three commits were cherry-picked with authorship preserved, and the follow-up commit includes a co-author trailer. A replacement PR is necessary because the contributor fork has maintainer edits disabled.

## Validation

- `swift test --filter UsageStoreWidgetSnapshotTests`
- `swift test --filter UsageStoreWidgetSnapshotAccountTests`
- `swift test --filter WidgetSnapshotTests`
- `make check`
- autoreview: clean, no accepted/actionable findings

The first focused test attempt hit the documented local Sparkle framework rpath issue. After applying the worktree-local PackageFrameworks symlink, all three focused suites passed.

### DIFF
--- a/Sources/CodexBar/UsageStore+BackgroundRefresh.swift
+++ b/Sources/CodexBar/UsageStore+BackgroundRefresh.swift
@@ -50,6 +50,7 @@ extension UsageStore {
             self.kiloScopeSnapshots = []
         }
         if provider == .claude {
+            self.widgetUsagePreservationBlockedProviders.insert(provider)
             self.clearClaudeSwapAccountState()
         }
         self.clearTokenSnapshot(for: provider)
--- a/Sources/CodexBar/UsageStore+Refresh.swift
+++ b/Sources/CodexBar/UsageStore+Refresh.swift
@@ -556,6 +556,7 @@ extension UsageStore {
             }
             self.lastKnownResetSnapshots[provider] = backfilled
             self.snapshots[provider] = backfilled
+            self.widgetUsagePreservationBlockedProviders.remove(provider)
             if provider == .deepseek {
                 self.clearDeepSeekProfileTransition()
             }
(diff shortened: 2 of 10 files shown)


steipete:
Merged after full verification. Claude widget quota rows are preserved across partial refreshes (merging prior quota with fresh token history), now scoped to a stable OAuth profile identity.

The ownership hardening matters: the original keyed ownership as `nil` for default OAuth, and legacy snapshots are also `nil`, so `nil == nil` compared equal and preserved quota across what should be an ownership boundary. With profile-scoped credential caching in place (#2484), that meant one Claude profile's quota could display under another. Legacy ownerless snapshots now fail closed — a one-time stale widget row is far better than showing the wrong account's numbers.

Authored by @ChenZiHong-Gavin (#2480), recreated here only because the fork disallowed maintainer pushes for the fix. Exact-head CI green.

PR #2507, opened 2026-07-29
fix: retain normalized ZoomMate parent cookies

weddle:
## Summary

Fixes a live ZoomMate regression introduced by the host-partitioned cookie routing added during review of #2344.

SweetCookieKit normalizes Chromium parent-domain `.zoom.us` cookies to `zoom.us` before creating `HTTPCookie` values. The current partitioning logic interprets that bare value as host-only, dropping the shared Zoom SSO cookies for both approved API hosts and preventing any browser session from being imported.

This patch:

- treats the known shared `zoom.us` parent domain as valid for both `ai.zoom.us` and `zoommate.zoom.us`;
- keeps leaf-host cookies exact-host only;
- adds regression coverage for the normalized parent-domain form.

## Live reproduction

Using the same signed-in Chrome session:

| Build | Result |
|---|---|
| Pre-merge ZoomMate snapshot (`979fd6c`) | Successfully fetched credits |
| Current build | `No ZoomMate session is cached and no session cookies were imported from Chrome` |

## Validation

- `swift build -c debug --target CodexBarCore` passes.
- The focused test command is currently blocked by an unrelated existing compile error in `ClaudeCLIBackgroundAvailabilityTests` (missing `includePrepaidBalance` argument).

### DIFF
--- a/Sources/CodexBarCore/BrowserCookieAccessGate.swift
+++ b/Sources/CodexBarCore/BrowserCookieAccessGate.swift
@@ -345,6 +345,28 @@ extension BrowserCookieClient {
             throw error
         }
     }
+
+    public func codexBarRecords(
+        matching query: BrowserCookieQuery,
+        in store: BrowserCookieStore,
+        logger: ((String) -> Void)? = nil) throws -> [BrowserCookieRecord]
+    {
+        guard BrowserCookieAccessGate.cookieStoreAccessDecision(
+            homeDirectories: self.configuration.homeDirectories) == .allowed
+        else {
+            throw BrowserCookieStoreAccessSuppressedError()
+        }
+        guard BrowserCookieAccessGate.shouldAttempt(store.browser) else { return [] }
+        guard BrowserCookieAccessGate.claimExplicitRetryCookieReadIfNeeded(for: store.browser) else { return [] }
+        do {
+            let records = try self.records(matching: query, in: store, logger: logger)
+            BrowserCookieAccessGate.recordAllowed(for: store.browser)
+            return records
+        } catch {
+            BrowserCookieAccessGate.recordIfNeeded(error)
+            throw error
+        }
+    }
 }
 #else
 public enum BrowserCookieAccessGate {
--- a/Sources/CodexBarCore/Providers/ZoomMate/ZoomMateChromiumCookieScopeReader.swift
+++ b/Sources/CodexBarCore/Providers/ZoomMate/ZoomMateChromiumCookieScopeReader.swift
@@ -1,0 +1,247 @@
+import Foundation
+
+#if os(macOS)
+import SQLite3
+import SweetCookieKit
+
+private let zoomMateSQLiteTransient = unsafeBitCast(-1, to: sqlite3_destructor_type.self)
+
+enum ZoomMateChromiumCookieScopeReader {
+    enum CookieScope: Hashable, Sendable {
+        case domain
+        case hostOnly
+    }
+
+    struct ScopedCookie: Sendable {
+        let record: BrowserCookieRecord
+        let scope: CookieScope
+    }
+
+    struct CookieMetadata: Sendable {
+        let hostKey: String
+        let topFrameSiteKey: String
+        let name: String
+        let path: String
+    }
+
+    struct Source: Sendable {
+        let label: String
+        let cookies: [ScopedCookie]
+    }
+
+    enum ReadError: LocalizedError {
+        case missingDatabase(label: String)
+        case sqliteFailed(label: String, details: String)
+
+        var errorDescription: String? {
+            switch self {
+            case let .missingDatabase(label):
+                "\(label) has no Chromium cookie database."
+            case let .sqliteFailed(label, details):
+                "\(label) cookie scope read failed: \(details)"
+            }
+        }
+    }
+
+    private struct CookieKey: Hashable {
+        let domain: String
+        let name: String
+        let path: String
+    }
+
+    private struct ScopeResolution {
+        var scopes: Set<CookieScope> = []
+        var hasPartitionedRecord = false
+    }
+
+    private static let rawDomains = [
+        ".zoom.us",
+        "zoom.us",
+        ".ai.zoom.us",
+        "ai.zoom.us",
+        ".zoommate.zoom.us",
+        "zoommate.zoom.us",
+    ]
+
+    static func read(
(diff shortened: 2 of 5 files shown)


weddle:
## Live verification for `4acee6b2`

I verified this revision against a signed-in ZoomMate account using Chrome and
the packaged CodexBar CLI.

### Browser behavior

Chrome DevTools Protocol reported the following routing decisions for the real
request destinations:

| Chromium storage scope | `ai.zoom.us` | `zoommate.zoom.us` |
| --- | --- | --- |
| `.zoom.us` domain cookie | sent | sent |
| host-only `zoom.us` cookie | `DomainMismatch` | `DomainMismatch` |
| `.ai.zoom.us` domain cookie | sent | `DomainMismatch` |
| host-only `ai.zoom.us` cookie | sent | `DomainMismatch` |

A controlled replay to the login endpoint also showed that the `.zoom.us`
parent-domain group returned a NAK, while the AI-leaf group alone did not. This
confirms that the parent-domain cookies are material to the login bootstrap.

SweetCookieKit normalizes both `.zoom.us` and host-only `zoom.us` to `zoom.us`,
so the normalized domain cannot reproduce Chrome's routing by itself. This
revision reads the non-secret Chromium scope metadata from the same copied
database snapshot used for decryption, joins it to the decrypted records by
domain/name/path, and then applies RFC 6265 matching:

- host-only cookies require an exact host match;
- domain cookies allow the exact domain or a dot-boundary suffix;
- missing, ambiguous, or partitioned scope metadata fails closed;
- the final headers remain restricted to the two fixed ZoomMate request hosts.

### Automated verification

The focused suite passes:

```text
$ swift test --filter ZoomMateCookieScopeTests
3 tests passed
```

It covers parent-domain routing, host-only and leaf isolation, dot-boundary
matching, hostile suffix rejection, root-path enforcement, and fail-closed
handling for missing/ambiguous/partitioned metadata.

Current PR checks:

- [macOS test shard 0 — passed](https://github.com/steipete/CodexBar/actions/runs/30497420828/job/90729397086)
- [macOS test shard 1 — passed](https://github.com/steipete/CodexBar/actions/runs/30497420828/job/90729397081)
- [lint — passed](https://github.com/steipete/CodexBar/actions/runs/30497420828/job/90729363612)
- [Linux x64 — passed](https://github.com/steipete/CodexBar/actions/runs/30497420828/job/90729363609)
- [Linux musl — passed](https://github.com/steipete/CodexBar/actions/runs/30497420828/job/90729397047)
- GitGuardian — passed

The remaining [Linux ARM64
failure](https://github.com/steipete/CodexBar/actions/runs/30497420828/job/90729363652)
is in the unrelated `CLICardsClaudeSwapTests` fake-executable test
(`NSCocoaErrorDomain error 256`). The CLI release build succeeded on ARM64, and
the identical test passed on Linux x64. The implementation and focused tests in
this revision are macOS-gated.

### Packaged live CLI verification

```console
$ CodexBar.app/Contents/Helpers/CodexBarCLI usage \
    --provider zoommate --source web --format json
```

Sanitized result:

```text
exitCode: 0
provider: zoommate
source: web
loginMethod: Cookie
primary credits payload: present
```

The packaged bundle identified commit `4acee6b2`, and
`codesign --verify --deep --strict` passed. This confirms that the revised
importer can read the live Chrome session, complete the ZoomMate login
bootstrap, and fetch current credits.

For redaction: this comment includes no account identifier, usage amount,
cookie name or value, cookie header, bearer/NAK value, authorization header,
response body, or credential. The underlying CDP capture retained only routing
metadata (including cookie names and scopes); it never emitted or saved cookie
values, token values, authorization headers, or response bodies.

weddle:
@clawsweeper re-review

steipete:
Confirmed @weddle's diagnosis and followed the single cookie-access-path policy.

The exact loss happens between SweetCookieKit's Chromium reader and ZoomMate's destination partitioning. Chromium stores the shared SSO cookies (including the `_zm_*` family and `cf_clearance`) with `host_key = .zoom.us`. SweetCookieKit 0.5.1 normalizes that to `BrowserCookieRecord.domain = zoom.us`, and `HTTPCookie` also exposes the bare form. ZoomMate then interprets a bare domain as host-only, so those shared cookies are dropped for both `ai.zoom.us` and `zoommate.zoom.us`. A genuinely host-only `zoom.us` cookie becomes indistinguishable, so treating every normalized `zoom.us` record as shared would cross the cookie security boundary.

SweetCookieKit 0.5.1 does **not** retain or expose this scope. Its schema-v24 work validates and strips the encrypted value's domain hash, but `BrowserCookieRecord` still loses the leading-dot distinction.

I opened https://github.com/steipete/SweetCookieKit/pull/19 with the library-level fix. It adds explicit host-only/domain scope to normalized records, carries it through Safari, Chromium, and Gecko imports, and includes Chromium fixtures for `.zoom.us` versus host-only `zoom.us`. The change is commit https://github.com/steipete/SweetCookieKit/commit/7342f04c45dc7ef117cc810306b3ef774f868334 and credits Ryan's diagnosis/co-authorship.

Validation is green: SweetCookieKit focused tests, its full 69-test suite, strict lint, and autoreview; CodexBar `make check`, all focused ZoomMate suites, and `make test` also pass on current main.

What remains: merge and release SweetCookieKit (no tag/release was made here), bump CodexBar to that release, then update ZoomMate to use `BrowserCookieRecord.scope` for exact-host versus RFC domain matching and add the provider-level regression fixture. I am leaving #2507 open for that follow-through.

steipete:
The library-path fix is up in #2627. It bumps SweetCookieKit to 0.5.2 and uses the preserved browser cookie scope so parent `.zoom.us` cookies reach both ZoomMate API hosts without leaking host-only cookies. This issue will be closed when the PR merges.

steipete:
Fixed on main via the library path: SweetCookieKit 0.5.2 now preserves host-only vs domain scope on normalized records (SweetCookieKit#19), and #2627 bumps CodexBar and makes ZoomMate route by real scope — parent `.zoom.us` domain cookies reach both API hosts, host-only cookies stay host-only, with regression fixtures reproducing exactly the scenario you diagnosed.

Your diagnosis was precisely right; the only change from this PR's approach is where the fix lives — scope belongs in the shared cookie library rather than a provider-local Chromium reader, so every cookie provider benefits. Changelog credits you. Closing in favor of the landed pair; ships in 0.47.1. Thanks @weddle!

PR #2525, opened 2026-07-31
fix: preserve fork accuracy during bounded Codex cost catch-up

xx205:
## Summary

- Preserve accurate Codex fork and subagent accounting while large rollout files are scanned in bounded, resumable slices.
- Route fork-parent baseline work through the same persistent scanner path instead of requiring a full parent read inside the child resolver.
- Persist compact parent token snapshots, sparse checkpoints, append-validation anchors, partial JSONL state, and unresolved child events so later passes can resume without re-reading verified prefixes.
- Add background catch-up progress plus adaptive scheduling: 20% target duty cycle on AC power, 5% on battery, and automatic pauses for Low Power Mode or serious/critical thermal pressure.
- Add Spend Dashboard controls to finish indexing immediately, return it to background mode, or stop without discarding saved progress.
- Keep the last populated cost report visible and explicitly marked stale while an incompatible parser cache or explicit rebuild catches up.

## Problem

#2452 correctly replaced a permanent giant-file skip with resumable bounded parsing, but fork baseline resolution still has a separate whole-parent admission path.

When a child references a parent larger than the 256 MiB per-file slice, the resolver cannot admit the full parent and returns an unresolved baseline. The unresolved fallback skips the first cumulative total and only counts later positive deltas, which can permanently omit child usage already present in that first total. Re-reading a smaller parent directly inside the resolver also bypasses the ordinary persistent catch-up path and can extend a foreground refresh.

### Accurate, resumable fork baselines

- Parent files discovered from children are queued into the same `CodexScanBudget` and cache path as ordinary rollout files.
- Each parent stores compact token snapshots plus sparse accumulator checkpoints.
- Timestamp-monotonic indexes use binary search and early stop; non-monotonic files retain the existing ordered fallback.
- A SHA-256 prefix anchor validates append-safe reuse. Replacement or prefix rewrite invalidates incompatible progress.
- Ordinary forks retain compact unresolved token lines and publish no inherited usage until the parent baseline is available.
- After a parent becomes complete, buffered children are replayed in the same refresh without re-reading their JSONL.
- Parents outside the requested history window remain dependency work and are retained until their children resolve.

### Bounded scanner work

- The initial app refresh and every subsequent automatic catch-up pass install a 2-second JSONL parser deadline in addition to the existing 256 MiB per-file and 512 MiB per-refresh byte budgets. Direct library/CLI callers keep their existing behavior unless they explicitly configure a deadline.
- The reader checks deadline/cancellation at least every 256 KiB and persists state even when a slice ends inside an oversized JSONL record.
- Progress is stored as processed/total bytes and completed/total files.
- Newest rollouts remain first, so recent usage becomes visible before historical catch-up completes.

### Adaptive catch-up and UI

- Automatic mode targets 20% active time on AC and 5% on battery.
- Low Power Mode and serious/critical thermal states pause new automatic passes.
- Accelerated mode uses one scanner task at utility priority and removes duty-cycle sleeps; it does not add parallel file parsing.
- Stop is graceful at the bounded-pass boundary and keeps the saved offset.
- Spend Dashboard shows byte/file progress and controls for accelerated, background, stop, and resume states.
- Parser-producer upgrades and explicit force rebuilds retain the last populated report with its original freshness timestamp until the bounded rebuild completes.
- The retained report is presentation-only: parser-dependent offsets and totals are never reused, and ordinary warm-cache incremental refreshes are not frozen.

## Real-corpus validation

A privacy-safe validation selected a stable child/parent fork pair from a real local corpus. The parent was 488,985,173 bytes, so it could not fit in one 256 MiB slice. The copied files used anonymous names and the report emitted no paths, session IDs, dates, model names, token totals, prompts, responses, or session content.

| Check | Unbounded reference | Bounded scanner |
|---|---:|---:|
| Files / total bytes | 2 / 489,147,332 | 2 / 489,147,332 |
| Passes | 1 | 21 |
| Scanner run time | 53.860 s | 119.172 s |
| Unresolved buffers after completion | — | 0 |
| Incomplete files after completion | — | 0 |
| `CostUsageDailyReport.data` | reference | exact match |
| `CostUsageDailyReport.summary` | reference | exact match |

The bounded passes ran back-to-back without automatic scheduler sleeps. The 2-second deadline covers JSONL reading/parsing; cache load/save and aggregation remain outside that parser deadline.

## Upgrade-cache convergence validation

A separate privacy-safe synthetic fixture seeded a populated cache under an incompatible parser producer, then rebuilt an oversized parent/child fork corpus with 1 KiB per-file and per-refresh slices.

- The prior daily report and summary remained visible on every pending pass.
- Catch-up status carried the prior cache timestamp so the UI can label the report as stale.
- No old parser offsets, file totals, or accumulator state were reused.
- The final bounded report exactly matched an unbounded reference.
- The stale report was removed as soon as catch-up completed.
- The existing warm-cache active/archive overlap regression passed in the full suite and in three repeated isolated runs, confirming ordinary incremental refreshes remain immediate.

## Verification

- `make test`
  - 760 test selections in 64 groups;
  - 64/64 groups passed on the first attempt;
  - 0 failures, retries, or timeouts;
  - 936.2 seconds total.
- Focused cache/fetcher/catch-up run:
  - 53 tests across 5 suites passed.
- Upgrade transition regressions:
  - incompatible-cache bounded convergence passed;
  - ordinary warm-cache overlap passed in the full suite and 3/3 isolated repeats.
- Fresh Debug package build:
  - SwiftPM and Xcode compilation were both capped at 2 jobs;
  - the signed app bundle passed deep validation and opened the Usage & Spend settings pane.
- `make check`
  - parser hash, package checks, locale coverage, SwiftFormat, and strict SwiftLint passed;
  - 22 localized catalogs match English.
- Privacy-safe real-corpus fork validation:
  - passed in 173.866 seconds;
  - bounded output exactly matched the unbounded reference.
- Freshly built and signed Debug app:
  - automatic/background mode showed live byte progress with `Finish now` and `Cancel`;
  - accelerated mode showed live byte progress with `Continue in background` and `Cancel`.

## Screenshots

Privacy-safe Spend Dashboard screenshots for automatic/background and accelerated indexing were captured from the freshly built app. They contain no local paths, account identity, session IDs, prompts, responses, or session content.

Automatic/background catch-up:

Spend Dashboard automatic background indexing

Accelerated catch-up:

Spend Dashboard accelerated indexing

## Scope and limitations

- Account, rate-limit, and remote provider probing are unchanged.
- Automatic mode remains intentionally slower than accelerated mode.
- The 2-second parser deadline does not yet bound directory enumeration or full cache decode/encode/write.
- Progress totals can grow when a live file appends or a parent outside the original window is discovered.
- This validation covers a real 466 MiB oversized-parent case, not a full 200–276 GiB cold-corpus benchmark; the PR does not promise a fixed completion ETA.


steipete:
Completed the #2538 CPU-spin catch-up on top of @xx205's PR in `7fe0534527`.

The cost cache now owns a persisted, resumable session-discovery index. Missing parent IDs are stored with an inventory generation derived from the relevant roots/partitions, so a cached unresolved child reuses the same negative dependency key and remains idle until that inventory changes. Root traversal and session-head reads reserve and consume the existing `CodexScanBudget`; partial head reads and directory traversal resume from the on-disk state instead of restarting. Creating the parent or changing the root/partition inventory rotates the generation and replays the child's buffered events once.

I also included the bounded mechanism-1 follow-up from #2538/#2558: newest-first sorting snapshots each file's metadata once, eliminating `fstatat` calls from the comparator.

Regression proof:

- 250-file synthetic corpus: the cold absent-parent lookup scans once; the second warm refresh performs **0 session-head parses**.
- 32-byte discovery budget: head parsing resumes across cache round trips and eventually records the negative lookup.
- Creating the parent invalidates the miss, publishes buffered child usage, and a following refresh remains bit-for-bit stable.
- Adding a partition file rotates the discovery generation while reparsing at most the new head.
- `CostUsagePerformanceGateTests`: 24/24 passed.
- `CostUsageScannerBreakdownTests`: 95/95 passed, including out-of-window symlink parents.
- Focused fork/cache/JSONL suites passed.
- `make check`: clean.
- `make test`: all 761 selections passed in 64/64 groups; zero retries or timeouts.
- Autoreview: `--mode local`, clean with no accepted/actionable findings.

The 0.46.1 changelog credits @xx205 and the #2538 investigation by @Helmi and @kiranmagic7. This now covers both identified refresh-amplification mechanisms without merging the PR.

steipete:
Refreshed PR #2525 onto current `main` in merge commit `a0d9d52f5a` (parents `7fe0534527` and `5b0b9fa3e3`).

- Reconciled the only textual conflict by regenerating `CodexParserHash` from the combined source (`aa0b0865c496e548`); #2519's Fast-mode `Submission` SQLite filters/parser and #2525's persisted fork-parent discovery/budget path are both intact. The changelog entry now sits in the existing 0.46.1 `### Fixed` section.
- The earlier Linux arm64 failure was not the SDK exit-11 flake and not scanner-related: `CLICardsClaudeSwapTests` failed to launch an atomically-written temporary executable. Current `main` contains the portable non-atomic `Data.write` fixture fix, and that test passed after the merge.
- Proof: `make check` clean; 180 focused scanner/performance/fork/cache/JSONL/#2519 tests passed; `make test` groups 1–35 passed, group 36 hit only the explicitly waived #2509 Kiro wall-clock race, and groups 37–66 were run individually with zero failures.
- Autoreview initially refused unrelated PNGs inherited from `main`; reviewing the exact merged tree against `origin/main` via review-only commit `b493ee77aad1` produced one 319 KB pass and finished clean with no accepted/actionable findings.

No merge of the PR was performed.

PR #2528, opened 2026-07-31
fix(cookies): let background refresh reuse an already-authorized Safe Storage ACL (#2451)

### LINKED ISSUE #2451: Background Chromium cookie import is skipped without preflighting, even when the Safe Storage ACL already allows it
CodexBar refuses to import Chromium cookies on any background refresh, even when the Chrome Safe Storage keychain item already grants it access. Cookie-backed providers therefore only update when you open the menu.

`BrowserCookieAccessGate.shouldAttempt()` returns false for any non-user-initiated fetch before it ever checks whether a prompt would actually occur:

```swift
// Sources/CodexBarCore/BrowserCookieAccessGate.swift:86
public static func shouldAttempt(_ browser: Browser, now: Date = Date()) -> Bool {
    guard browser.usesKeychainForCookieDecryption else { return true }
    guard !KeychainAccessGate.isDisabled else { return false }
    guard ProviderInteractionContext.current == .userInitiated else {
        self.log.info(
            "Skipping background Chromium cookie import to avoid a Keychain prompt",
            metadata: ["browser": browser.displayName])
        return false
    }
```

The check that would answer "would this actually prompt?" already exists in the same file. `chromiumKeychainRequiresInteraction(for:)` at line 272 calls `KeychainAccessPreflight.checkGenericPassword` and returns false when the item comes back `.allowed`, which is the case for anyone who has added CodexBar to the Chrome Safe Storage access list. On my machine both CodexBar and CodexBarCLI are in that list, the cookies are `v10`, and a user-initiated fetch decrypts them without any dialog. Background fetches still never try.

Two consequences I hit:

Claude on the web source only refreshes when the menu is opened, so it sits stale in between and has to be refreshed by hand.

OpenCode Go silently falls back to the `opencodego.local` probe, which reads `~/.local/share/opencode/opencode.db` and only knows what that one machine did. It showed 0.5% weekly while the account was actually around 60%. The UI gives no indication that the number came from local logs instead of the account, so it reads as a plain wrong number rather than a fallback.

To reproduce: sign into claude.ai in Chrome, set the Claude usage source to Web API, then run `codexbar usage --provider claude --source web -v --log-level debug`. The CLI runs in background interaction context, so it logs the skip and reports "No Claude session key found in browser cookies" even though the cookie is there.

Suggested change: in `shouldAttempt`, when the interaction is not user-initiated, call `chromiumKeychainRequiresInteraction(for:)` and allow the import when it returns false, keeping the hard skip for `.interactionRequired`. That preserves the no-surprise-prompt behaviour and lets background refresh work for anyone who has already granted access.

Workaround for anyone else running into this: Safari, Firefox and Zen return false from `usesKeychainForCookieDecryption` (`BrowserCookieImportOrder.swift:37`), so they bypass the gate entirely. Signing into the provider in one of those keeps background refresh working, at the cost of using a browser you may not otherwise open.

CodexBar 0.45.2, macOS 26.5.2, repo at cc8da27.

LeoLin990405:
Fixes #2451.

## Problem

`BrowserCookieAccessGate.shouldAttempt` returns `false` for **every** non-user-initiated Chromium request before it ever reaches the existing no-UI Keychain preflight:

```swift
guard ProviderInteractionContext.current == .userInitiated else {
    // ...logs and returns false, before chromiumKeychainRequiresInteraction() ever runs
    return false
}
```

So even when a browser's Safe Storage entry is **already authorized** (the ACL grants access without any prompt), a scheduled/background refresh can't use it — cookie-backed usage only updates when the menu is opened. The gate short-circuits *before* the preflight that would have told it the read is safe.

## Fix

Move the background-context decision until **after** the same strictly no-UI Safe Storage preflight the user-initiated path already relies on (`KeychainAccessPreflight.checkGenericPassword`, which uses `kSecUseAuthenticationUIFail` + a non-interactive `LAContext`, so it can never surface a prompt):

- **`.allowed`** → the ACL already grants access with no prompt → background read proceeds.
- **`.interactionRequired`** → would prompt → skip (no-surprise boundary preserved).
- **`.notFound` / `.failure`** → not an explicit grant → skip. (Stricter than the user-initiated path, which is optimistic here — a background refresh should never take a chance.)
- An active **per-browser or Chromium-family denial cooldown** is still honored, and is checked **before** the preflight so a suppressed family never probes (this keeps the existing family-suppression test's preflight-count invariant intact).

The user-initiated path is byte-for-byte unchanged. Because the fix reuses the *same* no-UI preflight the interactive path already trusts, it introduces no new Keychain-prompt surface — it just stops discarding the already-authorized case in the background.

## Tests

- Two `BrowserDetectionTests` cases were characterization tests that pinned the old *"skip before preflight"* behavior (one was literally named `background cookie import skips chromium before keychain preflight` and asserted `preflightCount == 0`). Updated to the fixed behavior and renamed.
- New `BrowserCookieAccessGateTests` covers the background matrix: allowed → proceed, interaction/not-found/failure → skip, an active cooldown suppresses even an allowed ACL (and lapses after 6h), user-initiated unchanged, and non-Keychain browsers (Safari) bypass the gate.

Acceptance criteria:

- `swift test --filter BrowserCookieAccessGateTests` ✅ (6/6)
- All `shouldAttempt`-touching suites green under the by-suite (`--no-parallel`) runner (288 tests / 15 suites).
- `./Scripts/lint.sh lint` clean (swiftformat + swiftlint --strict).

I also reverse-verified the guard: reverting only the source change makes the new "background proceeds on `.allowed`" test fail, confirming it pins the fix rather than tautologically passing.

Per `AGENTS.md`, verification stays on the existing no-UI preflight override and never reads the real login Keychain.

## Out of scope

The issue also floats surfacing the selected source/fallback in the UI when a provider shows local data instead of account usage. That's a separate rendering concern; this PR is scoped to the gate defect + tests. Happy to follow up if you'd like it in the same issue.

---

## Real-behavior proof (macOS, real Chrome Safe Storage)

Ran the fixed gate against the **real** `Chrome Safe Storage` login-keychain item on macOS, in a genuine `.background` interaction context, via the strictly no-UI preflight (`kSecUseAuthenticationUIFail`, attributes-only — cannot show a prompt):

```
preflight   service="Chrome Safe Storage" account="Chrome"  ->  allowed
background       shouldAttempt(.chrome)  =  true     # ← the fix; pre-fix this was false
userInitiated    shouldAttempt(.chrome)  =  true     # unchanged
```

- **Already-authorized ACL → background proceeds**, prompt-free: on real Keychain state reporting `.allowed`, the background gate now returns `true`; pre-fix it returned `false` for every non-user-initiated request regardless of the ACL.
- **No secret touched**: attributes-only preflight, never `kSecReturnData`; reuses the existing no-UI seam per `AGENTS.md`.
- Reverse-verified: reverting only the source change makes the "background proceeds on `.allowed`" test fail. Full write-up in the PR comment thread.

### DIFF
--- a/Sources/CodexBarCore/BrowserCookieAccessGate.swift
+++ b/Sources/CodexBarCore/BrowserCookieAccessGate.swift
@@ -87,10 +87,7 @@ public enum BrowserCookieAccessGate {
         guard browser.usesKeychainForCookieDecryption else { return true }
         guard !KeychainAccessGate.isDisabled else { return false }
         guard ProviderInteractionContext.current == .userInitiated else {
-            self.log.info(
-                "Skipping background Chromium cookie import to avoid a Keychain prompt",
-                metadata: ["browser": browser.displayName])
-            return false
+            return self.shouldAttemptInBackground(browser, now: now)
         }
         if self.deniedBrowsersForTesting?.contains(browser) == true {
             return self.isExplicitRetryAllowed(for: browser)
@@ -277,6 +274,69 @@ public enum BrowserCookieAccessGate {
                 return false
             case .interactionRequired:
                 return true
+            case .notFound, .failure:
+                continue
+            }
+        }
+        return false
+    }
+
+    /// Background (non-user-initiated) refreshes must never surface a Keychain prompt. Rather than
+    /// skipping unconditionally, reuse the strictly no-UI Safe Storage preflight: when the ACL already
+    /// grants access (an explicit `.allowed`), a scheduled refresh can read cookies without any prompt,
+    /// so background usage stays in sync instead of only updating when the menu is opened. Anything else
+    /// — interaction required, not found, or a query failure — keeps the no-surprise boundary and skips.
+    /// An active per-browser or Chromium-family denial cooldown is still honored.
+    private static func shouldAttemptInBackground(_ browser: Browser, now: Date) -> Bool {
+        if self.deniedBrowsersForTesting?.contains(browser) == true {
+            return false
+        }
+        if self.hasActiveDenialCooldown(for: browser, now: now) {
+            self.log.debug(
+                "Skipping background Chromium cookie import; denial cooldown active",
+                metadata: ["browser": browser.displayName])
+            return false
+        }
+        guard self.chromiumKeychainAccessIsAllowed(for: browser) else {
+            self.log.info(
+                "Skipping background Chromium cookie import to avoid a Keychain prompt",
+                metadata: ["browser": browser.displayName])
+            return false
+        }
+        self.log.debug(
+            "Background Chromium cookie import allowed by no-UI Keychain preflight",
+            metadata: ["browser": browser.displayName])
+        return true
+    }
+
+    /// Read-only check for an active per-browser or Chromium-family denial cooldown. Mirrors the
+    /// suppression window enforced on the user-initiated path without mutating persisted state, so a
+    /// scheduled refresh stays side-effect free.
+    private static func hasActiveDenialCooldown(for browser: Browser, now: Date) -> Bool {
+        self.lock.withLock { state in
+            self.loadIfNeeded(&state)
+            if let blockedUntil = state.deniedUntilByBrowser[browser.rawValue], blockedUntil > now {
+                return true
+            }
+            if let familyBlockedUntil = state.chromiumFamilyDeniedUntil, familyBlockedUntil > now {
+                return true
+            }
+            return false
+        }
+    }
+
+    /// Returns true only when the no-UI Safe Storage preflight explicitly reports `.allowed` for one of
+    /// the browser's labels before any label requires interaction. Symmetric with
+    /// `chromiumKeychainRequiresInteraction`; `.notFound`/`.failure` are skipped and never treated as a
+    /// grant, so only an already-authorized ACL enables a background read.
+    private static func chromiumKeychainAccessIsAllowed(for browser: Browser) -> Bool {
+        let labels = browser.safeStorageLabels.isEmpty ? self.safeStorageLabels : browser.safeStorageLabels
+        for label in labels {
+            switch KeychainAccessPreflight.checkGenericPassword(service: label.service, account: label.account) {
+            case .allowed:
+                return true
+            case .interactionRequired:
+                return false
             case .notFound, .failure:
                 continue
             }
--- a/Tests/CodexBarTests/BrowserCookieAccessGateTests.swift
+++ b/Tests/CodexBarTests/BrowserCookieAccessGateTests.swift
@@ -1,0 +1,109 @@
+import Foundation
+import Testing
+@testable import CodexBarCore
+
+#if os(macOS)
+import SweetCookieKit
+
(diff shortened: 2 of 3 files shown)


LeoLin990405:
CI note: the only real failure is `build-linux-musl-cli`, which flaked at the **"Install Swift Static Linux SDK"** step (SDK download) before compiling anything. `lint-build-test` then trips solely because musl is a required job — its own log says:

> Lint and macOS Swift test shards passed.
> Linux musl build gate/result mismatch: required=true result=failure

This change is entirely `#if os(macOS)`-gated (plus test files), so it can't affect the Linux/musl build — both `build-linux-cli` (arm64 + x64) jobs, lint, and both macOS test shards are green. A re-run of the musl job should clear it; I don't have rerun rights on the repo. Thanks!

LeoLin990405:
## Real-behavior proof (macOS, real Chrome Safe Storage)

Ran the fixed gate against the **real** `Chrome Safe Storage` login-keychain item on macOS, in a genuine `.background` interaction context (the same context the scheduled app refresh uses). The strictly no-UI preflight — `kSecUseAuthenticationUIFail` + a non-interactive `LAContext`, attributes-only (never `kSecReturnData`), so it physically cannot show a prompt — reports the ACL is already authorized, and the background gate now proceeds:

```
preflight   service="Chrome Safe Storage" account="Chrome"  ->  allowed
background       shouldAttempt(.chrome)  =  true     # ← the fix; pre-fix this was false
userInitiated    shouldAttempt(.chrome)  =  true     # unchanged
```

What this demonstrates on real Keychain state:

- **Already-authorized ACL → background proceeds.** With the real ACL reporting `.allowed`, `shouldAttempt(.chrome)` in `.background` now returns `true`. Pre-fix (the early return *before* the preflight) it returned `false` for every non-user-initiated request regardless of the ACL — that is exactly the bug this PR fixes.
- **Prompt-free.** The whole decision runs through the existing no-UI preflight; the check returned instantly with no Keychain dialog. This is the "confirm the real supported browser-label path remains prompt-free" item: on an already-authorized ACL the read is granted without UI, and any non-`.allowed` result is skipped rather than prompted.
- **No secret touched.** The preflight reads generic-password *attributes* only, never the secret payload, so no cookie/secret value is read or logged. Per `AGENTS.md`, this reuses the existing no-UI seam rather than a real secret read.

### Logic coverage (deterministic, no real Keychain)

`BrowserCookieAccessGateTests` (serialized) covers the full background matrix by stubbing the preflight outcome:

| preflight (background) | result |
|---|---|
| `.allowed` | proceed |
| `.interactionRequired` | skip |
| `.notFound` / `.failure` | skip |
| `.allowed` **but** active denial cooldown | skip (cooldown honored) |
| `.allowed` (user-initiated) | proceed — unchanged |
| non-Keychain browser (Safari) | bypass gate |

I also reverse-verified the guard: reverting only the source change makes the "background proceeds on `.allowed`" case fail, confirming the test pins the fix rather than passing tautologically.

CI is green (lint, both macOS `swift-test` shards, all Linux/musl builds).

PR #2536, opened 2026-07-31
Add `codexbar hooks watch` so hooks fire without the macOS app

### LINKED ISSUE #2535: [BUG] Hooks can be configured on Linux/headless but can never fire
Hook rules are configurable everywhere, but `HookRunner.dispatch` is called from exactly one place in the tree — `Sources/CodexBar/UsageStore+Hooks.swift`, the macOS app.

On Linux, headless, or a server install that means a user can enable hooks, add rules, see them in `codexbar hooks list`, and none of them will ever fire.

What exists today on the CLI:

- `hooks list` / `enable` / `disable` — configuration only.
- `hooks test <event>` — fires a *synthetic* event on demand, not a real transition.
- `serve` — its `--refresh-interval` is a cache TTL that bounds staleness when someone asks; there is no background poll, so nothing observes a transition.

So `Sources/CodexBarCore/Hooks/` (HookEvent, HookRule, HookRateLimiter, HookRunner) is fully implemented and unreachable off macOS.

Repro on Linux:

```
codexbar hooks enable
# add a quota_low or refresh_failed rule to the config
codexbar hooks list     # rule is present and enabled
# ...nothing ever runs it
```

`hooks test` runs the command, which makes it look wired up — but that path only ever fires the sample event.

Proposal: a `codexbar hooks watch` subcommand that polls the enabled providers, detects real quota/status transitions, and dispatches through the existing `HookRunner`. Edge-triggered against the previous poll so a persisting condition does not re-fire, in-memory baselines matching the app (first sample establishes state without firing), and read-only like `guard` so it never prompts for credentials.

Happy to send a PR — I have it building and tested on Linux. Wanted to file the gap first in case you would rather solve it a different way (e.g. wiring the app's detection into Core and sharing it, which I deliberately avoided so this change cannot regress macOS hook behavior).

OfficialAbhinavSingh:
Closes #2535.

## Problem

`HookRunner.dispatch` is called from exactly one place in the tree — `Sources/CodexBar/UsageStore+Hooks.swift`, the macOS app. So on Linux/headless a user can enable hooks and add rules, and none of them can ever fire. `hooks test` only fires a synthetic event on demand, and `serve --refresh-interval` is a cache TTL rather than a poller, so nothing observes a transition.

That leaves `Sources/CodexBarCore/Hooks/` fully implemented and unreachable off macOS.

## Change

**`HookTransitionDetector` (CodexBarCore)** — platform-neutral, side-effect free: it decides *what fired* and never fetches or runs commands. No new math; it uses what Core already ships (`QuotaLowHookThreshold.crossedRules`, `QuotaWarningWindow`, `HookRule`, `HookRateLimiter`).

It preserves the app's deliberate semantics:

- first sample of a lane establishes a baseline without firing (`UsageStore+Hooks.swift:145-147`)
- `unknown` / `maintenance` never flip provider status, so a hiccuped probe cannot fire an outage hook (`:79-88`)
- a rule change clears baselines, so a re-enable cannot fire for a crossing that happened while hooks were off (`:189-194`)
- disappearing lanes are pruned so a reappearance starts fresh (`:198-210`)

Everything is edge-triggered, which is what `HookRunner` already assumes when it declines to rate-limit quota events ("quota events already dedupe upstream").

**`codexbar hooks watch` (CodexBarCLI)** — drives it. `--interval` (default 300s, minimum 60s), `--provider` (repeatable), `--once`, JSON output. Read-only like `guard`, so it never prompts for credentials. Stops on SIGINT/SIGTERM/SIGHUP via the existing `CLITerminationSignalMonitor`.

**The macOS app is intentionally untouched.** The crossing math was already shared; only the thin wiring is macOS-only. Rewiring `UsageStore` would be architecturally cleaner but cannot be verified without a Mac, in a feature whose failure mode is running the wrong shell command. Noted as a follow-up below.

## Proof (Linux, `swift:6.3.3`)

A hook command actually executing headlessly — the thing that was impossible before:

```
$ codexbar hooks watch --once --provider codex
refresh_failed codex status=error

$ cat /tmp/hook.log
HOOK RAN event=refresh_failed provider=codex status=error
  stdin: {"event":"refresh_failed","provider":"codex","status":"error","timestamp":"2026-07-31T15:13:01Z"}
```

Correct `CODEXBAR_*` environment and correct JSON stdin payload.

JSON mode:

```
$ codexbar hooks watch --once --provider codex --json
{"event":"refresh_failed","provider":"codex","timestamp":"2026-07-31T15:13:48Z","status":"error"}
```

Argument validation runs before configuration state, so a bad `--interval` reports the real problem instead of being masked by disabled hooks:

```
$ codexbar hooks watch --interval 5 --once
--interval must be at least 60 seconds.
$ codexbar hooks watch --interval abc --once
Invalid --interval value: abc
$ codexbar hooks watch --provider nosuchprovider --once
Unknown provider: nosuchprovider
```

The floor is rejected rather than clamped: unlike `serve`'s cache TTL, each tick originates traffic against every selected provider.

Tests, format, lint:

```
$ swift test --filter HookTransitionDetectorLinuxTests
✔ Test run with 18 tests in 1 suite passed after 0.002 seconds.

$ swift test
✔ Test run with 335 tests in 49 suites passed after 0.397 seconds.

$ swiftformat --lint <changed files>
0/4 files require formatting.

$ swiftlint --strict
Done linting! Found 0 violations, 0 serious in 1659 files.
```

The 18 detector tests cover each invariant individually: first-sample-no-fire, crossing fires once and does not re-fire while persisting, `quota_reached` upward edge only, reset via boundary advance and via usage drop, `unknown`/`maintenance` ignored, outage→recovery pair, config change clears baselines, per-account independence, synthetic-placeholder lanes ignored, disappearing lane resets, and a refresh failure not disturbing quota baselines.

## Notes

- `refresh_failed` reports a coarse category (`timeout`, `offline`, `auth_required`, `network_error`, `cancelled`, `error`). The raw provider error is never forwarded, since it can embed response-body previews.
- A failed refresh returns early without touching baselines, so the next good poll compares against the last real sample instead of firing a phantom transition.
- Extracted the `hooks` descriptor into its own function in `CLIEntry`; adding the subcommand inline pushed `commandDescriptors()` past the `function_body_length` limit.

## Follow-ups (deliberately out of scope)

- `--state-file` for baselines that survive a restart.
- Unify the macOS app's hook wiring onto `HookTransitionDetector`.
- `ProviderStatusIndicator` is currently duplicated between `Sources/CodexBar/UsageStoreSupport.swift:4` and `Sources/CodexBarCLI/CLIPayloads.swift:121`.


OfficialAbhinavSingh:
**Re: `build-linux-musl-cli` red** — traced the log: it's a `Signal 11` crash *inside* `swift sdk install` itself (SwiftPM's static-SDK download step, in `libcurl`→`libssl`→`libcrypto`→`URLSession._MultiHandle.deinit` teardown), not a build or test failure. `Package.swift` is untouched in this PR. `lint-build-test` only failed as the aggregator (`LINUX_MUSL_RESULT: failure` in its own summary) — everything else, including both `swift-test-macos` shards, is green. Looks like a toolchain-download flake; should clear on rerun.

**Addressed all 3 Codex review findings in `b21b6950`**, each verified and each with a regression test:

- **Filtered `quota_low` dispatch** — you were right, dispatching against the full config meant `HookRunner`'s `matches()` re-check (`usage >= threshold`) re-fired any rule whose threshold sat below the one that actually crossed. `HookTransitionDetector.evaluate` now returns `HookDispatch` (event + optional rule subset); `quota_low` carries the pre-computed crossed rules, mirroring the app's own `HooksConfig(events: crossed)` pattern.
- **`quota_reached` restricted to the session lane** — confirmed against `docs/configuration.md:82` ("the primary session quota crosses into depletion") and the app's `emitQuotaReachedHook(sessionWindow:)` call site. Weekly depletion now only produces `quota_reset`.
- **Interruptible sleep** — confirmed `CLITerminationSignalMonitor` only flips a flag, doesn't cancel the running task. Replaced the single long `Task.sleep` with a 200ms-tick loop (`sleepInterruptibly`) that checks the stop flag between ticks, so SIGINT/SIGTERM/SIGHUP is honored promptly instead of after the full interval.

Full suite: 340/340 (up from 335, +5 new tests), `swiftformat --lint` clean, `swiftlint --strict` 0/1659.

OfficialAbhinavSingh:
Fixed the P2 in `6d361708`.

You were right, and my earlier reorder didn't actually close it — I moved the argument decoding above the `hooks.enabled` checks but left it *below* `loadConfig`, which is the call that exits on a malformed config. So `hooks watch --interval 5` against a broken config still reported the decode failure.

Rather than reorder again, I removed the coupling: `decodeHooksWatchProviderNames` now validates explicit `--provider` names with no config at all and returns `nil` to mean "fall back to the enabled set", and a new `hooksWatchProviders` resolves that fallback once the config is loaded. Both command-only decoders run before `loadConfig`, so the ordering defect is structurally impossible rather than corrected by hand.

Verified end-to-end against a deliberately malformed config:

```
### malformed config + --interval 5
--interval must be at least 60 seconds.

### malformed config + --interval abc
Invalid --interval value: abc

### malformed config + unknown provider
Unknown provider: nosuchprovider

### malformed config + VALID args (config error must still surface)
Error: Failed to decode CodexBar config: The operation could not be completed. The data isn't in the correct format.
```

That last case is the negative control: the fix is scoped to argument precedence and does not swallow genuine config errors.

Added `CLIHooksWatchArgumentPrecedenceLinuxTests` (9 tests) covering the precedence regression you asked for, including the config-free decoding paths and the enabled-set fallback. Full suite 349/349, `swiftformat --lint` clean, `swiftlint --strict` 0/1659.

On the **P1 product decision** — that one is genuinely yours to make, and I flagged the same concern when filing #2535. Two things that may help scope it: `watch` is opt-in and does nothing unless a user has already enabled hooks and configured rules, and `--interval` has a 60s floor that is rejected rather than clamped (default 300s), so it cannot be turned into an unattended hammer by a typo. `--once` also covers anyone who would rather drive it from cron/systemd than run a long-lived process. Happy to narrow the scope further, gate it behind a flag, or drop the polling loop entirely and ship only `--once` if you'd prefer the smaller surface.

steipete:
Gate update: merged current `origin/main` and removed the stateless one-shot mode from `hooks watch`.

The reason is state ownership, not scheduler preference: every invocation creates a fresh transition detector and rate limiter, and the first observation intentionally establishes a baseline without firing. Repeated cron/systemd-style one-shots therefore cannot observe quota/status edges or preserve event throttling across polls. The sound path is now the only path: one continuous `codexbar hooks watch` process. `--once` was removed from the Commander signature, runtime loop, help, docs, examples, and changelog; the parser now rejects it before config/provider access. The Unreleased entry credits @OfficialAbhinavSingh.

Proof on head `b29225ff9f`:

- `swift test --filter 'HookTransitionDetector|CLIHooksWatch'` — 33 tests passed across the detector, argument parsing, and interruptible sleep suites
- `make check` — passed; SwiftFormat clean, SwiftLint 0 violations in 1,691 files
- black-box CLI: `hooks --help` documents continuous state lifetime and contains no `--once`; `hooks watch --once` exits 1 with `Unknown option --once`
- structured Codex autoreview against `origin/main` — clean, no accepted/actionable findings

Linux-native coverage remains delegated to CI. No PR was merged.

steipete:
Final moving-main follow-up: `main` advanced during the long CI window and briefly made this PR conflicting. I merged the new tip (`aeab80e099`), resolved the CHANGELOG-only conflict by preserving both Unreleased entries, and pushed final head `90a49a5f8`.

Re-proof after that merge:

- `swift test --filter 'HookTransitionDetector|CLIHooksWatch'` — 33 tests passed
- `make check` — passed; SwiftFormat clean, SwiftLint 0 violations in 1,699 files
- structured Codex autoreview against the new `origin/main` — clean
- replacement run `30772591632` — fully green across Linux x64/arm64/musl, both macOS shards, lint, GitGuardian, and `lint-build-test`

GitHub reports the final head mergeable. No PR was merged.

PR #2549, opened 2026-08-01
Add Notion AI provider

n0ah37:
## Summary

Adds a **Notion AI** provider that tracks the two usage-allowance windows Notion shows in **Settings → Notion AI → Usage**:

- **Rolling** — the 6-hour window, with its reset countdown
- **Monthly** — the billing-period window, with its period end

Notion starts enforcing this allowance on **August 3, 2026**. Before that date the same endpoint already returns real usage numbers with `"enforcement": "preview"`, so the gauges are accurate either way — which is why it seemed worth having in the menu bar now rather than after people start hitting the cap.

No existing issue or PR covers Notion, so this is a fresh integration rather than a follow-on.

## Data source

Two cookie-authenticated POSTs to `app.notion.com` per refresh, both endpoints the Notion web app itself calls:

| Endpoint | Purpose |
| --- | --- |
| `/api/v3/getSpaces` | Resolves the signed-in account (email, name) and its workspaces, including `plan_type` and `subscription_tier`. Drives automatic workspace selection and the identity line. |
| `/api/v3/getCreditRateLimitStatus` | The allowance itself, for one `spaceId`. |

```json
{
  "status": "within_limit",
  "window": { "creditType": "basic_ai_credits", "scope": "per_user", "window": "6h", "used": 85.31, "limit": 100 },
  "resetsInSeconds": 20834,
  "billingPeriodWindow": {
    "creditType": "basic_ai_credits", "scope": "per_user", "cadence": "billing_period",
    "used": 16.89, "limit": 100, "periodEndMs": 1787342274000
  },
  "enforcement": "preview"
}
```

`window` → `primary` (window length parsed from the `6h` token, reset from `resetsInSeconds`), `billingPeriodWindow` → `secondary` (reset from `periodEndMs`). Usage is scaled against the returned `limit` rather than assumed to be a percentage, so a future non-100 limit keeps working; over-quota values are preserved for downstream display clamping.

## Behaviour notes

- **Plan gating.** Only Business and Enterprise workspaces carry an allowance. Anything else answers `{"status":"not_applicable"}`, which surfaces as a clear provider error naming the workspace instead of a 0% gauge.
- **Multi-workspace accounts.** Defaults to the first workspace on a plan that has an allowance. A specific one can be pinned via the **Workspace ID** settings field or the existing `workspaceID` provider config field (no config-schema change) — both dashed and undashed UUID forms are accepted.
- **Cookie import requires `token_v2`.** A browser profile that has Notion cookies but no session cookie is skipped rather than used for a request that would 401.
- **Not read:** Notion *credits* (Custom Agents, Workers) are a separate meter on `getAIUsageEligibilityV2`. Out of scope here.
- Follows T3 Chat's shape throughout — descriptor + single web strategy, fetcher, snapshot mapping, settings store, manual Cookie header / cURL capture path. Like T3 Chat, it is not yet wired into widgets or the debug-log pane.

## Provider count

Bumped 66 → 67 in the places `Scripts/check-site-locales.mjs` enforces: `README.md`, `docs/providers.md`, `docs/social.html`, `docs/llms.txt`, `docs/index.html`, and the three counted keys across all 23 locales in `docs/site-locales.mjs`.

`Sources/CodexBarCore/Generated/CodexParserHash.generated.swift` is regenerated because the `.notion` case lands in `CostUsageScanner.swift`, which the hash covers.

## Docs

- `docs/notion.md` — setup, data source, mapping table, troubleshooting.
- `docs/providers.md` — strategy-table row and provider section.
- `docs/configuration.md` — notes Notion as a `workspaceID` consumer.

## Commands run

```
swift build
swift test --filter NotionUsageFetcherTests   # 10 tests, all passing
make check                                    # SwiftFormat + SwiftLint clean, 0 violations
make test                                     # full sharded suite, 64/64 groups green
```

## Verification

The endpoints, payload shape, and values were verified against a live Business workspace: the API's `85.31` / `16.89` and `resetsInSeconds: 20834` match what the Notion settings pane renders as "85% used / Resets in 5 hours 49 mins" and "17% used / Resets on Aug 22". `getSpaces` plan gating was confirmed by calling the endpoint for both a Business and a free personal space — the latter returns `not_applicable`, and an unknown space id returns 403.

Parsing, window mapping, workspace selection, and the id-normalization path are covered by unit tests using that captured payload shape.

Per `AGENTS.md`, I did not run a live browser-cookie import or an end-to-end `codexbar usage notion` probe, since those can raise a macOS Keychain prompt and need to be explicitly requested — so the Swift cookie-import leg specifically is unexercised here, and there are no menu screenshots. Happy to add both if you'd like them before merge.


n0ah37:
Closing — not ready for review yet. Will reopen once it clears my own quality bar.

PR #2555, opened 2026-08-01
Menu bar: add a pipe separator token to the layout editor

johnlarkin1:
Adds `|` as a second separator token in the menu bar layout editor, alongside the existing `·`.

## Why

The Structure palette offers a middot and a space. A vertical bar reads better in some strips — `5h 25% | W 60%` — and today that has to be faked with spaces. This makes it a first-class token.

## Changes

- New `MenuBarLayoutToken.separatorPipe`, rendered as `|` by `MenuBarLayoutRenderer` with the same attributes as the dot separator.
- Structure palette now offers `·`, `|`, space, in that order. The pipe uses `rectangle.split.2x1` (SF Symbols has no plain vertical-bar glyph — happy to swap it) and has its own VoiceOver label.
- `menu_bar_layout_token_separator_pipe_accessibility` added to all 23 catalogs.
- Renderer test token table covers the pipe.

Purely additive: the enum gains a case, so stored `separatorDot` layouts decode exactly as before. No preset or migration changed.

## Verification

`swift build` and `node Scripts/check-app-locales.mjs` clean against this branch on current `main`.

The non-English strings for the new accessibility key are machine-generated — worth a skim if the wording matters to you.

### DIFF
--- a/Sources/CodexBar/MenuBarLayout.swift
+++ b/Sources/CodexBar/MenuBarLayout.swift
@@ -19,6 +19,7 @@ enum MenuBarLayoutToken: Codable, Hashable, Sendable {
     case costToday
     case cost30d
     case separatorDot
+    case separatorPipe
     case space
 }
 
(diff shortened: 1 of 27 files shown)


johnlarkin1:

PR #2556, opened 2026-08-01
Menu bar: add pace tokens to the layout editor

johnlarkin1:
Adds **Session pace**, **Weekly pace**, and **Auto pace** tokens to the menu bar layout editor, next to the existing percent tokens.

Each renders the pace headline the usage card already computes — `On pace`, `23% in reserve`, `12% in deficit` — so the strip can show how much headroom is left against the window, not just how much quota is gone.

## Design

- **No new pace logic.** `.runsOut` already evaluates pace and keeps only the projection (`rightLabel`). That same evaluation now also yields the headline (`leftLabel`) for `.pace` — one computation, two tokens.
- **Window vocabulary matches `.percent`.** `.pace(window:)` mirrors `.percent(window:)`, including the `S` / `5h` / `W` prefixes, so a strip carrying both reads consistently. Provider-aware window resolution is inherited from the existing resolvers; no new notion of "weekly" is introduced.
- **Auto falls back.** Providers without a session-length pace (`UsagePaceText.sessionPace` gates to a handful) show the weekly figure rather than collapsing to the unavailable placeholder.
- **Purely additive.** No default layout changes; the tokens do nothing until dragged in. `MenuBarLayoutToken` uses synthesized `Codable`, so existing stored layouts decode unchanged.

## Verification

`swift build` and `node Scripts/check-app-locales.mjs` clean against this branch on current `main`. Renderer tests extended to cover the pace tokens and the codable round trip.

One inherited quirk worth naming: `weeklyPace` derives from `weekly ?? automatic`, the same window `.runsOut` already uses. For a provider with no weekly window, a token labelled "Weekly pace" therefore shows a figure computed from the automatic window. That behaviour predates this PR — this just surfaces it under a more specific label. Happy to tighten it if you'd rather it render unavailable instead.

### DIFF
--- a/Sources/CodexBar/MenuBarLayout.swift
+++ b/Sources/CodexBar/MenuBarLayout.swift
@@ -12,6 +12,7 @@ enum MenuBarLayoutToken: Codable, Hashable, Sendable {
     case providerName
     case accountLabel
     case percent(window: PercentWindow)
+    case pace(window: PercentWindow)
     case usageBar
     case resetCountdown
     case resetAbsolute
--- a/Sources/CodexBar/MenuBarLayoutEditor.swift
+++ b/Sources/CodexBar/MenuBarLayoutEditor.swift
@@ -246,6 +246,9 @@ struct MenuBarLayoutEditor: View {
                     .percent(window: .weekly),
                     .percent(window: .automatic),
                     .usageBar,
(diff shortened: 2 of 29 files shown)


johnlarkin1:

PR #2557, opened 2026-08-01
Menu bar: optional color-coded usage icon (off by default)

### LINKED ISSUE #2325: v0.45.0 custom menu-bar layout renders provider icons as tiny white dots on a light menu bar
## Summary

After updating to CodexBar 0.45.0 and using the new customizable menu-bar layout, the provider icons beside the usage values are extremely small and render in white. On a light/white macOS menu bar they look like indistinct white circles or dots, so it is effectively impossible to tell which provider each value belongs to.

The percentage and reset text remain readable; the provider identity icons are the affected part.

## Environment

- CodexBar: 0.45.0 (build 107)
- macOS: 15.7.3 (24G419)
- macOS appearance: Light
- Menu bar style: `Icon & percent`
- Layout: `All providers`
- Metric: `Percent + reset`
- `Improve visibility on inactive displays`: off

## Steps to reproduce

1. Update to CodexBar 0.45.0.
2. Open **Settings > Menu Bar**.
3. Select **Icon & percent**.
4. Configure the customizable layout for **All providers** with **Percent + reset** (including the Icon token in the menu-bar strip).
5. View the resulting items on a light/white macOS menu bar.

## Expected behavior

Each provider icon should retain a recognizable minimum size and automatically use enough contrast against both light and dark menu-bar backgrounds. It should be possible to identify the provider at a glance.

## Actual behavior

The icons are tiny white marks that visually collapse into small white circles/dots against the light menu bar. With multiple providers shown, the values can no longer be reliably associated with their provider.

## Regression

This appeared after the update that introduced the customizable drag-and-drop menu-bar layouts. The previous menu-bar rendering was readable.

johnlarkin1:
Tints the meter icon green → amber → red as usage rises. One toggle in **Settings ▸ Menu Bar ▸ Icon**, **off by default**.

This is a narrower, rebuilt version of the idea in #1207, split down to icon rendering only.

## Addressing the #1207 review

| Prior blocker | This branch |
|---|---|
| changed the existing-user colour default | New key `menuBarUsageColorsEnabled`, absent-means-`false`. No migration, no `registerDefaults`. Existing installs render byte-identical icons. |
| template-image tint path that does not solve the reported macOS behaviour | No `contentTintColor` anywhere. A template image keeps only its alpha mask and is recoloured by AppKit — which is exactly why tinting one does nothing on macOS 26. When tinted, the geometry is drawn **in** the colour and the bitmap marked non-template, the same approach as the existing `quotaWarningFlashImage`. |
| treats every secondary quota as weekly | **No time window is introduced.** The tint is a function of the primary bar value already produced by `resolvedMenuBarIconPercents`, which is provider-aware today. No `weekly`, no `secondary` in this diff. |
| unlocalized controls | Two keys across all 23 catalogs; `check-app-locales.mjs` clean. |

## Why there is no appearance observer

The palette is fixed sRGB, not `NSColor.system*`. Since `baseFill` becomes the tint outright, **no dynamic colour is ever resolved into the bitmap** — not the track fill, not the stroke, not the status overlay (hence threading `color:` into `drawStatusOverlay`). The render is a pure function of its inputs, so there is no stale-bitmap-on-appearance-flip bug to fix, no KVO observer, and the cache key is honest. A test pins that invariant.

Works identically on macOS 14–26. No `#available` gates.

## Scope boundaries

- **Icon + Percent is untouched.** That path renders the provider brand logo through `MenuBarLayoutRenderer.attachmentImage`, which exists to fix #2325. The toggle is disabled in that style. Brand-logo tinting would be a separate change.
- Morph/loading animations never tint.
- In-menu icon previews stay template so they invert on highlight.
- The quota-warning flash still composites on top and wins.

## Known trade-off

A non-template image does not invert under the status-item highlight — same as `quotaWarningFlashImage` today. Fine against macOS 26's translucent capsule, less crisp on 15's solid fill. Off by default, one toggle to revert.

## Verification

`swift build` and `node Scripts/check-app-locales.mjs` clean against this branch on current `main`. Tests cover the tint ramp, the non-dynamic-colour invariant, and the off-by-default no-op.

The ramp's anchors are green `0.14` → orange `0.85` → red `0.80`, so red dips slightly on the orange → red leg; the test asserts monotonic green end-to-end and monotonic red only on the green → orange leg. Palette values were picked for legibility against light, dark, and translucent menu bars rather than to form a monotonic ramp — say the word if you'd rather they were tuned differently.

### DIFF
--- a/Sources/CodexBar/IconRenderer.swift
+++ b/Sources/CodexBar/IconRenderer.swift
@@ -36,6 +36,7 @@ enum IconRenderer {
         let style: Int
         let indicator: Int
         let hideCritters: Bool
+        let tint: Int
     }
 
     private final class IconCacheStore: @unchecked Sendable {
@@ -120,13 +121,16 @@ enum IconRenderer {
         wiggle: CGFloat = 0,
         tilt: CGFloat = 0,
         statusIndicator: ProviderStatusIndicator = .none,
-        hideCritters: Bool = false) -> NSImage
+        hideCritters: Bool = false,
+        tint: NSColor? = nil) -> NSImage
     {
         let shouldCache = blink <= 0.0001 && wiggle <= 0.0001 && tilt <= 0.0001
         let render = {
-            self.renderImage {
-                // Keep monochrome template icons; Claude uses subtle shape cues only.
-                let baseFill = NSColor.labelColor
+            self.renderImage(isTemplate: tint == nil) {
+                // Untinted icons stay monochrome templates; Claude uses subtle shape cues only. A tint is drawn
+                // as real RGB instead, because a template image keeps only its alpha mask and is recolored by
+                // AppKit -- which is why tinting one via `contentTintColor` has no effect on macOS 26.
+                let baseFill = tint ?? NSColor.labelColor
                 let trackFillAlpha: CGFloat = stale ? 0.18 : 0.28
                 let trackStrokeAlpha: CGFloat = stale ? 0.28 : 0.44
                 let fillColor = baseFill.withAlphaComponent(stale ? 0.55 : 1.0)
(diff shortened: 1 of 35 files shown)


johnlarkin1:

Issue #2538, opened 2026-07-31
cost-usage-scan thread spins at ~91% CPU on 0.46.0 despite #2452 bound — many-file corpora (3.5k files, 4.5 GB)


Helmi:
## Summary

The `com.steipete.codexbar.cost-usage-scan` dispatch queue pins a full CPU core at ~91% continuously on 0.46.0. Disabling `tokenCostUsageEnabled` drops CPU to 0.0%, confirming the cost scan thread as the sole consumer. The bounding fix in #2452 does not appear to cover this corpus shape.

## Environment

- CodexBar 0.46.0 (110)
- macOS 26.5.2 (25F84), Apple Silicon
- Codex sessions: 3,559 JSONL files, 4.5 GB total
- Largest file: 156 MB; 5 files > 50 MB, 1 file > 100 MB
- 9 providers tracked in merged overview
- `historicalTrackingEnabled = true`, `tokenCostUsageEnabled = true`
- `refreshFrequency = fiveMinutes`

## What happens

Fresh launch (no prior process): CPU climbs to ~91% within 30s and stays there indefinitely. Sustained over 3+ days, the process accumulated 55+ hours of CPU time. Load averages reached 47/75/85.

Killing and relaunching reproduces immediately — this is not a slow leak, it is a tight loop from startup.

## Sample data

`sample <pid> 5` consistently shows the `cost-usage-scan` thread dominating CPU time. Two distinct hot paths observed across restarts:

1. **String matching loop** — `StringProtocol.contains()` → `StringProtocol.range(of:)` → `_StringCompareOptionsIterable._range()` — thousands of calls per sample window
2. **JSON parsing loop** — `newJSONString` → `__CFStringCreateImmutableFunnel3` → `__CFStringDecodeByteStream3` — continuous JSON string decoding

The main thread and all other queues are idle. Only `com.steipete.codexbar.cost-usage-scan` is active.

## Why #2452 does not catch this

PR #2452 added per-file byte budgets and resumable parsing for oversized individual rollouts (tested against a 276 GB / 1,322-file corpus with very large files). My corpus is the opposite shape: many relatively small files (3.5k files averaging ~1.3 MB). The per-file byte budget never triggers because no individual file is oversized — but the aggregate scan across all files still runs continuously without sleeping between iterations.

kiranmagic7:
[...]

PR #2520, opened 2026-07-30
Expose Codex scan completeness in cost JSON

NickGuAI:
## Summary

- persist whether the requested Codex history window was fully scanned
- propagate that state through the existing `CostUsageTokenSnapshot.historyCoverageIsEstablished` field
- emit `historyCoverageIsEstablished` in `codexbar cost --format json`
- keep unfinished recent-lookback files selected until their bounded parse resumes to completion
- gate cached completeness on the same native-cache validity checks as cached native totals

## Contract

```text
╔══════════════════════════╗
║ Bounded Codex log scan   ║
╚════════════╤═════════════╝
             │ partial/deferred work?
       ┌─────┴─────┐
      yes          no
       │            │
       ▼            ▼
 coverage=false  coverage=true
       └─────┬──────┘
             ▼
 CostUsageTokenSnapshot
             ▼
 cost JSON payload
```

Consumers can distinguish a cold, partial scan from a complete zero or complete total without parsing diagnostic logs. A successful non-Codex cost snapshot reports `true`; provider error rows omit the field.

Completeness remains fail-closed while a recently modified session outside the date partition is only partly parsed. A Pi-only or mixed cached result also cannot inherit completeness from a native Codex cache that is invalid for the requested window.

Related consumer work: https://github.com/NickGuAI/monorepo-g/issues/2010

## Real CLI behavior

Built and ran `CodexBarCLI` from this PR's exact head (`1b2f75a`) on Swift 6.2 / Ubuntu 24.04. Provider-native session files were mounted read-only; the transcript projects only public state and redacts usage values.

First, a cold scan of one completed real Codex session:

```console
$ codexbar cost --provider codex --days 30 --format json --json-only --refresh \
    | jq 'map({provider, hasUsage:(.last30DaysTokens>0), historyCoverageIsEstablished})'
[
  {
    "provider": "codex",
    "hasUsage": true,
    "historyCoverageIsEstablished": true
  }
]
```

Then, using that cache, the same provider roots were expanded to the real 30-day active and archived corpus. The bounded work exceeded one refresh, and the previously complete state was not reused:

```console
$ codexbar cost --provider codex --days 30 --format json --json-only --refresh \
    | jq 'map({provider, hasUsage:(.last30DaysTokens>0), historyCoverageIsEstablished})'
[
  {
    "provider": "codex",
    "hasUsage": true,
    "historyCoverageIsEstablished": false
  }
]
```

[...]

PR #2559, opened 2026-08-01
fix(codex): re-scan sessions resumed in older date partitions on a warm cache

### LINKED ISSUE #2558: [BUG] Codex sessions resumed in an older date partition are never re-scanned on a warm cache (silent cost/token under-count)
Resuming a Codex session whose rollout file lives in an older date partition permanently under-reports its cost and tokens on every warm refresh. The usage is not delayed — it never appears until something forces a full rescan.

## Why

`loadCodexDaily` builds its candidate file set from exactly three sources:

1. `listCodexSessionFilesByDatePartition` — walks only `<root>/YYYY/MM/DD` directories **inside the scan window** (`CostUsageScanner.swift:1462`).
2. the flat root listing.
3. `cachedCodexSessionFiles` — requires the path to already be in `cache.files`.

The mtime-based lookback that would otherwise catch a file outside the window, `listCodexRecentlyModifiedFiles`, is gated behind:

```swift
let shouldRunColdCacheLookback = cache.files.isEmpty || plan.rootsChanged
```

(`CostUsageScanner.swift:3407`, used at `:3423`)

That is cold-cache only. A rollout that was never scanned (created before the history window) and is not in `cache.files` is therefore unreachable by all three sources on a warm refresh. When the user resumes that session, Codex appends to the original file in its original partition — and nothing ever looks at it again.

Recovery requires `forceRescan`, a roots change, or deleting the cache. `pricingChanged` / `needsProjectMetadataMigration` set `forceFullScan` but do **not** add the file to the candidate list, so they do not recover it either.

## Repro

Verified on `78523f4a` (current main) with a pure-filesystem test on Linux — no network, no credentials:

1. Write an in-window session so the cache is warm (`cache.files` non-empty).
2. Write a second rollout into an **older** partition, outside the history window.
3. Scan (warm-up).
4. Append a new turn to the old rollout, as resuming does.
5. Scan again, and compare against a `forceRescan`.

```
warm refresh reported 110 tokens but a forced rescan reported 4455;
the resumed session in an older partition was never re-scanned
```

The warm path sees 110 tokens. The forced rescan sees 4455. The 4,345 tokens from the resumed turn are silently missing, and stay missing on every subsequent refresh.

## Impact

Codex cost and token totals read low across every surface that consumes the daily report — `codexbar cost`, `serve /cost`, `/dashboard/v1/snapshot`, the macOS Usage & Spend pane, and model analytics. It is silent: no error, no diagnostic, just numbers that are too small. The effect is worse the shorter the configured `costUsageHistoryDays` window, because more sessions fall outside it.

Long-running sessions resumed over days or weeks are exactly the ones that accumulate the most usage, so the sessions most likely to be mis-reported are the expensive ones.

## Note

`listCodexRecentlyModifiedFiles` already exists and does precisely the right thing — it is simply unreachable on the hot path. The fix looks like letting it also run when the requested window could not have enumerated a modified file, rather than only on a cold cache.

I have the failing test above and a fix; happy to send a PR. Filing first in case you would rather solve the discovery gap a different way, since this file is currently busy (#2520, #2525, #2538).

OfficialAbhinavSingh:
Closes #2558.

## Problem

A Codex session whose rollout file lives in a date partition **outside** the scan window is unreachable by every candidate source once the cache is warm:

1. `listCodexSessionFilesByDatePartition` walks only `YYYY/MM/DD` directories **inside** the window (`CostUsageScanner.swift:1462`).
2. the flat root listing does not cover partitioned files.
3. `cachedCodexSessionFiles` requires the path to already be in `cache.files` — and it never got there.

The mtime lookback that exists precisely to catch active sessions was gated behind:

```swift
let shouldRunColdCacheLookback = cache.files.isEmpty || plan.rootsChanged
```

i.e. cold cache only. So resuming such a session appends usage that is never scanned again. Recovery needs `forceRescan`, a roots change, or deleting the cache — `pricingChanged` / `needsProjectMetadataMigration` set `forceFullScan` but do not add the file to the candidate list, so they do not recover it either.

Silent: no error, no diagnostic, just Codex cost and token totals that read low across `codexbar cost`, `serve /cost`, `/dashboard/v1/snapshot`, Usage & Spend, and model analytics.

## Fix

Run the lookback on every refresh rather than only cold ones.

The important part is the cost. `listCodexRecentlyModifiedFiles` does two things: a **bounded** partition lookback (`codexActiveSessionLookbackDays` extra day directories) and an **unbounded recursive walk** of the whole sessions root. Only the bounded half now runs when the cache is warm — the recursive walk stays cold-cache only, so a warm refresh does not gain a cost that scales with corpus size. That matters here given #2538.

17 lines in one function plus the call site.

## Proof (Linux, `swift:6.3.3`)

Pure filesystem — no network, no credentials. Write an in-window session so the cache is warm, write a second rollout into an older partition, scan, append a turn to the old rollout as resuming does, scan again, compare against `forceRescan`.

**Before (stock `78523f4a`):**
```
✘ warm refresh reported 110 tokens but a forced rescan reported 4455;
  the resumed session in an older partition was never re-scanned
```

**After:**
```
✔ Test "a session resumed in an older partition keeps being counted on a warm cache" passed
```

I reverted the fix and re-ran to confirm the test genuinely fails without it, rather than trusting a green run.

Full Linux suite 318/318, `swiftformat --lint` clean, `swiftlint --strict` 0 violations / 1668 files.

## Scope note

The fix covers resumes within `codexActiveSessionLookbackDays` of the window start, which is what the bounded lookback is designed for and matches the test. A session resumed after a gap longer than that still needs the recursive walk, and I deliberately did not enable that on warm refreshes because of its cost profile. Happy to extend if you'd prefer broader coverage over the cheaper refresh.

This file is currently busy (#2520, #2525, #2538) — I kept the change as small as I could to minimise conflicts, and I'm happy to rebase whenever.


steipete:
Rebased onto `2920019bc1` (post-#2525) and force-pushed the maintainer-editable contributor branch. I did not merge.

Reconciliation with #2525:

- The warm path now checks only the extra 30 older date partitions; the corpus-wide recursive fallback remains cold-cache-only.
- Each older-partition directory check is charged to #2525's shared `CodexScanBudget`, including its byte and wall-clock limits.
- The persisted missing-parent discovery state is unchanged. The quiescence regression still proves that an unchanged missing parent performs zero repeated head parses, while a newly created parent is published once.
- The Codex parser producer hash was regenerated to `5b3a8f9c68c15741`.

The earlier CI failure was not a product defect or the #2573 spend-dashboard clock fixture. The substantive failure was `codex project metadata migration drops unscanned legacy files`: its synthetic “old, inactive” rollout had a brand-new filesystem mtime, so the new active-session lookback correctly classified it as recently modified. The fixture now pins that rollout's mtime to its old session date. The aggregate `lint-build-test` failure was downstream of that macOS shard failure.

Proof on the rebased head:

- `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --filter CodexWarmCacheResumeLinuxTests` — 2/2 passed, including warm-total-equals-cold/forced reference and explicit shared-budget deferral/catch-up.
- `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --filter 'CostUsage(ScannerBreakdown|PerformanceGate)Tests'` — 119/119 passed.
- `make check` — parser hash current; SwiftFormat clean; SwiftLint 0 violations.
- `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 make test` — 782/782 selections; 66/66 groups passed first attempt; zero retries, recoveries, timeouts, or Kiro waiver.
- Source-blind public CLI fixture with isolated `CODEX_HOME` and cache homes — pre-resume 110 tokens; warm post-resume 4,455; independent cold scan 4,455.
- Autoreview, local reconciliation and complete branch diff against `origin/main` — clean, no accepted/actionable findings.

Commits:

- `24ac2b9c27` — contributor implementation rebased onto post-#2525 main.
- `d35be70f36` — shared-budget integration, regression reconciliation, and regenerated parser hash.
- `542ed75d48` — make the shared-budget proof portable across Linux and macOS.
- `44274374df` — persist bounded lookback progress/candidates and reconcile file-path aliases.
- `6fc7a3ab21` — make persisted-cursor assertions use the exact cache artifact on Linux.

CI/review follow-up: the first rebased run exposed one Linux-only assertion in the new budget test. A later ClawSweeper review correctly found that a budget smaller than the partition offset would restart at the oldest day and could starve later partitions. The final implementation persists per-root day cursors and discovered candidate paths, keeps cold recursive discovery confined to its original cycle, and reconciles `/var` versus `/private/var` cache aliases by file identity. Two repeated 10-unit passes now prove progress into the old partition before a larger pass converges to the cold reference. Linux assertions decode the exact persisted artifact rather than the migration-aware current-cache view; the production second pass still proves reload and advancement. The final focused suites, 782-selection full suite, `make check`, source-blind CLI contract, and complete-branch autoreview are green on `6fc7a3ab21`.

steipete:
Exact-head follow-up for `186029684f` (not merged):

- CI run [30775618181](https://github.com/steipete/CodexBar/actions/runs/30775618181) is fully green: lint, Linux x64, Linux ARM64, Linux musl, both macOS shards, and the aggregate gate all passed.
- The preceding run's x64 and ARM64 jobs both failed at the same persisted-artifact assertion because Foundation `replaceItemAt` removed the existing cache destination without publishing its replacement on Linux. `186029684f` replaces that double-staging path with the repository's cross-platform `Data.write(..., .atomic)` pattern.
- The Codex parser hash is current at `dcab8c067490a89e`.
- Post-fix local proof: warm-cache regressions 2/2; broader CostUsage suites 347/347; `make check` clean; `make test` 782/782 selections and 66/66 groups on the first pass with zero retries, timeouts, or Kiro waiver.
- Final local autoreview was clean with no accepted/actionable findings. ClawSweeper also re-reviewed exact head `186029684f` and found no actionable defect; exact-head CI was its only remaining merge gate.

PR #2563, opened 2026-08-01
Fix idle OpenCode WAL database reads

### LINKED ISSUE #2544: OpenCode Go local reader errors on SQLite open when opencode is idle (missing WAL files)
When the opencode CLI is not running, the OpenCode Go local usage read fails every refresh with:

```
os_unix.c:51044: (2) open(/Users/yuhan/.local/share/opencode/opencode.db-wal) - No such file or directory
unable to open database file in "SELECT 1 FROM sqlite_master WHERE type = 'table' AND name = ? LIMIT 1"
```

`opencode.db` runs in WAL journal mode. `OpenCodeGoLocalUsageReader` opens it with `SQLITE_OPEN_READONLY`, but SQLite deletes the `-wal`/`-shm` sidecar files when opencode exits cleanly. A read-only open of a WAL database with no sidecar files fails outright (SQLITE_CANTOPEN, "unable to open database file").

Reproduced locally: copy `opencode.db`, remove `-wal`/`-shm`, read-only open fails. Drop in a 0-byte `-wal` and it reads fine.

Suggestions:
- fall back to `immutable=1` when the read-only open fails
- or copy the db to a temp file and read the copy
- or at least degrade gracefully instead of logging errors on every refresh cycle

steipete:
## Summary

- Read idle OpenCode Go WAL-mode history when SQLite sidecars are absent.
- Keep active WAL databases on the normal read-only path and avoid creating sidecars.
- Add focused regression coverage and a changelog entry.

## Root cause

`opencode.db` retains WAL mode in its database header after a clean shutdown removes the `-wal` and `-shm` sidecars. A normal read-only connection can open the database handle, but its first schema statement fails with `SQLITE_CANTOPEN` because SQLite cannot open the missing sidecars.

## Fix approach

The reader keeps its existing `SQLITE_OPEN_READONLY` path. If reading fails with `SQLITE_CANTOPEN` and both sidecars are absent, it retries through a read-only `immutable=1` SQLite URI. Active WAL databases continue through the normal path, while missing, non-WAL, and corrupt databases retain their existing behavior. The regression fixture also verifies that the read creates no sidecars.

## Proof

- `make check` — clean.
- `swift test --disable-automatic-resolution --filter OpenCodeGoLocalUsageReaderTests` — 10 passed.
- `swift test --disable-automatic-resolution --filter OpenCodeGoProviderStrategyTests` — 9 passed.
- `make test` — groups 1–34 passed. Group 35 hit known `KiroStatusProbeTests` subprocess/PTY wall-clock timing flakes on the loaded runner. Maintainer repeat runs failed different timing tests, confirming load sensitivity; see #2509, which scales those timing budgets.

Closes #2544

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Menu: the compact multi-account layout now covers every stacked multi-account list — token accounts on any provider and Codex accounts (flat lists; workspace-grouped Codex lists keep their sections).
 
 ### Fixed
+- OpenCode Go: read idle WAL-mode local history without creating SQLite sidecars (#2544). Thanks @Astro-Han for the report!
 - Menu: switching provider tabs no longer flashes. The sibling-tab warmup now runs off a tracking-safe timer (the previous Task-based warmup never fired while the menu was open, which is the only time it matters), and provider tabs share one stable menu height via an invisible spacer, so a switch is a single-frame content swap with no window resize. Verified frame-by-frame with a new env-gated self-probe (`CODEXBAR_FLICKER_PROBE_DIR`).
 
 ### Changed
--- a/Sources/CodexBarCore/Providers/OpenCodeGo/OpenCodeGoLocalUsageReader.swift
+++ b/Sources/CodexBarCore/Providers/OpenCodeGo/OpenCodeGoLocalUsageReader.swift
@@ -66,21 +66,45 @@ public struct OpenCodeGoLocalUsageReader: Sendable {
     }
 
     private func readRows() throws -> [UsageRow] {
+        do {
+            return try self.readRows(immutable: false)
+        } catch let failure as SQLiteReadFailure {
+            // A clean WAL shutdown can remove both sidecars while leaving the database header in WAL mode.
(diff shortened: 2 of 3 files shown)


steipete:
Exact test proof:

- `make check`
  - Passed all portable checks.
  - SwiftFormat: 0 of 1669 files require formatting.
  - SwiftLint: 0 violations in 1668 files.
- `swift test --disable-automatic-resolution --filter OpenCodeGoLocalUsageReaderTests`
  - 10 tests passed.
- `swift test --disable-automatic-resolution --filter OpenCodeGoProviderStrategyTests`
  - 9 tests passed.
- `make test`
  - Groups 1–34 passed.
  - Group 35 failed only in known `KiroStatusProbeTests` subprocess/PTY wall-clock timing tests on the loaded runner. Maintainer repeat runs failed different timing tests, confirming that the failures are load-sensitive and unrelated to OpenCode. Tracked by #2509.
- `swift test --skip-build --no-parallel --filter KiroStatusProbeTests`
  - 55 tests ran: 53 passed and 2 unrelated timing tests failed (`fetch cancellation while waiting for account probe is preserved` and `tty runner hard stops a process that ignores SIGTERM`).

PR #2568, opened 2026-08-02
Fix premature Codex weekly reset confetti

### LINKED ISSUE #2481: Codex weekly reset confetti can fire before the prior reset boundary

steipete:
## Summary

- Keep confirmed Codex weekly lows private when the confirmation was captured more than two minutes before the previously published reset boundary.
- Preserve the prior snapshot, reset-window baseline, history, diagnostics, and celebration detector state until publication is admitted.
- Retain genuine reset publication at and after the two-minute tolerance edge.
- Add the 0.46.1 changelog entry with credit to @gmkbenjamin.

## Root cause

The existing admission path verified that the weekly reset boundary advanced and that two low samples reported equivalent new boundaries. It never checked whether the previously published boundary was actually due, so an early rolling `now + 7 days` boundary could satisfy both checks and publish a false reset.

## Why this supersedes #2482

#2482 identified the right publication gap and supplied the reported regression. This version preserves @gmkbenjamin's initial fix and good test coverage, but replaces the later `EarlyBoundaryPattern` classifier and delayed third fetch with one prior-boundary guard. Every classifier branch ultimately preserved the same pre-boundary result; removing it avoids extra latency and an additional provider request without changing the intended outcome.

## Test plan

- `make check`
- `swift test --filter 'CodexWeeklyResetConfirmationTests|CodexWeeklyResetPublicationTests|UsageStorePlanUtilizationTests'`
- `make test`
- Structured autoreview of the final branch against `origin/main`

Closes #2481

References #2482


steipete:
Exact local proof for the pushed head:

- `make check` — passed; SwiftFormat reported 0 files requiring formatting and SwiftLint reported 0 violations across 1,668 files.
- `swift test --filter 'CodexWeeklyResetConfirmationTests|CodexWeeklyResetPublicationTests|UsageStorePlanUtilizationTests'` — passed: 109 tests in 3 suites, including the reported rolling-boundary regression, exact two-minute tolerance edge, full pre-publication surface preservation, genuine post-boundary publication, and reset/confetti detector coverage.
- `make test` — passed: 763/763 selections in 64/64 groups; all groups passed on the first attempt, with 0 retries and 0 timeouts.
- `/Users/steipete/Projects/agent-skills/skills/autoreview/scripts/autoreview --mode branch --base origin/main --stream-engine-output` — clean: no accepted/actionable findings reported.

No Kiro waiver was needed.

PR #2600, opened 2026-08-03
Classify Codex rate windows by duration (5h/weekly/30-day)

### LINKED ISSUE #2592: Codex: 30-day (43200-min) window now lands in `primary` and renders as the "Session" card (resets in 24 days)
## Summary

Codex can now return a **30-day (43200-minute) window in `primary`**, with the 7-day weekly window moved to `secondary`. CodexBar renders `primary` as the **"Session"** card unconditionally, so the provider card shows a "Session" quota that resets in **24 days**, while the real weekly quota is pushed down into the "Weekly" slot.

The numbers are correct — only the label is wrong. But because the card title and the menu-bar `Automatic` metric both assume `primary == 5-hour session`, the monthly lane wins the "closest to its limit" comparison and the weekly lane (the one that actually constrains a Plus account) becomes invisible at a glance.

This is the same root assumption already reported in #2324 ("Switcher rows `Weekly progress` uses the primary/session quota"), which was closed without changing the assumption. #2324 documented an account where Codex had **no `primary` and a 7-day `secondary`**. This report documents the opposite direction: a **30-day `primary`**. Between the two, `primary` clearly cannot be treated as "the session window".

## Evidence

Raw `rate_limits` payloads captured from `~/.codex/sessions/**/rollout-*.jsonl` on one account (usage percentages removed):

| Date | `primary.window_minutes` | `secondary` |
|---|---|---|
| 2026-07-15 | `10080` (7 days) | `null` |
| 2026-07-20 | `10080` (7 days) | `null` |
| 2026-07-26 | `10080` (7 days) | `null` |
| 2026-08-02 | **`43200` (30 days)** | `null` (`plan_type: "free"`) |

Example of the new shape:

```json
"rate_limits": {
  "limit_id": "codex",
  "primary":   {"used_percent": <redacted>, "window_minutes": 43200, "resets_at": 1787842532},
  "secondary": null,
  "credits":   {"has_credits": false, "unlimited": false, "balance": null},
  "plan_type": "free"
}
```

`resets_at: 1787842532` = 2026-08-27 22:55 (+08:00), i.e. a reset **24 days out**, rendered under a card titled "Session".

Cross-check against CodexBar's own state — `defaults read com.steipete.codexbar sessionLimitResetDetectorStates` for the Codex account has `resetBoundary = 809535262` (Apple epoch; `+978307200` → `1787842462`), which is **70 seconds** from the payload's `primary.resets_at`. So the Session card is definitively fed by `primary`.

After the account resubscribed to Plus, `secondary` appeared as a genuine 7-day window: `weeklyLimitResetDetectorStates` gives `resetBoundary = 807983340` → 2026-08-09 23:49 (+08:00), exactly 7 days after the subscription became active.

## How the 30-day window appears

Reproduction seems to require a subscription gap:

1. A Plus subscription lapses; the account falls back to the free tier (here: ~9 days).
2. During the free-tier period the backend opens a **30-day** usage window and reports it in `primary` (window start 2026-07-28 in this case).
3. The account resubscribes. A fresh 7-day window is created in `secondary`, but the 30-day window is **not** reset — it keeps its original `resets_at`, and its `used_percent` is simply recomputed against the larger Plus allowance.

So a continuously-subscribed Plus account probably never sees this. The window here expires 2026-08-27, after which `primary` may or may not revert to the weekly shape.

ChatGPT's own web UI does not surface this 30-day window at all — it only shows the limit that binds the current plan — so CodexBar showing it as "Session" reads as invented data to the user, which is how this was first noticed.

## Impact

- Provider card shows a "Session" quota with a multi-week reset countdown.
- `Settings > Menu Bar > Menu bar metric = Automatic` picks the lane closest to its limit. The stale 30-day lane carries leftover free-tier usage while the fresh weekly lane sits at 0% used, so `Automatic` pins the menu bar to the monthly window and the weekly quota is never shown.
- Workaround that works today: set the Codex menu bar metric to `Secondary` (`defaults write com.steipete.codexbar menuBarMetricPreferences -dict codex secondary`). It is fragile — if the weekly window ever moves back to `primary` and `secondary` becomes `null`, that slot goes empty.

## Suggested fix

Derive the card title from `window_minutes` instead of from field position:

- `~300` → Session / 5h
- `10080` → Weekly
- `43200` → Monthly (the `monthly` metric kind already exists in the binary and `MONTHLY` strings are already localized)
- otherwise → a generic label derived from the duration

The same resolution should feed `MenuBarMetricWindowResolver` and the `Automatic` metric, so "weekly" always means a real 7-day window (falling back only when none exists), which would also close out the open question left in #2324.

## Environment

- CodexBar 0.46.0 (Homebrew cask, current latest)
- macOS 27.0 (26A5388g), Apple silicon
- codex-cli 0.144.1
- ChatGPT Plus, `auth_mode: chatgpt`

Yuxin-Qiao:
## Summary
- Classify Codex windows by duration instead of payload slot: 300 min -> Session, 10080 min -> Weekly, 43200 min -> Monthly (new `RateLane.monthly`).
- A 30-day primary no longer renders as the "Session" card; the provider card, widget rows, and provider submenu all show "Monthly" with the real reset countdown.
- `Automatic` menu-bar metric now prefers the genuine weekly lane when a longer window exists, so a stale 30-day primary cannot hide the binding weekly quota. Explicit Primary/Secondary preferences are unchanged.
- Menu-bar layout preview and editor use the same automatic resolution.
- The monthly lane intentionally carries no session/weekly pace, workday, or quota-warning markers.

Closes #2592. Related: #2324 (inverse shape, weekly in `secondary`); the switcher already prefers real 7-day windows, so this closes the remaining card/menu-bar gap.

## Reproduction
Focused tests fail on current main before this change and pass after:
- `CodexConsumerProjectionTests.thirty day primary window maps to a monthly lane instead of session`
- `CodexConsumerProjectionTests.automatic menu bar metric prefers the weekly window over a thirty day primary`
- `MenuCardModelCodexProjectionTests.codex card titles follow window duration instead of slot position`

## Open question for maintainers
Unknown durations (anything other than 300/10080/43200) still fall back to slot semantics ("Session"/"Weekly") to preserve existing behavior (including the drifted-weekly guard). #2592 also proposes a generic duration-derived title for those; that part needs a product decision, so it is left out of this PR.

## Verification
- `make test`: 794 selections, 67 groups, 0 failures
- `make check`: clean (SwiftFormat + SwiftLint strict)


Yuxin-Qiao:
@clawsweeper re-review

Addressed the chart-history finding: `PlanUtilizationHistoryChartMenuView` now derives Codex visible series from the same duration classifier, and legacy 43,200-minute histories stored under session/weekly fold into the monthly series. Added focused chart + recording regression tests.

Verification on the rebased branch (now on current main):
- `make test`: 799 selections, 67 groups, 0 failures
- `make check`: clean

Note on real behavior proof: a live screenshot of the Monthly card requires an account whose OpenAI rate_limits currently report the 43,200-minute primary window (subscription-gap state); I have not probed any live account per repo policy. Source-level reproduction and regression coverage are in the PR.

Yuxin-Qiao:
## Runtime proof (after-fix, head cbbc44d6)

Captured from the PR build running on macOS with an injected 43,200-minute Codex snapshot via a temporary local-only env probe (`CODEXBAR_PROOF_30DAY`), removed after capture. The live account on this machine currently reports the standard 5h/weekly shape, so the Monthly state cannot be produced by the live backend here; the injected snapshot renders through the exact production card/menu-bar code paths.

Codex card

OCR of the card (rendered by the PR build):
- `Monthly` / `45% left` / `Resets Aug 28 at 1:26 AM` (30-day primary)
- `Weekly` / `95% left` / `Resets Aug 11 at 1:26 AM` (7-day secondary)

Menu bar metric

Menu bar metric (Automatic) with the same snapshot: `95% • in 7d` - the weekly lane is selected instead of the stale 30-day lane.

@clawsweeper re-review

Yuxin-Qiao:
@clawsweeper re-review

Addressed the submenu pace finding (1ccd7c4):
- `UsagePaceText.sessionPace` now rejects non-300-minute Codex windows, matching the existing antigravity/kimi duration guards.
- Added a provider-submenu regression test: a 43,200-minute primary renders "Monthly: 10% left" with no "Pace:" line beneath it. Verified the test fails without the guard with exactly the reported shape: `Monthly: 10% left / Resets in 2h / Pace: 10% in reserve · Lasts until reset / Weekly: 95% left`.

Verification:
- `make test`: 799 selections, 67 groups, 0 failures
- `make check`: clean

Yuxin-Qiao:
@clawsweeper re-review

Addressed the reset-backfill finding (2e7d47b):
- `codexMergedResetBackfillSnapshot` now also collects a 43,200-minute monthly candidate across all slots, placing it in an empty slot or replacing the older session/weekly candidate when its future reset is farther out.
- Regression tests: a fresh reset-less 30-day primary inherits its cached reset through the merged baseline; a newer monthly shape wins the primary slot over a stale session candidate.

Verification:
- `make test`: 799 selections, 67 groups, 0 failures
- `make check`: clean

steipete:
Landed. Verification before merge (independent maintainer-agent review; two real defects found and fixed on the branch before landing):

- Core classification verified: exact 300/10080/43200-minute durations map to session/weekly/monthly lanes; unknown durations deliberately fall back to slot semantics, preserving the drifted-weekly guard. All surfaces (provider card, widget rows, submenu) derive titles from the shared helper. The "Automatic" menu-bar change is provably a no-op unless a visible window is longer than weekly — i.e. only the #2592 shape changes.
- Fixed P1 (commit `352e00b4c`): `codexBackfillingResetWindows` rebuilt raw slots only from session/weekly lanes, so a monthly-classified primary vanished (or was replaced by a stale cached session window) whenever a reset-backfill baseline existed. Both failure modes reproduced in tests, then fixed to preserve monthly windows in place; helpers extracted to `UsageStore+CodexResetBackfill.swift`.
- Fixed P2: the new pace guard suppressed pace for *any* non-300-min Codex window, killing existing 540-min/nil-duration pace text; now only weekly/monthly-classified windows are suppressed, with the duplicate exact-duration gates removed so the central guard is the single authority.
- All four regression tests fail on the unfixed head and pass after. Full sharded suite: 799 selections, 67 groups, 0 failures; `make check` clean. CI green at merge.

Fixes #2592.

PR #2611, opened 2026-08-03
fix: restore background Chromium cookie imports without Keychain prompts

tangzhu599:
## Summary

Restore automatic Chromium cookie imports for background refreshes without reintroducing Keychain prompts. Cookie-only providers such as Kimi currently report "No available fetch strategy for kimi." on every automatic refresh and only work after a manual refresh.

## Root cause

#2225 (17c8729b) blocked all background Chromium cookie access by returning `false` from `BrowserCookieAccessGate.shouldAttempt` whenever the interaction context is not `.userInitiated`. Periodic and menu-open refreshes run in `.background`, so a provider whose only credential path is browser cookies (no API key, no fresh CLI credential) reports `noAvailableStrategy` until the user performs a manual (`.userInitiated`) refresh. The #2225 review even flagged this: "users who rely on Chromium cookies for automatic provider refresh may receive stale usage or authentication failures until they perform a manual refresh."

#2225's own follow-up prescribed the fix: SweetCookieKit was hardened so hosts can suppress interactive Safe Storage promotion, and "once released, CodexBar should bump that dependency."

## Change

- **SweetCookieKit 0.4.1 → 0.5.1** — adds `BrowserCookieKeychainAccessGate.withUserInteractionDisallowed`, the hardening #2225 planned for.
- **`BrowserCookieAccessGate.shouldAttempt`** — background path now: honors recorded denial cooldowns, runs the prompt-free attribute preflight (skips when interaction is required), and allows the import otherwise, instead of blanket-skipping every Chromium browser.
- **`codexBarRecords`** — every cookie read goes through `withCookieReadInteractionPolicy`: background reads run with `withUserInteractionDisallowed` (SweetCookieKit can never promote to an interactive prompt; a key that needs UI simply yields no cookies), user-initiated reads keep the interactive recovery path.
- **`recordDenied`** — only user-initiated actions write denial cooldowns, so a background probe can never suppress the user's next manual refresh for 6h.
- Tests updated for the new background contract; regression tests added for the no-cooldown-poisoning and non-interactive-read guarantees.

## Safety

- No background path can surface a Safe Storage prompt: preflight is non-interactive, and the read itself is wrapped with `withUserInteractionDisallowed`.
- Background probes never write denial cooldowns (a background probe must not suppress the user's next manual refresh).
- Safari / Firefox / Zen behavior unchanged; user-initiated Chromium behavior unchanged.

## Validation

- `swift build --target CodexBarCore` and `CodexBarCLI` pass.
- Gate behavior verified against the real module with a scratch harness: 10/10 checks pass (background admits Chromium when the preflight allows, background reads disallow keychain interaction, background probes don't poison manual refresh, user-initiated denial/cooldown semantics preserved).
- SwiftFormat clean. Full Swift Testing suite runs in CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.45.3 — Unreleased
 
 ### Fixed
+- Browser cookies: restore automatic Chromium cookie imports for background refreshes when the Safe Storage Keychain item is readable without interaction, keeping the read strictly non-interactive (SweetCookieKit 0.5.1). Cookie-only providers such as Kimi no longer report “No available fetch strategy” until a manual refresh, and background probes never write denial cooldowns that would suppress the next manual refresh.
 
 ## 0.45.2 — 2026-07-19
 
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "d5ef2ec180d58ea5f869b40e5024f2e23d1ce02e999305b2e78d1b5c3783b83b",
+  "originHash" : "6c75943d484e197ee5463361d945dccbba8776ba5a340830e90099119bc1b18a",
   "pins" : [
     {
       "identity" : "commander",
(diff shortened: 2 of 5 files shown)


tangzhu599:
Closing as duplicate: upstream #2528 (a3883384) already implements the same fix — background refreshes reuse an already-authorized Safe Storage ACL. This branch also carried two hardening deltas (read-site withUserInteractionDisallowed wrapper, userInitiated-only denial recording) that can be re-proposed as a focused follow-up if wanted.

PR #2615, opened 2026-08-03
Add focused daily spend ledger

sahilaidev:
## Summary

This is the focused follow-up requested in #2485: the daily spend ledger only.

- add a read-only daily ledger below the existing spend chart
- aggregate spend, tracked tokens, and requests per provider and day
- fill covered days with zero usage instead of dropping them
- restrict combined rows to the exact common coverage window across providers
- keep unavailable cost, request, or token data explicit rather than presenting partial totals

This PR does **not** add currency conversion, exchange-rate polling, preference changes, packaging changes, or fork customization. It uses the app-wide currency presentation already present upstream.

## Aggregation semantics

Request totals are derived from complete daily rows when a snapshot does not provide `last30DaysRequests`, which covers sources such as Codex and OpenCode Go. If a covered row omits its request count, arithmetic overflows, or a provided aggregate contradicts the daily rows, request totals fail closed to unavailable.

The ledger uses the intersection of every provider's established coverage. It renders every day in that intersection, including zero-usage days, and renders no combined ledger when provider coverage is disjoint. Sources without a proven aggregate cost remain unavailable instead of becoming zero-spend rows. The existing chart remains unchanged and the ledger is read-only, avoiding selection snapping from chart-only days outside common ledger coverage.

## Review fixes

- require every ledger source to have a proven summary cost before emitting daily rows
- include `last30DaysRequests` in the snapshot revision so cached/provider-derived aggregate changes rebuild the ledger
- cover both paths with focused regressions, including a same-publication-revision controller reload

## Verification

- `make check`
  - 22 localization catalogs checked
  - SwiftFormat clean
  - SwiftLint: 0 violations
  - packaging, documentation, repository-size, and shell checks passed
- focused post-fix suites: 32 tests passed
  - `SpendDashboardDailyLedgerTests`: 7 tests
  - `SpendDashboardControllerTests` and revision coverage: 25 tests across 3 suites
- `make test`: 810/810 selections passed in 68 groups
  - no failures, retries, or timeouts
- rebased onto current `upstream/main` before final verification

## Runtime proof

The original implementation was verified in a signed local macOS build with 30 days of Codex history. After the review fixes, the production SwiftUI ledger component was rendered again with synthetic Claude and OpenAI data so no personal spend history is exposed.

Synthetic daily spend ledger showing provider icons, tracked tokens, requests, and estimated spend

The synthetic capture shows combined-provider totals, single-provider days, token and request columns, and the app's native dark Settings presentation.


sahilaidev:
@clawsweeper re-review

sahilaidev:
@clawsweeper re-review

PR #2620, opened 2026-08-03
Compact usage details into one line

steipete:
## Summary

- move each usage window's used percentage onto the title row and keep its reset time right-aligned
- preserve the progress bar and expected-pace tick while folding deficit/reserve, outcome, risk, and session-equivalent detail into one meta line
- share the line presentation with the provider preview and make menu-height fingerprints depend on row shape instead of changing numeric text
- adapt the useful presentation and refresh-geometry seams from #2182; thanks @jack24254029

## Screenshots

The GitHub CLI does not expose attachment upload in this environment, so the sanitized usage-only crops are preserved locally at:

- `.build/one-line-shots/before.png`
- `.build/one-line-shots/after.png`

The raw captures were deleted after cropping because they contained account identity.

## Proof

- `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --filter 'UsagePaceTextTests|UsageMenuCardLayoutTests|MenuCardHeightFingerprintTests|HistoricalUsagePaceTests'` — 86 tests passed
- `make check` — SwiftFormat, SwiftLint, locales, package/signing, and repository checks passed
- `make test` — 800 selections in 67 groups passed; no retries or timeouts
- `./Scripts/package_app.sh` — signed app, widget, helpers, and Sparkle framework validated
- source-blind packaged-app validation — title percentages, right-aligned resets, bars/ticks, one-line pace detail, and restart/reopen geometry passed
- autoreview — clean, no accepted/actionable findings

## Reference

Supersedes the layout direction in #2182 without closing it.


bassner:
Any chance we can make this configurable and allow to switch back? I liked the previous layout much better, it was easier to scan. The percentages are now no longer vertically aligned after they have been cramped into the title, and are harder to visually distinguish from the title. @steipete 
Thank you!

Going back to the previous version for the time being

PR #2643, opened 2026-08-04
Retain priced Codex models when Auto Review is unpriced

akshayprabhu200:
## Summary

Keep priced Codex model rows visible when the same local history also contains an unpriced `codex-auto-review` routing row. The dashboard labels this as a partial model breakdown and removes ranking; it does not invent a price for Auto Review.

## Root cause

Codex local `token_count` records do not contain a billable model. The surrounding context can name `codex-auto-review`, which is a routing label with no pricing entry. The scanner therefore leaves that row unpriced. The spend dashboard previously treated a single unpriced model row as reason to discard every Codex model row, even when the daily stored cost exactly matched the sum of the other priced rows.

## Behavior

- Retains the directly priced Codex rows when their sum exactly matches the daily stored cost.
- Shows the routing row without a fabricated cost.
- Keeps the group partial and unranked.
- Continues to fail closed for malformed or inconsistent histories, including named rows with a present but invalid cost.

## Validation

- `make check`
- `swift test --filter SpendDashboardModelTests` (30 tests passed)
- `make test` (complete suite passed with the App Store Xcode toolchain before the final focused guard)
- `git diff --check`
- The partial-attribution fixtures use clearly fictional model names; they do not expose non-public model identifiers.
- A local `codexbar cost --provider codex --format json --days 30` structural scan found priced rows alongside unpriced `codex-auto-review` rows, with a valid aggregate cost.

## Native runtime proof

Native Usage & Spend proof — model names retained, values redacted

This native Usage & Spend capture shows priced Codex model rows, the partial-breakdown warning, dashed unranked markers, and `codex-auto-review` with an em dash instead of a fabricated cost. Model names remain visible; financial and usage quantities are redacted.

The capture was made at signed parent `39610a6cb8ae6e75761d64d5760222321762069f`. Follow-up signed head `063fc189fed45a0f6f544a53d0d1d301c9f7d246` adds only a fail-closed guard and regression test for malformed named costs; it does not alter the displayed valid-data behavior.


akshayprabhu200:
Native Usage & Spend proof with model names retained:

Native Usage & Spend proof

The capture shows the priced Codex model rows, partial-breakdown warning, dashed unranked markers, and `codex-auto-review` with an em dash rather than a fabricated cost. Financial and usage quantities are redacted; model identifiers remain visible.

Follow-up signed commit `063fc189fed45a0f6f544a53d0d1d301c9f7d246` closes the malformed-cost fail-closed case identified in the review, with a focused regression test.

@clawsweeper re-review

steipete:

- Retention logic verified strict: gated to the Codex provider; named priced rows must sum exactly to the entry's stored cost, unpriced rows must have an absent (not malformed) cost plus valid tokens, unnamed rows must be proven-zero, and at least one priced row must exist. A present-but-invalid cost fails closed (test-pinned).
- "No ranking on partial" confirmed in the presentation layer: retained-partial groups get `.incomplete` completeness, rendering `circle.dashed` + warning label instead of a rank, em dash for the unpriced row's nil cost. Non-Codex behavior untouched (pre-existing Claude partial tests pass unchanged).
- Added a guard test (`f1c6fbac9`) pinning that genuinely-unpriced-everything (routing rows only, no priced rows) stays `.unavailable` — previously only true by unstated invariant.
- Post-merge-with-main housekeeping: extracted model-breakdown helpers to `SpendDashboardModel+ModelBreakdown.swift` (`2a3150801`) to stay under the SwiftLint type-body cap after #2649 landed; byte-for-byte mechanical move, autoreview clean.
- Local: `swift test --filter SpendDashboardModelTests` — 32/32; `make check` — 0 violations. CI green at merge.

PR #2648, opened 2026-08-04
fix: keep Codex fork catch-up progress across appends

xx205:
## Summary

- Resume an appended unresolved-fork rollout from its validated cached offset instead of restarting
  a bounded full scan at byte zero.
- Charge only the appended suffix against per-file and per-refresh budgets, while preserving compact
  fork events and merging newly appended token snapshots.
- Restrict that resume path to ordinary unresolved forks; appended subagent buffers still take the
  correctness-preserving full-rescan path.
- Preserve unchanged same-size buffered retries as zero disk work for both ordinary forks and
  subagents, so an exhausted refresh budget cannot defer a pure in-memory replay.
- Keep a temporarily unresolved dependency key unset so a parent discovered in the same refresh can
  trigger the existing in-memory child retry.
- Accept the immediate predecessor parser producer because the cache schema and stored token-delta
  semantics are unchanged, avoiding a one-time 20+ GiB rebuild just to adopt this fix.

## Problem

#2525 made oversized Codex histories accurate and resumable, and persisted compact unresolved-fork
events. One completed-fork append path was still too restrictive, however. Its in-memory retry was
eligible only when the cached parent dependency key was `nil` **and** the cached offset already
equaled the current file size.

As soon as an active fork appended, its old offset was smaller than the new size. A stable
missing-parent lookup could also produce a non-nil dependency key. Either condition disabled the
fast path. Ordinary append parsing intentionally excludes forks, so the scanner fell back to a
correctness-preserving bounded full rescan from byte zero.

That fallback is safe for accounting, but it replaces a later completed offset with the first
bounded slice. On a large active rollout this makes Spend Dashboard progress move backward and can
re-read the same prefix after later appends. `Finish now` makes the rollback more visible because it
runs catch-up passes continuously; it does not itself clear the cache.

There was a related transition bug when a previously missing parent appeared. If the first replay
ran before the newly discovered parent had been scanned, retaining the old missing-parent dependency
key suppressed the post-parent retry and left the child's buffered usage unpublished.

The appended-subagent safety boundary also needs a distinct same-size case. A fully scanned
subagent can be retried before its queued parent is available without reading JSONL, because the
unchanged file cannot contain new lineage. Treating that retry like an appended subagent reports the
whole file as pending work and can defer the in-memory replay behind an exhausted byte budget.

### Validated append resume

A completed buffered fork can resume at its previous offset only when all of these checks pass:

- the cache contains a fork ID and non-empty compact unresolved-fork events;
- the previous scan completed and has no partial JSONL-record resume state;
- the current file identity matches the cached identity;
- the previous parsed offset is positive, not past EOF, and equals the token-index anchor offset;
- the existing 64 KiB prefix-boundary anchor still matches.

The scanner then replays compact events in memory and reads only the suffix. Appended token snapshots
are merged with the existing snapshot index. A same-size dependency retry preserves the existing
snapshots because it performs no disk read.

The validation is split by operation: an unchanged same-size replay accepts either ordinary-fork or
subagent buffers, while an appended suffix resume requires an empty subagent buffer. Any failed
validation keeps the existing full-rescan behavior. In particular, a completed subagent rollout
that may be reclassified by newly appended lineage metadata still receives a full rescan.

### Work accounting and parent transition

`pendingCodexScanWorkBytes` now reports `currentSize - cachedOffset` for a validated buffered fork,
so the suffix, not the whole file, is admitted against the 256 MiB / 512 MiB limits.
For a validated same-size buffered retry it reports zero, allowing the pure in-memory replay even
when the refresh byte budget is already exhausted.

During buffered replay, a newly unresolved dependency now remains `nil` instead of inheriting the old
missing-parent key. Once the queued parent finishes scanning, the existing same-refresh retry can
publish the child's compact events without rereading its JSONL.

### Cache compatibility

The generated parser hash still rotates. The immediately preceding producer
`codex:cu:p843ca061c36bbea1` is explicitly compatible because this change adds a resume path but does
not change the cache layout or token-delta semantics. Older pre-interleave-containment producers
remain incompatible.

## Privacy-safe validation

A current local cache was searched without emitting identifiers for one completed ordinary fork
with a valid prefix anchor, a non-empty unresolved-fork buffer, no subagent buffer, and real bytes
appended after its cached offset. Exactly one entry matched. Its rollout and single cache entry were
copied into an isolated temporary root; the local-only harness and clone were removed from the
validation workspace afterward.

```text
REAL_APPEND_PROOF source=real-local-rollout-copy
REAL_APPEND_PROOF cached_completed_offset=763410606
REAL_APPEND_PROOF file_size_after_append=773963499
REAL_APPEND_PROOF appended_bytes=10552893
REAL_APPEND_PROOF admitted_work=10552893
REAL_APPEND_PROOF after_fix_processed_offset=773963499
REAL_APPEND_PROOF after_fix_scan_complete=true
REAL_APPEND_PROOF warm_retry_offset=773963499
REAL_APPEND_PROOF warm_retry_stable=true
REAL_APPEND_PROOF privacy=paths-ids-dates-models-tokens-content-redacted
```

The scan slice was only 64 KiB larger than the 10,552,893-byte suffix and therefore could not cover
the 773,963,499-byte full file. Admitting exactly the suffix and reaching EOF proves the fixed path
resumed from the cached offset. An unchanged warm retry preserved the completed offset.

As supporting observation, the CodexBar 0.47.0 process did not restart while one account-scoped
cache indexing the same 20+ GiB corpus reported:

```text
1,415,520,432 -> 823,980,119 -> 1,082,720,169 processed bytes
```

The exact regression case is synthetic and checked in as `CodexForkAppendResumeTests`:

| Field | Before fix | After fix |
| --- | ---: | ---: |
| Initial completed file | 18,390 B | 18,390 B |
| Appended suffix | 391 B | 391 B |
| Refresh budget | 512 B | 512 B |
| `parsedBytes` after append | 452 B | 18,781 B |
| `codexScanComplete` | `false` | `true` |
| token snapshots | 0 | 2 |
| pending work after append | 18,781 B | 391 B |

Because the 512-byte budget cannot cover the 18,390-byte prefix, reaching 18,781 bytes proves that
the fixed path resumed from the cached offset rather than rereading the file.

The adjacent subagent boundary regression starts with a completed buffered subagent larger than the
512-byte budget. With the file unchanged and the dependency temporarily unresolved, pending work is
zero and an already exhausted budget still admits `.allow(0)`. After appending lineage and token
events, pending work becomes the entire file and the bounded refresh remains incomplete at or below
512 bytes. This proves the same-size retry is free without weakening the appended-subagent full
rescan rule.

## Verification

- `make check`
  - parser hash is current;
  - repository/package/locale checks passed;
  - SwiftFormat: 0 files require formatting;
  - SwiftLint: 0 violations in 1,762 files.
- Focused CostUsage validation:
  - 142 tests in 5 suites passed;
  - includes cache compatibility, performance gates, fork/subagent integration, scanner breakdown,
    missing-parent appearance, the append regression, and the subagent full-rescan boundary.
- GPT-5.6 pricing isolation check:
  - 2 tests passed when `CFFIXED_USER_HOME` pointed to a fresh temporary directory;
  - without isolation, those tests read the developer machine's live `models.dev` cache and fail
    four bundled-price assertions. The pricing production and test files are unchanged from base.
- Full `make test` with an isolated user cache and at most two Swift compile jobs:
  - 813/813 selections in 68/68 groups passed on the first attempt;
  - 0 failed groups, retries, or timeouts;
  - 1,361.4 seconds total (134.6 seconds discovery, 1,226.8 seconds execution).

## Scope

- No UI strings, controls, accounting formulas, or catch-up duty-cycle settings change.
- This fixes repeated prefix work and progress rollback inside each cache. The default cache and an
  account-scoped Dashboard cache can still scan the same corpus independently; cross-cache work
  sharing is a separate optimization.
- Append reuse relies on Codex rollout files being append-only. File replacement, truncation, or a
  failed boundary anchor continues to invalidate the resume path.

## Screenshots

Not applicable. This is a scanner/cache correctness and performance fix with no UI change; the
privacy-safe validation is represented by deterministic cache-state assertions.


steipete:
Landed. Verification before merge (independent maintainer-agent review, line-by-line with a security lens):

- Truncation/rewrite safety: the append-resume path requires inode match, completed prior scan, no partial-resume state, valid offset bounds, token-anchor offset equality, and a SHA-256 match of the 64 KiB window ending at the cached offset. Truncate-then-regrow and in-place rewrites fail the hash and fall back to the bounded full rescan — strictly more validation than the old same-size retry path.
- Budget accounting: admission charges exactly `size - offset` for a validated resume using the same validators as the scan path, so admission and actual work cannot disagree. The checked-in test proves a 512-byte budget still reaches EOF on an 18 KiB-prefix file via resumption.
- Subagent boundary: append resume requires an empty subagent buffer; appended subagent files keep the full correctness-preserving rescan (test-pinned).
- Producer whitelist of the predecessor cache key verified justified by inspection: no stored schema or token-delta semantics change, spares large corpora a full rebuild.
- Parser hash reproduces deterministically via `Scripts/regenerate-codex-parser-hash.sh`.
- Local: 183 scanner/cache tests + broader CostUsage sweep of 483 tests in 38 suites, 0 failures; `make check` clean. CI green at merge.

PR #2649, opened 2026-08-04
OpenCode Go: per-model cost breakdown by day

kentoku24:
## Summary
- Extracts each local assistant message's `modelID` from `opencode.db` (the real model behind the constant `opencode-go` Zen-proxy `providerID`) and groups cost/request counts by `(day, model)` instead of just by day.
- Populates `modelsUsed` / `modelBreakdowns` on each `CostUsageDailyReport.Entry` (previously always `nil` for OpenCode Go), so the shared Cost history chart now shows a per-model breakdown for OpenCode Go the same way it already does for Claude/Codex — no UI changes needed, since `CostHistoryChartMenuView`, the inline dashboard, and the CLI cost command are all provider-agnostic and already render `modelBreakdowns` whenever it's populated.
- Rows with no `modelID` fall back to an `"unknown"` bucket instead of being dropped, so day totals stay correct either way.

## Why
Requested: mimic the per-model cost breakdown that Claude Code already shows in the "Cost" tab, but for OpenCode Go, listing model spend per date.

## Verification against a real installation
The `modelID` field name was originally inferred from OpenCode's known message schema rather than confirmed against real data, since the existing local test fixtures never included a model field. That's now been checked against a real `~/.local/share/opencode/opencode.db`:

- Read-only `sqlite3` query against real `opencode-go` assistant messages confirms `modelID` is a real top-level key on the message JSON, alongside `providerID`, `cost`, `tokens`, `time` — e.g. distinct values seen: `deepseek-v4-flash`, `deepseek-v4-pro`, `qwen3.7-plus`, `minimax-m3`, `glm-5.2`, `kimi-k3`.
- Ran the actual `OpenCodeGoLocalUsageReader().fetch()` (the exact code path this PR changes) against that real database (14-day window):
  - `daily_entries=10`, `days_with_breakdown=10`, `days_without_breakdown=0` — every recent day produced a populated model breakdown; none fell back to the `"unknown"` bucket.
  - `distinct_models=["deepseek-v4-flash", "deepseek-v4-pro", "kimi-k3", "minimax-m3", "qwen3.7-plus"]`
  - Most recent day: `last_day_model_count=3`, `last_day_models=["deepseek-v4-flash", "deepseek-v4-pro", "qwen3.7-plus"]`, `last_day_request_counts=[1, 3, 1046]`
- No dollar costs, timestamps, or message content were captured or shared — only model identifiers and integer counts.

This confirms the `modelID` extraction and per-day/per-model grouping work correctly end-to-end against a real OpenCode Go install, not just the synthetic SQLite fixtures in the test suite.

## Post-review fix
- ClawSweeper caught a real normalization bug: the grouping key used the untrimmed `modelID`, even though emptiness was checked against the trimmed value — so a model id with incidental leading/trailing whitespace would form its own separate bucket instead of merging with the clean value for the same model. Fixed by grouping on the trimmed value consistently, with regression tests for both a whitespace-only id (falls back to `unknown`) and a whitespace-padded id (merges with the unpadded bucket).

## Reviewer notes
- No token-count extraction was added — OpenCode Go's local rows don't reliably carry token counts across both the message-only and message+part query paths, and `ModelBreakdown.totalTokens` is optional, so the cost-only breakdown degrades gracefully in the shared formatter exactly like it does for other cost-only providers.

## Test plan
- [x] `swift build` succeeds
- [x] `swift test --filter OpenCodeGoLocalUsageReaderTests` — all 13 tests pass (10 existing + 3 new covering multi-model days, part-inherits-message-model, and the unknown-model fallback)
- [x] Full `swift test` suite exits 0; all 8 OpenCode Go test suites pass. Some unrelated timing-sensitive tests (`SubprocessRunner`, `CLIServeRouter`, `Antigravity`, `DeepSeek`) were flaky under heavy concurrent load on this machine but are unrelated to this change.
- [x] Verified against a real `opencode.db` install (see "Verification against a real installation" above)

### DIFF
--- a/Sources/CodexBarCore/Providers/OpenCodeGo/OpenCodeGoLocalUsageReader.swift
+++ b/Sources/CodexBarCore/Providers/OpenCodeGo/OpenCodeGoLocalUsageReader.swift
@@ -122,7 +122,8 @@ public struct OpenCodeGoLocalUsageReader: Sendable {
             let cost = sqlite3_column_double(stmt, 1)
             guard createdMs > 0, cost >= 0, cost.isFinite else { continue }
             let requestCount = max(1, Int(sqlite3_column_int64(stmt, 2)))
-            rows.append(UsageRow(createdMs: createdMs, cost: cost, requestCount: requestCount))
+            let model = sqlite3_column_text(stmt, 3).map { String(cString: $0) } ?? ""
+            rows.append(UsageRow(createdMs: createdMs, cost: cost, requestCount: requestCount, model: model))
         }
         return rows
     }
@@ -165,7 +166,8 @@ public struct OpenCodeGoLocalUsageReader: Sendable {
         SELECT
           CAST(COALESCE(json_extract(data, '$.time.created'), time_created) AS INTEGER) AS createdMs,
           CAST(json_extract(data, '$.cost') AS REAL) AS cost,
-          1 AS requestCount
+          1 AS requestCount,
+          COALESCE(json_extract(data, '$.modelID'), '') AS modelID
         FROM message
         WHERE json_valid(data)
           AND json_extract(data, '$.providerID') = 'opencode-go'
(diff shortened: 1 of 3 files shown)


steipete:

- SQL verified in both query variants: message-only and message+part UNION branches emit matching 4-column shapes; step-finish parts inherit `modelID` from the parent-message CTE.
- Schema drift safe: `modelID` is read via `json_extract`, so an old `opencode.db` without the key yields the "unknown" bucket rather than a prepare failure; day totals provably unchanged for rows without modelID (fallback test pins full cost landing in the day entry).
- Whitespace-normalization regression (padded ids forming separate buckets) fixed in `b25c008dd` with tests.
- Local: `swift test --filter OpenCodeGoLocalUsageReaderTests` — 15/15 pass. CI green.

Known UX note: a DB predating `modelID` now shows an all-"unknown" breakdown instead of none (disclosed in the PR).

Issue #2634, opened 2026-08-04
Claude usage stops updating on 0.47.0; refresh fails with SessionError


Astro-Han:
macOS, CodexBar 0.47.0, Claude Code 2.1.220, Claude source = OAuth, Keychain prompt policy = only on user action.

The Claude card stopped updating right after the 0.47.0 upgrade. Clicking Refresh does nothing. Logs repeat every cycle:

```
Claude OAuth delegated refresh touch did not update Claude keychain
Claude OAuth delegated refresh touch failed (errorType=SessionError)
```

The last good snapshot was taken by the 0.46.0 process before I restarted; the first 0.47.0 launch produced the errors above.

I dug through the diff between 0.46.0 and 0.47.0 and the regression looks intentional. e17ba24b ("Stop reading Claude-owned credentials", merged Aug 1) changed `keychainAccessAllowed` from:

```swift
return !KeychainAccessGate.isDisabled
```

to a hard `return false`, with a comment about Claude Code's ACL rewrite causing recurring password dialogs. The prompt-storm motivation is fair (see #2380), but with current Claude Code this leaves no working credential path:

1. `~/.claude/.credentials.json` is stale. Claude Code 2.1.x keeps credentials in the keychain only; the file is no longer updated on login (mine is from Jul 25 and returns 401 on `GET api.anthropic.com/api/oauth/usage`). It also lacks `expiresAt`, so `isExpired` is always true.
2. The keychain item is valid. The same endpoint returns 200 with the token from `Claude Code-credentials` (expiresAt 1785838980484 = 2026-08-04 18:23 CST, refresh token good until Aug 24).
3. The delegated refresh touch cannot repair anything. It launches `claude` in a PTY; on 2.1.220 `/status` is gone, and `/usage` prints "Error: Failed to load usage data" and never exits, so the 8s capture times out as SessionError.

Expired file, unreadable keychain, hanging CLI session. All three paths dead, while 0.46.0 worked on the same setup because path 2 was open.

Repro:

1. CodexBar 0.47.0 + Claude Code 2.1.220, OAuth source.
2. Let the file credential go stale (or the access token expire).
3. Click Refresh. Same SessionError, no new rows in usage history.

Workaround that works here: copy the `claudeAiOauth` object from the `Claude Code-credentials` keychain item into `~/.claude/.credentials.json`, including `expiresAt` and `scopes`. The card comes back immediately (verified via `codexbar usage --provider claude`). It lasts until the access token rotates.

[...]

PR #2650, opened 2026-08-04
Report unreadable Claude OAuth refresh as terminal

kes02:
Fixes part of #2634.

## Problem

When the delegated Claude CLI refresh completes cleanly but the Keychain fingerprint does not move, `performAttempt` reports `.attemptedFailed("Claude keychain did not update after Claude CLI touch.")`. On 0.47.0 that verdict is wrong in a specific and permanent way.

Success is defined as the fingerprint changing inside the touch window (`min(timeout, 2)`s, `delays = [0.2, 0.5, 0.8]`). If Claude Code already refreshed its own item on its own schedule, the CLI has nothing to do, the fingerprint does not move, and the attempt fails. Since `keychainAccessAllowed` is an unconditional `return false` in release, the refreshed credential is unreadable anyway — so on a profile with no `~/.claude/.credentials.json`, no amount of retrying can succeed.

The background path then printed `... delegated Claude CLI refresh failed: Claude keychain did not update after Claude CLI touch. Run 'claude login', then retry.` Both halves mislead: retrying cannot help, and `claude login` refreshes the Keychain item that CodexBar does not read.

## Change

The verdict is carried on an internal `AttemptResult` rather than a new `Outcome` case:

```swift
struct AttemptResult: Sendable, Equatable {
    let outcome: Outcome
    let isUnreadableAfterRefresh: Bool
}
```

`Outcome` ships in the `CodexBarCore` library product, so adding a case would break downstream exhaustive switches on source update. The public enum and `attempt(now:timeout:environment:)` are unchanged; the fetcher reads the detail through a new internal `attemptDetailed`. `git diff upstream/main` adds and removes no `public` line.

The state is detected after the touch, not before:

```swift
guard !configuration.keychainReadAllowed else { return false }
guard !configuration.hasSelectedProfileOAuthCredentialsFile else { return false }
return !ClaudeOAuthCredentialsStore.hasSelectedProfileOAuthCredentialsFile(
    environment: configuration.environment)
```

- The touch still runs. On older Claude Code it can create the credentials file, and the post-touch re-check picks that up.
- `keychainReadAllowed` is captured in `AttemptConfiguration` because `performAttempt` runs in a detached task, where task-local overrides do not apply.
- The attempt records the long cooldown instead of the short one, so background polling stops relaunching `claude` every 20s for a state that cannot change. User-initiated refresh still bypasses the cooldown.
- A joined attempt reuses the verdict rather than retrying; readability does not depend on interaction.

The message change is scoped to the background path. `shouldPreserveOwnerCLIHandoff` returns `true` only for `.userInitiated`, so an explicit Refresh still rethrows the typed `.refreshDelegatedToClaudeCLI` handoff and reaches the owner-CLI pipeline exactly as before. Only background refreshes fall through to the generic catch, and those are the ones that were printing "Run `claude login`, then retry" for a state where neither helps.

Two presentation helpers moved to `ClaudeUsageFetcher+DelegatedRefreshMessages.swift` to stay under the file-length limit; visibility widened from `private static` to `static`, no logic change.

## Scope and limits

- **This does not fix the `SessionError` path in the report.** When the touch itself errors, the outcome stays `.attemptedFailed` with the underlying error, on the short cooldown. Masking a real touch error behind a terminal verdict would hide the CLI-hang defect, which looks like a separate problem. Only the completes-but-unobservable variant is addressed here.
- **This does not restore usage.** It reports the state accurately and stops the retry churn. Actually recovering OAuth needs the credential-ownership decision, which is still open on the issue.
- The post-touch re-check runs inside the detached task, so under tests it reads `isolatedTestCredentialsURL` — a per-process random temp path that is never created — rather than an overridden URL. The captured pre-touch value is what the tests steer, and a stray file there would fail the assertion loudly rather than pass wrongly.

## On after-fix runtime output

I tried to capture it and could not; the obstacle is structural. `Scripts/package_app.sh` signs ad hoc, so a locally built bundle has a different code identity than the installed notarized one and cannot read CodexBar's own cached-credential Keychain item:

```
Keychain cache item is unusable by this executable
(oauth.claude.profile.22c64993…); treating as missing
```

It therefore starts with no cached credentials at all — a different state from the bug, which needs an *expired* cache. It never entered the OAuth path (zero delegated-refresh log lines) and fell through to the CLI scrape. The build holding the reproducing state is the notarized one, which lacks the patch; the build with the patch cannot reach the state. Closing that needs a Developer ID signed build.

The pre-fix side is live rather than fixture-based: on this setup the retained log holds no `touch succeeded` and no `touch failed`, only `touch did not update Claude keychain` → `attemptedFailed`, with the Keychain `mdat` sitting 31 minutes after the cached token's expiry.

Also worth stating: the changed message is background-only. `ClaudeProviderDescriptor` sets `allowBackgroundDelegatedRefresh: false`, `performAttempt` returns `.skippedByPromptPolicy` for background work unless the prompt policy is `always`, and `shouldPreserveOwnerCLIHandoff` keeps the existing handoff for `.userInitiated`. On the "only on user action" configuration from the linked issue the message path is unreachable by design, so no local capture would surface it either.

## Tests

`ClaudeOAuthDelegatedRefreshUnreadableResultTests`:

- unreadable Keychain + no credentials file + clean touch → `isUnreadableAfterRefresh`, and the touch is asserted to have run once
- same conditions with a credentials file present → plain `.attemptedFailed`, flag clear

Both use testing overrides and a temp credentials URL; no real Keychain access.

## Commands run

- `swift build` and `swift build -c release`
- `swift test --filter ClaudeOAuthDelegatedRefresh` — 27 tests, all green
- `make test` — full suite, 813 selections
- `./Scripts/lint.sh format` and `./Scripts/lint.sh` — 0 violations

Two pre-existing failures showed up on my machine during the full run; both reproduce identically on pristine `main`:

- `ProviderPluginExtensionParityTests` — `PerplexityUsageSnapshot.promoExpiryFormatter` never sets `locale`, so `MMM` resolves through `Locale.current` and renders `1월` on a Korean-locale host. Unrelated to this change, fixed separately in #2651.
- `ClaudeOAuthDelegatedRefreshProfileIsolationTests / legacy cooldown migrates only to the default credentials profile` — fails when run alongside sibling suites, passes in isolation, and does not reproduce on every run.

### DIFF
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeOAuth/ClaudeOAuthDelegatedRefreshCoordinator.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeOAuth/ClaudeOAuthDelegatedRefreshCoordinator.swift
@@ -28,6 +28,11 @@ public enum ClaudeOAuthDelegatedRefreshCoordinator {
         case cliUnavailable
         case attemptedSucceeded
         case attemptedFailed(String)
+        /// The Claude CLI can still refresh its own credentials, but CodexBar has no permitted way to observe the
+        /// result: production never reads the foreign Keychain item, and this profile has no credentials file to
+        /// fall back on. The touch still runs first, in case it writes that file; when it does not, report this
+        /// rather than a retryable failure so background polling can back off.
+        case unreadableAfterRefresh
     }
 
     private static let log = CodexBarLog.logger(LogCategories.claudeUsage)
@@ -71,7 +76,8 @@ public enum ClaudeOAuthDelegatedRefreshCoordinator {
             switch outcome {
             case .attemptedFailed, .skippedByCooldown, .skippedByPromptPolicy, .cliUnavailable:
                 return await self.attempt(now: now, timeout: timeout, environment: environment)
-            case .attemptedSucceeded:
+            case .attemptedSucceeded, .unreadableAfterRefresh:
+                // Retrying cannot make an unreadable refresh readable; reuse the joined verdict.
                 return outcome
             }
         case let .joinDifferentProfileThenRetry(id, task, state):
@@ -99,6 +105,7 @@ public enum ClaudeOAuthDelegatedRefreshCoordinator {
         let readStrategy: ClaudeOAuthKeychainReadStrategy
         let promptMode: ClaudeOAuthKeychainPromptMode
         let keychainAccessDisabled: Bool
+        let keychainReadAllowed: Bool
         let hasSelectedProfileOAuthCredentialsFile: Bool
         #if DEBUG
         let cliAvailableOverride: Bool?
@@ -147,6 +154,7 @@ public enum ClaudeOAuthDelegatedRefreshCoordinator {
             // from the user's stored preference, not the strategy-adjusted mode used by our own reads.
             promptMode: ClaudeOAuthKeychainPromptPreference.storedMode(),
             keychainAccessDisabled: KeychainAccessGate.isDisabled,
+            keychainReadAllowed: ClaudeOAuthCredentialsStore.keychainAccessAllowed,
             hasSelectedProfileOAuthCredentialsFile: ClaudeOAuthCredentialsStore
                 .hasSelectedProfileOAuthCredentialsFile(environment: environment),
             cliAvailableOverride: self.cliAvailableOverrideForTesting,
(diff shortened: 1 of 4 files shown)


kes02:
@clawsweeper re-review

steipete:

- Confirmed the PR's scoping is correct, not papering-over: only the provably terminal state (touch completed cleanly, Keychain unreadable, no credentials file before/after) is reported terminal with the long cooldown; touch *errors* deliberately stay retryable because on older Claude Code a retried touch can still produce `~/.claude/.credentials.json`. An attempted "improvement" extending terminal to error cases broke the existing joined-retry test for exactly that reason and was reverted with evidence.
- `AttemptResult` wrapper adds/removes no public API; tests use overrides with no real Keychain access.
- Pushed `032e7e7b9`: decision-site comment + regression test locking the SessionError variant from #2634 to `.attemptedFailed` with the message preserved.
- Local: targeted suites green; the flaky `ClaudeOAuthDelegatedRefreshProfileIsolationTests` and shared-UserDefaults parallel flakes reproduce identically on the pristine main baseline — pre-existing, not from this PR. Codex autoreview clean. CI green at merge.

Root-cause context for the wider #2634 regression (why `.attemptedSucceeded` is structurally unreachable on release 0.47.0) is posted on the issue: https://github.com/steipete/CodexBar/issues/2634#issuecomment-5186926685 — the remaining fix is an ownership-policy decision, not code mechanics.

PR #2655, opened 2026-08-04
fix: show Codex credit usage in menu bar

McBrosa:
## Summary

- show Codex credit-limit usage in the menu bar's automatic percentage and usage-bar tokens when the account has credits but no usable session or weekly rate window
- preserve explicit session and weekly token semantics; only the existing Codex credits fallback can replace the automatic metric
- honor the global "As used" / "As remaining" direction, so the reported 1,193.6 of 42,750 credits renders as 3% used or 97% remaining
- keep the layout editor preview on the same projection path and document the fix in the changelog

## Root cause

The legacy status-item renderer already had a Codex credits fallback, but the stored menu-bar layout renderer only received `RateWindow` values. Credit-only Business accounts therefore supplied valid credit-limit data to the menu card while `Auto %` received no window and rendered the standard en-dash placeholder.

This carries the existing `CodexConsumerProjection.menuBarFallback` decision into layout render data as a used percentage. The renderer consumes it only for automatic usage tokens, leaving session and weekly quotas unavailable when those windows do not exist.

## Proof

- `swift test --filter 'MenuBarLayoutRendererTests|StatusItemAnimationCodexCreditsTests'` — 16 tests passed
- `make check` — locale, documentation, packaging, SwiftFormat, and strict SwiftLint checks passed with zero violations
- `make test` — 812 selections across 68 groups passed

The status-item regression uses synthetic credit-limit data matching the report (`1,193.6 / 42,750`) and asserts that the stored default layout renders `3%` instead of `–`. No live provider probe, Keychain access, or bundle-level screenshot was used.


steipete:
Thanks @McBrosa. Closing as superseded by the implementation now on `main`, rather than merging this older fallback path.

[7350a1e1cf92](https://github.com/steipete/CodexBar/commit/7350a1e1cf929816cd88c218e0ff8c560d96c7e2) added the monthly-credit projection, and [30a389d6e664](https://github.com/steipete/CodexBar/commit/30a389d6e66406c2a08bc7c8ff18d6649c75b9f0) completed the case where the usage snapshot has not loaded. Both are reachable from current main. The earlier August 12 review predates these fixes.

The [current projection](https://github.com/steipete/CodexBar/blob/4e1f7a3ef36241b5bccf8499a78a91ad2f4b9d77/Sources/CodexBar/Providers/Codex/CodexConsumerProjection.swift#L470) provides a monthly automatic meter for credit-only accounts, without inventing Session/Weekly windows or replacing available rate windows. Status-item render data and the live editor path use that automatic projection; Auto % and Usage bar honor used/remaining direction. This retains the later optional-credit visibility setting as well.

Fresh verification: `env -u CODEXBAR_ALLOW_TEST_KEYCHAIN_ACCESS CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --filter 'CodexConsumerProjectionTests|MenuBarLayoutRendererTests|StatusItemAnimationCodexCreditsTests'` passed **75 tests in 3 suites**. The relevant source and tests are unchanged from main. Existing regressions cover missing usage, windowless usage, automatic selection, and rate-window precedence. The full 933-selection suite also passed during this maintainer batch.

This closure is based on current source and executed regression tests, not a new real-account UI capture. Thank you for identifying the gap and preparing the original regression work.

PR #2660, opened 2026-08-04
feat: resizable settings sidebar with persisted width

steipete:
## Summary

Makes the settings sidebar resizable by dragging the divider, replacing the fixed 260pt width. The width clamps to 200–380pt and persists across launches (`settingsSidebarWidth` in UserDefaults).

## Approach

The `HStack` layout from `main` is untouched — the sidebar keeps its `.frame(width:)` (now state-backed) and the hairline stays the original `Divider()`. Dragging is handled by an **input-only** 12pt AppKit strip (`SidebarResizeHandle`) overlaid on the detail pane's leading edge, which sits exactly on the sidebar/content boundary. That placement keeps the strip out of the `HStack`'s width negotiation (a sibling `NSView` there breaks it), off the sidebar list's scroller, and means zero visual change to the chrome.

AppKit `mouseDown`/`mouseDragged` tracking is used instead of a SwiftUI `DragGesture`: window-coordinate deltas stay stable while the strip itself moves mid-drag, and no per-event relayout jitter.

The hover cursor is belt-and-braces: `.pointerStyle(.columnResize)` registers with SwiftUI's pointer engine on macOS 15+, and a tracking area asserts `NSCursor.resizeLeftRight` directly (the only mechanism on the macOS 14 floor). Notable gotcha uncovered while debugging: under `NSHostingView` the strip's `visibleRect` spans the whole window, so an `.inVisibleRect` tracking area never fires — the tracking rect must be explicit `bounds`.

Window sizing stays static: the 800pt window minimum minus the 380pt max sidebar still leaves a 420pt detail pane, so no dynamic min-width plumbing is needed (a test asserts this invariant).

## Screenshots

| Dragged narrow (200pt min) | Dragged wide |
| --- | --- |
| narrow | wide |

## Testing

- `swift test` — full suite green (three pre-existing timing-flaky tests in `CLIServeRouterTests`/`ClaudeOAuthPromptCoalescingTests` fail only under full-suite load; all pass in isolation).
- New unit tests: drag clamping/re-anchoring on `SidebarResizeHandleView`, sidebar bounds invariants.
- Live verification via scripted drags on the running app: drag +80pt persisted exactly 340; min/max clamps land on exactly 200/380; persisted width restored after relaunch; drags do not disturb controls under the strip.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 - CLI: expose Codex cost-history completeness in JSON and add an experimental provider-native-only scan mode (#2520). Thanks @NickGuAI!
 - Usage & Spend: add an accessible daily, weekly, and cumulative token-activity heatmap backed by the existing persistent cost scan cache (#2548). Thanks @Yuxin-Qiao!
+- Settings: the sidebar is now resizable — drag the divider between 200–380pt; the width persists across launches.
 - z.ai/GLM: route BigModel aliases and relay-file keys only to China endpoints, reject canonical cross-region overrides before bearer auth, and keep Kimi browser import disabled when Cookie Source is Off (#2351). Thanks @Leehow!
 - Kimi: enrich Code API and CLI usage with the monthly membership pool from a signed-in Kimi Desktop session, using WAL-safe read-only cookie access (#2351). Thanks @Leehow!
 - Sessions: discover live pi and OMP sessions through one Pi-family scanner, with dialect-aware metadata, PID-only startup rows, and mixed-version CLI/remote support (#2529). Thanks @wdmitchelluk!
--- a/Sources/CodexBar/PreferencesView.swift
+++ b/Sources/CodexBar/PreferencesView.swift
@@ -22,6 +22,9 @@ enum SettingsPane: Hashable {
     static let windowMinWidth: CGFloat = 800
     static let windowMinHeight: CGFloat = 540
     static let sidebarWidth: CGFloat = 260
+    static let sidebarMinWidth: CGFloat = 200
+    static let sidebarMaxWidth: CGFloat = 380
+    static let sidebarWidthDefaultsKey = "settingsSidebarWidth"
     static let detailMaxWidth: CGFloat = 780
 
     var title: String {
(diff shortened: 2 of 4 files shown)


steipete:
## Cursor fix — verified on a separate Mac, measured not eyeballed

The hover cursor was flashing and reverting. Root cause: the cursor was set imperatively (`NSCursor.set()` on mouse enter/move). SwiftUI re-renders the settings panes continuously, and every render resets the cursor, so an imperatively-set cursor survives only until the next frame. `NSWindow.disableCursorRects()` (the popular workaround) is the wrong lever — the reset does not come from cursor rects.

The fix is the declarative mechanism: `resetCursorRects()` + `addCursorRect(bounds, cursor: .resizeLeftRight)`, with `invalidateCursorRects(for:)` on `layout()` so the rect follows the strip as the sidebar resizes. AppKit re-asks the view for its cursor rects on every invalidation, so the cursor holds by construction. A `.cursorUpdate` tracking area backs it up.

### Measurement

Screenshots are a weak signal for cursor state, so this was measured instead. A probe harness hosts the real `SidebarResizeHandle` in a `PreferencesView`-shaped layout under a 20 Hz re-render loop (reproducing the condition that resets the cursor), samples `NSCursor.currentSystem` at 25 Hz, and is driven by an external pointer driver. Run on a separate Mac mini (macOS 26.6) so nothing touched the dev machine.

| | hover | drag |
| --- | --- | --- |
| before (imperative `NSCursor.set()`) | 0% resize | 40% resize (the flash) |
| after (cursor rects) | 100% resize | 100% resize |

Run-length encoding of the cursor state across a full hover → drag → leave sequence, twice:

```
run 1: [arrow x145, RESIZE x277, arrow x374]  transitions: 2
run 2: [arrow x153, RESIZE x416, arrow x227]  transitions: 2
```

Exactly two transitions — into the resize cursor on entry, back to arrow on exit — with no flicker in between, and the window never left with cursor rects disabled. Drag resized 260 → 330 pt in both runs.

Also worth recording, since it cost real debugging time: under `NSHostingView` a child `NSView`'s `visibleRect` spans the whole window rather than the view, so `.inVisibleRect` tracking areas never fire. The tracking rect must be an explicit `bounds`.

PR #2490, opened 2026-07-29
feat: multi-currency cost display with live exchange rates (Issue #2449)

Yuxin-Qiao:
## Summary

Adds support for displaying cost estimates in non-USD currencies (GBP £, EUR €, CNY ¥, JPY ¥, etc.), addressing [Issue #2449](https://github.com/steipete/CodexBar/issues/2449).

### New: CurrencyExchange service
- Cross-currency conversion via USD pivot (e.g., GBP→CNY)
- Live rates from open.er-api.com with 24h caching
- Hardcoded fallback rates for offline use

### Updated: UsageFormatter
- Smart `convertedCostString` with auto-follow or explicit currency modes
- All cost display paths now respect user's preferred currency

### Updated: Settings
- `preferredCurrencyCode` in SettingsStore with UserDefaults persistence
- Currency Picker in Preferences → General (Auto / USD / GBP / EUR / CNY / JPY / CAD / AUD / HKD / TWD / SGD / INR)
- Rate fetch triggers on app launch and settings change

### Updated: Views
- Provider cost sections (extra usage, API spend) now convert to preferred currency
- Menu bar layout cost strings follow currency preference
- Menu descriptor usage summaries (OpenAI, Claude Admin, OpenRouter, Poe)
- OpenRouter quota detail, Kiro overage cost, Amp credits

### Localization
- en + zh-Hans strings for `currency_title`, `currency_subtitle`, `currency_auto`

### Tests
- 43 tests pass including new cross-currency conversion tests

## How it works

When set to **Auto**, the app follows each provider's native currency (e.g., Anthropic extra usage in GBP, OpenAI in USD). When set to a specific currency, all cost values are converted through the exchange rate service.

```
API returns GBP → convert GBP→USD→CNY → display ¥
```

PR #2669, opened 2026-08-05
feat: add KRW to the preferred currency picker

kes02:
## Summary

Adds `KRW (₩)` to the preferred currency picker in Settings → General. KRW was the one major Asian currency missing from the list introduced in #2490 (refs #2449), so won-billed users convert USD estimates by hand to compare them against a card statement.

## Change

Three lines of source, no new exchange-rate logic:

- `CurrencyExchange.swift:18` — `"KRW"` joins `supportedCurrencies`, next to CNY/JPY rather than appended, matching the existing regional grouping.
- `CurrencyExchange.swift:30` — `"KRW": 1428.90` joins the hardcoded fallback table. Existing fallback values are untouched.
- `PreferencesGeneralPane.swift:85,105` — `PreferredCurrencyOption.krw` and its `"KRW (₩)"` label.

`fetchLatestRatesIfNeeded` already merges *every* rate in the `open.er-api.com/v6/latest/USD` payload, and that payload carries `KRW`. Listing the code in `supportedCurrencies` is what opens the `requiresLiveRates` gate (`CurrencyExchange.swift:116`), so selecting KRW triggers the same live-fetch and 24h-cache path as the existing currencies; the hardcoded rate is only read before the first successful fetch.

No public API surface change: `git diff origin/main -- '*.swift'` adds and removes no `public` line — `supportedCurrencies` gains an element, its signature is unchanged.

## Formatting

KRW is a zero-decimal currency, and `UsageFormatter.currencyString` delegates to `.currency(code:)`, which resolves ICU's fraction digits — so no formatter change is needed: $10 renders as `₩14,289`, $0.42 as `₩600`.

That path pins `Locale(identifier: "en_US")`, so the rendering is host-independent and the new no-fractional-separator assertion holds on a localized machine too — the failure mode #2651 fixed for Perplexity does not apply here.

## Validation

- `make check` — swiftformat clean, swiftlint 0 violations across 1809 files
- `make test` — 817 selections / 69 groups, 0 first-pass failures, 0 retries, 0 timeouts
- `swift test --filter UsageFormatter` — 46 tests pass, including the new assertions
- `git diff --check` — clean

New coverage in `UsageFormatterTests.swift`:

- USD → KRW conversion equals the table rate (`convert(usdAmount:to:)`)
- `₩` is rendered and no fractional separator is emitted
- `requiresLiveRates("KRW") == true`, so the live fetch is actually triggered rather than the currency silently sitting on the hardcoded rate

## Real behavior proof

Fresh bundle built from this branch at `44393a0c` (`make start` → release package → relaunch), macOS 26.5 (arm64).

**1. Settings → General — Preferred Currency resolves to `KRW (₩)`**
CodexBar Settings → General with Preferred Currency set to KRW (₩)

**2. Menu card — provider costs render as won**

Today `₩98,810`, 30d cost `₩1,884,614`, daily-chart peak `₩300,527` — thousands separated, no fractional separator, matching the zero-decimal expectation asserted in `UsageFormatterTests`.
CodexBar menu card showing today and 30-day costs in Korean won

**Rate provenance.** Selecting KRW opens the `requiresLiveRates` gate, so the displayed won comes from a live rate rather than the hardcoded `1428.90` fallback: the app fetched and cached `CodexBar.CurrencyExchangeRates["KRW"] = 1428.890254` with `CodexBar.CurrencyExchangeLastFetch = 2026-08-05 14:01:55 UTC`. Earlier in the same session, before that fetch returned, the Today row still rendered `$62.75`; once the rate landed it became `₩89,659` — exactly `62.75 × 1428.890254`. The screenshot above was taken a few minutes later, after more usage had accrued.


steipete:
Landed. Verified before merge: wiring matches existing currencies (membership-driven live-rate gate, pinned en_US_POSIX-style formatting via ICU so KRW renders zero decimals); tests assert both the ₩ symbol and absence of decimal separators. make check clean; UsageFormatter suite green. Partially addresses #2449 (adds the currency; billing-alignment remains open). Thanks @kes02!

PR #2692, opened 2026-08-06
Preserve MiniMax menu usage fill setting

VACInc:
## Summary

- adds regression coverage for MiniMax menu usage bars in both "As remaining" and "As used" modes
- verifies both 5-hour and weekly quota lanes

## Why

The former MiniMax-specific menu renderer hard-coded used-percentage fill, while the menu bar icon honored the global setting. The recent provider-detail migration removed that special renderer and now routes MiniMax through the shared menu metric path, correcting the behavior on `main`. This test locks in that fix so MiniMax cannot silently regress to the inverted menu display.

## Impact

MiniMax menu usage bars remain aligned with the Usage Fill setting, matching the menu bar icon.

## Validation

- `swift test --filter MiniMaxMenuCardModelTests`
- `make check`
- `make test` (821 selections, 69/69 groups passed)

No screenshot: test-only regression coverage for existing behavior on `main`.

### DIFF
--- a/Tests/CodexBarTests/MenuCardModelTests.swift
+++ b/Tests/CodexBarTests/MenuCardModelTests.swift
@@ -561,6 +561,71 @@ struct FactoryMenuCardModelTests {
 
 struct MiniMaxMenuCardModelTests {
     @Test
+    func `minimax menu metrics respect usage bar fill setting`() throws {
+        let now = Date()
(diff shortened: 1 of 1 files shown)


VACInc:
Closing because MiniMax now uses the shared menu-metric path, whose usage-fill behavior is already covered generically; provider-specific coverage would be redundant.

Issue #2637, opened 2026-08-04
CodexBar 0.46.0 main process grows past 7 GiB and requires daily restart


HaroldZhong:
## Summary

The main `CodexBar` process steadily grows in memory during normal menu-bar use. I have to fully quit and relaunch CodexBar about once a day to recover the memory.

Apple diagnostic reports captured repeated growth across separate launches. A later JetsamEvent named CodexBar as the largest process, using 7.31 GiB at capture time with an 8.87 GiB lifetime peak.

No diagnostic files are attached. The relevant measurements are transcribed below with machine-specific identifiers removed.

## Expected behavior

CodexBar's memory footprint should settle or remain reasonably bounded during normal background refreshes.

## Actual behavior

Memory keeps increasing over several hours until CodexBar consumes multiple gigabytes. Fully quitting and relaunching resets it, but the growth returns.

## Environment

- CodexBar 0.46.0 (build 110), installed through Homebrew
- macOS 26.5.2
- Apple Silicon Mac with 24 GB RAM
- OpenAI Web Access enabled
- Token-cost history enabled
- Adaptive refresh enabled
- Refresh providers when opening the menu enabled

CodexBar 0.47.0 was released after these measurements and has not yet been tested.

## Reproduction

1. Fully quit and relaunch CodexBar.
2. Leave it running with normal adaptive background refreshes and occasional menu opens.
3. Monitor the main `CodexBar` process over the following hours.
4. Its footprint continues to grow rather than plateauing.
5. Fully quitting and relaunching recovers the memory, but the issue repeats.

I have not isolated the growth to one provider or manual action.

## Evidence

Apple-generated diagnostic reports recorded:

| Capture | Report duration | Footprint | Maximum |
|---|---:|---:|---:|
| Launch 1, early | 7,855 s | 121.89 MB → 1,079.80 MB | 1,487.55 MB |
| Launch 1, middle | 14,492 s | 1,029.80 MB → 1,187.64 MB | 2,408.28 MB |
| Launch 1, later | 31,991 s | 1,180.59 MB → 2,532.62 MB | 6,551.66 MB |
| Launch 2 | 78,619 s | 176.34 MB → 5,621.97 MB | 8,628.98 MB |

A subsequent JetsamEvent recorded:

- `largestProcess: CodexBar`
- Current footprint: 7.31 GiB
- Lifetime maximum: 8.87 GiB
- Process state: active

[...]

PR #2703, opened 2026-08-06
Align cost cache save overshoot with documented load cap

Yuxin-Qiao:
## Summary

Follow-up to the review on #2637's merged fix (#2646, commit `90c6faed0`). Addresses all points raised in the review comment and in the ClawSweeper re-review:

1. **`dayDate` duplicated `parseDayKey` and skipped calendar normalization** — removed the private `dayDate` helper; `previousReportForCatchUp` now calls the existing `CostUsageScanner.parseDayKey`, which normalizes via `localGregorianCalendar(matching:)` and anchors at noon. A day key like `"2026-06-01"` now parses to 2026 even when `Calendar.current` is non-Gregorian (e.g. Buddhist calendar would previously resolve to 1483).
2. **`save` overshoot invariant now matches the documented load cap** — updated the `maxCacheLoadBytes` doc comment and the guard-fail comment so the behavior matches the contract: `save` bounds artifacts to `maxCacheFileBytes`, and when protected entries (resuming sessions, fork parents) cannot be trimmed further it may overshoot only up to the load cap. Anything above the cap is dropped as legacy/foreign.
3. **Weak constant-only test replaced with a real round trip** — `codex load cap keeps headroom over the save budget with a real save load round trip` now exercises `save`/`load` and asserts the overshoot artifact is written and readable within the load cap. Also added `catch up report honors a non-gregorian system calendar` as a regression guard for the `parseDayKey` fix.
4. **Preserve the preceding producer keys (ClawSweeper P2)** — the generated parser hash rotated to `d3df003a9bff32af`, so both the current-main key (`codex:cu:p1cd29792d9ca2b11`) and the prior bounded-cache key (`codex:cu:p37aedd661c4272a8`) were added to `compatibleCodexProducerKeys`. Upgrading from either preceding build no longer rejects valid local caches or forces a rebuild. `current codex cache accepts calendar normalization predecessors` covers both keys through default `load` and `loadCodexForMigration` with a real save/load round trip.

Also regenerated `CodexParserHash.generated.swift` (repo requires it after touching `Vendored/CostUsage`) and updated the hardcoded line anchors in `ProviderArchitectureGatekeeperTests` for the shifted source.

## Real behavior proof (existing local cache upgrade)

Ran an after-fix upgrade check against the existing local Codex cost cache (11.7 MB artifact, 304 session files, 23 days of usage data). The artifact was stamped with the current-main producer key (`codex:cu:p1cd29792d9ca2b11`) to model the upgrade source, then loaded with this head's default `load` and `loadCodexForMigration`:

```
[UPGRADE-PROOF] sourceArtifactBytes=11665828 sourceFiles=304 sourceDays=23
  loadedFiles=304 loadedDays=23 lastScanUnixMsPreserved=true
  migrationFiles=304 migrationIncompatibleCache=nil
  foreignKeyRefused=true
```

- Default `load` reuses the full artifact: all 304 files and 23 days survive, `lastScanUnixMs` is preserved — no rebuild.
- Migration `load` also reuses it (`incompatibleCache == nil`).
- Control: an artifact stamped with a producer key outside the compatibility set is still refused (`foreignKeyRefused=true`), confirming the allowlist is what grants reuse rather than unconditional acceptance.

## Commands run

- `swift test --filter CostUsageCacheTests` — 39/39 pass
- `swift test --filter ProviderArchitectureGatekeeperTests` — 38/38 pass
- `make test` — full suite passes
- `make check` — SwiftFormat/SwiftLint clean (0 violations)
- `Scripts/regenerate-codex-parser-hash.sh`

## Notes

Branch rebased onto current `origin/main` (`005a71f55`); PR is mergeable. The calendar reachability for `dayDate` in the original review was upstream-normalized in the app path (`saveCodexCache` passes `range.calendar`, already Gregorian); this change removes the latent bug at the API level anyway, matching the review's recommendation. No behavior change for Gregorian-calendar users.


Yuxin-Qiao:
@clawsweeper re-review

Addressed the P2: `codex:cu:p37aedd661c4272a8` is now in `compatibleCodexProducerKeys` and the upgrade-load regression test covers default `load` plus `loadCodexForMigration` with a real save/load round trip. PR body now includes the runtime proof.

Yuxin-Qiao:
@clawsweeper re-review

Rebased onto current main (`005a71f55`) and added both preceding producer keys (`p1cd29792d9ca2b11`, current main; `p37aedd661c4272a8`, prior bounded-cache build) to `compatibleCodexProducerKeys`. The upgrade regression test now round-trips both keys through default `load` and `loadCodexForMigration`. PR body has the runtime proof and command outputs.

Yuxin-Qiao:
@clawsweeper re-review

Added the real upgrade proof: took the existing local Codex cost cache (11.7 MB, 304 session files, 23 days), stamped it with the current-main producer key, and loaded it with this head via default `load` and `loadCodexForMigration`. Both reuse the full artifact (`loadedFiles=304 loadedDays=23 lastScanUnixMsPreserved=true migrationIncompatibleCache=nil`); a foreign-key control artifact is refused. Details in the PR body.

steipete:
Landed. Verified before merge: confirmed the (256 MiB, 320 MiB] overshoot band is reachable only via protected entries and that this PR’s alignment is documentary — the never-persist-refusable invariant from #2646 is untouched and now pinned by a real save/load round trip. The calendar fix is a genuine latent bug (day keys misresolved under non-Gregorian system calendars). Maintainer fix pushed before merge (`589e89d5a`): the producer-compat list was missing two shipped predecessor hashes (`6c0f1f…`, `aa27d2…`), which would have forced pointless rebuilds for post-#2646/pre-#2691 builds; added both, regenerated the hash, extended the compat regression to all four keys. CostUsageCacheTests 39/39, gatekeeper 38/38, autoreview clean. Thanks @Yuxin-Qiao!

PR #2710, opened 2026-08-06
Fix Codex quota refresh after manual reset

endless7:
## Summary

- accept a confirmed early weekly reset when the same Codex reset credit transitions from `available` to `redeemed`
- preserve the existing premature-reset guard when there is no matching redemption evidence
- add regression coverage for manual reset credit redemption before the previous natural weekly boundary

## Root cause

`CodexWeeklyResetConfirmation` rejected every reset observed more than two minutes before the previously scheduled weekly boundary. A manual reset credit intentionally moves that boundary early, so the app kept preserving the stale pre-reset snapshot even after two matching low-usage samples.

The merged confirmation path treats a provider-reported redemption of the same stable reset credit ID as evidence that the early boundary change is legitimate. All other boundary and two-sample confirmation checks remain in place.

## User impact

For providers that retain the consumed credit as `redeemed`, Codex quota usage refreshes automatically after a user redeems a manual reset credit, without weakening protection against transient or premature zero-usage responses.

## Validation for the merged change

- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift test --filter CodexWeeklyResetConfirmationTests` (16 tests passed)
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer make check`
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer make test` (821 selections across 69 groups, 0 failures)

No UI changes; screenshots are not applicable.

## Post-merge provider finding

This PR was merged at head `288380d52eb75e49c93fdc7754f877a25b0ac2aa` before the following provider trace was recovered. The follow-up described here is **not part of this merged PR**.

Observed on CodexBar 0.47.0 with a signed-in Codex OAuth account. Account identity, tokens, and reset-credit IDs are omitted.

| Captured at (UTC) | Weekly used | Reset boundary (UTC) |
|---|---:|---|
| 2026-08-06 08:59:18 | 98% | 2026-08-10 01:00:26 |
| 2026-08-06 09:28:18 | 99% | 2026-08-10 01:00:26 |
| 2026-08-06 09:33:18 | 0% | 2026-08-13 09:33:18 |

The successful reset-credit inventory persisted with the post-redemption observation was:

```json
{"availableCount":0,"credits":[]}
```

A later read-only live OAuth fetch returned the same successful empty inventory. In this observed provider shape, the consumed credit is omitted instead of retained as a `redeemed` row, so the merged same-row transition remains conservative and may still preserve the stale snapshot.

Follow-up commit `endless7/CodexBar@78035f4f` handles both representations, but only when two independent low-usage inventories corroborate consumption and the available count decreased. It has 18 focused tests, a clean `make check`, and 821 full-suite selections across 69 groups with no failures or retries. The account has no reset credit remaining, so this validation replays the exact redacted timeline through production decision code; it is not presented as a second live redemption.

### DIFF
--- a/Sources/CodexBar/Providers/Codex/CodexWeeklyResetConfirmation.swift
+++ b/Sources/CodexBar/Providers/Codex/CodexWeeklyResetConfirmation.swift
@@ -143,7 +143,12 @@ struct CodexWeeklyResetConfirmation: Sendable {
                previousWeekly,
                capturedAt: previous.updatedAt)
         {
-            if confirmation.updatedAt < previousBoundary.addingTimeInterval(-2 * 60) {
+            let confirmsManualReset = Self.confirmsManualResetCreditRedemption(
+                previous: previous,
+                confirmation: confirmation)
+            if confirmation.updatedAt < previousBoundary.addingTimeInterval(-2 * 60),
+               !confirmsManualReset
+            {
                 return .preservePrevious
             }
             guard initialBoundary.timeIntervalSince(previousBoundary) >= Self.resetEquivalenceToleranceSeconds,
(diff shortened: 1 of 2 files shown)


steipete:
Landed. Verified before merge (independent maintainer-agent review): root cause confirmed in `CodexWeeklyResetConfirmation.confirmationDecision` (the pre-boundary guard rejects legitimately early boundaries after a manual reset); the exemption is narrow — same normalized credit ID transitioning available→redeemed with non-regressing inventory timestamps — and self-limiting (cannot fire twice off one redemption). All other safeguards intact; missing credit inventory falls back to the conservative preserve path. Checked against the #2600 duration-based lane classification: no stale primary=session assumptions. Local: CodexWeeklyResetConfirmationTests 16/16, gatekeeper 38/38, make check clean. CI green at merge. Thanks @endless7!

endless7:
@clawsweeper re-review

Issue #2668, opened 2026-08-05
CodexBar keeps asking to access key "Claude Code-credentials" in my keychain


XIIIIK:
Hi developers, I am using a MacBook Pro with macOS version Tahoe 26.4.1. I updated the CodexBar to the latest version. It keeps asking to access key "Codex Code-credentials" in my keychain every time I log in to my laptop.

viniciusls-dev:
Here it keeps me asking every 30min otherwise it doesn't show the consumption. I've tried disabled and re-enabling the Keychain integration (because without that it doesn't work at all) but none solved.

LukebethamStonehaven:
<img width="310" height="161" alt="Image" src="https://github.com/user-attachments/assets/c4a52f91-4cb0-4692-afcc-5651e03f420c" />

Yeh same since last update I can't seem to authenticate with claude anymore - I am logged in in terminal, browser and claude code desktop app and it continually just says:

OAuth credentials not found

On Tahoe 26.6

gaborpelesz:
I found a workaround (or maybe the required solution) which is to set the **Keychain prompt policy** from `Only on user action` to `Always allow prompts` in the CodexBar Claude settings.

<img width="1005" height="200" alt="Image" src="https://github.com/user-attachments/assets/6f611db9-a291-40a2-bd80-2cae19f472c5" />

viniciusls-dev:
I found another workaround - changing `Usage source` from `OAuth` to `CLI` 😅 It has been working well since yesterday

steipete:
Fixed on main by #2675 (commit 0954a74a0b9de421c73142147f8e1afc5510f92f), implementing #2634. Claude Code credential access now requires explicit, default-off consent, and automatic refresh can fall back to the Claude CLI without reading the protected Keychain item. ClaudeOAuthDirectKeychainReadConsentTests covers both default-off consent and blocked access when disabled. This shipped in v0.49.0 and remains included in v0.55.1. Closing as completed; please reopen with a redacted current-version reproduction if this still occurs.

Issue #2588, opened 2026-08-02
Continual prompting for password


mdgld:
On 16” macbook pro M3 pro, MacOS 26.5.2, as of latest update I keep getting prompts to enter keychain password for claude. I’ve selected “always allow” multiple times but keep getting prompted.

mikirejf:
Same here. Lately it has been getting even worse.

steipete:
Fixed on main by #2585: CodexBar no longer reads Claude-owned credentials from the Keychain and routes refresh through the owning Claude CLI, which removes the recurring ACL-prompt mechanism (Claude Code replacing its item/ACL on refresh invalidated every grant). Please reopen with the requesting-process and item names if a build containing that merge still prompts.

mdgld:
@steipete thanks Peter, will give that a try.

PR #2718, opened 2026-08-06
Add Claude setting to suppress Keychain access explanation

MGidnian:
## Summary

- Add a Claude setting, `Show Keychain access explanation`, with an explicit opt-out.
- Suppress CodexBar's explanatory pre-alert for Claude OAuth Keychain access when the setting is off.
- Preserve the existing explanation-by-default behavior for installations that have not set the preference.
- Add a separate result-capable prompt callback so the app can report whether an explanation was actually shown,
  without changing the public `KeychainPromptHandler.handler` callback signature.
- Preserve explicit legacy handler overrides ahead of the default result-capable route.
- Preserve the native macOS Keychain prompt and all OAuth/Keychain behavior.
- Add persistence and prompt-gating tests and document the distinction between the two prompts.

## Motivation

When refreshing Claude usage, CodexBar can show its explanatory alert before macOS asks for access to the Claude OAuth Keychain item. After the user has already granted access with “Always Allow”, seeing the explanatory alert again is cumbersome. The existing cooldown is temporary and does not provide a durable preference for an already-authorized setup.

This change makes the common already-authorized flow quiet after the user disables the explanation in Claude settings,
while preserving the current explanation-by-default behavior for upgrades. That avoids silently changing an existing
installation's security/education flow; new or existing users can opt out explicitly once they understand the prompt.

## Scope and safety

This only suppresses CodexBar's explanatory alert for the `.claudeOAuth` prompt context. It does not disable Keychain
access, change `ClaudeOAuthKeychainPromptMode`, bypass native macOS authorization, or alter OAuth refresh behavior.
The native macOS authorization prompt remains controlled by macOS. The additive result callback keeps the existing
six-hour cooldown semantics correct: only a displayed CodexBar explanation acknowledges the cooldown, while existing
CodexBarCore clients can continue assigning the public `Void` callback.

## Follow-up UX option

If maintainers prefer an inline acknowledgement, the same preference could later be exposed directly in this alert with an option such as “Don't show this again.” I kept this PR settings-first so the preference is explicit, discoverable, and consistent with the existing Claude authentication controls.

## Verification

- `make check`
- `make test` (823 selections, 69 groups, all passed)
- `swift test --filter 'KeychainPromptCoordinatorTests|SettingsStoreCoverageTests|ClaudeOAuthKeychainPreAlertGateTests'`
  (59 tests, all passed)
- The focused tests cover the unset/default state, explicit opt-out state, persistence, and the false handler result
  used by suppression, plus precedence for legacy public handler overrides; the full suite covers the unchanged native
  Keychain integration paths.
- The production app compilation path built successfully during validation. Complete packaging was attempted with
  `./Scripts/compile_and_run.sh`, but the separate Widget extension `xcodebuild` stalled in the locked CI-like local
  environment; no Keychain credentials or native authorization prompts were touched.
- UI screenshot capture was not possible because the Mac was locked during the runtime check. No live provider or
  Keychain reads were performed.

## Screenshots

Not included: this reuses the existing provider settings UI and changes the visibility of the existing explanatory flow
rather than adding a new visual component. The focused and full test commands above are the available redacted terminal
evidence for both preference states.

## Related

Related to #2588 and #2668.

### DIFF
--- a/Sources/CodexBar/CookieHeaderStore.swift
+++ b/Sources/CodexBar/CookieHeaderStore.swift
@@ -72,7 +72,7 @@ struct KeychainCookieHeaderStore: CookieHeaderStoring {
         if case .interactionRequired = KeychainAccessPreflight
             .checkGenericPassword(service: self.service, account: self.account)
         {
-            KeychainPromptHandler.handler?(KeychainPromptContext(
+            KeychainPromptHandler.notify(KeychainPromptContext(
                 kind: self.promptKind,
                 service: self.service,
                 account: self.account))
--- a/Sources/CodexBar/CookieHeaderStore.swift
+++ b/Sources/CodexBar/CookieHeaderStore.swift
@@ -72,7 +72,7 @@ struct KeychainCookieHeaderStore: CookieHeaderStoring {
         if case .interactionRequired = KeychainAccessPreflight
             .checkGenericPassword(service: self.service, account: self.account)
         {
-            KeychainPromptHandler.handler?(KeychainPromptContext(
+            _ = KeychainPromptHandler.handler?(KeychainPromptContext(
                 kind: self.promptKind,
                 service: self.service,
                 account: self.account))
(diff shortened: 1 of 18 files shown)


MGidnian:
@clawsweeper re-review

MGidnian:
@clawsweeper re-review

MGidnian:
@clawsweeper re-review

PR #2726, opened 2026-08-07
fix: make Codex cost catch-up append-linear

xx205:
## Summary

Make bounded Codex usage catch-up append-linear while preserving the accounting and fork-safety guarantees added by #2452, #2525, and #2648.

This change:

- persists token accumulator state and append-only usage events in a SQLite sidecar;
- publishes immutable usage-row generations through the existing JSON cache authority;
- validates append reuse with file identity, ctime, size, and prefix anchors;
- preserves fail-closed deferred replay for unresolved forks and subagents;
- consumes published row references and suffix rows instead of rebuilding all historical workspace rows on every pass;
- retires replaced SQLite row generations from the moment they become unreferenced, so readers that selected the previous JSON publication retain a safe grace window;
- serializes app, account, and CLI cache writers with a cross-process lock; and
- keeps Finish Now progress monotonic while retrying transient contention or no-progress passes.

The existing automatic limits remain unchanged: at most 256 MiB per file and 512 MiB per pass, followed by the existing background duty cycle.

## Problem

#2452 bounded scans of very large Codex session corpora. #2525 restored accurate fork accounting during bounded catch-up, and #2648 preserved ordinary-fork progress when a session file was appended.

The remaining slowdown came from older cache architecture shared by those paths: after reading each new bounded JSONL suffix, the scanner rebuilt increasingly large in-memory token snapshots, checkpoints, fingerprints, and workspace usage rows from all accumulated history. Work therefore grew with every pass and approached quadratic behavior. Finish Now exposed that behavior by running passes back-to-back; it did not create it.

There were also narrow resume gaps around metadata-free files, predecessor producer generations, and bounded metadata parsing. Those could force redundant work or make an otherwise reusable cache generation unavailable.

### Append-linear token index

The token sidecar stores the accumulator state needed at the next byte offset plus append-only usage events. A pass reads and parses only the new JSONL suffix when the stored prefix is still valid.

Append reuse is accepted only after validating the physical file identity, ctime/size relationship, and sampled prefix anchors. Ambiguous ownership or content changes still use the conservative byte-zero reconciliation path.

### Transactional usage-row publication

The workspace sidecar stores immutable usage-row generations. Compact JSON cache state remains the publication authority, so a crash may leave SQLite ahead of JSON but cannot expose an unpublished partial generation.

Per-entry producer keys allow compatible cached rows to survive a global producer migration. The first append after migration imports the predecessor generation and then processes only the JSONL suffix.

### Reader-safe generation retirement

Garbage collection now records when a generation first becomes absent from the complete published JSON reference set. It deletes that generation only after it has remained unreferenced for the full 24-hour grace period, rather than measuring age from generation creation time.

The SQLite schema migrates from v1 to v2 in place while the refresh lease is held. A generation selected by a concurrent reader therefore remains loadable across a replacement publication, and republishing a retired generation clears its prior retirement deadline before the next sweep.

### Fork and subagent safety

Resolved forks reuse validated inherited baselines. Missing-parent forks and subagent buffers remain deferred and fail closed: the scanner does not publish incomplete suffix accounting and does not silently fall back to counting inherited cumulative totals as new usage.

### Finish Now coordination

Finish Now bypasses the normal scanner debounce, retries transient lock contention and no-progress passes, and identifies metadata-only progress without resetting the visible byte count. Cross-process locking also provides a stable cache-clear barrier.

## Privacy-safe representative validation

I ran Finish Now-style bounded catch-up against an isolated projection of real local Codex JSONL data. The projection was copied before the run, and the original session files were not modified.

| Check | Result |
| --- | --- |
| Representative corpus | approximately 3 GiB across 10–99 files |
| Covered shapes | current session family, large file, resolved fork, missing parent, subagent, and small files |
| Bounded Finish Now catch-up | 38 passes in 79 s |
| Independent reference scan | 6 passes in 78 s |
| Accounting parity | bounded result = persisted-restart result = independent reference result |
| Compact 478 MiB large-prefix case | bounded: 2 passes / 17 s; one-shot: 2 passes / 18 s; exact result match |
| Sustained throughput | first quartile 38 MiB/s; last quartile 39 MiB/s; ratio 1.04 across 33 samples |
| Stable EOF replay | zero JSON body work, zero usage-row work, and unchanged sidecars |
| Peak derived storage | SQLite 16 MiB; JSON cache below 1 MiB |

No paths, usernames, session IDs, dates, model names, token totals, prompts, responses, or JSONL content are included in this proof.

The representative run validates the core indexing and publication design. After the final narrow edge-case fixes, the exact submitted tree was also checked with deterministic regressions for:

- metadata-free partial files using 1 KiB slices: 12 monotonic passes to EOF, followed by a zero-work replay;
- predecessor SQLite row generations: stable reads import with zero JSON body work, the first append is suffix-only and matches a cold control, and the following pass does zero work;
- nil-session rowless EOF state: current generations do zero work and predecessor generations migrate once; and
- oversized leading metadata records plus bounded reads ending exactly at immutable EOF.

The reviewer-reported reader/GC interleaving also has a deterministic, privacy-safe before/after proof using two row-store instances sharing a synthetic database:

| Step | Before fix | After fix |
| --- | --- | --- |
| Reader selects generation A; writer publishes B and immediately sweeps | A deleted (`deletedGenerationCount = 1`) | A retired but retained (`deletedGenerationCount = 0`) |
| Reader imports its already-selected A reference | `needsRebuild` | exact synthetic row match |
| Sweep after A has been unreferenced for more than 24 hours | already missing | A deleted (`deletedGenerationCount = 1`) |

Additional regressions verify that v1-to-v2 migration preserves published rows and that republishing a retired generation restarts its grace window. All identifiers and token values in these tests are synthetic.

## Verification

- `swift test --jobs 2 list` prebuild, followed by the repository's isolated sharded test runner: 834 test selections in 70 groups; 70/70 passed on the first attempt; 0 failures, retries, or timeouts; 806.8 s total.
- 142 focused row-store, cache, workspace-sidecar, scanner, and token-index regression tests passed.
- `ProviderArchitectureGatekeeperTests`: 38/38 passed, including the exact-anchor and provider-reference fingerprint scan added on current `main`.
- `make check`: portable repository checks, parser hash, SwiftFormat, and strict SwiftLint all passed; 0 SwiftLint violations across 1,821 files.
- `git diff --check` passed.
- Compile and test concurrency was capped at two jobs.

## Scope and limitations

- A cold corpus must still read every relevant JSONL byte once.
- Automatic byte limits and duty-cycle behavior are unchanged.
- This does not change token accounting or pricing formulas.
- The representative proof intentionally uses approximately 3 GiB rather than retaining or publishing a full 42+ GiB local corpus.
- Ambiguous duplicate physical ownership still chooses correctness over speed and performs byte-zero reconciliation.
- If SQLite is transiently unavailable, publication is deferred instead of publishing a partial suffix.
- The progress denominator may legitimately grow when an active file is appended or a previously missing parent becomes discoverable.

No UI strings or layouts changed, so screenshots are not applicable.


steipete:
Decision on this PR — and it's the good kind of decline. Your analysis convinced the owner to go *further* than the sidecar: the verified review of this PR (quadratic catch-up confirmed at the old head, your linear-work regression proof, invariants preserved) triggered the call to migrate Codex cost storage to **a single SQLite database entirely**, deleting the JSON artifact and its whole bounding/pruning/producer-key machinery rather than maintaining a dual store. The monolithic decode/encode is the root of every incident this subsystem produced this month (#2637, #2646, #2703), and your append-state design showed the way out.

So this PR won't merge as-is — not because anything in it is wrong, but because "JSON stays publication authority" is the part we're removing. The migration series is starting immediately (tracking issue follows shortly, linking here); your accumulator schema and linearity test design are being carried into it with credit. You're the person who understands this subsystem best from the outside — review on the series would be genuinely valued, and if you'd rather build any phase of it yourself, say so on the tracking issue and it's yours.

Thanks for a rigorous PR — fourth one in this area, and each one has moved the architecture forward.

PR #2735, opened 2026-08-07
Fix custom menu bar reset countdown mismatch

hyuntaedotkim:
## Summary

  - use the exact display time for custom menu bar reset countdowns so they match the opened menu
  - preserve attributed-title cache reuse by keying reset values on their formatted display text
  - add regression coverage for the one-minute mismatch and the unchanged-title cache path
  - document the fix in the changelog

  ## Root cause

  The opened menu formatted reset times from the current instant, while the custom menu bar layout rounded that instant down to the start of the wall-clock minute before calling the same formatter.

  Because the countdown formatter rounds remaining seconds up, this could make the menu bar show one additional minute. For example, 5m 59s remaining appeared as `in 6m` in the menu but as `in 7m` in the menu bar.

  Passing the exact `Date` directly through the existing renderer cache key would invalidate the cache on animation ticks. The cache now keys reset values on their formatted countdown and absolute strings, so it changes only when the visible reset text changes.

  ## User impact

  Custom reset countdown tokens in the menu bar now stay aligned with the opened menu while retaining the existing refresh scheduling and rendering cache behavior.

  ## Validation

  - `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --filter MenuBarLayoutRendererTests` — 14 tests passed
  - `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --filter 'cross provider case clusters are derived or specifically justified'` — 1 test passed
  - `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 make check` — passed
  - `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 make test` — 824 selections across 69 groups; 0 failures, retries, or timeouts

  ## Screenshots

  - Before: 
Screenshot 2026-08-07 at 8 44 51 PM

  - After: 
Screenshot 2026-08-07 at 10 04 00 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 0.48.1 — Unreleased
 
+### Fixed
+- Menu bar: keep custom reset countdowns aligned with the opened menu instead of rounding their clock back to the start of the wall minute.
+
 ## 0.48.0 — 2026-08-06
 
 ### Added
--- a/Sources/CodexBar/MenuBarLayoutRenderer.swift
+++ b/Sources/CodexBar/MenuBarLayoutRenderer.swift
@@ -49,14 +49,35 @@ struct MenuBarLayoutRenderOptions: Hashable {
     let showUsed: Bool
     let appearanceName: String
     let isDebugApp: Bool
-    /// Minute-granularity clock. Countdown tokens refresh without invalidating cached titles every tick.
+    /// Exact display clock. The cache keys on the resulting reset strings, so animation ticks that keep
+    /// the same visible countdown still reuse the cached title.
     let now: Date
 }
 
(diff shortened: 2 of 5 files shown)


steipete:
Landed. Verified before merge: the custom-layout `.resetCountdown` token rounded 'now' back to the wall-minute boundary while the opened menu used the exact instant — a one-minute visible divergence. Fix changes only the display clock, preserves the #2600 duration-aware lane selection, and keys the render cache by rendered reset strings so unchanged minutes still hit cache. 119 renderer/reset/window/gatekeeper tests green, make check clean, CI green. Thanks @hyuntaedotkim — clean first contribution!

PR #2741, opened 2026-08-07
Kimi: hide Code 7-day row when it duplicates the weekly lane

Yuxin-Qiao:
## Summary
- Kimi Code accounts were showing the same 7-day quota twice: once as the primary "Weekly" lane and again as a "Code 7-day" extra window.
- Verified against a live Kimi account (read-only): the `GetUsages` FEATURE_CODING detail and the `GetSubscriptionStats` `ratelimitCode7d` field report the same 7-day Code quota (identical usage counts and reset timestamps, e.g. 29/100 requests resetting at the same instant).
- Hide the Code 7-day extra window only on positive evidence that it duplicates the primary weekly lane: the primary lane must be reliable (unreliable counters withhold `windowMinutes`), used percent must be within 1pt, AND both lanes must report reset timestamps within 5 minutes. Two missing resets no longer count as a match, so genuine separate quotas still render.

## Review follow-up (353531681)
- Addressed both ClawSweeper findings on the suppression helper:
  - [P2] Preserve the Code quota for unreliable weekly counters: a numeric weekly limit with absent/malformed counters produces a 0% primary window with no duration; that unreliable lane can no longer suppress the subscription Code quota.
  - [P2] Require reset evidence before treating quota lanes as duplicates: when either lane omits a reset timestamp the Code row is kept, so coincidentally similar percentages cannot hide an independent quota.
- Regression tests:
  - numeric-limit fallback with missing counters (unreliable 0% primary lane, matching 0% ratio, no resets on either side) keeps the Code row;
  - matching percentages with both reset timestamps missing keeps the Code row.

## Test
- `swift test --filter "code weekly window"`: 3 passed (hide-on-match, keep-on-unreliable, keep-on-missing-resets).
- `swift test --filter KimiProviderTests`: 73 passed.
- `make check`: 0 violations.
- Real-app proof: rebuilt app shows only `kimi-monthly` as the extra window; the duplicate Code 7-day row no longer renders.

## Official naming alignment (36e8981f)
- Kimi lanes now match the official Kimi "Usage Progress" descriptions and order: `5-hour usage` (secondary), `7-day usage` (primary), `Total usage` (subscription pool). The menu already rendered in this order; the previous generic "Rate Limit"/"Weekly"/"Monthly" labels are replaced with the official terms, including translations in all app locales.
- Confirmed the official `GetSubscriptionStats` API returns the total-quota breakdown: `subscriptionBalance.amountUsedRatio` is the total quota usage (official UI: 93.81%) and `subscriptionBalance.kimiCodeUsedRatio` is the Code share of that total (official UI tooltip: Kimi Code usage 73.68%). `KimiSubscriptionBalance` now decodes `kimiCodeUsedRatio` and the parsing test asserts it (0.2854 from the live fixture).
- Updated menu-card, widget, and CLI golden expectations for the new labels; 217 focused tests pass and `make check` reports 0 violations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 0.48.1 — Unreleased
 
+### Fixed
+- Kimi: hide the Code 7-day window when it reports the same percentage and reset time as the primary weekly quota, so the menu no longer shows one allowance twice.
+
 ## 0.48.0 — 2026-08-06
 
 ### Added
--- a/Sources/CodexBar/Resources/ar.lproj/Localizable.strings
+++ b/Sources/CodexBar/Resources/ar.lproj/Localizable.strings
@@ -1430,3 +1430,6 @@
 "Enter every endpoint setting to review the exact network origins." = "أدخل جميع إعدادات نقاط النهاية لمراجعة أصول الشبكة الدقيقة.";
 "Approve" = "موافقة";
 "Install" = "تثبيت";
+"5-hour usage" = "الاستخدام خلال 5 ساعات";
+"7-day usage" = "الاستخدام خلال 7 أيام";
+"Total usage" = "إجمالي الاستخدام";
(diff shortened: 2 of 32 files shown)


steipete:
Landed. Verified before merge: the suppression predicate requires reliable primary counters, percentage agreement within one point, AND both resets within five minutes — distinct windows (different reset or >1pt gap) keep the Code row. 88 Kimi tests across 14 suites green, localizations checked in en/zh/de, make check clean, CI green. Thanks @Yuxin-Qiao!

PR #2743, opened 2026-08-07
Fix remaining percentage display

VACInc:
## Summary

- Display the percentage already normalized for the selected style.
- Keep **As used** unchanged and make **As remaining** show the remaining value.
- Add regression coverage for both styles.

## Root cause

`Metric.percent` is already expressed in the selected style, but `linePresentation` inverted remaining values a second time.

## Screenshots

| Before | After |
| --- | --- |
| Before: As remaining labels show 32% and 24% | After: As remaining labels show 68% and 76% |

## Checks

- `make check`
- `make test` — 824 selections across 69 groups, 0 failures


steipete:
Reviewed and verified — the double-inversion diagnosis was correct. While this was in review, #2744 landed the same fix with the explicit left/used suffixes and changelog coverage, so closing this as superseded rather than asking for a rebase. Thanks @VACInc — your IBM Bob provider PR (#2690) still has the host-validation ask open and is the one to focus on.

VACInc:
Thanks for reviewing and closing this as superseded. For clarity, we believe the host-validation ask on #2690 is resolved in `92fec1d0`: it validates the parsed host before forwarding credentials and includes hostile-domain regression coverage. I followed up on #2690 to confirm whether you see any specific remaining gap.

Issue #2733, opened 2026-08-07
Claude usage: opt-in live window feed from Claude Code's statusLine JSON — needs a ruling on AGENTS.md:47


luisgonzaleznf:
## Summary

Claude is the only provider whose live sources are both budget-gated, which makes its card structurally stale rather than occasionally wrong. Claude Code ≥ 2.1.80 exposes a server-authoritative `rate_limits` object that costs nothing to read, but the only channel it comes through is the `statusLine` command — and `AGENTS.md:47` says never to rely on the status line for usage parsing.

I think there's a real distinction between what that rule forbids and what this would do, but it's your rule and your call. **I'd rather get a ruling before writing code than send you a surprise PR that contradicts a documented line.** Happy to implement whichever way you want it, or drop it.

## The problem is staleness, not blanking

#2247 (`ff52c6d`) fixed the blanking half of #2241 — transient parse failures now preserve the last-good snapshot and its timestamp. That part works.

What's left is freshness, and it's structural rather than a parser bug. Both live sources carry the same 5-minute cooldown:

- `ClaudeOAuthUsageRateLimitGate.swift:6` — `defaultCooldown: TimeInterval = 60 * 5`, keyed per access token, armed whenever `/api/oauth/usage` 429s
- `ClaudeCLIRateLimitGate.swift:5` — same cooldown for the PTY path

Both gates are correct given what they're gating. `api.anthropic.com/api/oauth/usage` has a tight budget and 429s under polling; the PTY path pays a subprocess plus session-artifact cleanup per probe. The issue history is what that looks like from the outside: #1679 (rate limited), #2117 (slow refresh), #2251 (session-list pollution, since mitigated by the `ClaudeProbe` artifact cleanup), and the 8.6% probe success rate measured in #2241.

CodexBar has no Claude source that is cheap enough to read often. Every other fix is scheduling against a fixed budget.

## The one free channel

Claude Code ≥ 2.1.80 passes a JSON object on stdin to whatever `statusLine` command is configured. It contains:

```json
{
  "rate_limits": {
    "five_hour": { "used_percentage": 42.7, "resets_at": 1786099200 },
    "seven_day": { "used_percentage": 63.1, "resets_at": 1786608000 }
  }
}
```

This is piggybacked on Messages API responses the session already made, so reading it spends nothing against the usage endpoint's budget. It appears only for Claude.ai-subscriber sessions, and only after the session's first API response.

[...]

PR #2747, opened 2026-08-07
Claude: report the terminal verdict when no refresh can restore the profile

luisgonzaleznf:
Fixes the user-facing half of #2733: a Claude profile that no refresh can restore is told to click Refresh, which sends the user round a loop that always lands back in the same place.

## The loop

CodexBar already has the right message for this state — `delegatedRefreshFailureMessage` returns *"Claude OAuth credentials expired and CodexBar cannot read them back… Switch Claude Usage source to Web/CLI."* But under the default prompt policy the user rarely sees it, and never keeps it:

1. A background cycle throws at `ClaudeUsageFetcher.swift:404` (`assertDelegatedRefreshAllowedInCurrentInteraction`, with `allowBackgroundDelegatedRefresh` hardcoded `false` at `ClaudeProviderDescriptor.swift:676`) with *"background repair is suppressed… Click Refresh in the CodexBar menu to retry."*
2. The user clicks Refresh. That interaction is `.userInitiated`, so delegation runs and reaches the accurate terminal verdict.
3. The next background poll throws the retry message again, overwriting it.

So the actionable guidance is transient and the misleading guidance is the steady state. Clicking Refresh cannot fix this profile — there is nothing for a refresh to hand back.

## Why the profile is unrecoverable

`keychainAccessAllowed` is hardcoded `false` outside `#if DEBUG` (`ClaudeOAuthCredentials.swift:2855-2871`) — production never reads Claude Code's Keychain item, deliberately, because Claude Code rewrites that item on every refresh and resets its ACL. With no `~/.claude/.credentials.json` either, a delegated refresh has no channel to return a credential through.

That is precisely what `isRefreshResultUnreadable` already detects — but only *after* the touch, which background cycles never reach.

## The change

When the prompt-policy branch is about to suppress background delegation, and the profile is provably unreadable, throw the terminal message instead of the retry suggestion.

Deliberately message-only:

- **Delegation behavior is unchanged.** Background still skips it; `.userInitiated` still runs it. That matters because the coordinator notes *"on older Claude Code a retried touch can still create the credentials file"* — so the terminal verdict must not pre-empt the touch. It doesn't; this only picks the message on the path where delegation was already being skipped.
- **No prompt policy or Keychain gate is relaxed.** Nothing about when CodexBar reads or prompts changes.
- `unreadableCredentialsMessage` is extracted from `delegatedRefreshFailureMessage` and shared, so both paths stay in sync by construction.

## Testing

- New `ClaudeUnrecoverableOAuthGuidanceTests` — 3 cases: unrecoverable profile gets the terminal message, recoverable profile keeps the retry message, user-initiated is never suppressed.
- Verified the key test **fails without the fix**, producing exactly the reported symptom: `"…Click Refresh in the CodexBar menu to retry."` where the terminal message is expected.
- New `appOAuthBackgroundReportsUnrecoverableProfileInsteadOfSuggestingRefresh` covering the same via the full fetcher path.
- `make check` → 0 violations across 1806 files.
- No regressions vs `main` at `22b24b8`. Remaining failures under a wide filter are pre-existing order-dependent suites (`ClaudeOAuthPromptCoalescingTests`, `ClaudeCLISessionTests`) — they fail on `main` too, the set shifts between runs of the same commit, and both pass 12/12 twice in isolation.

**Two existing tests changed, and I want to flag why rather than bury it.** `appOAuthBackgroundRespectsPlatformKeychainPromptPolicy` and `oauth delegated retry only on user action background suppresses delegation` both assert the retry wording while using `environment: [:]` with no credentials-file override. My change makes that message depend on whether a credentials file exists, so as written they would have become host-dependent. Both now pin a credentials file explicitly and keep their original assertions unchanged — the retry wording is still correct for a profile that *is* recoverable. Their behavioral assertions (`attempts == 0`, `delegatedCounter == 0`) are untouched.

## Context

Reported from a real stuck profile: credential ~44h expired, `hasRefreshToken=true`, `owner=claudeCLI`, `source=cacheKeychain`, delegating every cycle and never recovering, on Developer ID signed 0.48.0 while `claude` itself worked throughout. The user spent a day on it because the message pointed at a Keychain setting; the fix was to stop using the OAuth source, which the app already knew and only said transiently.

This supersedes #2739, which I closed as a no-op — that one un-gated a sync that `keychainAccessAllowed` disables in production anyway. This change is on a path I've confirmed executes in production, which is what that one wasn't.

### DIFF
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeUsageFetcher+DelegatedRefreshMessages.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeUsageFetcher+DelegatedRefreshMessages.swift
@@ -2,6 +2,20 @@ import Foundation
 
 /// Split out of `ClaudeUsageFetcher.swift` to keep that file within the file-length limit.
 extension ClaudeUsageFetcher {
+    /// Not "run `claude login`, then retry": that refreshes Claude Code's own Keychain item, which this build
+    /// never reads, so the same expired cache comes back.
+    static let unreadableCredentialsMessage =
+        "Claude OAuth credentials expired and CodexBar cannot read them back. Claude Code owns the "
+            + "Keychain item and no credentials file is present for this profile, so refreshing will not "
+            + "restore usage. Switch Claude Usage source to Web/CLI."
+
+    /// True when no refresh can restore this profile: CodexBar never reads Claude Code's Keychain item in
+    /// production, so with no credentials file there is nothing a delegated refresh could hand back.
+    static func isDelegatedRefreshProvablyUnreadable(environment: [String: String]) -> Bool {
+        guard !ClaudeOAuthCredentialsStore.keychainAccessAllowed else { return false }
+        return !ClaudeOAuthCredentialsStore.hasSelectedProfileOAuthCredentialsFile(environment: environment)
+    }
+
     static func delegatedRefreshOutcomeLabel(
         _ outcome: ClaudeOAuthDelegatedRefreshCoordinator.Outcome) -> String
     {
@@ -30,11 +44,7 @@ extension ClaudeUsageFetcher {
         }
 
         if result.isUnreadableAfterRefresh {
-            // Not "run `claude login`, then retry": that refreshes Claude Code's own Keychain item, which this
-            // build never reads, so the same expired cache comes back.
-            return "Claude OAuth credentials expired and CodexBar cannot read them back. Claude Code owns the "
-                + "Keychain item and no credentials file is present for this profile, so refreshing will not "
-                + "restore usage. Switch Claude Usage source to Web/CLI."
+            return Self.unreadableCredentialsMessage
         }
 
         switch result.outcome {
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeUsageFetcher.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeUsageFetcher.swift
@@ -191,7 +191,7 @@ public struct ClaudeUsageFetcher: ClaudeUsageFetching, Sendable {
         self.configuration.browserDetection
     }
 
-    private struct ClaudeOAuthKeychainPromptPolicy {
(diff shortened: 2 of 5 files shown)


steipete:
Thanks for digging into this — the stuck-card loop you're describing is real, but this implementation can't land in its current shape, and the reasoning is worth spelling out.

The patch computes "provably unreadable" **before** the delegated touch (Keychain unavailable + credentials file absent) and reports terminal guidance from that. That widens terminal classification in exactly the way #2650's review deliberately rejected: on older Claude Code versions, the touch itself can *create* `~/.claude/.credentials.json` — which is why the coordinator re-checks the file **after** the touch and keeps touch errors retryable (there's a regression test locking that invariant). Pre-touch absence is not proof that no refresh can restore the profile, and a background poll here can mask a retryable touch error with permanent-sounding guidance.

The underlying experience you're fixing is the target of #2675 (consented direct read + CLI usage fallback), which resolves the stuck card without weakening the classifier. Once that lands, if persistent terminal guidance across background polls is still desirable, the right construction is to persist the actual post-touch terminal verdict and surface that — happy to review a follow-up in that shape. Closing this one in the meantime rather than leaving it in limbo; the analysis in your PR body genuinely helped validate #2675's direction.

luisgonzaleznf:
Understood, and the Codex finding was right — thanks for the review time.

My mistake was reasoning about the code path and not the user path. Background still skipped delegation exactly as before, so I convinced myself the older-Claude-Code touch was preserved. But dropping "Click Refresh" from the message removes the prompt to *perform* that touch, so a recoverable profile could be told to abandon the one action that would have fixed it. Worse than the bug I was aiming at.

A corrected version would gate on an actual post-touch `isUnreadableAfterRefresh` verdict, persisted per profile so background polls can reuse it, rather than the pre-touch heuristic — terminal guidance only after a real touch has proven it.

Worth me putting that together, or would you rather leave this alone? Happy either way — no reply needed if it's a no, I won't follow up.

The observed-behaviour writeup in #2733 stands on its own regardless.

Issue #2760, opened 2026-08-08
Migrate Codex cost-usage persistence to a single SQLite database (replaces the JSON artifact)


steipete:
## Decision

Owner decision (2026-08-08): Codex cost-usage persistence moves to **a single SQLite database**, replacing the JSON artifact entirely. The monolithic decode/encode shape is the root cause of this month's incident cluster (#2637 memory growth, the #2646 bounding machinery it necessitated, #2703's overshoot contract) and the quadratic catch-up analyzed in #2726. The dual-store sidecar approach from #2726 is superseded by full unification; its accumulator design and linearity proof carry into this series with credit to @xx205.

## Frozen design

- **One file**: `cost-usage.sqlite` (WAL mode, `busy_timeout`) in the same directory as the current artifact. Holds everything the JSON artifact holds today: per-file scan state (offsets, inode/size/SHA-256 validation anchors, token snapshots, fork/lineage and subagent state, buffered retry lines), day aggregates and model breakdowns, discovery/lookback state, plus persisted terminal accumulators (the #2726 concept) so appends process only deltas.
- **Derived data, rebuild-first migration**: the DB is a cache over session files. `PRAGMA user_version` gates the schema; unknown/old version or corruption → drop and rebuild bounded (exactly today's producer-key rotation semantics — users already experience periodic rebuilds). **No JSON→SQL migrator**: on first run the legacy JSON artifact is deleted and the DB is built fresh via the existing bounded scan. The parser-hash mechanism keys `user_version` bumps instead of JSON producer keys.
- **Budgets become DB-native**: entry budgets → row caps; the 256/320 MiB byte machinery is replaced by a file-size cap enforced with window-aware retention DELETEs + `incremental_vacuum`. The load-refusal cap disappears — there is no monolithic decode to protect against.
- **Invariants preserved**: the #2648 session-file validation ladder is untouched (it guards source files, not the artifact). Window-aware retention semantics from #2646 are re-expressed as SQL retention with equivalent tests. Append-linearity gets the #2726-style regression test (one appended row → exactly one processed/repriced/hashed row; stable EOF → zero work).
- **Concurrency**: existing scan-queue serialization stays the single-writer; WAL makes app+CLI concurrent readers safe, replacing whole-file atomic-replace choreography.
- **Platforms**: `CSQLite3` already ships (macOS app, Linux glibc/musl CLI).

## Phases (each a CI-green PR, landed sequentially)

[...]

PR #2764, opened 2026-08-08
fix: keep recently modified files out of cost-store window pruning (refs #2760)

Yuxin-Qiao:
## Summary

- `retentionCandidates` (the window prune, also reached from `enforceBudgets` when the store is over budget) now mirrors the JSON path's `isRecentlyActive` exemption: a session file whose mtime falls inside the retained window is not pruned, even when all its usage coverage days are out of window.
- The exemption uses **inclusive calendar-day bounds**: from 00:00 of `sinceDay` through 24:00 of `untilDay` (computed via `CostUsageScanner.CostUsageDayRange.localGregorianCalendar`, then passed to SQL as `AND (f.mtime_ms < ? OR f.mtime_ms >= ?)`). This addresses the review findings:
  - Start boundary: `parseDayKey` anchors at noon, which would have pruned files modified during the morning of `sinceDay`. The cutoff now starts at day start.
  - End boundary: files modified after `untilDay` are candidates again instead of being retained indefinitely.
- The budget-deletion path (`deleteOldestRetainedFile`) keeps its own narrower `zeroDayRecentlyActive` guard on purpose: hard caps still need an escape valve when every remaining file is recently active.

## Real behavior proof (production refresh path, synthetic corpus)

Drives the production refresh path (`CostUsageScanner.loadDailyReport` for `.codex` with a synthetic `CODEX_HOME` + injected cache root, no credentials, no real user data): one still-active session whose usage rows are all out of window (coverage 2026-05-10, mtime 2026-08-02) plus 30 idle stale sessions, then an over-budget `enforceBudgets` (1-file row budget, the same function `saveCodexCache` calls on every refresh).

```text
[retention-proof] stale-coverage file retained after over-budget prune: /private/var/folders/.../codex-home/sessions/2026/08/02/stale-active.jsonl
[retention-proof] warm refresh reused the cached row, headParses=0
```

The 30 idle files are pruned; the active stale-coverage file survives the over-budget pass, and the warm refresh performs zero head parses (the row's validation anchor is reused instead of re-parsing the file).

## Test

- Boundary regression tests (explicit UTC calendar, deterministic):
  - `retention keeps recently modified file with stale coverage` (mtime = 06:00 on `sinceDay`)
  - `retention prunes stale file modified before the window`
  - `retention prunes stale file modified after the window`
  - `retention keeps stale file modified at the window edges` (00:00 of `sinceDay`, 23:59:59 of `untilDay`)
- Refresh-path proof: `over budget prune retains stale coverage file modified inside the window` (in `CostUsagePerformanceGateTests`)
- `swift test --filter "CostUsageStoreTests|CostUsagePerformanceGateTests"` (73 tests, green)
- `make check` (0 violations)


Yuxin-Qiao:
@clawsweeper re-review

Yuxin-Qiao:
@clawsweeper re-review

steipete:
Landed. Verified before merge: complementary to #2767, not duplicative — the shared activeWindowMs boundary was noon-anchored via parseDayKey, leaving first-day-morning files unprotected and bleeding into the morning after the final day; this pins protection to local midnight boundaries with tests on both edges plus an over-budget refresh proving cached-row reuse. Thanks @Yuxin-Qiao!

Issue #2778, opened 2026-08-08
Flaky: OpenRouter plugin parity suites fail with script.primary == nil under CI load (runtime timeout swallowed?)


steipete:
## Summary

The OpenRouter plugin parity tests (`ProviderPluginParityTests`, `ProviderPluginDetailsParityTests`) fail intermittently across unrelated PR branches with a single signature: `script.primary == nil` — the JS plugin evaluation returns nothing, so every golden expectation fails at once. Observed today on at least three unrelated branches (the SQLite cutover, its review fix, and the pricing fix), on Linux glibc jobs and now both macOS test shards, always under heavy parallel CI load; the same tests pass locally and on quiet reruns.

## Hypothesis to verify

`nil` instead of a thrown error suggests the plugin runtime swallows a startup/evaluation timeout (worker spawn or Sucrase transpile exceeding the runtime deadline under CPU starvation) and the tests read the nil result as a golden mismatch instead of surfacing "runtime timed out". If so, two fixes: (1) parity tests must fail with the underlying runtime error, not a nil-comparison cascade — no more ambiguous signatures; (2) either serialize these suites, scale the plugin-runtime timeout for test environments, or warm the worker once per suite.

## Ask

Root-cause properly (reproduce under induced CPU load if needed), then fix both the diagnosability and the flake. This signature has consumed multiple rerun cycles today and erodes trust in every red CI run.

PR #2787, opened 2026-08-08
fix: prevent QuickJS worker executor-isolation traps (refs #2778)

steipete:
## Root cause

The macOS 26 runner's Swift 6.2.4 runtime traps when the dedicated QuickJS thread invokes a block-based thread entry closure that retained its caller's executor isolation. The crash artifact identifies the `CodexBar QuickJS provider plugin` thread and this frame chain:

```
_dispatch_assert_queue_fail
_swift_task_checkIsolatedSwift
closure #1 in QuickJSSerialWorker.init(name:stackSizeBytes:)
```

This is a Swift concurrency isolation failure at the serial-worker boundary, not a QuickJS JIT, executable-memory, or library-validation failure.

## Fix

- replace block-based QuickJS thread entries with `Thread` subclasses whose `main()` methods are statically nonisolated
- apply the same fix to the one-shot QuickJS TypeScript transpiler thread
- keep every queued worker job and synchronous operation explicitly `@Sendable`
- carry opaque QuickJS teardown state in an unchecked-Sendable holder while preserving worker-thread confinement
- retain the runner crash/backtrace diagnostics that produced the IPS evidence

## Proof

- exact affected plugin group: 49/49 passed
- full local suite after the final fix: 828 selections, 69/69 groups passed; zero retries or timeouts
- `make check`: clean
- autoreview: clean

Refs #2778.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,7 +120,9 @@ jobs:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
       - name: Runner info
-        run: printf 'macOS %s | %s\n' "$(sw_vers -productVersion)" "$(sysctl -n machdep.cpu.brand_string)"
+        run: |
+          printf 'macOS %s | %s\n' "$(sw_vers -productVersion)" "$(sysctl -n machdep.cpu.brand_string)"
+          defaults write com.apple.CrashReporter DialogType server
 
       - name: Select Xcode 26.3 or 26.2
         run: |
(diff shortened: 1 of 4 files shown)


steipete:
Closing as superseded by #2788 and merge commit 5e8797b14b6cb7aef11f44e253e851d961408880. The QuickJS worker and TypeScript-transpiler Thread-subclass repair is already on `main`, while this branch is now conflicted and its remaining failure-only CI diagnostic coverage is incomplete.

If broader QuickJS crash diagnostics become necessary again, they should return as a narrow current-main CI change covering both macOS shards and all affected plugin parity suites rather than reviving this mixed branch.

PR #2793, opened 2026-08-09
fix: derive QuickJS JS stack limit from worker stack (repair main CI crash)

steipete:
## Fixes the QuickJS plugin crash on CI macOS (main is currently red on this)

The JavaScript stack budget passed to `JS_SetMaxStackSize` was a fixed 1 MiB regardless of the native worker thread it ran on. On GitHub macOS runners the Swift async/swift-testing baseline consumes enough of the 4 MiB worker stack that, after JavaScript spends its 1 MiB budget, too little native stack remains for QuickJS to *construct* the stack-overflow `RangeError` — so the throw path itself faults (SIGSEGV/SIGTRAP with no CodexBar frames). This is the quickjs-ng/zipline#1130 unreliable-guard-at-thin-margins class; it reproduced only on CI's deeper baseline, not on the dev Mac.

**Fix:** derive the JS limit as a quarter of the *actual* worker stack (so JS can never approach the native ceiling at any thread size) and grow the production worker to 8 MiB (~7 MiB native headroom). The regression test now runs at a deliberately starved 512 KiB geometry — which crashed the fixed-limit code and now throws a clean script error (verified 5/5 locally, grouped plugin filter 3/3).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,6 +159,10 @@ jobs:
       - name: Swift Test
         # Keep small batches to reduce SwiftPM process overhead without returning to one aggregate run.
         timeout-minutes: 50
+        env:
+          # Force in-process backtraces so a signal-5 crash prints its frames inline (the OS crash
+          # reporter leaves no .ips once Swift's handler engages). Temporary diagnostic.
+          SWIFT_BACKTRACE: enable=yes,interactive=no,threads=all
         run: |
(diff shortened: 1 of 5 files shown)


steipete:
Superseded by the clean-branch version; that branch picked up unrelated rebase content.

steipete:
🧪 Image-upload API test (will be cleaned up): asset uploaded via `uploads.github.com/user-attachments/assets` with a plain `gh auth token` — no browser involved.

test gradient

PR #2803, opened 2026-08-09
fix: stop the cost store's executor assumption trapping on macOS 26+

steipete:
## Summary

CodexBar died on launch on macOS 26 and 27 with:

```
CostUsageStore.swift:117: Fatal error: Incorrect actor executor assumption;
Expected same executor as CodexBarCore.CostUsageStore.
```

`CostUsageStore` is an actor running on a custom `DispatchQueue`-backed `SerialExecutor`, and its `sync*` bridges call `assumeIsolated` from inside `queue.sync`. The executor implemented `checkIsolated()`, which satisfied older runtimes. Swift 6.2+ runtimes consult the newer `SerialExecutor.isIsolatingCurrentContext()` **first**, and the stdlib's default implementation cannot see through `DispatchQueue.sync` — so the assumption failed even though the work genuinely was on the store's queue.

The fix tags the queue with a `DispatchSpecificKey` and answers from it, gated `@available(macOS 26.0, *)`. The macOS 14 deployment target is unaffected; older runtimes keep using `checkIsolated()`.

## Why this matters

`assumeIsolated` traps in **release** builds too, so this was not debug-only — it would take down shipped builds on macOS 26, which is already GA. CI did not catch it because the *runtime*, not the compiler, decides which verification path is taken.

This is the same class of macOS-26+ platform change as the QuickJS/`@MainActor` crash fixed earlier in this release.

## Testing

- `swift test --filter CostUsageStore` — 78 tests across 5 suites pass.
- Crash report before the fix confirms the path: `OS_dispatch_queue.sync` → `closure #1 in CostUsageStore.syncLoadCodexCache` → `Actor.assumeIsolated` → `_assertionFailure`.
- App verified by launch: it died immediately before the change and now starts and stays up with zero fatal errors.

## A caveat on the added test, stated plainly

The new test exercises both sync bridges from a non-actor thread, which is real coverage of that path, but **it does not by itself reproduce the trap**. Whether `assumeIsolated` traps depends on the calling context's current executor, and no test-harness context reproduced it — a plain `Thread`, a normal `@Test` (which runs inside a Task), and `@MainActor` all passed with the fix reverted. It is committed as a contract/documentation test, not as proof. The regression proof is the app launch plus the crash report above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Fireworks: track 30-day rated billing spend with an API key and account slug (#2687). Thanks @x0mh0x!
 
 ### Fixed
+- Cost store: give the store actor's custom serial executor an `isIsolatingCurrentContext()` implementation — macOS 26+ runtimes consult it before `checkIsolated()`, and its default cannot see through `DispatchQueue.sync`, so every synchronous store bridge tripped "Incorrect actor executor assumption" and the app died on launch.
 - Codex: reject Standard/Fast pricing rows that exceed canonical fork-deduplicated usage, preventing copied fork rows and their Fast surcharge from inflating cost estimates (#2754). Thanks @1328189205 for the report and @Yuxin-Qiao for the initial fix and regression-test approach!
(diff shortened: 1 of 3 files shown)


nburatovich:
This bug also affects macOS 15. Had to downgrade to 0.48.1.

PR #2811, opened 2026-08-09
Bound PTY hard-stop cleanup latency

steipete:
## Summary

Bound PTY idle/deadline hard-stop latency without abandoning detached processes that still retain the PTY.

The original PR routed a live root through the scoped abort path, but that could miss true double-forked/session-escaped holders. This reconstruction keeps the immediate bounded abort while moving holder discovery and cleanup into a small, deadline-bounded lease.

## Design

- `launchPTY` reserves one CLOEXEC duplicate of the PTY master before `posix_spawn`. Reservation failure prevents child creation, so cleanup never needs to allocate under `EMFILE`/`ENFILE` pressure.
- A lock-protected one-shot owner transfers that descriptor to cleanup, discards it before output-overflow close-master-first aborts, or closes it during ordinary teardown.
- Early-stop cleanup always uses the bounded path, even if the root exits during the settle window. Normal exited-before-deadline #2807 pre-drain cleanup remains unchanged.
- The owned root/process group is synchronously aborted first. Holder discovery starts afterward, so processes created by root TERM handlers are visible.
- The deferred worker captures no `SpawnedProcessGroup`: only immutable output identities, exclusions, grace values, and the reserved master descriptor.
- The reserved descriptor prevents PTY identity reuse. A production 15-second lease closes/disarms it exactly once; expired scan results cannot signal.
- Holder cleanup sends TERM, then repeatedly discovers current PID-safe holders, sends SIGKILL, waits, and rescans until the PTY is clear or the lease expires.

Deliberately unchanged:

- the `/exit` write for ordinary non-overflow cleanup;
- output-overflow ordering: discard reserve, close master, scoped abort;
- normal exited-before-deadline synchronous cleanup and #2807 EOF/EIO drain bookkeeping;
- non-PTY subprocess behavior.

## Regression proof

The composite deterministic fixture covers four escaped generations:

1. a double-forked holder whose intermediate parent exits and whose grandchild calls `setsid()`;
2. a holder forked by that process's TERM handler;
3. a third-generation holder created only after cleanup records its first kill snapshot;
4. a separate session-escaped holder created by the root's TERM handler during the synchronous abort.

A second fixture proves an early-stopped root can exit during settle while a detached holder remains, without returning to synchronous system-wide discovery.

Wall-clock assertions use the repository standard: 3 seconds locally and 9 seconds on loaded CI. Synthetic discovery begins one second beyond that budget in either environment, so an implementation that waits for discovery still fails. A DEBUG-only lease allowance keeps the artificial delay separate from production's fixed 15-second cleanup window.

## Validation

- Composite four-generation regression repeatedly returned in 1.36–1.52 seconds locally and removed every PID.
- Settle-exit regression repeatedly returned in 1.04–1.05 seconds locally and removed the detached holder.
- `KiroStatusProbeTests`: 57/57 passed.
- `SpawnedProcessGroupTests`: 18/18 passed, including descriptor ownership, reservation failure, and lease expiry.
- `TTYCommandRunnerTests`: 29/29 passed.
- `BoundedChildProcessProofTests`: 3/3 across repeated runs.
- Release `CodexBarCore` build passed with DEBUG test APIs excluded.
- `make check` passed.
- Full local matrix: 841 selections across 71 groups passed with zero retries or timeouts.
- Structured Codex autoreview through P2 and TruffleHog passed with no accepted/actionable finding.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.49.4 — Unreleased
 
 ### Fixed
+- PTY probes: bound hard-stop latency while still cleaning session-escaped output holders.
 - Cost history: honor the preferred display currency throughout the detail chart and refresh converted labels when exchange rates change (#2887). Thanks @vAhyThe!
 
 ## 0.49.3 — 2026-08-12
--- a/Sources/CodexBarCore/Host/PTY/TTYCommandRunner.swift
+++ b/Sources/CodexBarCore/Host/PTY/TTYCommandRunner.swift
@@ -677,7 +677,11 @@ public struct TTYCommandRunner {
                 launchedProcess.abortSynchronously()
             } else {
                 if !didTerminateSynchronously {
-                    launchedProcess.terminateSynchronously()
+                    if launchedProcess.isRunning {
+                        launchedProcess.hardStopLivePTYRootSynchronously()
+                    } else {
+                        launchedProcess.terminateSynchronously()
+                    }
                 }
                 try? primaryHandle.close()
             }
--- a/Sources/CodexBarCore/Host/PTY/TTYCommandRunner.swift
+++ b/Sources/CodexBarCore/Host/PTY/TTYCommandRunner.swift
@@ -678,11 +678,8 @@ public struct TTYCommandRunner {
                 launchedProcess.abortSynchronously()
             } else {
                 if !didTerminateSynchronously {
-                    if launchedProcess.isRunning {
-                        launchedProcess.hardStopLivePTYRootSynchronously()
-                    } else {
-                        launchedProcess.terminateSynchronously()
-                    }
+                    // An early-stopped root may exit during settle while detached PTY holders remain.
+                    launchedProcess.hardStopLivePTYRootSynchronously()
                 }
                 try? primaryHandle.close()
             }
--- a/Sources/CodexBarCore/Host/PTY/TTYCommandRunner.swift
+++ b/Sources/CodexBarCore/Host/PTY/TTYCommandRunner.swift
@@ -673,12 +673,13 @@ public struct TTYCommandRunner {
                 // Once the bounded buffer overflows, do not spend seconds sweeping every process's
                 // descriptors before signaling. Closing the master unblocks a child stuck writing,
                 // and the scoped abort escalates within its fixed grace window.
+                launchedProcess.discardReservedPTYPrimaryDescriptor()
                 try? primaryHandle.close()
                 launchedProcess.abortSynchronously()
             } else {
                 if !didTerminateSynchronously {
                     if launchedProcess.isRunning {
-                        launchedProcess.hardStopLivePTYRootSynchronously(primaryFileDescriptor: primaryFD)
+                        launchedProcess.hardStopLivePTYRootSynchronously()
                     } else {
                         launchedProcess.terminateSynchronously()
                     }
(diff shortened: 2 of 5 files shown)


steipete:
@clawsweeper re-review

The branch is reconstructed on current main at `33034b46daf4c924965859c50d13cd4c2ee5fbac`. The prior escaped-holder finding is addressed with a bounded PTY lease and a convergent holder-cleanup loop. The regression now covers a true double-fork/session escape, a TERM-spawned holder, and a third-generation holder created after the first kill snapshot.

steipete:
@clawsweeper re-review

Final head is `da3b7a45d52328995f988773ead3b705770d0ba0`. The first CI attempt exposed a timing-fixture issue: the 2.5-second synthetic scan woke before the three-second assertion and competed on a loaded runner. The injected discovery now starts at four seconds, beyond the unchanged `<3s` caller deadline. Five local runs returned in 1.346–1.491 seconds and still removed all three escaped holder generations.

steipete:
@clawsweeper re-review

Final reconstructed head is `b01946c9d37097f55c40a967f934b3f8f5508196`. The prior review findings are covered by deterministic tests: post-abort discovery catches a holder created by the root TERM handler; a DEBUG snapshot barrier forces a third-generation holder to appear after the first kill snapshot; early-stop settle exit remains on bounded cleanup; descriptor reservation happens before spawn.

steipete:
@clawsweeper re-review

Owner decision: accept the bounded lease design. A system-wide holder scan already inside kernel enumeration may finish after the 15-second lease expires, but expiry closes the reserved descriptor, disarms all TERM/SIGKILL effects, and the worker retains only immutable identity values—not the process group or PTY resources.

The branch was mechanically rebased onto current `main`; new head is `654ec81fb4628e3db466014eac299f2e9aeb111a`, patch-equivalent to the previously reviewed head.

steipete:
Landed and verified.

Behavior and ownership:

- Early PTY idle/deadline cleanup now aborts the owned root/process group promptly, then performs holder discovery through a 15-second lease.
- The PTY master cleanup descriptor is reserved before spawn; reservation failure creates no child, and cleanup performs no allocation under FD pressure.
- The deferred worker retains only immutable identities and the reserved descriptor—not `SpawnedProcessGroup`—and expiry closes/disarms it before any late signal.
- Holder cleanup converges through TERM, PID-safe SIGKILL, wait, and rescan cycles.
- Output overflow still discards the reserve, closes the master, then uses scoped abort. Normal #2807 pre-deadline exit/drain behavior remains unchanged.

Proof:

- Composite real-process fixture covers a double-fork/session escape, TERM-created child, post-snapshot third generation, and a separate holder created by the root TERM handler.
- Separate fixture covers the root exiting during early-stop settle while a detached holder remains.
- Local caller times were approximately 1.04–1.52 seconds; every recorded PID was removed.
- `KiroStatusProbeTests`: 57/57; `SpawnedProcessGroupTests`: 18/18; `TTYCommandRunnerTests`: 29/29; repeated `BoundedChildProcessProofTests`: 3/3.
- Release `CodexBarCore` build, `make check`, full 841-selection/71-group matrix, P2 autoreview, and TruffleHog passed.
- Two early CI runs exposed timing-fixture weaknesses; the final fixture uses the repository's load-aware timing budget while keeping discovery beyond the assertion. Final exact-head CI passed: https://github.com/steipete/CodexBar/actions/runs/31681276162
- Owner accepted the remaining operational tradeoff: a kernel enumeration already in progress may finish after lease expiry, but it retains only immutable values and cannot retain PTY/process resources or signal afterward.

Merged as `8518b14e2b7a2d6200fc6095a9a868bc5e97a415`. Final-main CI passed: https://github.com/steipete/CodexBar/actions/runs/31683344912

PR #2813, opened 2026-08-09
fix: gate CLI background usage fetch on confirmed-absent OAuth creds

### LINKED ISSUE #2708: "No available fetch strategy" for Claude provider
If I re-login Claude - it works for a short while and then goes back to this error message "No available fetch strategy".

Image

axisrow:
## Summary
Fixes two independent causes of the intermittent **"Claude OAuth credentials not found. Run `claude` to authenticate."** flicker (#2708), both of which surface the same misleading error while the user is actually still authenticated:

**1. CLI-background deadlock (original scope).** After a signed-out/failed background Claude CLI attempt, the background Auto usage-fetch path could stay permanently blocked (or, depending on the exact failure sequence, get silently re-permitted) because the OAuth-absence deadlock-breaker did not consistently account for revoked or unidentified profiles. A fresh sign-in would work briefly and then the provider would fall back to "No available fetch strategy" again.

- Add `directCredentialIsMissingOverride` TaskLocal to `ClaudeOAuthFetchStrategy` for deterministic test overrides.
- Thread an `oauthCredentialsConfirmedAbsent` check into `ClaudeCLIBackgroundAvailability.allowsBackgroundAutoUsageFetch` so the CLI background fallback only fires when direct OAuth credentials are confirmed absent.
- Require a non-nil, non-revoked background-availability marker before consulting the OAuth-absence fallback, so:
  - a marker revoked by a failed foreground fetch is not re-permitted on the next background tick (must wait for the next successful foreground fetch), and
  - a profile CodexBar cannot identify (missing/malformed/unreadable account config) never reaches the fallback at all, since a failed attempt for such a profile could never be recorded as a revocation.

**2. Background web-cookie recovery (added in this update).** `ClaudeWebAPIFetcher.fetchUsageSerialized` — the Claude web/cookie usage path — clears a stale cached session cookie on an auth failure and unconditionally falls through to a browser cookie read to recover, the exact same shape as the bug fixed for Ollama in #2814 (`ollama.com`'s equivalent report was #2072). In a background context `BrowserCookieAccessGate` typically denies that read (no interactive Keychain prompt outside a user-initiated action), so the browser lookup finds nothing and a generic "no session key found" error replaces the original, more informative cached-auth error (surfacing as this same flicker). Fixed with the identical pattern used for Ollama:

- Still attempt browser-cookie recovery after clearing the stale cache, even in a background context — `BrowserCookieAccessGate` already gates that read on its own no-UI preflight (Safari never needs Keychain decryption, and a Chromium browser with a prior "Always Allow" Keychain grant is read without a prompt), so a background attempt is not unconditionally blocked.
- Only when that recovery attempt also fails do we now surface the original cached-auth error instead of the misleading generic one.
- Added `ClaudeWebBackgroundRecoveryTests` covering: background recovery that finds nothing (original error surfaces), background recovery that succeeds without a prompt (still works, proving the fix doesn't block legitimate no-UI recovery), and the equivalent user-initiated case.

Reviewed over 3 local review cycles (`/review` + a Codex adversarial-review companion) before opening; two real logic gaps were found and fixed, remaining findings were verified as either duplicates of already-accepted deliberate behavior or non-issues.

Closes #2708

## Test plan
- [x] `swift test --filter ClaudeCLIBackgroundAvailabilityTests` (14/14 passing)
- [x] `swift test --filter ClaudeWebBackgroundRecoveryTests` (3/3 passing)
- [x] `swift test --filter ClaudeWebCookieRenewalTests` (15/15 passing, no regressions)
- [x] `make lint` (0 violations)
- [ ] Full `swift test` suite (large parallel run is flaky on this machine due to unrelated timing-sensitive tests under CPU load — e.g. `AdaptiveRefreshTimerTests`, `ClaudeOAuthCredentialsStoreNeverPromptCacheTests`' subprocess-based tombstone tests; verified these same suites fail identically on the pre-fix commit, i.e. unrelated to this change. All suites touched by this PR pass cleanly in isolation.)


axisrow:
@clawsweeper re-review

PR #2814, opened 2026-08-09
fix: stop background Ollama refresh from spuriously requiring re-sign-in

### LINKED ISSUE #2072: No Ollama session cookie found. Please sign in at https://ollama.com/signin in your browser.

axisrow:
## Summary
Closes #2072.

`ollama.com` rotates its session cookie independently of the user's signed-in state. When a background refresh (the 5-minute timer) sees the cached cookie fail with an auth error, `OllamaStatusFetchStrategy.fetchAutomatic` cleared the cache and unconditionally fell through to a browser cookie read to recover — but `BrowserCookieAccessGate` denies that read in a background context (no interactive Keychain prompt outside a user-initiated action). The browser read then returned no candidates, surfacing as the misleading **"No Ollama session cookie found. Please sign in at https://ollama.com/signin in your browser."** error, even though the user was still signed in and the cache had simply been cleared a moment earlier by this same call. The cache stayed empty, so every subsequent background tick repeated the same failure until the next manual refresh repopulated it — this is the reported flicker (also matches `SuperSalsa20`'s repro in the issue thread: background fails → interactive refresh recovers → next background refresh fails again).

`fetchAutomatic` still attempts browser-cookie recovery after clearing the stale cache, even in a background context — `BrowserCookieAccessGate` already gates that read on its own no-UI preflight (Safari never needs Keychain decryption, and a Chromium browser with a prior "Always Allow" Keychain grant is read without a prompt), so a background attempt is not unconditionally blocked. Only if that recovery attempt itself fails does the strategy fall back to surfacing an error: a generic, non-actionable browser failure (e.g. no candidates found) is swapped for the original, more informative cached-auth error, while a specific actionable browser-access diagnosis (Safari needs Full Disk Access, a Chromium Keychain prompt was declined/disabled) is preserved and re-thrown as-is, since that is always more useful than the stale cached error it would otherwise replace.

Root-cause writeup with the full execution chain: https://github.com/steipete/CodexBar/issues/2072#issuecomment-5231739349

Note: `ClaudeWebAPIFetcher.fetchUsageSerialized` uses a similar cache-invalidate-then-fallback-to-browser pattern; not touched in this PR to keep the change scoped to the reported Ollama issue, but worth a follow-up if an equivalent gap is independently confirmed there.

## Test plan
- [x] `swift test --filter OllamaUsageFetcherRetryMappingTests` (33/33 passing) — covers background recovery after cache invalidation, preserving actionable browser-access errors, and the pre-existing user-initiated cached-cookie-replacement behavior.
- [x] `make lint` (0 violations)
- [x] Full `swift test` suite passing

## Verification

Live-verified against this repo's own signed-in `ollama.com` session, run through a locally packaged, adhoc-signed `CodexBar.app` built from this branch (`Scripts/package_app.sh release`) so the CLI gets real Keychain persistence (an unsigned/unbundled `swift build` binary intentionally falls back to an in-memory-only cache — see `KeychainCacheStore.swift`, "Unbundled processes ... must never touch the shared cache item"). Full trace kept locally, redacted excerpts below (cookie *names* only ever appear in these logs — never values; usernames/emails scrubbed).

**1. Cold background fetch after a forced cache miss** — the background path honors its no-UI boundary: with no prior "Always Allow" Keychain grant for Chrome/Chromium on this machine, it correctly declines to prompt and surfaces the actionable sign-in message rather than silently hanging or crashing:
```
debug cookie-cache: provider=ollama Cookie cache miss
info  keychain-preflight: service=Chrome Safe Storage — requires interaction for the current process
info  browser-cookie-gate: browser=Chrome — Skipping background Chromium cookie import to avoid a Keychain prompt
info  keychain-preflight: service=Chromium Safe Storage — requires interaction for the current process
info  browser-cookie-gate: browser=Chromium — Skipping background Chromium cookie import to avoid a Keychain prompt
Error: No Ollama session cookie found. Please sign in at https://ollama.com/signin in your browser.
```

**2. User-initiated recovery (`cookie refresh --provider ollama --allow-keychain-prompt`)** — real browser-cookie import against the live session, cache populated:
```
info  browser-cookie-gate: browser=Chrome — Explicit browser cookie retry allowed
info  browser-cookie-gate: browser=Chrome — Explicit browser cookie retry succeeded
debug cookie-cache: provider=ollama source=Chrome Default — Cookie cache refresh staged
ollama: ✅ Browser cookie refreshed.
```

**3. Warm-cache reuse in a separate process invocation** — proves the cookie from (2) persisted to the real Keychain (not just in-process memory) and is reused without re-hitting the browser:
```
debug keychain-preflight: service=com.steipete.codexbar.cache — Keychain preflight allowed
debug cookie-cache: provider=ollama Cookie cache hit
== Ollama (web) ==
Session: 100% left ...
```

`fetchAutomatic`'s recovery and error-classification logic does not read `ProviderInteractionContext` — only `BrowserCookieAccessGate`'s no-UI preflight does — so the background and user-initiated code paths exercised above run the identical classification code covered by the stale-cached-cookie and actionable-browser-error unit tests. The one branch not live-exercised here is recovery from a *stale-but-present* cached cookie (vs. the cache-miss case (1) actually hit): reproducing that live would require writing a crafted expired value into this machine's real login-Keychain entry, which isn't a step I'm willing to script against real credentials — it's covered instead by `OllamaUsageFetcherRetryMappingTests`' injected-failure tests, which exercise the exact same code path with a controlled input.

### DIFF
--- a/Sources/CodexBarCore/Providers/Ollama/OllamaProviderDescriptor.swift
+++ b/Sources/CodexBarCore/Providers/Ollama/OllamaProviderDescriptor.swift
@@ -172,11 +172,21 @@ struct OllamaStatusFetchStrategy: ProviderFetchStrategy {
                     clearCached(cached)
                     // The ollama.com session cookie rotates independently of the user's signed-in state, so a
                     // background refresh can see a cached cookie go stale even when the user never signed out.
-                    // BrowserCookieAccessGate denies a background browser read (no interactive Keychain prompt
-                    // outside a user-initiated action), so falling through to `fetchBrowser` here would only
-                    // trade this accurate auth error for a misleading "no session cookie" one. Defer recovery
-                    // to the next user-initiated refresh, which already has the explicit-retry path.
-                    guard ProviderInteractionContext.current == .userInitiated else { throw error }
+                    // BrowserCookieAccessGate already gates browser reads on its own no-UI preflight — it does
+                    // NOT always deny a background attempt (Safari never needs Keychain decryption, and a
+                    // Chromium browser with a prior "Always Allow" Keychain grant is also read without a
+                    // prompt) — so still attempt `fetchBrowser` here rather than assuming it will fail. Only
+                    // when that attempt itself fails do we surface the original auth error (not a misleading
+                    // "no session cookie" one) instead of finalizing on a user-initiated refresh's explicit
+                    // opt-in, mirroring the `shouldTryBrowserCandidates` recovery below.
+                    let cachedError = error
+                    do {
+                        let resolved = try await fetchBrowser()
+                        storeResolved(resolved)
+                        return resolved.snapshot
+                    } catch {
+                        throw cachedError
+                    }
                 } else if self.shouldTryBrowserCandidates(afterCachedFailure: error) {
                     let cachedError = error
                     do {
--- a/Sources/CodexBarCore/Providers/Ollama/OllamaProviderDescriptor.swift
+++ b/Sources/CodexBarCore/Providers/Ollama/OllamaProviderDescriptor.swift
@@ -170,6 +170,13 @@ struct OllamaStatusFetchStrategy: ProviderFetchStrategy {
             } catch {
                 if self.shouldInvalidateCachedCookie(after: error) {
                     clearCached(cached)
+                    // The ollama.com session cookie rotates independently of the user's signed-in state, so a
+                    // background refresh can see a cached cookie go stale even when the user never signed out.
+                    // BrowserCookieAccessGate denies a background browser read (no interactive Keychain prompt
+                    // outside a user-initiated action), so falling through to `fetchBrowser` here would only
+                    // trade this accurate auth error for a misleading "no session cookie" one. Defer recovery
+                    // to the next user-initiated refresh, which already has the explicit-retry path.
+                    guard ProviderInteractionContext.current == .userInitiated else { throw error }
                 } else if self.shouldTryBrowserCandidates(afterCachedFailure: error) {
                     let cachedError = error
                     do {
--- a/Sources/CodexBarCore/Providers/Ollama/OllamaProviderDescriptor.swift
+++ b/Sources/CodexBarCore/Providers/Ollama/OllamaProviderDescriptor.swift
@@ -179,23 +179,15 @@ struct OllamaStatusFetchStrategy: ProviderFetchStrategy {
                     // when that attempt itself fails do we surface the original auth error (not a misleading
                     // "no session cookie" one) instead of finalizing on a user-initiated refresh's explicit
                     // opt-in, mirroring the `shouldTryBrowserCandidates` recovery below.
-                    let cachedError = error
-                    do {
-                        let resolved = try await fetchBrowser()
-                        storeResolved(resolved)
-                        return resolved.snapshot
-                    } catch {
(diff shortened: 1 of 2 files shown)


axisrow:
## 🔍 Local review (cycle 3/3, final) — round 9284d2d6-d87e-4e30-a172-67877e4caf36
Reviewed locally (`/review` + Codex companion), no bots pinged.

| Verdict | Reviewer | Finding | Location |
|---|---|---|---|
| SKIP | codex | The shared browser-recovery helper masks any browser-side error under the original cached-auth error, but this catch-and-preserve pattern predates this PR and was only extended to one more call site as intended. | OllamaProviderDescriptor.swift:206 |
| Confirmed, no defect | claude | Re-verified that the background browser-cookie retry this PR relies on cannot trigger an interactive Keychain prompt in any code path. | OllamaProviderDescriptor.swift |

No FIX or UNVERIFIED findings this round. Cycle complete.

axisrow:
## 📋 Review summary — all cycles

| Cycle | Reviewer | Finding | Verdict | Resolution |
|---|---|---|---|---|
| 1 | codex | Background auth-failure recovery unconditionally re-threw instead of ever attempting a browser-cookie read, even when that read would have succeeded without a prompt. | FIX | Fixed in 61ce42c7 |
| 1 | claude | (no issues found) | — | — |
| 2 | claude | The new recovery branch duplicated an adjacent branch nearly verbatim; extracted into one shared helper. | FIX | Fixed in 169b0841 |
| 2 | codex | Cross-account cookie mixing on a passive refresh is a real gap in the cache, but it predates this PR and already existed via other reachable code paths. | SKIP | Pre-existing, out of scope |
| 2 | codex | An in-flight background browser fetch can overwrite a cache entry the user cleared moments earlier, but this race already existed in the pre-existing recovery paths this PR reuses. | SKIP | Pre-existing, out of scope |
| 2 | codex | A background access failure can extend an existing multi-hour cooldown on further browser attempts, but the cooldown mechanism itself is unchanged by this PR. | SKIP | Pre-existing, out of scope |
| 3 | codex | The shared recovery helper always re-surfaces the original cached-auth error over any browser-side error, but that masking behavior predates this PR by many commits. | SKIP | Pre-existing, out of scope |
| 3 | claude | Re-confirmed that the background browser-cookie retry this PR relies on cannot trigger an interactive Keychain prompt in any reachable code path. | Confirmed, no defect | Re-verified |

**Totals:** 2 FIX (both resolved), 4 SKIP (pre-existing architectural properties, out of scope for this issue), 0 UNVERIFIED.

Review complete after 3 cycles — no FIX or UNVERIFIED findings remain. Per the review-only-on-merge policy, merging is a manual step for the PR author. The SKIPped findings (cross-account cookie binding, unconditional cache overwrite race, background cooldown lockout) describe a real, broader architectural gap in the automatic cookie-resolution system worth a follow-up issue, independent of this PR.

axisrow:
@clawsweeper re-review

Issue #2815, opened 2026-08-09
[Codex cost] Large active rollout remains incomplete in 0.48.1; consider ccusage fallback/reference


Yoroin:
### Summary

CodexBar 0.48.1 still undercounts local Codex usage when one active rollout is very large and continues appending. The bounded scanner remains in catch-up even after processed bytes equal total bytes, while the Spend Dashboard can show no partial data.

This looks closely related to #2452, #2525, and #2648, but it is reproducible on the released 0.48.1 build.

### Environment

- CodexBar 0.48.1 (113)
- macOS 27.0, arm64
- TokenRemain 1.2.11 with bundled ccusage 20.0.19 used only as a same-machine comparison

### Privacy-safe evidence

No session paths, IDs, prompts, responses, or account identifiers are included.

- Active Codex rollout/corpus displayed by CodexBar: about 1.11 GB
- Persisted CodexBar scan state: 17 files total, 15 complete
- `codexScanCatchUpPending = true`
- `codexScanProcessedBytes == codexScanTotalBytes`
- Spend Dashboard initially showed `0 / 7` coverage and no spend/token data, with no **Finish now** control.
- Restarting CodexBar made **Finish now** appear.
- Selecting **Finish now** switched briefly to accelerated mode, then returned to background mode after one pass while the same 2 files remained incomplete.

Same-machine result for 2026-08-09:

| Scanner | Codex tokens | Estimated cost | Runtime |
|---|---:|---:|---:|
| CodexBar CLI 0.48.1 cached/incomplete report | 17,532,866 | $13.699906 | ~0.7 s |
| TokenRemain bundled ccusage 20.0.19 | 55,300,931 | $62.396858 | 0.28 s |

The ccusage command used by TokenRemain was equivalent to:

```sh
ccusage daily --json --by-agent --offline --no-color \  --timezone Asia/Shanghai --since 2026-08-09
```

The cost difference can include pricing differences, but the token-count gap is large. I am presenting the ccusage result as a comparison/reference, not claiming it is automatically ground truth.

### Possible implementation direction

TokenRemain avoids maintaining its own Codex JSONL reducer for this view. It bundles a universal `ccusage` helper and invokes `daily --json --by-agent --offline --since ...` with a 30-second timeout:

- [TokenRemain CCUsageService.swift](https://github.com/Carstin520/token-remain/blob/main/Sources/UsageDock/Services/CCUsageService.swift)
[...]

PR #2817, opened 2026-08-09
Add opt-in ccusage fallback for incomplete Codex history

Yoroin:
## Summary

- Add an opt-in Codex-only ccusage bridge for native history scans whose coverage is not yet established.
- Preserve multi-account isolation by forwarding the selected CODEX_HOME.
- Keep native results on missing helpers, timeouts, invalid JSON, non-zero exits, or lower-token fallback output.
- Add optional universal helper packaging, architecture checks, signing, tests, and MIT provenance documentation.

## Changes after review

- Preserve native completeness independently: fallback results now keep `historyCoverageIsEstablished=false` while exposing `historyFallbackCoverageIsEstablished=true`.
- Dashboard and CLI can use fallback coverage without claiming that the native archive catch-up has completed.
- Packaged helpers now require an explicit version and SHA-256 digest, verify the digest before signing, and record verified provenance beside the helper.
- Added regression coverage for native/fallback flags, dashboard behavior, CLI JSON, and digest mismatch handling.

## Validation

- CCUsageCodexBridgeTests: 6/6 passed.
- CostUsageFetcherTests: 15/15 passed.
- Targeted CLI and dashboard tests passed.
- Package helper tests passed.
- Shell syntax, git diff check, and SwiftFormat lint passed.
- Real-archive smoke run completed against a real local Codex archive with ccusage 20.0.19. The uploaded evidence is intentionally redacted:
  - provider=codex
  - historyCoverageIsEstablished=false
  - historyFallbackCoverageIsEstablished=true
  - dailyDays=2
  - fallback totals present=true
- No credentials, account identifiers, session identifiers, prompts, raw logs, or local filesystem paths are included here.

## Design notes

- CodexBar does not download the helper or search PATH. Use CODEXBAR_CCUSAGE_PATH locally or CODEXBAR_CCUSAGE_SOURCE when packaging.
- Native catch-up status remains authoritative; the fallback is only used after an incomplete native scan and a successful, non-regressive ccusage result.
- This PR remains Draft so maintainers can decide whether the preferred long-term direction is this optional bridge, native semantic-stall recovery, or both.

Related to #2815.

### DIFF
--- a/Scripts/lint.sh
+++ b/Scripts/lint.sh
@@ -49,6 +49,10 @@ check_package_signing() {
   "${ROOT_DIR}/Scripts/test_package_signing.sh"
 }
 
+check_package_ccusage_helper() {
+  "${ROOT_DIR}/Scripts/test_package_ccusage_helper.sh"
+}
+
 check_package_info_plist() {
   "${ROOT_DIR}/Scripts/test_package_info_plist.sh"
 }
@@ -110,6 +114,7 @@ run_portable_checks() {
   check_package_product_paths
   check_package_strip
   check_package_signing
+  check_package_ccusage_helper
   check_package_info_plist
   check_release_dsym_paths
   check_sparkle_signing_paths
--- a/Scripts/package_app.sh
+++ b/Scripts/package_app.sh
@@ -382,7 +382,10 @@ verify_binary_arches() {
 install_optional_ccusage_helper() {
   local source="${CODEXBAR_CCUSAGE_SOURCE:-}"
   local required="${CODEXBAR_REQUIRE_CCUSAGE:-0}"
+  local version="${CODEXBAR_CCUSAGE_VERSION:-}"
+  local expected_sha256="${CODEXBAR_CCUSAGE_SHA256:-}"
   local destination="$APP/Contents/Helpers/ccusage"
+  local provenance="$destination.provenance"
 
   case "$required" in
     0|1) ;;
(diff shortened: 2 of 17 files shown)


Yoroin:
@clawsweeper re-review

Updated the PR body with redacted real-archive smoke evidence and pushed the follow-up fixes for native coverage separation and version/SHA-256 helper provenance.

steipete:
Status update: the native scanner root causes behind the incomplete-history reports were fixed on main in https://github.com/steipete/CodexBar/pull/2918 (unbounded/non-durable indexing, the Finish-now 30-day cap, and stalled semantic-progress detection). With those landed, bundling a third-party executable as a fallback becomes a product/supply-chain decision rather than a correctness need, so I'd like to hold this PR and first see whether the reports on #2815 clear up on the next release.

Your diagnostics on #2815 were genuinely useful in pinning the indexing issues down — thank you @Yoroin. If large-rollout gaps persist after 0.49.5, a bounded ccusage *comparison mode* (as a test oracle rather than a data source) could be the right follow-up shape.

steipete:
Thanks @Yoroin for the implementation and diagnostics. The native scanner now has frozen-target catch-up (#3314), removed-source discovery recovery (#3292), and guarded current-window publication (#3669). Closing this fallback proposal and keeping the remaining native diagnostics/accounting discussion in #2815.

Even with an explicit executable path and offline arguments, this adds helper packaging/signing and a separate coverage contract. At the current PR head, acceptance compares only the total token sum, so a larger fallback total can still omit a known native day. Daily aggregate repricing also cannot preserve per-request long-context/priority evidence.

We are not adopting this code or asking for a revision. A separately scoped test-only comparison could be considered later, without treating larger ccusage totals as proof of correctness.

PR #2818, opened 2026-08-09
fix: show dock icon while Settings is open so update dialogs can front

steipete:
## Problem

CodexBar is packaged as an LSUIElement accessory app. Opening Settings did not change the activation policy, so Sparkle's Check for Updates and automatic new-version windows could appear behind other apps or fail to become key.

## Root cause

The app remained on the accessory activation policy while presenting real windows. Calling activate alone is insufficient for reliable Dock presence and frontmost window behavior, and reverting too early can pull an update dialog away when Settings closes first.

## Behavior

- Promote CodexBar to the regular activation policy and activate it before opening Settings or invoking Sparkle.
- Recognize the actual PreferencesView hosting window as well as SwiftUI's settings identifier.
- Promote for automatically presented titled Sparkle windows.
- Stay regular while any real visible keyable window remains.
- Ignore the lifecycle keepalive window, status-bar windows, tiny keepalive windows, and utility windows that cannot become key.
- Re-evaluate after close on the next main-actor turn, then return to accessory mode only after the last real window closes.

## Test evidence

- `swift test --filter '(DockIconPolicyDecisionTests|SettingsWindowOpeningTests)'` — 6 tests passed.
- `make check` — formatting, SwiftLint, manifests, package checks, docs, and repository gates passed.
- `CODEXBAR_TEST_SUITE_TIMEOUT=600 make test` through the structured review helper — all 833 selections in 70 groups passed; no retries or timeouts.
- Structured autoreview — clean, no accepted/actionable findings.

Live source-blind UI validation was not run because app startup can exercise real Keychain migration/read paths, which repository safety rules prohibit without an explicit live-test request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 0.49.1 — Unreleased
 
+### Fixed
+- Settings: show CodexBar in the Dock while Settings or an update dialog is open, so Check for Updates and new-version prompts reliably appear in front.
+
 ## 0.49.0 — 2026-08-09
 
 ### Added
--- a/Sources/CodexBar/CodexbarApp.swift
+++ b/Sources/CodexBar/CodexbarApp.swift
@@ -132,6 +132,7 @@ struct CodexBarApp: App {
 
     private func openSettings(pane: SettingsPane) {
         self.preferencesSelection.pane = pane
+        DockIconController.shared.promote()
         NSApp.activate(ignoringOtherApps: true)
         let outcome = SettingsWindowOpener.live().open(preferred: .appKit)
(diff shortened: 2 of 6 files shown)


steipete:
Live-tested on macOS 26 with the signed release-packaged bundle from this branch:

- At rest: `lsappinfo` reports `ApplicationType=UIElement` (no dock icon).
- Opening Settings promotes to `Foreground` — dock icon appears, window fronts.
- "Check for Updates" in About now visibly presents the Sparkle dialog ("You're up to date! CodexBar 0.49.0…"), which was the reported bug.
- Dismissing the dialog and closing Settings demotes back to `UIElement`; only the lifecycle keepalive window remains.

PR #2825, opened 2026-08-10
Preserve Codex cost usage window under byte pressure

### LINKED ISSUE #2823: Codex cost-usage byte budget (256 MiB) silently prunes in-window sessions; older days vanish from Spend Dashboard
### Summary

The Codex cost-usage store's **256 MiB file-size budget silently sacrifices in-window sessions**, so the Spend Dashboard permanently under-reports older days for large corpora. This is the byte-budget counterpart of #2760's row-budget logic: the row budget protects the requested window (`protectRequestedWindow: true`), but the byte budget explicitly does not.

### Reproduction (CodexBar 0.49.0, macOS arm64)

Environment: ~3,860 Codex rollout files, ~590 MB of JSONL in the scan window. One week of heavy Codex CLI usage.

1. Restore a complete, freshly scanned cost-usage DB (~590 MB, all 3,860 files, `scan_complete`).
2. Launch CodexBar 0.49.0. On the first save cycle, `enforceBudgets` runs with `defaultFileBudgetBytes = 256 * 1024 * 1024` (`CostUsageStore+CodexCache.swift:5`).
3. Observed result: the DB file is pruned to exactly 268,279,808 bytes (≈256 MiB), ~2,300 files are deleted, and the earliest in-window days (e.g. 2026-07-10…07-24) disappear from the Spend Dashboard. Daily totals for those days drop to 0 / missing.

### Root cause

`CostUsageStore+Retention.swift` → `enforceBudgets`:

```swift
// The row budget mirrors the former entry budget: it only ever removes files the
// requested window does not read. In-window and recently active files are never
// dropped just for a row overage; the byte budget below is the sole authority
// that may sacrifice in-window data.
while fileBytes > byteLimit {
    guard try Self.deleteOldestRetainedFile(
        database,
        sinceDay: sinceDay,
        untilDay: untilDay,
        calendar: calendar,
        protectRequestedWindow: false)   // ← explicitly allows in-window deletion
    ...
```

The byte loop calls `deleteOldestRetainedFile` with `protectRequestedWindow: false`, deleting the oldest-updated in-window files until the DB fits under 256 MiB. With heavy Codex usage (each large rollout accumulates hundreds of MB of JSONL), 256 MiB is routinely exceeded, so the app permanently drops older in-window days — exactly the "missing days" symptom users report.

Note the row-budget loop right above it passes `protectRequestedWindow: true` — the asymmetry is intentional per the comment, but it means the byte cap can silently erase requested-window data that the row cap is explicitly forbidden from touching.

### Expected behavior

- In-window days should never be sacrificed to the byte budget; the byte loop should either respect `protectRequestedWindow: true` or be skipped entirely when the window is covered (files are re-scanned on demand anyway, so deletion is pure data loss for the displayed window).
- Alternatively, make `defaultFileBudgetBytes` configurable (or raise it well above realistic corpora, e.g. 8 GiB) — a 256 MiB cap on a corpus that legitimately exceeds 500 MiB is untenable.

### Suggested fix

At minimum, change the byte-budget loop to `protectRequestedWindow: true` and fall back to `stripOldestRebuildableDetail` (already implemented) before ever dropping in-window files. That preserves the cap's purpose (bounded file size) without silently corrupting the displayed window.

### Impact

Any user whose Codex corpus exceeds ~256 MiB sees incorrect, permanently-degraded history with no indication. Verified end-to-end on 0.49.0 (released 2026-08-09): raw JSONL files remain intact on disk and re-scan fine, but the app prunes them from the DB to fit the cap on every save.

Whiteknight07:
## Summary

- protect completed files that touch the requested Codex cost-usage window from both row- and byte-budget deletion
- keep the 256 MiB database cap best-effort once only requested-window or recently active data remains
- remove the fallback that stripped protected scan detail and immediately scheduled it for reconstruction
- add repeated-enforcement, report-integrity, fork-parent, out-of-window pruning, and non-Gregorian calendar coverage
- move the post-0.49.1 cost-cache CPU fix into the correct 0.49.2 changelog section

## Root cause

The row-budget path protected the requested reporting window, but the byte-budget path explicitly disabled that protection. Deleting a file cascaded into `file_day_aggregates`, so rebuilding global aggregates permanently removed older requested days from Spend Dashboard.

The initial PR patch correctly enabled requested-window protection, but then fell back to `stripOldestRebuildableDetail`. That reset protected files to incomplete, deleted their token/row/accumulator state, and marked catch-up pending. The next refresh would reconstruct the same detail, exceed the same cap, strip it again, and potentially repeat indefinitely.

The maintainer revision uses a stable policy: delete only safe out-of-window completed files. Once every remaining file is protected, stop enforcement without mutating scan state. The current report and fork baselines remain authoritative, catch-up does not restart, and the database may remain above its best-effort byte cap.

## Storage-policy decision

`VISION.md` requires explicit sign-off for data-storage behavior changes. This PR deliberately prioritizes requested-window report fidelity and stable refresh behavior over an absolute 256 MiB SQLite cap.

**Recommendation:** approve the fidelity-first policy. A hard cap cannot be maintained for arbitrarily large requested windows without either deleting displayed history or introducing a larger durable aggregate-only storage design.

## Validation

- `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --filter CostUsageStoreTests` — 60 tests passed
- `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --filter CostUsagePerformanceGateTests` — 25 tests passed
- `make check` — passed
- `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 make test` — 834 selections in 70 groups passed; zero failures or retries
- structured autoreview — secret scan clean; no accepted/actionable findings
- regression coverage uses the real `CostUsageStore`/SQLite path and proves two consecutive over-cap enforcement passes retain in-window rows and report aggregates without setting catch-up pending

Fixes #2823

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 - PTY probes: bound hard-stop latency while still cleaning session-escaped output holders.
 - Cost history: honor the preferred display currency throughout the detail chart and refresh converted labels when exchange rates change (#2887). Thanks @vAhyThe!
+- Codex: preserve requested Spend Dashboard history when the local cost cache exceeds its byte budget, rather than deleting in-window sessions or repeatedly rebuilding protected data (#2823). Thanks @WillStark for the report and @Whiteknight07 for the initial fix!
 
 ## 0.49.3 — 2026-08-12
 
--- a/Sources/CodexBarCore/Vendored/CostUsage/CostUsageStore+Retention.swift
+++ b/Sources/CodexBarCore/Vendored/CostUsage/CostUsageStore+Retention.swift
@@ -283,10 +283,9 @@ extension CostUsageStore {
             }
 
             var catchUpRequired = false
-            // The row budget mirrors the former entry budget: it only ever removes files the
-            // requested window does not read. In-window and recently active files are never
-            // dropped just for a row overage; the byte budget below is the sole authority
-            // that may sacrifice in-window data.
+            // The row and byte budgets only remove files outside the requested window. Once
+            // only protected data remains, preserving report fidelity takes precedence over
+            // forcing the database under its best-effort byte cap.
             while try Self.rowCount(database) > Int64(rowLimit) {
                 guard try Self.deleteOldestRetainedFile(
                     database,
@@ -305,15 +304,8 @@ extension CostUsageStore {
                     sinceDay: sinceDay,
                     untilDay: untilDay,
                     calendar: calendar,
-                    protectRequestedWindow: false)
-                else {
-                    guard try Self.stripOldestRebuildableDetail(database) else { break }
-                    catchUpRequired = true
-                    try Self.markCatchUpRequired(database)
-                    try Self.reclaimFreePages(database)
-                    fileBytes = Self.fileSize(at: self.databaseURL)
-                    continue
-                }
+                    protectRequestedWindow: true)
+                else { break }
                 catchUpRequired = true
                 try Self.reclaimFreePages(database)
                 fileBytes = Self.fileSize(at: self.databaseURL)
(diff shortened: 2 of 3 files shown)


Whiteknight07:
@clawsweeper re-review

steipete:
Rebased onto current `main` (`1f5bdb856e`) and refreshed the implementation, head is now `8ee63cc790`.

Changes vs the previous head:

- Re-applied the retention fix on top of the new identical-content save path that landed with #2852: the tightened-budget assertion in `identical codex cache save skips content rewrites but advances the scan timestamp` now expects `catchUpRequired == false` with the store staying over the best-effort cap and in-window rows intact, matching the protected-window policy.
- Hardened `byte budget preserves fork parent detail required by a surviving child`: the fork parent is now out-of-window and stale, so the test exercises the fork-dependency protection clause rather than passing via requested-window protection alone.
- Moved the changelog entry into the current `0.49.4 — Unreleased` section and dropped the stray duplicate of the 0.49.2 CPU-fix bullet.

Proof: `swift test --filter CostUsageStoreTests` (71 tests), `swift test --filter CostUsagePerformanceGateTests` (28 tests), and `make check` all pass locally; full `make test` running.

PR #2827, opened 2026-08-10
Expose all provider accounts from serve

VACInc:
> **Draft / WIP.** Authored against `main`; the change is a close-read patch that has **not been compiled or test-run on macOS yet** (see "What was not tested"). Opening as a draft to gather direction before running the full build+test matrix.

### High Level TLDR

`codexbar serve` exposed multiple Codex and claude-swap accounts, but ordinary token-account providers only returned the
active account. This change makes `/usage` fetch every configured account and makes dashboard snapshots project those
results into one provider row with `accounts[]`, using the same configured account order and active selection as the app.

## Root Cause

The app menu reads all `ProviderTokenAccountData.accounts`, but the headless resolver returned only the clamped active
account unless an interactive CLI caller explicitly passed `--all-accounts`. The dashboard also disabled its existing
all-account mode and only attached nested accounts through the claude-swap adapter.

Before behavior was established by source and observed route output: Codex `/usage` returned multiple rows, while a
generic provider with multiple configured accounts returned only one active row and no dashboard `accounts` field.

## What changed

- Generalized serve account collection to every token-account provider.
- Kept Codex on its reconciled live, managed, and profile-home account projection.
- Grouped account usage into `providers[].accounts`, with the active account retaining provider-level fields.
- Kept single-account providers on the existing shape.
- Preserved claude-swap precedence and separated its cache/in-flight account mode.
- Added opaque account IDs, identity redaction, per-account errors, and collection-level `accountsError` handling.

## Real behavior proof

Linux static proof completed:

- `git diff --check a0d77c37f9cc85d8703eb0a501c6764112f5a6e8..HEAD` passed.
- The final patch exactly matches commit `ee0ae94e312007e63ee42b8226a7ff2e3f0e3ef8`.

macOS compilation, focused tests, and live fixture route output are not available from this Linux environment. The
exact macOS commands and route matrix are documented in `REPORT-IMPL.md`; their copied output must be added before merge.

## Verification

Added `DashboardMultiAccountSnapshotTests` for generic projection, active selection, redaction, opaque IDs, row-local
failure, collection failure, single-account compatibility, and headless all-account resolution. Updated serve cache-mode
coverage and retained the existing claude-swap snapshot suite.

Pending on macOS:

- `swiftformat Sources Tests`
- `swiftlint --strict`
- focused dashboard, claude-swap, and serve-router tests
- `swift build`
- `make test`
- `make check`
- before/after route performance samples

## What was not tested

- Compilation or tests, because this Linux host has no Swift toolchain.
- AppKit/macOS runtime behavior.
- Real provider, Keychain, bridge, or network account calls.
- Before/after latency and provider call volume.


steipete:
Thanks @VACInc. Dashboard account grouping, opaque IDs, and row-local errors are useful, but separate from cross-device cost aggregation. This patch also changes ordinary generic `/usage` from the selected account to every configured account through the default `includeAllAccounts` path. Keeping this open pending the dashboard/API decision and consolidation with #3791. The recommended contract preserves generic `/usage` cardinality and enables all-account collection specifically for dashboard snapshots. #3791 overlaps this work but is not a landed replacement on the reviewed baseline.

steipete:
Thanks @VACInc. At `ee0ae94e3120`, all-account collection defaults on, changing generic `/usage` from selected-account output to multiple rows. That conflicts with the required compatibility boundary. Public IDs also hash internal cache keys rather than durable account-source metadata. #3890 addresses those design points, pending verification. The recommendation is to close this implementation without adopting it.

Issue #2824, opened 2026-08-09
cost-usage-store spins a full core on 0.49.0 — ~41% of process CPU in NSISO8601DateFormatter/ICU locale init (110 MB corpus)


changliu98:
## Summary

On 0.49.0 the `com.steipete.codexbar.cost-usage-store` queue pins a full CPU core indefinitely, and **~41% of total process CPU is spent inside `NSISO8601DateFormatter.dateFromString:`** — specifically in ICU locale construction on every parse.

This is a much smaller corpus than #2538 (122 files / 110 MB vs 3,559 files / 4.5 GB), and the dominant queue is now `cost-usage-store` rather than `cost-usage-scan`, so I believe this is a distinct path rather than a recurrence of that issue.

## Environment

- CodexBar **0.49.0 (114)** — current latest release
- macOS 26.6.1 (25G76), Apple M4 Max (12P + 4E)
- Providers enabled: `codex`, `claude` (all others disabled)
- Codex corpus: **122 JSONL files, 110 MB total** in `~/.codex/sessions`; largest single file 20 MB, only 4 files > 3 MB
- Claude corpus: 37 JSONL files, 19 MB total
- `~/Library/Caches/CodexBar/cost-usage/cost-usage.sqlite` present (2.3 MB)

## Measurements

Sustained ~99% of one core for the entire 11 minutes the app was up — 10:23 of CPU time in 11:04 of wall clock (94%). Activity Monitor **Energy Impact 83**.

Sampled `ps` CPU time every 10s for a minute:

```
17:45:48  cpu_time=6:53.26   sqlite_mtime=17:45:48
17:45:58  cpu_time=7:03.18   sqlite_mtime=17:45:58
17:46:08  cpu_time=7:13.09   sqlite_mtime=17:46:08
17:46:18  cpu_time=7:22.99   sqlite_mtime=17:46:18
17:46:28  cpu_time=7:32.90   sqlite_mtime=17:46:28
17:46:38  cpu_time=7:42.81   sqlite_mtime=17:46:38
```

Exactly **9.9s of CPU per 10s of wall clock**, six readings, no variance — a flat spin, not a workload converging. Note `cost-usage.sqlite` mtime tracks the wall clock: the store is being **rewritten every second** (possibly related to #2495).

## Sample analysis

`sample <pid> 5` — 3,922 samples/thread. Attribution:

| Thread | Samples | Share |
|---|---|---|
[...]

PR #2832, opened 2026-08-10
Skip unchanged cost snapshot conversions

kiranmagic7:
Refs #2824.

`persistFile` converted every retained Codex token timestamp and JSON-encoded every usage row before deciding whether the SQLite cursor was unchanged. During bounded catch-up, that repeated the hottest ISO-8601 work across the full historical prefix even when the stable-cursor branch wrote no rows.

This moves conversion and encoding behind the persistence decision. Stable files do none of that work, append-safe files convert only the new suffix, and replacement behavior stays the same. The database schema, metadata, aggregates, event indexes, and row indexes are unchanged.

Tests:

- `CI=1 CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --disable-keychain --disable-netrc --filter 'CostUsageStore|CostUsagePerformanceGateTests'` (103 tests passed)
- `CI=1 CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 make check` (passed)
- `CI=1 CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 make test` does not complete locally because two unrelated `AdaptiveRefreshTimerTests` time out with `CancellationError()`. The exact clean base, `1ba79680364dddbc611d710fbf59dddf76062627`, reproduces both failures at lines 39 and 78.

Real-corpus proof (redacted):

I ran the exact pre-fix base and this head against the same completed 125.8 MB local Codex session-day subset, using a fresh isolated cache for each run. The base carried only the same counter hook around token-snapshot conversion; it did not change persistence decisions.

```text
pre-fix base  files=62 snapshots=1866 cold_conversions=1866 stable_conversions=1866 cold_seconds=4.801 stable_seconds=0.234
patched head  files=62 snapshots=1866 cold_conversions=1866 stable_conversions=0    cold_seconds=4.802 stable_seconds=0.118
```

The patched run asserted that stable conversions were zero and passed. The near-identical cold times are a useful comparability check; the conversion counts are the deterministic signal. No filenames, paths, session contents, report output, project names, or account/model/cost data were captured.

This is a narrow reduction in the repeated store work reported in #2824, not the broader delta-persistence redesign. Compatibility risk should be low because the write branches and schema are unchanged. Review focus: the append-only suffix indexes and the stable-cursor no-conversion invariant.

### DIFF
--- a/Sources/CodexBarCore/Vendored/CostUsage/CostUsageStore+CodexCache.swift
+++ b/Sources/CodexBarCore/Vendored/CostUsage/CostUsageStore+CodexCache.swift
@@ -216,13 +216,8 @@ extension CostUsageStore {
         baseline: PersistedFileBaseline,
         calendar: Calendar)
     {
-        let snapshots = (usage.codexTokenSnapshots ?? []).enumerated().map {
-            Self.tokenSnapshot(path: path, eventIndex: $0.offset, snapshot: $0.element, calendar: calendar)
-        }
-        let rows = (usage.codexRows ?? []).enumerated().compactMap { index, row -> CostUsageStoreUsageRow? in
-            guard let payload = try? JSONEncoder().encode(row) else { return nil }
-            return CostUsageStoreUsageRow(path: path, rowIndex: index, payload: payload)
-        }
+        let tokenSnapshots = usage.codexTokenSnapshots ?? []
+        let usageRows = usage.codexRows ?? []
         let details = StoredFileDetails(
             lastTotals: usage.lastTotals,
             projectPath: usage.projectPath,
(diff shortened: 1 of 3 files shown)


steipete:
Thanks for the careful diagnosis and real-corpus proof. This independently found the same eager-conversion hot path that was merged shortly afterward in #2834 / `17afef7b84eba42ad27563f5b5e02f9ec141eb09`, so I’m closing this as superseded rather than asking you to resolve the current conflicts.

I compared the implementations against current `main`:

- both choose stable reuse, suffix-only append, and full replacement under the same cursor/count conditions;
- both perform zero snapshot conversion and row encoding for stable collections and materialize only absolute-indexed suffixes for safe appends;
- the merged `CostUsagePersistencePlanner` centralizes that decision for both collection types and has broader coverage for stable transforms, append indexes, SQLite round-trips, and replacement after file-identity changes;
- the process-global conversion observer unique to this PR is no longer needed for regression coverage.

Your 125.8 MB before/after run remains useful corroboration of the root cause, and the unchanged cold-run timing is a good comparability check. No criticism of the patch itself—this was simply a parallel fix race. Thanks again for the strong evidence.

kiranmagic7:
That makes sense, thanks for checking the overlap. I agree #2834 covers the stable, append, and replacement cases, and the planner shape is the better place for the shared decision logic. I will leave this closed.

PR #2856, opened 2026-08-11
fix: stop the CLI version lookup walking past the filesystem root

Manwholikespie:
## Summary

`codexbar --help` never returns when run from a plain SwiftPM build output on affected macOS 15 Foundation versions. `containingAppVersion` walks ancestors looking for a containing `.app`, but `deletingLastPathComponent()` can produce a distinct, growing `/..`-style URL at the filesystem root. The old path-equality termination therefore never fires, so the standalone CLI spins and allocates indefinitely.

The final repair extracts one monotonic ancestor step: the candidate parent is accepted only when it has strictly fewer path components. `containingAppVersion` loops through that helper and returns `nil` as soon as no decreasing ancestor exists. Version-source precedence is unchanged: adjacent `VERSION`, enclosing `.app` `Info.plist`, then the existing bundle-version fallback.

## Regression coverage

The deterministic regression injects synthetic parent providers that return a distinct URL with the same component count and a URL with a greater component count. The helper rejects each candidate after one provider call, so removing the monotonic guard produces an immediate assertion failure without relying on host Foundation root behavior, memory growth, or an outer suite timeout.

The integration coverage still asserts that a standalone executable outside an app and a root input return `nil`. Existing coverage continues to verify enclosing-app resolution, adjacent `VERSION` precedence, and symlinked invocation behavior.

## Red / green proof

- Red: temporarily returning the injected candidate without the component-count guard made `test_nextAncestorRejectsNonDecreasingParents` fail with two `XCTAssertNil` failures in 0.060s.
- Green: after restoring the guard, the same deterministic test passed in 0.001s. The red state was not committed.

## Validation

- `swift test --filter CLIEntryTests` — 37 tests, 0 failures; repeated after the final rebase.
- Raw `CodexBarCLI` built with SwiftPM and exercised under a 10s watchdog: `--help` exited 0 in 0.014s with 2,612 bytes of output; `--version` exited 0 in 0.010s with 9 bytes of output.
- `make check` — passed; SwiftFormat required no changes and SwiftLint reported 0 violations across 1,842 files.
- `make test` — 839 selections across 70 groups, exit 0. 69 groups passed first attempt; the existing heavy cost/performance shard hit its 180s aggregate timeout, then all 12 isolated selections passed and the group recovered. Total: 0 unrecovered failures, 1 recovered group, 1 timed-out aggregate group, 12 isolated retries, 1,238.2s.
- `git diff --check` — passed.
- `autoreview --mode branch --base origin/main --max-priority P2` — clean; no accepted/actionable findings.

## Original macOS 15 evidence

Contributor reproduction host: macOS 15.7.4 (24G517), arm64, Xcode 26.2 / Swift 6.2.3.

- Before the repair, `CodexBarCLI --help` under an RSS watchdog grew from 166 MB to 1,181 MB in 11 seconds, was killed at the ceiling, and produced no output. After the repair it peaked at 1 MB and exited immediately with help output.
- A pre-fix `sample` attributed 4,176 of 4,176 samples to `containingAppVersion` at `CLIIO.swift:112-122`.
- A `JetsamEvent` from a pre-fix suite run recorded `CodexBarCLI` as the largest process, with three concurrent instances at 65.21 GB, 46.51 GB, and 11.48 GB.
- Test discovery dropped from 153.8s to 9.8s because discovery itself invokes the CLI.

## Linked issue

None open that matches this failure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 0.49.3 — Unreleased
 
+### Fixed
+- CLI: stop the version lookup from walking past the filesystem root — `deleteLastPathComponent()` appends `..` at `/` rather than staying put, so the loop's path-equality guard never tripped and a binary outside a `.app` spun a core while allocating without bound. `codexbar --help` from a plain build reached 1.2 GB in 11 seconds and never printed; the CLI test suites were timing out on it.
+
 ## 0.49.2 — 2026-08-10
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,6 @@
 
 ## 0.49.3 — Unreleased
 
-### Fixed
-- CLI: stop the version lookup from walking past the filesystem root — `deleteLastPathComponent()` appends `..` at `/` rather than staying put, so the loop's path-equality guard never tripped and a binary outside a `.app` spun a core while allocating without bound. `codexbar --help` from a plain build reached 1.2 GB in 11 seconds and never printed; the CLI test suites were timing out on it.
-
 ## 0.49.2 — 2026-08-10
 
 ### Fixed
--- a/Sources/CodexBarCLI/CLIIO.swift
+++ b/Sources/CodexBarCLI/CLIIO.swift
@@ -109,8 +109,6 @@ extension CodexBarCLI {
         var currentURL = executableURL.deletingLastPathComponent()
         let fileManager = FileManager.default
 
-        // `deleteLastPathComponent()` appends ".." at the filesystem root instead of staying
-        // put, so a path-equality guard never trips. Component counts strictly decrease.
         while currentURL.pathComponents.count > 1 {
             if currentURL.pathExtension == "app" {
                 let infoURL = currentURL
--- a/Sources/CodexBarCLI/CLIIO.swift
+++ b/Sources/CodexBarCLI/CLIIO.swift
@@ -109,7 +109,9 @@ extension CodexBarCLI {
         var currentURL = executableURL.deletingLastPathComponent()
         let fileManager = FileManager.default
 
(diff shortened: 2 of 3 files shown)


steipete:
Maintainer repair is now at `415996e5f7d3f019718b2f428749f7ed27acfdba`: one monotonic ancestor helper, deterministic injected non-decreasing-parent coverage, unchanged version-source precedence, and no changelog entry. Red proof failed immediately in 0.060s with the guard removed; final focused tests are 37/37, raw `--help` and `--version` exited in 0.014s/0.010s under a 10s watchdog, `make check` passed, the full 839-selection suite exited 0 after one heavy shard recovered via 12 isolated retries, and P2 branch autoreview is clean. @clawsweeper re-review

steipete:
Landed as squash merge `84f7cfe7de68c713066ba2644d6c800522d6dd52`.

Local proof at reviewed head `415996e5f7d3f019718b2f428749f7ed27acfdba`: `swift test --filter CLIEntryTests` passed 37/37; raw SwiftPM `CodexBarCLI --help` and `--version` both exited 0 under the 10s watchdog with output (0.014s/2,612 bytes and 0.010s/9 bytes); `make check` passed with 0 SwiftLint violations; and `make test` passed all 839 selections, with the one heavy shard recovered by the repository's 12 isolated retries. The configured P2 branch autoreview was clean (TruffleHog clean, no accepted/actionable findings, overall confidence 0.97).

Exact-head CI [run 31481364137](https://github.com/steipete/CodexBar/actions/runs/31481364137) completed green across both macOS shards, Linux x64/arm64, musl, lint, and the aggregate gate. ClawSweeper's durable review at the same SHA reported no findings and sufficient proof. The original bug reproduces on affected macOS 15 Foundation behavior; deterministic injected-parent coverage protects every CI host from regressions in the termination invariant.

Thanks for the report and fix, @Manwholikespie!

PR #2863, opened 2026-08-11
fix: preserve Codex request-tier pricing across forks

### LINKED ISSUE #2858: Codex cost nearly doubles when canonical day/model fallback applies >272K long-context pricing to aggregated tokens
# Codex cost nearly doubles when canonical day/model fallback applies >272K long-context pricing to aggregated tokens

## Summary

CodexBar 0.49.2 appears to apply GPT-5.6 Sol's `>272K input tokens` long-context pricing threshold to an **aggregated day/model token total** when row-derived pricing is rejected and the report falls back to canonical fork-deduplicated usage.

This can nearly double the displayed API-equivalent cost even though no individual Codex request crossed the 272K threshold.

In the affected real-world example below:

- Canonical token accounting appears correct.
- Project token totals sum exactly to the top-level daily total.
- Project-level cost totals approximately **$1,639.61**.
- The same day's top-level cost is **$3,190.68**.
- The ~1.95× difference is consistent with GPT-5.6 Sol's 2× long-context input/cache pricing being applied to a day-level aggregate.

This appears related to the canonical fallback introduced around #2754 / #2782, but is a different failure mode.

## Environment

- CodexBar: **0.49.2**
- Provider: Codex
- Model: `gpt-5.6-sol`
- Date: `2026-08-11`
- macOS
- High-volume Codex workload with subagents/forks
- Codex effective context limit in this configuration: approximately **256K**, below the GPT-5.6 Sol API long-context threshold of **>272K input tokens**
- Cache hit rate: approximately 97%

## Reproduction / observed output

Command:

```bash
codexbar cost --provider codex --days 1 --refresh --format json --pretty
```

Top-level result:

```text
inputTokens:       2,434,223,488
cacheReadTokens:   2,367,130,752
outputTokens:          5,773,638
totalTokens:       2,439,997,126
totalCost:             $3,190.675589
```

Model breakdown:

```text
gpt-5.6-sol:
  totalTokens: 2,392,035,065
  cost:        $3,190.675589

codex-auto-review:
  totalTokens: 47,962,061
```

The same report's project breakdowns are:

```text
Card-Manager
  totalTokens: 2,360,377,121
  totalCost:   $1,601.016303

the
  totalTokens:    77,811,286
  totalCost:      $36.921728

saw
  totalTokens:     1,808,719
  totalCost:       $1.669774
```

These project token totals reconcile exactly:

```text
2,360,377,121
 + 77,811,286
 +  1,808,719
---------------
2,439,997,126
```

which is exactly the top-level daily `totalTokens`.

However, their costs sum to:

```text
$1,601.016303
 +  $36.921728
 +   $1.669774
----------------
$1,639.607805
```

while the top-level daily report gives:

```text
$3,190.675589
```

Difference:

```text
$1,551.067784
```

Ratio:

```text
$3,190.675589 / $1,639.607805 ≈ 1.946x
```

The exact same canonical token total therefore receives almost twice the cost depending on whether it is evaluated through the top-level day/model path or through the more granular project/session paths.

## Independent pricing sanity check

GPT-5.6 Sol base API rates are:

```text
uncached input: $5 / 1M
cached input:   $0.50 / 1M
output:         $30 / 1M
```

`inputTokens` includes cached input, so for the whole report:

```text
uncached input
= 2,434,223,488 - 2,367,130,752
= 67,092,736
```

Naively treating every reported token as Sol gives an upper-bound base-rate calculation of:

```text
67.092736M × $5       = $335.46
2367.130752M × $0.50  = $1,183.57
5.773638M × $30       = $173.21
---------------------------------
                          ~$1,692.24
```

That calculation still includes approximately 47.96M `codex-auto-review` tokens that the report does not assign Sol cost to, so the expected Sol cost should be somewhat below $1,692.

The project-level result of **$1,639.61** is therefore very plausible.

The top-level **$3,190.68** is not.

## Suspected cause

The current report path first tries row-derived pricing and checks that its Standard/Fast token ownership reconciles with the canonical fork-deduplicated total.

Conceptually:

```swift
let rowCostIsTrusted = rowCost?.isTrusted(
    canonicalTotalTokens: totalTokens
) ?? true
```

When the physical pricing rows exceed canonical ownership, the #2754 / #2782 guard rejects them and falls back to canonical pricing:

```swift
let canonicalCost = CostUsagePricing.codexCostUSD(
    model: model,
    inputTokens: input,
    cachedInputTokens: cached,
    outputTokens: output,
    ...
)

let cost = rowCostIsTrusted
    ? rowCost?.totalCostUSD ?? authoritativeCost ?? canonicalCost
    : canonicalCost
```

At this point, however, `input`, `cached`, and `output` are the **aggregated values for the whole day/model**, not one API request.

`codexCostUSD` then chooses long-context pricing from the supplied `inputTokens`:

```swift
let totalInput = max(0, inputTokens)

let usesLongContextRates = pricing.thresholdTokens.map {
    totalInput > $0
} ?? false
```

For GPT-5.6 Sol, `$0` is `272_000`.

This is correct when `codexCostUSD` receives **one request's token counts**.

It is not correct when it receives:

```text
inputTokens = 2,xxx,xxx,xxx
```

representing thousands of separate requests aggregated across a day.

The aggregate inevitably exceeds 272K, so the fallback appears to classify the entire day as long-context usage.

## Why long-context pricing should not apply here

The GPT-5.6 Sol public API rate has a long-context surcharge for individual requests with more than 272K input tokens.

In this Codex workload, the effective Codex context limit is approximately 256K.

Therefore no individual request can reach the API's `>272K` threshold, even though the sum of thousands of requests over a day is billions of tokens.

A per-request pricing threshold cannot safely be evaluated after those requests have been collapsed into a day/model aggregate.

## Why the ~1.95× inflation is consistent with this

GPT-5.6 Sol long-context rates are approximately:

```text
input/cache: 2×
output:      1.5×
```

This workload is overwhelmingly cached input.

The observed:

```text
$1,639.61 → $3,190.68
```

is a **1.946×** increase.

That is almost exactly what would be expected if an input/cache-heavy workload were accidentally repriced using the 2× long-context tier.

## Expected behaviour

Request-scoped pricing properties must survive fork deduplication and report aggregation.

In particular:

```text
longContext = request.inputTokens > 272_000
```

must be evaluated per request/turn before aggregation.

The canonical fallback should not infer:

```text
longContext = dayModel.inputTokens > 272_000
```

because a daily aggregate says nothing about whether any constituent request crossed the threshold.

The cost should also be invariant under report grouping:

```text
sum(project/session costs for a day)
≈ day-level cost for the same canonical usage
```

subject only to rounding or explicitly excluded/unpriced usage.

A 1.95× difference for the exact same token population should not be possible.

## Suggested fix

Possible approaches:

1. Preserve enough request-level pricing metadata when canonical fork ownership is established so canonical costs can still be summed at the request level.

2. Store canonical owned cost alongside canonical owned token classes before collapsing into day/model aggregates.

3. Preserve a long-context/base-context split in canonical aggregation, for example:

```text
base:
  input
  cached
  output

longContext:
  input
  cached
  output
```

4. If request-level tier information has been lost, do **not** infer long-context status from the aggregated day/model token count.

For a safe fallback, it would be preferable to mark the cost as incomplete/estimated rather than confidently applying the >272K tier to an aggregate.

## Regression test suggestion

Construct multiple independent short-context GPT-5.6 Sol requests, each below 272K, whose combined canonical input exceeds 272K.

For example:

```text
request A: 200K input
request B: 200K input
```

Both requests should use base rates.

Then force the report through the canonical fallback path used when row-derived fork pricing is rejected.

Assert:

```text
cost(A + B canonical aggregate)
==
cost(A) + cost(B)
```

and specifically verify that the 400K aggregate does **not** activate long-context pricing.

A second fixture could contain one genuine >272K request plus one short request and verify that only the genuine long-context request receives the higher tier.

## Related

- #1796 / #1797: cached-input pricing inflation and long-context-threshold handling
- #2754: Standard/Fast cost split could re-inflate fork usage
- #2782: canonical fallback when row ownership does not reconcile
- #2772: identical token counts producing substantially different cost depending on pricing path

This report appears to be a distinct interaction between the #2782 fallback and request-scoped long-context pricing.

## Raw CLI excerpt

```text
CodexBar 0.49.2

Today:
  2,439,997,126 tokens
  $3,190.675589

Projects:
  Card-Manager  2,360,377,121 tokens  $1,601.016303
  the              77,811,286 tokens     $36.921728
  saw               1,808,719 tokens      $1.669774

Project sum:
  2,439,997,126 tokens
  $1,639.607805
```

No prompts, responses, session IDs, account identifiers, or private source contents are required to reproduce the pricing invariant.

steipete:
## Problem

Fork children can retain copied parent-prefix usage rows even when canonical per-file/day/model token ownership has already removed that prefix. When the physical rows exceeded canonical totals, the report rejected row pricing and priced the whole canonical day/model aggregate as one request. For GPT-5.6 Sol, that let many sub-272K requests cross the 272K long-context threshold only after aggregation and nearly doubled the estimate.

Synthetic red proof on `e5528d4`: two independent 200K-input requests should cost `$2.006`; the fallback priced their 400K aggregate at `$4.009`, a `$2.003` overcount.

## Root cause

`buildCodexReportFromCache` concatenated physical pricing rows across files, then used only a global token-total guard. Once that guard rejected fork-inflated rows, request-scoped threshold pricing received day/model aggregates. Project and session rollups happened to price smaller groups, so grouping changed the result. Existing SQLite rows already retained the full request boundaries needed to repair the report on read.

## Fix

- Reconcile persisted rows independently for each cached file/day/model against canonical `usage.days` ownership, using persisted array/SQLite `row_index` order. Exact token-bearing sets retain order; inflated sets may retain only the newest contiguous exact suffix. Zero-owned groups use the empty suffix.
- Fail closed for irreconcilable groups, incomplete pricing evidence, unstable request identity, ambiguous copied cost-only prefixes, or Standard/Fast ownership mismatches. Genuine retained long requests and zero-token authoritative costs inside an owned suffix remain authoritative.
- Trust request rows only when every positive-token row prices and Standard+Fast tokens exactly equal canonical tokens. Preserve all retained row metadata and the Standard/Fast split.
- Price aggregates only when exact: linear models always qualify, and thresholded models qualify only when aggregate input is at or below the threshold. Above-threshold aggregates without request boundaries remain unavailable.
- Build each project's primary totals from one project-scoped canonical cache so unresolved ownership for the same day/model cannot disappear through per-file merging. Source detail remains file-scoped, and distinct-model partial-cost semantics are unchanged.
- Adopt the shipped `b975eb705f905b9a` predecessor store in place only when its derived version, `quick_check`, and incremental auto-vacuum all validate. Validation and metadata/version rotation run under one `BEGIN IMMEDIATE` transaction, so concurrent openers cannot observe a torn adoption pair. Unknown, mismatched, or corrupt stores still rebuild.
- Preserve price-on-read. There is no SQLite schema change, no new persisted field, and no persisted catalog-derived USD.

## Proof

- Red: `swift test --filter CostUsageScannerForkSplitTests` failed on current main with the synthetic `$2.006` request sum versus `$4.009` aggregate result.
- Green: `swift test --filter CostUsageScannerForkSplitTests` — 14/14 tests.
- Green: `swift test --filter CostUsagePricingTests` — 48/48 tests.
- Green: `swift test --filter CostUsageScannerPriorityTests` — 17/17 tests.
- Green: `swift test --filter CostUsagePerformanceGateTests` — 26/26 tests, including zero session-head reparses during compatible warm adoption.
- Green: `swift test --filter CostUsageStoreTests` — 64/64 tests, including 16 concurrent predecessor openers with zero rebuilds and preserved tables/metadata.
- Green: pricing-race, cutover, failure-injection, scanner breakdown, project/session, subagent, fetcher, and round-trip suites.
- Green: `make check` and current generated parser-hash check.
- Green: `make test` — 840 selections, 70/70 groups, no retries, failures, or timeouts.
- Green: `git diff --check` and configured Codex branch autoreview through P2.
- Warm SQLite reload matches the cold in-memory report while retaining full `usage_rows` request payloads without a schema change.

Fixes #2858


steipete:
@clawsweeper review

Please review exact head `2b773e990a146531ba9a931f7b82a3429ab6b5eb` through P2. Focus on request-scoped threshold pricing, per-file fork-prefix suffix reconciliation, stable row identity, Standard/Fast ownership, safe aggregate fallback, and warm SQLite `usage_rows` parity. Do not merge.

steipete:
@clawsweeper re-review

Please review exact head `61d46c66e8e81cb71423e807c8250fd755927533` through P2. Focus on persisted-order fork suffix ownership, cost-only and incomplete-pricing evidence, safe aggregate thresholds, project same-model propagation, and atomic compatible-store adoption. Do not merge.

steipete:
Resolved the remaining SQLite lock finding at exact head `b392c56691`.

- Current-schema validation now runs in one deferred read transaction, so version/hash/integrity are observed from one snapshot without taking a writer lock.
- Compatible predecessor adoption alone uses `BEGIN IMMEDIATE`, re-reads compatibility under that lock, and adopts only if it is still eligible.
- `ensureDatabase` now propagates `SQLITE_BUSY`/`SQLITE_LOCKED` through the existing preserve-and-retry path instead of rebuilding.
- Real held-lock regressions prove a current store remains readable and a locked predecessor is preserved, returns a safe fallback, then adopts intact after retry.
- The tests remain extensions of the existing `CostUsageStoreTests` suite, preserving the proven 840-selection / 70-group shard layout.

Final proof: `CostUsageStoreTests` 66/66; lock regressions 2/2; focused pricing suites green; `make check` green; full `make test` 840/840 in 70/70 groups; `git diff --check` green; configured branch autoreview through P2 clean with no actionable findings.

@clawsweeper re-review

steipete:
Landed as `4befde4bbdf0cabc6db048bcc93e9bb28fb2d766`.

The red repro was two 200K short requests: request-level expected cost was $2.006, while the old day-aggregate fallback produced $4.009. The final path reconciles copied fork prefixes from canonical request rows, preserves valid request-tier pricing, and refuses ambiguous long-context pricing on aggregates.

Proof: `make test` passed 840/840 selections in 70/70 groups; `make check` and `git diff --check` were clean; exact-head [CI run 31522518443](https://github.com/steipete/CodexBar/actions/runs/31522518443) was green; final P2 autoreview (TruffleHog clean, overall 0.86) and exact-head ClawSweeper reported no actionable findings.

Compatible shipped-predecessor SQLite caches retain complete rows/history without reparse while sidecar generation rotates. Under external writer locks, current stores remain readable and predecessor stores are preserved for safe retry/adoption. There is no schema change and no persisted estimated USD.

PR #2864, opened 2026-08-11
Fix Finish now history catch-up

### LINKED ISSUE #2861: [Codex cost] “Finish now” accelerates only 30 days instead of the configured history window
## Summary

The **Finish now** control in **Usage & Spend** does not accelerate the full configured Codex history window.

The normal Codex catch-up task uses the user setting `costUsageHistoryDays`, which supports 1–365 days. The separate Spend Dashboard catch-up launched by **Finish now** instead hard-codes `SpendDashboardSource.scanDays` (30 days). The progress card and completion state come from the shared cache, which can retain a wider configured window, so the button can accelerate a 30-day request while the displayed wider task remains pending.

A second issue compounds this: `catchUpPending` includes semantic discovery, buffered retry, and active-lookback work, but the current `progressKey` hashes only tracked-file parsing state. A bounded pass can advance discovery without changing that key, causing both catch-up schedulers to report `noProgress` and stop.

## Environment

- CodexBar 0.49.2 (116)
- macOS 27.0, arm64
- History window: 365 days
- macOS Low Power Mode: off
- AC power
- Large local Codex history: approximately 9.1 GB

## Reproduction

1. Start with a fresh CodexBar cost-history cache and a Codex history extending beyond 30 days.
2. Set **History window** to a value greater than 30, for example 123, 248, or 365 days.
3. Open **Settings → Usage & Spend** and wait for local estimated history catch-up to begin.
4. Select **Finish now**.
5. Observe that the dashboard accelerator still invokes the scanner with 30 days, while the shared wider cache remains pending. If a pass advances only discovery/lookback state, it may also stop as `noProgress`.

## Actual behavior

- After several hours, the dashboard remained at **Refreshing** and **0 / 7** coverage.
- Rounded progress moved from about **3.62 / 9.12 GB** to **3.96 / 9.08 GB**, confirming that the process was extremely slow rather than completely deadlocked.
- All tracked files touching the recent 30-day window were already scan-complete and all seven displayed days had aggregate rows, but the wider retained cache was still pending.
- **Finish now** is wired to accelerated mode, but its context is fixed at 30 days rather than the configured history window.
- Semantic cursor progress can be misclassified as a stall because it is absent from `progressKey`.

Only aggregate counts and rounded byte progress were inspected. No session content was read or included.

## Expected behavior

- **Finish now** should accelerate the full effective scope represented by its progress card:
  - at least 30 days, because the dashboard offers a 30-day view;
  - otherwise the configured History window, up to 365 days.
- Configured values such as 123, 248, and 365 days should reach the accelerated scanner unchanged.
- Accelerated catch-up should continue while any persisted cursor that contributes to `catchUpPending` is advancing.
- A truly unchanged pass should still stop safely as `noProgress`.
- Changing the History window while catch-up is active should replace the stale catch-up context.

## Root cause

- [Spend Dashboard scan window is fixed at 30 days](https://github.com/steipete/CodexBar/blob/main/Sources/CodexBar/SpendDashboardController.swift#L129).
- [Dashboard catch-up stores that fixed value in its task context](https://github.com/steipete/CodexBar/blob/main/Sources/CodexBar/UsageStore%2BSpendDashboardCodexCostCatchUp.swift#L54).
- [The general Codex catch-up correctly uses the configured value](https://github.com/steipete/CodexBar/blob/main/Sources/CodexBar/UsageStore%2BCodexCostCatchUp.swift#L39).
- [Pending state includes more than the per-file fields represented by the current progress key](https://github.com/steipete/CodexBar/blob/main/Sources/CodexBarCore/CostUsageFetcher.swift#L314-L339).

## Suggested native fix

1. Set the dashboard catch-up window to `max(SpendDashboardSource.scanDays, settings.costUsageHistoryDays)`.
2. Include the effective history length in catch-up task identity/current-context validation.
3. Extend the progress token to cover persisted file/buffer, discovery, and active-lookback cursors, without including timestamps or usage content.
4. Add synthetic regression tests for custom history windows and cursor-only progress.

## Related

- #2815 — overlapping semantic-stall behavior
- #2848 — automatic catch-up was intentionally changed to near-idle duty cycles in 0.49.2, making the scope mismatch much more visible

## Privacy

This report intentionally omits screenshots, usernames, account identities, filesystem paths, session IDs, prompts, responses, token/cost totals, and cache contents.

thomaschow19:
## Summary

- Make Usage & Spend **Finish now** accelerate the effective history window: `max(30, configured history days)`.
- Restart the active dashboard catch-up when that effective history scope changes.
- Expand the opaque progress fingerprint to cover durable file, buffered-retry, discovery, and active-lookback cursor metadata.

## Root cause

The dashboard catch-up context always passed 30 days, even when the shared cache represented a wider configured history window. Its no-progress check also fingerprinted only file-parsing fields, so persisted discovery, lookback, or buffered-retry cursor advancement could be mistaken for a stall.

## Implementation

- Forward arbitrary configured windows such as 123, 248, and 365 days unchanged while preserving the dashboard's 30-day floor.
- Include the effective history window in catch-up task identity and current-context validation.
- Fingerprint every persisted cursor that can keep catch-up pending.
- Return only an opaque in-process digest; raw path and session identifier values are not returned or logged.
- Leave automatic duty cycles, partial-result presentation, pricing, and scanner architecture unchanged.

## User impact

**Finish now** now accelerates the full effective configured-history scope. Catch-up continues while durable discovery/lookback/retry cursors advance, while a genuinely unchanged pass still stops safely.

## Validation

- Focused dashboard catch-up and cost-fetcher tests: 9 tests passed.
- Provider architecture gate: 38 tests passed.
- `git diff --check`
- `make check`
- `make test`: 840 selections across 70 groups, 0 failures and 0 retries.
- All validation used synthetic caches and test stores; no live provider or Keychain probes were run.

Fixes #2861

Refs #2815, #2848

### DIFF
--- a/Sources/CodexBar/UsageStore+SpendDashboardCodexCostCatchUp.swift
+++ b/Sources/CodexBar/UsageStore+SpendDashboardCodexCostCatchUp.swift
@@ -31,9 +31,11 @@ extension UsageStore {
             return
         }
 
-        let scopeSignature = accounts
+        let historyDays = max(SpendDashboardSource.scanDays, self.settings.costUsageHistoryDays)
+        let accountScopeSignature = accounts
             .map { "\($0.id)|\($0.cacheIdentity)" }
             .joined(separator: "\u{0}")
+        let scopeSignature = "\(historyDays)\u{0}\(accountScopeSignature)"
         if self.spendDashboardCodexCostCatchUpTask != nil,
            self.spendDashboardCodexCostCatchUpScopeSignature == scopeSignature
         {
@@ -51,7 +53,7 @@ extension UsageStore {
         let context = SpendDashboardCodexCostCatchUpContext(
             token: token,
             accounts: accounts,
-            historyDays: SpendDashboardSource.scanDays,
+            historyDays: historyDays,
             scopeSignature: scopeSignature,
             providerConfigRevision: self.settings.providerConfigRevision(for: .codex),
             costUsageSettingsRevision: self.settings.costUsageSettingsRevision)
@@ -245,6 +247,7 @@ extension UsageStore {
             && self.spendDashboardCodexCostCatchUpScopeSignature == context.scopeSignature
             && self.settings.providerConfigRevision(for: .codex) == context.providerConfigRevision
             && self.settings.costUsageSettingsRevision == context.costUsageSettingsRevision
+            && max(SpendDashboardSource.scanDays, self.settings.costUsageHistoryDays) == context.historyDays
             && self.settings.isCostUsageEffectivelyEnabled(for: .codex)
             && self.isEnabled(.codex)
             && context.accounts.allSatisfy(SpendDashboardSource.codexAuthFingerprintMatches)
--- a/Sources/CodexBarCore/CostUsageFetcher.swift
+++ b/Sources/CodexBarCore/CostUsageFetcher.swift
@@ -312,24 +312,119 @@ public struct CostUsageFetcher: Sendable {
         }
 
         let scoped = CostUsageScanner.codexCache(cache, scopedTo: roots)
-        var progressHasher = Hasher()
-        for (path, usage) in scoped.files.sorted(by: { $0.key < $1.key }) {
-            progressHasher.combine(path)
-            progressHasher.combine(usage.codexScanFileId)
-            progressHasher.combine(usage.parsedBytes)
-            progressHasher.combine(usage.size)
-            progressHasher.combine(usage.codexScanComplete)
-        }
         let hasIncompleteFile = scoped.files.values.contains { $0.codexScanComplete == false }
         let pending = cache.codexScanCatchUpPending == true || hasIncompleteFile
         return CodexScanCatchUpStatus(
             pending: pending,
-            progressKey: "\(scoped.files.count):\(progressHasher.finalize())",
+            progressKey: Self.codexScanCatchUpProgressKey(cache: cache, scoped: scoped),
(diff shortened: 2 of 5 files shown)


steipete:
Thanks for isolating the shared-cache scope bug and for preserving arbitrary configured history values. The core `max(30, configuredHistoryDays)` behavior is correct.

This draft cannot safely land independently of #2849, though. The proposed progress digest covers current `main`'s older lookback state but omits the new current-window, directory, flat-root, migration, and inventory cursors on #2849; a cursor-only pass could still be classified as stalled. Sorting/hashing every cached file, session mapping, and pending queue also conflicts with #2849's bounded-pass invariant. Finally, the history-change test invokes synchronization directly rather than proving that a visible dashboard observes the setting change and replaces work at a safe pass boundary.

Please keep this draft open. After the persisted-inventory ownership decision on #2849, the right integration is to retain the effective-window concept but use scanner-owned `madeDurableProgress` reporting (or the approved equivalent), include the real settings/dashboard lifecycle, and prove no corpus-wide status traversal. I won't ask you to chase a moving cursor contract in this branch before that decision.

steipete:
Fixed on main via https://github.com/steipete/CodexBar/pull/2918: "Finish now" now honors the configured cost-history window (`max(30, configured)`) with the window carried in the catch-up scope signature. Your fix and tests were folded in with `Co-authored-by` credit — thank you @thomaschow19! Closing as superseded; #2861 is closed by the merge.

PR #2869, opened 2026-08-11
feat(replicate): add billing through a provider plugin

Egnus:
Adds Replicate monthly inference spend and optional prepaid credit balance through a bundled JavaScript provider. Native code owns local credential selection and cache recovery; the script owns the billing-page bootstrap, account-specific invoice/credit requests, parsing and snapshot projection.

Manual Cookie headers bypass cache and browser import. Automatic mode tries the cached session, then Chrome candidates only after an authentication failure; pinned or unreadable cached credentials stop account fallback. Malformed or changed billing HTML preserves the selected credential. Cache writes retain the original observation and interactive-mutation generation, so late refreshes cannot overwrite newer credentials. Candidate requests use a cookie-storage-disabled session.

The required amount is the current monthly invoice's `total_cost_before_adjustments`. Missing/invalid spend fails rather than becoming zero; optional credit failure preserves spend. The result has billing details and cost data, with no invented quota or spending limit. A small shared host fix preserves bundled plugins' explicit HTML Accept header while retaining JSON enforcement for user plugins.

Validation: 66 focused tests in four suites passed, plus two offscreen card tests. Both plugin engines cover request headers, account paths, parsing/status errors, optional credit, and retry timing. Credential tests cover manual isolation, candidate retries, pinned/unreadable caches, concurrent cache replacement and cancellation. `make check` and independent Codex review passed. Standalone head `1152be3b37a` passed `make test`: 1,185 selections across 108 groups. One existing `CloudSyncSettingsTests` watcher case failed on the first attempt and passed its automatic group retry; there were no timeouts or remaining failures. [CI](https://github.com/steipete/CodexBar/actions/runs/35329987621) is still running.

Earlier contributor proof covered an Arc session through Manual Cookie on the native proposal; automatic Chrome import was not verified. Current-account confirmation of the rewritten path is still required. No real credentials were read for these tests.

Credit: Ignacio Bustos (@Egnus).

Synthetic snapshot comparison: the first image renders the proposal's legacy zero-percent snapshot through the current generic card to illustrate its placeholder quota; it is not an original UI capture. The second image renders the current plugin's reported spend and credit. Neither image is live-account proof.

replicate-before

replicate-after


steipete:
@Egnus I rewrote this PR in place at `1152be3b37a0a1056eef200ef6eaa09bb547c604`, preserving your credit. HTTP and parsing now live in a bundled plugin; the native layer handles manual/cache/Chrome sessions and account-safe recovery. The updated PR body describes the boundaries and synthetic proof.

66 focused tests, two card tests, `make check`, and independent review passed. The full suite and [CI](https://github.com/steipete/CodexBar/actions/runs/35329987621) are running.

Before landing, please test the current branch with your existing manual session and confirm monthly spend and credit balance against Billing. Include the tested commit and a redacted `codexbar usage --provider replicate --json --source web` result, with no cookies, tokens or account identifiers. Automatic Chrome import also needs current verification from someone who uses Chrome. Earlier Arc/manual evidence supports the endpoint contract but does not exercise this rewritten path.

steipete:
Merged as ca359cf9fce6007bc3bf3db87aa36c24a24ef33c, preserving the original PR history and contributor credit. The integration also retains the prepared 0.61.0 Unreleased notes.

Validation: `make check`; the original [exact-head CI](https://github.com/steipete/CodexBar/actions/runs/35329987621) passed; independent implementation and merge reviews were clean. The combined provider tree passed `make test`: 1,217 selections across 112 groups, with no failures, retries or timeouts. Generated provider catalogs, colors, metrics and icons were reconciled and independently checked. Synthetic card proof is in the PR.

Current authenticated Replicate usage was not available locally. This landing uses the recorded contributor service evidence plus transport/parser, integration and CI coverage; the remaining live-account coverage gap is documented rather than claimed complete.

PR #2873, opened 2026-08-11
fix: raise Azure v1 validation budget

### LINKED ISSUE #2867: Azure OpenAI GPT-5.6 Sol validation fails with `max_completion_tokens: 1`
## Problem

Azure OpenAI GPT-5.6 Sol fails CodexBar's validation probe with:

```text
HTTP 400:
Could not finish the message because max_tokens or model output limit was reached.
Please try again with higher max_tokens.
```

I am using:

```bash
AZURE_OPENAI_API_VERSION=v1
```

## Reproduction

CodexBar fails:

```bash
codexbar usage --provider azure-openai
```

But the same deployment succeeds when tested directly with a larger completion budget:

```json
{
  "messages": [
    {
      "role": "user",
      "content": "ping"
    }
  ],
  "max_completion_tokens": 64
}
```

Response:

```text
pong
```

The successful request used 5 completion tokens.

## Suspected cause

CodexBar's Azure v1 validation probe uses:

```json
"max_completion_tokens": 1
```

This is too low for GPT-5.6 Sol to complete even the validation request.

## Suggested fix

Increase the probe limit to a small safe value such as:

```json
"max_completion_tokens": 16
```

or higher.

Since this request is only used to validate the deployment, it should not require a one-token completion limit.

steipete:
## Problem

Azure OpenAI's `v1` Chat Completions validation used `max_completion_tokens: 1`. That total budget includes hidden reasoning tokens, so reasoning-capable deployments could exhaust it before producing a visible answer even though the deployment was reachable.

The reporter's live A/B on the affected deployment showed that a budget of 1 failed while 64 returned `pong` and consumed 5 completion tokens. This PR does not claim an independent live Azure run.

## Fix

- Add a named, bounded 64-token validation completion cap and use it only for the `v1` request payload.
- Keep dated Azure API versions unchanged at `max_tokens: 1`.
- Preserve the fixed prompt, endpoint validation, authentication headers, and response parsing.
- Document that each refresh is a real, potentially billable inference request and that 64 is a maximum rather than automatic consumption.

## Proof

- Regression red: the focused v1 request test expected 64 and captured the current payload value of 1.
- `swift test --filter AzureOpenAIUsageFetcherTests`
- `swift test --filter ProviderEndpointOverrideSecurityTests`
- `swift test --filter ProviderArchitectureGatekeeperTests`
- `swift test --filter CodexBarLinuxTests`
- `make check`
- `make test` (840 selections, 70/70 groups passed without retries)
- `git diff --check`
- Global Codex branch autoreview versus `origin/main` through P2: clean, no actionable findings.

Fixes #2867

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.49.3 — Unreleased
 
 ### Fixed
+- Azure OpenAI: allow enough v1 completion budget for reasoning-capable deployment validation while keeping the probe bounded (#2867). Thanks @yilinxia!
 - Codex: preserve request-level pricing tiers while reconciling forked usage, preventing day aggregates from triggering long-context rates (#2858). Thanks @thomaschow19!
 - CLI: stop standalone version lookup from walking past the filesystem root and hanging with unbounded memory on affected macOS versions (#2856). Thanks @Manwholikespie!
 - Cost store: prevent launch-time executor-assumption crashes on macOS 15 by keeping synchronous SQLite cache bridges on their validated serial queue (#2857). Thanks @Manwholikespie!
--- a/Sources/CodexBarCore/Providers/AzureOpenAI/AzureOpenAIUsageFetcher.swift
+++ b/Sources/CodexBarCore/Providers/AzureOpenAI/AzureOpenAIUsageFetcher.swift
@@ -94,6 +94,7 @@ private struct AzureOpenAIChatCompletionResponse: Decodable {
 public enum AzureOpenAIUsageFetcher {
     private static let timeoutSeconds: TimeInterval = 20
     private static let maxErrorBodyLength = 240
+    private static let v1ValidationCompletionTokenCap = 64
 
     public static func fetchUsage(
         apiKey: String,
@@ -218,7 +219,7 @@ public enum AzureOpenAIUsageFetcher {
         ]
         if self.usesV1API(apiVersion) {
             payload["model"] = deploymentName
-            payload["max_completion_tokens"] = 1
+            payload["max_completion_tokens"] = Self.v1ValidationCompletionTokenCap
         } else {
             payload["max_tokens"] = 1
         }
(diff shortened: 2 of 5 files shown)


steipete:
@clawsweeper review

Please review exact head `ed93630ec2074a3f5e0b29b3f5b92d8aa9ea4d6e` through P2. Focus on Azure OpenAI v1 Chat Completions validation request budgeting, the v1-only `max_completion_tokens: 64` cap, dated API isolation at `max_tokens: 1`, bounded billable exposure, and preservation of the existing endpoint validation, authentication headers, fixed prompt, and response parsing. Treat the live 1-to-64 A/B and 5-token observed use as reporter-provided evidence, not an independent run. Do not merge.

steipete:
The changelog entry is intentionally retained. This task explicitly requires the exact Azure OpenAI line under `0.49.3 — Unreleased` / `Fixed`, and the project policy requires changelog coverage for user-facing fixes. Editing an existing Unreleased section is not a release, tag, package, appcast, or publication operation.

The implementation finding set is otherwise empty, so no code change is needed.

@clawsweeper re-review

Please re-review exact head `ed93630ec2074a3f5e0b29b3f5b92d8aa9ea4d6e` through P2 with the explicit changelog acceptance criterion above. Preserve the prior Azure request-shape and bounded-billing focus. Do not merge.

steipete:
Landed as `7b5510b22847eebbac13baf9486480ce10c4d021`.

Reporter A/B: `max_completion_tokens: 1` failed; `64` succeeded and used 5 completion tokens. Focused Azure request tests passed 10/10 across 3 suites; full `make check` and `make test` were green. Exact-head CI: https://github.com/steipete/CodexBar/actions/runs/31531397707. P2 autoreview and exact-head ClawSweeper review were clean with no findings.

The change is v1-only: dated APIs remain at `max_tokens: 1`. Each v1 refresh is still a real, potentially billable request, now bounded at 64 completion tokens.

Thanks @yilinxia for the report and live validation.

PR #2882, opened 2026-08-12
fix: follow system dimming for menu bar icon, standard size and stepper-tuned vertical offset

luantu:
## Summary

Two related menu bar refinements for the custom layout title, plus a follow-up that makes the vertical nudge editable with a stepper.

### 1. Icon follows the system's active/inactive display tinting (no more wrong gray-on-secondary-display)

Before, a leading `.icon` token was baked into the attributed title as an `NSTextAttachment` (a pre-rendered bitmap). AppKit dims menu bar items on inactive displays automatically, but it only does that for **template images assigned to `NSStatusBarButton.image`** — pre-rendered attachments stay bright/gray regardless. On dual-screen setups this made CodexBar's icon look inconsistent with every other menu bar extra.

Now the leading icon token is surfaced as `button.image` (a template image) so AppKit applies the same active/inactive display tinting as native menu bar icons, while text tokens keep rendering through the attributed title.

### 2. Standard size + consistent optical centering

- Provider brand icon rendered at 18×18 (was 16×16), and single-line titles use the 18 pt face instead of 16 — closer to standard menu bar icon/text sizing.
- Single-line titles get a small optical baseline offset (-1) so they sit vertically balanced like the stacked (-3) layout already did.
- Status item length now accounts for the surfaced icon width.

### 3. Stepper-tuned vertical offset

The existing "vertical adjustment" preference (already added as a numeric input in the layout editor) is now a `TextField` + `Stepper` pair (range -20…20, default 0, ±1 per click), matching the pattern already used elsewhere in Preferences (e.g. cost history days). The label was shortened from "Vertical Adjustment" to "Vertical" and the whole row aligns with the adjacent Size/Gap pickers.

## Why this matters on dual-screen and custom-resolution displays

- **Dual screens:** the menu bar on the non-active display is dimmed by macOS. With a bitmap attachment the icon stayed at full brightness (or looked wrong); as a template `button.image` it now dims exactly like native icons when focus moves to the other display.
- **Custom resolutions / scaled displays:** the baseline offsets and the 18 pt face keep the title optically centered in the status bar button across different backing-scale factors, and the stepper lets users fine-tune ±1 px per display if the optical default is still off.

## Tests

- `MenuBarLayoutRendererTests` extended: leading-icon-as-image behavior, single-line baseline offset, vertical-adjustment baseline shift, stale (refresh-failed) title dimming, and `leadingIcon` size/template checks.
- Ran `swift test --filter MenuBarLayoutRendererTests` (18 tests, all pass) plus the MenuBarLayout/StatusItem/SettingsStore suites (369 tests pass). Lint-clean (`swiftlint`/`swiftformat`); the 42 repo-wide lint violations are pre-existing on upstream.

## Commits

- `feat: make menu bar vertical adjustment a numeric input`
- `fix: follow system dimming for menu bar icon and use standard size`
- `refactor: vertical adjustment uses stepper with visible value`

### DIFF
--- a/LOCAL_PATCHES.MD
+++ b/LOCAL_PATCHES.MD
@@ -1,0 +1,143 @@
+# LOCAL_PATCHES — 本地未上游修改记录
+
+> 记录本仓库相对于上游 `steipete/CodexBar` 的全部本地改动，供后续重放 / 提 PR / 迁移使用。
+> 生成时间：2026-08-13
+
+## 1. Git 分支状态（已同步上游）
+
+| 分支 | behind origin/main | ahead | 说明 |
+|---|---|---|---|
+| `main` | 0 | 5 | 4 个本地提交 + 1 个 merge commit（见 §3） |
+| `feature/menubar-icon-dimming` | 0 | 4 | 本地 4 个提交已 rebase 到上游 |
+| `feature/codex-quota-notifications` | 0 | 0 | 已 fast-forward |
+
+- 上游 `origin` = https://github.com/steipete/CodexBar.git
+- fork `fork` = https://github.com/luantu/CodexBar.git
+
+---
+
+## 2. 修复：Codex 周额度重置死锁（未提交，位于工作区）
+
+**状态**：未 commit，位于 `feature/menubar-icon-dimming` 工作区（`git diff` 可见）。
+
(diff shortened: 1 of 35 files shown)


luantu:
Closing to re-submit a clean PR without the unrelated weekly-reset fix and local journal mixed in.

PR #2894, opened 2026-08-13
Add DeepSeek rolling usage metrics

kyoubelyu:
## Summary

- add rolling 5-hour and 7-day DeepSeek token totals
- add rolling 5-hour and 7-day DeepSeek spend totals
- reuse the existing DeepSeek Platform session, declarative detail rows, and optional-usage gating
- keep rolling dashboard endpoint failures isolated from existing monthly usage details
- document the balance, monthly usage, and rolling usage request flow

## Implementation notes

DeepSeek's documented public API exposes account balance, but not rolling usage windows. The additional totals use the private Platform dashboard's `usage/by_api_key/amount` and `usage/by_api_key/cost` endpoints with exact Unix timestamp ranges. Responses are aggregated across API keys, models, and time buckets.

These values are absolute rolling totals. DeepSeek does not provide quota denominators or reset timestamps, so they are intentionally rendered as detail rows rather than percentage-based rate-limit bars.

The four rolling requests run concurrently and have their own bounded join. Missing, failed, or slow rolling responses degrade independently and do not discard the existing monthly detail response.

## Tests

- `swift test --filter DeepSeek` — passed (121 tests; one screenshot proof skipped unless its output directory is configured)
- `make check` — passed (0 SwiftLint violations in 1,848 files)
- `make test` — 60 groups passed; the `StatusMenuTests` group failed on `merged data tick keeps row count and card views stable` (`menu.items.count` 18 vs. 16). The same failure reproduces in an isolated rerun with only Codex and Claude enabled; it does not exercise DeepSeek code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ See [CLI configuration](docs/cli-configuration.md) for the full flow.
 - [Sakana AI](docs/sakana.md) — Manual Cookie header for 5-hour and weekly quota windows.
 - [Abacus AI](docs/abacus.md) — Browser cookie auth for ChatLLM/RouteLLM compute credit tracking.
 - [Mistral](docs/mistral.md) — Browser cookies for API spend, credit balance, and monthly-plan usage.
-- [DeepSeek](docs/deepseek.md) — API key for credit balance tracking (paid vs. granted breakdown).
+- [DeepSeek](docs/deepseek.md) — credit balance plus optional rolling 5-hour/weekly token and spend totals.
 - [Fireworks](docs/fireworks.md) — API key + account slug for 30-day spend from the billing summary API.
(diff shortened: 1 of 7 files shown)


steipete:
Thanks @kyoubelyu for the proposal and parser coverage. I've decided to keep DeepSeek reporting focused on the existing daily/monthly activity and provider-reported balances and costs.

The five-hour and seven-day values here are additional reporting intervals, not DeepSeek quota windows with provider-defined limits or resets. Four extra requests and another rolling aggregation path add more refresh cost and maintenance than those intervals justify for the menu.

Current main already uses the by-key dashboard endpoints, so this decision is about the additional rolling feature, not rejecting those endpoints or claiming the new rows already exist. Closing as not planned; thanks for exploring it.

PR #2924, opened 2026-08-14
Hide untouched Antigravity model families

steipete:
Replaces #2875 (fork disallows maintainer edits) — the contributor's commits are preserved verbatim with @urda as author, plus one correction on top.

## Summary

Antigravity's `RetrieveUserQuotaSummary` reports every model family the plan covers, so an account that only runs Gemini still carries Claude/GPT rows pinned at 0% that crowd out real lanes (especially the small widget's 2-row limit). Display surfaces (menu card, provider settings usage list, widget) now hide a family once every lane in it reports **known** zero usage.

## Correction over #2875

The original hid a family whose usage was *unknown* (nil) as if it were known-zero. `AntigravityQuotaFamilyVisibility` now requires `usageKnown` before treating a lane as idle, so a family that simply hasn't reported yet keeps rendering. The all-untouched fallback (never render an empty card) was already correct and is preserved.

## Proof

- 44 focused tests green (MenuCardAntigravityTests, UsageStoreWidgetSnapshotTests)
- `make check` clean; full `make test` 859/859 selections green
- Codex autoreview clean on both the corrective patch and the whole branch

Fixes the dead-row half of the Antigravity display complaints; ranked menu-bar selection is unaffected (untouched families never win there anyway).


urda:
Thanks for making this built in!

urda:
image

:shipit:

urda:
image

Amazing

PR #2926, opened 2026-08-14
Show Codex Business credit remaining in provider switcher

jey3dayo:
## Summary

Codex Business accounts can report an included monthly credit limit without the normal session or weekly rate-limit windows. The switcher therefore had no metric to draw below the Codex icon.

- Fall back to the projected monthly credit limit only when no normal Codex rate-limit window is available.
- Preserve the existing session/weekly rate-limit metric whenever it exists.
- Keep the quota indicator visible for the selected provider tab as well as inactive tabs.
- Add focused regression coverage for the projection, switcher metric, and selected-tab indicator.

## Validation

- `swift test --filter 'StatusMenuSwitcherClickTests|CodexConsumerProjectionTests'`\n- `make check`\n- `git diff --check`\n- `./Scripts/package_app.sh debug` and fresh app relaunch\n\n`make test` was also run locally. It still has the pre-existing environment-dependent failure in `BrowserDetectionTests.browser uninstall invalidates cookie source immediately`; the focused tests above pass.


jey3dayo:
Business account verification: the provider switcher now renders the Codex credit bar below the Codex tab.

codex-business-account

PR #2945, opened 2026-08-15
Enable settings window minimization

Yuxin-Qiao:
## Summary
- Enable the native minimize button on the SwiftUI Settings window.
- Add a regression test covering the window style mask and button state.

## Root cause
The Settings scene's AppKit window did not include `NSWindow.StyleMask.miniaturizable`. The traffic-light control was therefore present but disabled, so macOS rendered the middle button as a gray circle without the `–` glyph.

## Runtime proof (macOS)
Built a fresh arm64 debug bundle from this PR and opened the real SwiftUI Settings scene in an isolated `HOME` with Keychain access suppressed. The local-only launch hook used to open the scene was removed before pushing; it is not part of this change.

The live window reported:

```text
styleMask = 32783
isMiniaturizable = true
standardWindowButton(.miniaturizeButton).isEnabled = true
AXSubrole = AXMinimizeButton
AXEnabled = true
```

Redacted title-bar capture from that run (the yellow traffic-light is the enabled minimize control):

Settings window minimize control

## Validation
- `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --filter SettingsWindowAppearanceTests` — 13 passed
- `make check` — SwiftFormat and SwiftLint clean
- `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 make test` — 861 selections in 72 groups passed; 0 failures, retries, or timeouts

## CI classification (2026-08-16)

- Head `21d31e150deab023d3bae9b5639bb1d8609331a8` has green Linux, lint, and change-detection jobs.
- The aggregate `lint-build-test` failure is the repository's Draft-policy gate reporting that macOS Swift tests are deferred; it is not a test or implementation failure.
- The PR remains Draft and needs only maintainer review.


Yuxin-Qiao:
@clawsweeper re-review

steipete:
Thanks @Yuxin-Qiao! This landed via #2969, which keeps your style-mask change and test, and additionally fixes an interaction it exposed: minimizing Settings dropped the app back to accessory policy, which removed the window's Dock tile and stranded it. #2969 keeps regular policy while Settings is minimized and deminiaturizes on reopen. The changelog credits you.

PR #2953, opened 2026-08-15
fix: route codex RPC diagnostics to the logger

### LINKED ISSUE #2952: CLI: `codexbar` prints ungated `[codex notify]` debug lines to stderr
Running any `codexbar` command that probes Codex prints this to stderr:

```
[codex notify] remoteControl/status/changed
```

It comes from `debugWriteStderr` in `Sources/CodexBarCore/UsageFetcher.swift:848`, which despite the name has no debug gate. It fires on every JSON-RPC notification (line 995) and every child stderr line (line 946).

Harmless (the branch `continue`s, stdout is untouched), just noisy.

Fix: route both to `Self.log.debug`. The CLI already defaults to log level `.error` with a stderr destination (`CLIEntry.swift:416-421`), so the lines would disappear by default and return under `--verbose`. That also drops the `#if !os(Linux)` compile guard and routes untrusted child stderr through `LogRedactor`.

CodexBar 0.50.0, Codex CLI 0.147.0, macOS 26.6.1.

urda:
## Summary

`CodexRPCClient` wrote two diagnostic lines straight to stderr on every Codex
probe. No environment variable and no runtime setting controlled the output.
The `codexbar` CLI therefore printed protocol noise on a normal run:

```
[codex notify] remoteControl/status/changed
```

This change deletes `debugWriteStderr` and sends both call sites to the logger
that `CodexRPCClient` already owns.

## Why the logger is enough

The CLI already defaults to log level `.error`, and it already sends logs to
stderr (`CLIEntry.swift:416-421`). The lines therefore disappear on a normal
run. They return under `--verbose` or `--log-level debug`. No new environment
variable is needed.

Three further gains:

- The `#if !os(Linux)` block goes away. That guard came from `d90dbcf55`, a
  Linux build fix, not a debug gate. Linux now receives the diagnostics that it
  silently dropped.
- In the menu bar app, the lines land in the unified log under the `codex.rpc`
  category, where a developer can find them.
- Child stderr now passes through `LogRedactor` (`CodexBarLog.swift:118`).
  `codex app-server` stderr is untrusted text, so redaction helps.

`.debug` matches the convention in this package. `CodexBarCore` has 150
`log.debug` call sites and zero calls to `.trace` or `.verbose`.

## Runtime proof

Account values are redacted below. Both runs use `2>&1`, so the capture
includes stderr.

Normal run, the diagnostic is gone:

```
$ swift build --product CodexBarCLI && .build/debug/CodexBarCLI usage 2>&1 | head -20
Build of product 'CodexBarCLI' complete! (2.14s)
Error: Claude OAuth credentials not found. Run `claude` to authenticate.
== Codex 0.147.0 (oauth) ==
Weekly: <redacted>
Pace: <redacted>
Resets in <redacted>
Credits: <redacted>
Limit Reset Credits: <redacted>
Account: <redacted>
Plan: <redacted>
```

No `[codex notify]` line appears. The `Error:` line confirms that stderr still
reaches the terminal, so the absence is the fix and not a lost stream.

Verbose run, the diagnostic returns as a structured log:

```
$ .build/debug/CodexBarCLI usage --verbose 2>&1 | grep -i "codex rpc notification"
2026-08-15T02:55:32-0700 debug com.steipete.codexbar.codex-rpc: method=remoteControl/status/changed [CodexBarCore] Codex RPC notification
```

The line now carries a timestamp, a level, and the `codex-rpc` category. The
method name moves into structured metadata.

## Note on the third file

`ProviderArchitectureGatekeeperTests.swift` holds a line-anchored suppression
list. Removing six lines from `UsageFetcher.swift` moved the anchor for
`providerID: .codex,` from line 1486 to line 1480. The patch updates that
number. The gate itself is unchanged.

## Commands run

- `swift build`
- `make check` (0 violations in 1880 files)
- `make test` (72 of 72 groups on first pass, 863 selections, no retries)

Fixes #2952

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 - Vertex AI: match Cloud Monitoring quota usage without a `limit_name` to its unambiguous same-metric, same-location limit, restoring quota percentages (#2958). Thanks @MachApple!
+- Codex: stop printing ungated `[codex notify]` and `[codex stderr]` diagnostics from the CLI, routing them to the debug log instead (#2952).
 
 ## 0.50.0 — 2026-08-15
 
--- a/Sources/CodexBarCore/UsageFetcher.swift
+++ b/Sources/CodexBarCore/UsageFetcher.swift
@@ -845,12 +845,6 @@ private final class CodexRPCClient: @unchecked Sendable {
     private let initializeTimeoutSeconds: TimeInterval
     private let requestTimeoutSeconds: TimeInterval
 
-    private static func debugWriteStderr(_ message: String) {
-        #if !os(Linux)
-        fputs(message, stderr)
-        #endif
-    }
-
     init(
         executable: String = "codex", // Provider-specific by design: this RPC client launches Codex app-server.
         arguments: [String] = ["-s", "read-only", "-a", "untrusted", "app-server"],
@@ -943,7 +937,7 @@ private final class CodexRPCClient: @unchecked Sendable {
             }
             guard let text = String(data: data, encoding: .utf8), !text.isEmpty else { return }
             for line in text.split(whereSeparator: \.isNewline) {
-                Self.debugWriteStderr("[codex stderr] \(line)\n")
+                Self.log.debug("Codex RPC child stderr", metadata: ["line": String(line)])
             }
         }
     }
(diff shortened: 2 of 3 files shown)


urda:
@clawsweeper re-review

steipete:
Thanks @urda! This landed via #2964, which carries your Co-authored-by credit and a changelog thanks — the fork branch went CONFLICTING against main and isn't maintainer-editable, and #2964 additionally adds an end-to-end CLI regression test with a stubbed codex app-server. Closing as superseded.

PR #2960, opened 2026-08-15
fix: follow app privacy setting on serve dashboard

urda:
## Summary

The `codexbar serve` web dashboard ignored the app's "Hide personal information" setting. It always showed full account emails unless the operator passed `--identity redacted` at startup.

This change makes the serve dashboard follow that setting.

- When `--identity` is absent, each request reads the app's `hidePersonalInfo` preference. The toggle on redacts email local parts. The toggle off keeps full emails.
- The setting is read per request, so a change applies without a serve restart. Serve already reads `resetTimesShowAbsolute` and `weeklyProgressWorkDays` from the same defaults domain.
- An explicit `--identity redacted` or `--identity full` still wins, so a scripted client keeps the mode that it asked for.
- The resolved mode joins the response cache key. A body that the cache stored before a toggle cannot be replayed after it.
- `codexbar dashboard` keeps its full identity default. A one-shot JSON command must not change its output because of a GUI toggle.

`serveDashboardSnapshotRoute` moved out of `handleServeRequest`. The new lines pushed that function past the 150-line SwiftLint limit, so the route got its own function instead of a lint suppression.

## Commands run

- `swift build`
- `make check` (SwiftFormat and SwiftLint strict): 0 violations
- `make test`: 864 selections in 72 groups, 0 failures, 0 retries

## Tests

`Tests/CodexBarTests/CLIServeDashboardIdentityTests.swift` covers:

- An absent flag follows the privacy setting in both directions.
- An explicit flag overrides the privacy setting in both directions.
- Flag presence separates an explicit `full` from an absent flag.
- The operation key differs between identity modes.

These tests run no live provider probe and open no Keychain prompt.

## Notes

The raw `/usage` and `/cost` routes still return full emails. Only the dashboard snapshot follows the setting, and the web UI reads that snapshot. This preserves the documented shape of the JSON API routes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 - Vertex AI: match Cloud Monitoring quota usage without a `limit_name` to its unambiguous same-metric, same-location limit, restoring quota percentages (#2958). Thanks @MachApple!
+- Serve: follow the app's "Hide personal information" setting on the web dashboard when no `--identity` flag is given, resolving the mode per request so the toggle applies without a serve restart.
 
 ## 0.50.0 — 2026-08-15
 
--- a/Sources/CodexBarCLI/CLIDashboardCommand.swift
+++ b/Sources/CodexBarCLI/CLIDashboardCommand.swift
@@ -348,6 +348,26 @@ extension CodexBarCLI {
         }
     }
 
+    /// True when the caller passed `--identity` explicitly. `serve` uses this to tell an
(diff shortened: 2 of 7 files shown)


steipete:
Thanks @urda! Your branch went CONFLICTING against main after today's merge train and the fork isn't maintainer-editable, so I carried this forward as #2967 with your commits and a changelog credit. Closing this one in favor of that refresh — the code is unchanged.

PR #2963, opened 2026-08-16
fix: render one Antigravity lane per quota bucket on the dashboard

urda:
## Summary

The dashboard snapshot builder emits `primary`, `secondary`, `tertiary`, and every entry in `extraRateWindows`. That is correct for providers whose primary and secondary are real, distinct windows.

Antigravity is not one of those providers. It reports one bucket per quota family and window, and `AntigravityStatusProbe` then copies the most-used Gemini bucket into `primary` and the most-used Claude/GPT bucket into `secondary`, so the icon, the menu bar, and the notification paths have standard slots to read. Those two slots hold copies, not new data.

The dashboard printed the slots and the bucket list together, so a four-window account rendered six rows. The two extra rows carried the generic `sessionLabel` and `weeklyLabel` from provider metadata, which made the copies look like separate quotas.

Before, with four real windows:

| Row | Used | Resets |
| --- | --- | --- |
| Gemini Models | 7.3% | 4d 23h |
| Claude and GPT | 0% | 4h 59m |
| Gemini 5-hour | 3% | 2h 45m |
| Gemini weekly | 7.3% | 4d 23h |
| Claude/GPT 5-hour | 0% | 4h 59m |
| Claude/GPT weekly | 0% | 6d 23h |

Rows 1 and 4 are one window. Rows 2 and 5 are one window. The identical reset times show it.

After, one row per bucket:

| Row | Used | Resets |
| --- | --- | --- |
| Gemini 5-hour | 3% | 2h 45m |
| Gemini weekly | 7.3% | 4d 23h |
| Claude/GPT 5-hour | 0% | 4h 59m |
| Claude/GPT weekly | 0% | 6d 23h |

Every family stays visible. The dashboard is a detail surface, so it follows Preferences > Providers > Antigravity, which also shows all four lanes. Compact surfaces keep their existing behavior: the menu card and the widget still hide a family whose lanes all report zero, per #2875.

`GET /dashboard/v1/snapshot` and `codexbar dashboard` share this builder, so both transports change together. Antigravity snapshots without quota-summary lanes, and every other provider, take the existing path unchanged.

`AntigravityStatusSnapshot.isQuotaSummaryWindowID` becomes public so CodexBarCLI can identify the summary lanes. The check already exists and is already used inside CodexBarCore.

## Commands run

- `swift build`
- `make check` (SwiftFormat and SwiftLint strict): 0 violations
- `make test`: 864 selections in 72 groups, 0 failures, 0 retries

## Tests

`Tests/CodexBarTests/DashboardAntigravityWindowTests.swift` builds the four-bucket shape the probe produces, with `primary` and `secondary` set to copies of two buckets, and covers:

- The snapshot renders one lane per quota bucket.
- The generic representative labels do not appear as rows.
- An Antigravity snapshot without quota-summary lanes keeps the standard session and weekly rows.

These tests use a fixture snapshot. They run no live provider probe.

## Known related issue, not in this change

`codexbar cards` has the opposite defect on the same provider. `CLIRenderer` renders `primary` and `secondary`, then asks the descriptor for extra windows through `extraRateWindowSelector`. Antigravity declares no selector, so the default returns an empty array, and terminal cards show only the two copies and none of the real buckets.

The fix there is not symmetrical with this one: the same selector feeds `MenuDescriptor`, so adding it would change the app menu as well. That belongs in its own change with its own coverage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 - Vertex AI: match Cloud Monitoring quota usage without a `limit_name` to its unambiguous same-metric, same-location limit, restoring quota percentages (#2958). Thanks @MachApple!
+- Antigravity: render one lane per quota bucket on the serve dashboard and in `codexbar dashboard`, instead of repeating two buckets as extra "Gemini Models" and "Claude and GPT" rows.
 
 ## 0.50.0 — 2026-08-15
 
--- a/Sources/CodexBarCLI/DashboardSnapshotBuilder.swift
+++ b/Sources/CodexBarCLI/DashboardSnapshotBuilder.swift
@@ -286,6 +286,13 @@ enum DashboardSnapshotBuilder {
         usage: UsageSnapshot?) -> [DashboardWindowPayload]
     {
         guard let usage else { return [] }
+        // Provider-specific by design: Antigravity's primary and secondary are representatives
+        // copied out of its own quota-summary lanes so the icon and menu bar have standard slots
+        // to read. Emitting all three repeats two lanes under the generic session and weekly
+        // labels, so the dashboard renders the summary lanes alone, one per quota bucket.
+        if provider == .antigravity, let windows = self.antigravityQuotaSummaryWindows(usage) {
+            return windows
+        }
         let labels = self.rateWindowLabels(provider: provider, metadata: metadata, usage: usage)
         var windows: [DashboardWindowPayload] = []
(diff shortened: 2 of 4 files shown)


steipete:
Thanks @urda! Your branch went CONFLICTING against main after today's merge train and the fork isn't maintainer-editable, so I carried this forward as #2966 with your commits and a changelog credit. Closing this one in favor of that refresh — the code is unchanged.

PR #2973, opened 2026-08-16
Generate portable CLI release checksums

### LINKED ISSUE #2971: CLI release checksum files embed runner-local absolute paths
## Summary

The checksum sidecars published for standalone CLI tarballs contain the release runner's absolute temporary path instead of the asset basename.

The SHA-256 digest itself is correct, but the embedded filename is not portable. As a result, the standard verification commands fail after downloading an asset and its sidecar.

## Affected release

I verified every CLI checksum sidecar attached to **v0.50.0**:

- Linux glibc: `x86_64`, `aarch64`
- Linux musl: `x86_64`, `aarch64`
- macOS: `x86_64`, `arm64`

All six contain an absolute runner-local path.

For example, the Linux x86_64 sidecar contains:

```text
8011ad514a66e787a9c4a00a43f06d2780db96571c047c5a2952b3137862a8a6  /tmp/tmp.yw7UERoYdU/CodexBarCLI-v0.50.0-linux-x86_64.tar.gz
```

The macOS sidecars similarly contain paths under `/var/folders/.../T/tmp.*`.

## Steps to reproduce

```bash
curl -fLO https://github.com/steipete/CodexBar/releases/download/v0.50.0/CodexBarCLI-v0.50.0-linux-x86_64.tar.gz
curl -fLO https://github.com/steipete/CodexBar/releases/download/v0.50.0/CodexBarCLI-v0.50.0-linux-x86_64.tar.gz.sha256
sha256sum -c CodexBarCLI-v0.50.0-linux-x86_64.tar.gz.sha256
```

Observed:

```text
sha256sum: /tmp/tmp.yw7UERoYdU/CodexBarCLI-v0.50.0-linux-x86_64.tar.gz: No such file or directory
/tmp/tmp.yw7UERoYdU/CodexBarCLI-v0.50.0-linux-x86_64.tar.gz: FAILED open or read
```

Computing the digest directly confirms that the published digest is otherwise valid:

```bash
sha256sum CodexBarCLI-v0.50.0-linux-x86_64.tar.gz
```

## Expected behavior

The sidecar should contain the asset basename so it can be verified from the download directory:

```text
8011ad514a66e787a9c4a00a43f06d2780db96571c047c5a2952b3137862a8a6  CodexBarCLI-v0.50.0-linux-x86_64.tar.gz
```

Then the conventional command succeeds:

```bash
sha256sum -c CodexBarCLI-v0.50.0-linux-x86_64.tar.gz.sha256
```

On macOS, the equivalent is:

```bash
shasum -a 256 -c CodexBarCLI-v0.50.0-macos-arm64.tar.gz.sha256
```

## Root cause

The current `Package` step in `.github/workflows/release-cli.yml` invokes the checksum utility with the absolute archive path:

```bash
sha256sum "$OUT_DIR/$ASSET" > "$OUT_DIR/$ASSET.sha256"
# or
shasum -a 256 "$OUT_DIR/$ASSET" > "$OUT_DIR/$ASSET.sha256"
```

Both tools preserve the supplied filename in their output, so the runner's `mktemp` directory is written into the published sidecar.

## Suggested fix

Generate and verify the checksum from inside `OUT_DIR`:

```bash
(
  cd "$OUT_DIR"
  if command -v sha256sum >/dev/null 2>&1; then
    sha256sum "$ASSET" > "$ASSET.sha256"
    sha256sum -c "$ASSET.sha256"
  else
    shasum -a 256 "$ASSET" > "$ASSET.sha256"
    shasum -a 256 -c "$ASSET.sha256"
  fi
)
```

The verification step would also prevent a future release from publishing a malformed or unusable sidecar.

## Impact

This does not indicate an incorrect digest or a corrupted archive. It does, however, break the standard checksum-verification workflow and any automation that consumes the published sidecar without rewriting its filename field.

I can submit a focused PR for the workflow change and a regression check.

shockbladenull:
## Summary

- generate CLI release checksum sidecars from inside the output directory so they record only the asset basename
- validate both the recorded filename and digest before uploading release assets
- add a portable regression test that moves the asset and sidecar to a different directory before running `-c`

Closes #2971

## Testing

- `./Scripts/test_release_checksum.sh` — passes
- `./Scripts/lint.sh lint-linux` — portable checks (including the new regression test), JavaScript formatting/linting, and TypeScript checks pass; the final SwiftLint invocation is blocked because the downloaded binary requires newer glibc/libstdc++ than the b2 host provides
- `make check` — blocked on Linux because the existing app-localization check requires macOS `plutil`
- `make test` — blocked before discovery because Swift is not installed on the b2 host

## Real release-workflow proof

- Workflow run: [Release CLI run 31934142399](https://github.com/shockbladenull/CodexBar/actions/runs/31934142399)
- Trigger: `workflow_dispatch` on `shockbladenull:agent/fix-cli-checksum-paths`
- Reviewed head: `9f7d27b230e4b7d7f650d6e1ce8a7424cfeb184e`
- Result: all six real `build-cli` matrix jobs passed: Linux glibc x86_64/aarch64, Linux musl x86_64/aarch64, and macOS x86_64/arm64.
- Every workflow `Package` step generated and validated its sidecar; both macOS package jobs also reported `: OK`.
- I downloaded all six workflow artifacts onto b2, a different host and path from the GitHub runner temporary directories. For every sidecar, I first asserted that its recorded filename exactly matched the archive basename, then ran `sha256sum -c` from the downloaded artifact directory. All six passed.
- Since this was a manual proof run rather than a release event, the publish-only and Homebrew-update jobs were skipped; no release or tap update was made.

```text
CodexBarCLI-v0.50.0-pr2973-proof-linux-aarch64.tar.gz: OK
CodexBarCLI-v0.50.0-pr2973-proof-linux-musl-aarch64.tar.gz: OK
CodexBarCLI-v0.50.0-pr2973-proof-linux-musl-x86_64.tar.gz: OK
CodexBarCLI-v0.50.0-pr2973-proof-linux-x86_64.tar.gz: OK
CodexBarCLI-v0.50.0-pr2973-proof-macos-arm64.tar.gz: OK
CodexBarCLI-v0.50.0-pr2973-proof-macos-x86_64.tar.gz: OK
Verified 6 portable checksum sidecars
```

### DIFF
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -409,11 +409,7 @@ jobs:
           ASSET="CodexBarCLI-${SAFE_REF_NAME}-${{ matrix.asset-platform }}-${{ matrix.asset-arch }}.tar.gz"
           (cd "$OUT_DIR" && tar czf "$ASSET" \
             CodexBarCLI codexbar VERSION CodexBar_CodexBarCore.bundle)
-          if command -v sha256sum >/dev/null 2>&1; then
-            sha256sum "$OUT_DIR/$ASSET" > "$OUT_DIR/$ASSET.sha256"
-          else
-            shasum -a 256 "$OUT_DIR/$ASSET" > "$OUT_DIR/$ASSET.sha256"
-          fi
+          Scripts/generate_release_checksum.sh "$OUT_DIR/$ASSET"
 
           echo "out_dir=$OUT_DIR" >> "$GITHUB_OUTPUT"
           echo "asset=$ASSET" >> "$GITHUB_OUTPUT"
--- a/Scripts/generate_release_checksum.sh
+++ b/Scripts/generate_release_checksum.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+if [[ "$#" -ne 1 ]]; then
+  echo "Usage: $(basename "$0") <asset>" >&2
+  exit 2
+fi
+
+ASSET_PATH="$1"
+if [[ ! -f "$ASSET_PATH" ]]; then
+  echo "Release asset not found: $ASSET_PATH" >&2
+  exit 1
+fi
+
+OUT_DIR="$(cd "$(dirname "$ASSET_PATH")" && pwd)"
+ASSET="$(basename "$ASSET_PATH")"
+
+(
+  cd "$OUT_DIR"
+  if command -v sha256sum >/dev/null 2>&1; then
+    CHECKSUM=(sha256sum)
+  elif command -v shasum >/dev/null 2>&1; then
+    CHECKSUM=(shasum -a 256)
+  else
+    echo "sha256sum or shasum is required." >&2
+    exit 1
+  fi
+
+  "${CHECKSUM[@]}" "$ASSET" > "$ASSET.sha256"
+  read -r _ CHECKSUM_ASSET < "$ASSET.sha256"
+  if [[ "$CHECKSUM_ASSET" != "$ASSET" ]]; then
+    echo "Checksum file references an unexpected path: $CHECKSUM_ASSET" >&2
+    exit 1
+  fi
+  "${CHECKSUM[@]}" -c "$ASSET.sha256"
+)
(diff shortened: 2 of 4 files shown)


shockbladenull:
@clawsweeper re-review

PR #3016, opened 2026-08-17
Add custom-pricing overlay above models.dev and builtin rates

Yuxin-Qiao:
## Summary
- Load `custom-pricing.json` as an exact-match overlay: overlay > models.dev > builtin.
- Treat missing overlay fields as unknown (not free) and `0` as free.
- Invalidate the Codex pricing fingerprint when the overlay changes so caches rescan.

Stacked PR 2/5 (Wave C). Depends on #3015; review the custom-pricing commit once that lands.

## Review follow-up
- `codexAggregateCostUSD` consults the overlay (including free `0`) before bundled/models.dev rates.
- Overlay JSON is loaded once per scan and passed through `CodexReportDayPricingContext.customPricing`.
- Overlay cost helpers live in `CostUsagePricing+Overlay.swift` so `CostUsagePricing` stays under `type_body_length`.

## Test plan
Fixtures only. No Keychain, no live accounts, no developer Application Support overlay file.

```
CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1
make check
# Codex parser hash is current (b7726c3088c14277)
# Done linting! Found 0 violations, 0 serious in 1928 files.

CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --filter \
  'aggregate fallback consults the overlay before bundled rates'
```

```
✔ aggregate fallback consults the overlay before bundled rates
```

- [x] Overlay aggregate fallback test above
- [x] `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --filter CostUsageCustomPricingTests`
- [ ] Overlay fixture does not read the developer Application Support file
- [ ] Unknown model stays unpriced instead of showing $0

PR #3015, opened 2026-08-17
Add spend provenance, pinned day timezone, and Codex fork fail-closed

Yuxin-Qiao:
## Summary
- Introduce `CostProvenance` plus priced/unpriced/unmetered/estimated coverage counts for local spend history.
- Pin spend day buckets to a stored IANA timezone so travel/TZ changes cannot reshuffle midnight usage.
- Lift reasoning tokens into the public cost models and fail-close missing-parent Codex forks as unmetered instead of billing copied prefixes.

Stacked PR 1/5 (Wave A). Follow-ups: custom pricing, OpenCodex parser, dashboard UI, CLI JSON.

## Review follow-up
- Cached/catch-up Codex loads now pass the pinned `costUsageBucketCalendar` instead of the current zone.
- Snapshot `costProvenance` is explicit. Summaries no longer infer list-price from “has cost, no metered.”
- Windowed summaries only attach `meteredCostUSD` when `days >= historyDays`.
- Coverage counters (`unpriced` / `unmetered` / `estimatedRequestCount`) round-trip through the previous-report cache.
- Undated fork sessions use mtime; if the day still cannot be pinned they are excluded, not hung on `untilKey`.
- `CostUsageScannerBreakdownTests` includes token-class fields so `ModelBreakdown` Equatable matches CI.

## Test plan
Fixtures only. No Keychain, no live accounts, no `~/.opencodex`.

```
CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1
make check
# Codex parser hash is current (b7726c3088c14277)
# Done linting! Found 0 violations, 0 serious in 1928 files.

CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --filter \
  'cached snapshot reads keep the pinned timezone|vendor reported snapshots stay vendor metered|shorter summaries omit snapshot-wide metered|cached previous reports round-trip|codex daily report parses token counts'
```

```
✔ cached snapshot reads keep the pinned timezone instead of the current zone
✔ vendor reported snapshots stay vendor metered without meteredCostUSD
✔ shorter summaries omit snapshot-wide metered spend
✔ cached previous reports round-trip coverage counters
✔ codex daily report parses token counts and caches
```

- [x] Focused provenance/cache/breakdown tests above
- [ ] Confirm Settings → Usage & Spend still loads with cost tracking off
- [ ] Missing-parent fork sessions stay out of priced totals

PR #3018, opened 2026-08-17
Add a read-only OpenCodex usage.jsonl parser and independent cache

Yuxin-Qiao:
## Summary
- Parse `~/.opencodex/usage.jsonl` (or `OPENCODEX_HOME`) as a spend-only source, not a quota Provider.
- Persist an independent `opencodex-usage.sqlite` with requestId dedupe; never write Codex cost tables.
- Tests use fixtures only and do not read the developer home log.

Stacked PR 3/5 (Wave D). Depends on #3015 and #3016.

## Review follow-up
- Unreported/unsupported rows with no usage stay unpriced (not `$0`).
- Session/today totals use `entry(in:forLocalDayContaining:)` instead of `daily.last`.
- Unpriced estimated requests move into the unpriced bucket and still count as 1 request.
- Cache writes `INSERT OR REPLACE` and dedupe by `requestID` before insert.
- OpenCodex snapshots set `costProvenance: .listPriceEstimate`.

## Test plan
Fixtures only. No Keychain, no live accounts, no `~/.opencodex`.

```
CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --filter OpenCodexUsageParserTests
```

```
✔ unreported rows without usage stay unpriced instead of zero spend
✔ unpriced estimated requests count once
✔ session totals use the current day instead of the latest historical day
✔ does not resolve a default home while tests are running
✔ duplicate request ids replace instead of aborting the cache write
✔ parses persisted usage rows without reading the developer home
✔ skips malformed lines and keeps nil usage classes unset
✔ aggregates a fixture log into an independent snapshot
✔ Suite OpenCodexUsageParserTests passed (8 tests)
```

- [x] OpenCodexUsageParserTests (8/8)
- [ ] Missing usage.jsonl is a silent source absence, not a global error
- [ ] Cache file is `opencodex-usage.sqlite`


Yuxin-Qiao:
@clawsweeper re-review

PR #3021, opened 2026-08-17
feat: add TOON usage output

### LINKED ISSUE #2996: Feature request: --format toon for agent-compatible structured output
## Summary
Add `--format toon` to emit quota data in TOON structured format, making CodexBar 
directly consumable by agents without JSON parsing.

## Motivation
Today, agents that need quota data Parse CodexBar's JSON output themselves

TOON is human-readable AND machine-parseable — one output serves both the operator 
glancing at terminal and the agent routing on headroom.

## Proposed behaviour
```bash
codexbar usage --format toon
codexbar usage --provider zai --format toon
```

Output example (matching TOON spec):
```
providers[6]{provider,plan,source,status,refreshedAt}:
  claude,max,oauth,fresh,2026-08-17T02:00:00Z
  codex,pro,oauth,fresh,2026-08-17T02:00:00Z
  zai,pro,api,fresh,2026-08-17T02:00:00Z
windows[7]{provider,id,percentRemaining,resetsAt,pace}:
  claude,five_hour,93,2026-08-17T02:00:00Z,behind
  zai,primary,19,2026-08-17T03:56:41Z,farAhead
effective[N]{provider,scope,effectivePercentRemaining,boundedBy}:
  ...
```

## Why this fits CodexBar
- CodexBar has the richest provider coverage (73 providers)
- The data is already computed; this is just a formatter
- TOON is the emerging standard for agent-readable CLI output
- Reduces ecosystem fragmentation vs maintaining parallel tools

## Scope (MVP)
Single new formatter mapping existing JSON structure to TOON. No new data, 
no new providers — pure presentation layer.

## References
- TOON format spec: https://toon-format.github.io/

elijahfriedman:
Resolves #2996

## Summary

* Adds `codexbar usage --format toon`, which emits the same data as `--format json`, encoded as TOON v4.1—a denser, human-readable format intended for agents that would otherwise parse JSON.
* Adds `ToonFormatter`, a generic `Encodable` → TOON encoder rather than a hand-mapped schema. It round-trips values through a custom `Encoder` conformance into an order-preserving node tree instead of using `JSONSerialization`, which does not preserve key order, then serializes that tree according to the TOON spec.

  * Uniform arrays of scalar-only objects with identical ordered field sets collapse into table form: `rows[N]{label,value}:`.
  * Non-uniform arrays fall back to indented list form.
* Reuses the existing `--format json` fetch/render pipeline for `--format toon`, including the same credits, color, and account behavior. Only the final output step changes.
* Does not add `.toon` to the shared `OutputFormat` enum. That enum is used by several other commands with exhaustive format switches (`hooks`, `config`, `cost`, `cache`, and `cookie`), and adding full TOON support across those commands is outside the scope of this change.
* `usage --format toon` is the only command wired up, and TOON recognition is scoped to that command in the resolver as well, so no other command's `--format` handling changes. This intentionally keeps the implementation to the MVP described in #2996: a presentation-only formatter over the existing JSON schema, with no new data or providers.

### Addressed review feedback

* **TOON recognition is restricted to the `usage` command.**

  * The previous revision resolved `toon` inside the shared `CLIOutputPreferences.from(values:)`, which is also called by `cost`, `cache`, `config`, `hooks`, `diagnose`, `guard`, `plugins`, and `cookie`. Those commands advertise only `text | json`, so they silently accepted `--format toon` and emitted JSON where they previously emitted text.
  * `from(values:)` and both `resolveOutputFormat` overloads now take an `allowsToon` parameter that defaults to `false`. `resolveUsageOutputPreferences` is the only site that passes `true`.
  * The argv-level bootstrap scanner derives the same verdict from the command itself via `commandSupportsToon(argv:)`, which mirrors `effectiveArgv`: a leading `usage`, or a bare `codexbar --format toon` that resolves to the implicit `usage` command.
  * Every other command is back to its exact pre-TOON decoding: `toon` is an unrecognized `--format` value that falls through to text, or to JSON only when `--json` / `--json-only` is also passed—identical to how `--format xml` has always behaved.
  * `docs/cli.md` now states this explicitly: `usage --format toon` is the only command that supports it, and every other command accepts only `--format text|json`.

* **Early failures now render as TOON, both before and after argument parsing.**

  * `CLIOutputPreferences` carries a `toonRequested` flag.
  * Post-parse paths inside `runUsage`—including `Self.exit`, `printError`, and `loadConfig`—respect it.
  * `CLIOutputPreferences.from(argv:)` provides an argv-level bootstrap scanner for failures that happen before `ParsedValues` exists, such as `Program.resolve` rejecting an unknown option.
  * Both `--format toon` and `--format=toon` are recognized, so early exits remain consistent with the requested output format rather than falling back to JSON or plain stderr text.

* **`--help` advertises TOON.**

  * `CLIHelp.usageHelp` and `CLIHelp.rootHelp`—the hand-written help strings that `CLIEntry` actually prints ahead of Commander's descriptor—now advertise `text|json|toon` and include a `--format toon` example. `codexbar cost` and the other commands keep advertising `--format text|json`, which now matches their behavior exactly.

* **Tabular collapse no longer fabricates fields.**

  * CodexBar's JSON payloads use `encodeIfPresent` for optional fields, so an absent value means the key is omitted rather than encoded as `null`.
  * Table form now requires every row to have exactly the same fields in exactly the same order.
  * Otherwise, the formatter falls back to list form, where each row contains only the fields actually present. No synthetic `null` values are introduced.

* **Non-finite doubles are rejected.**

  * The TOON encoder rejects non-finite floating-point values, keeping the output valid and avoiding values that TOON consumers cannot reliably round-trip.

* **Commander does not reject `--format toon` during parsing.**

  * I checked the vendored Commander source: `CommandParser.parse` does not validate an option value against its declared type. It only rejects unknown option names and missing values.
  * The `@Option` property's `ExpressibleFromArgument` constraint is compile-time-only, and `CommanderError.invalidValue` has no call sites in the package.
  * This was also confirmed empirically: `usage --format toon` parses and renders TOON successfully.
  * As a result, the P1 concern that Commander rejects `--format toon` at parse time does not reproduce, and no parser change is required.

## Tests

* **`ToonFormatterTests`**

  * Flat objects
  * Tabular collapse for uniform scalar arrays
  * List-form fallback for non-uniform field sets
  * Omitted optional fields, including assertions that no `null` is fabricated and no `{...}` table header is emitted
  * Inline primitive arrays
  * Empty arrays
  * TOON quoting rules
  * ISO 8601 dates
  * End-to-end `ProviderPayload` encoding

* **`CLIOutputTests`**

  * `renderProviderPayloads` selects TOON vs. JSON based on `toonRequested`
  * The argv bootstrap scanner recognizes both `--format toon` and `--format=toon`
  * A later `--json` overrides an earlier TOON request
  * **Command boundary:** the same `ParsedValues` resolves to TOON through `resolveUsageOutputPreferences` and to plain text through the shared `from(values:)`
  * **Command boundary:** `decodeFormat` returns `.text` for `--format toon`, and `.json` only when `--json` is present—identical to its handling of an unrelated unsupported value such as `xml`
  * **Command boundary:** an argv sweep over `cost`, `cache`, `config`, `hooks`, `diagnose`, `guard`, and `serve` asserts no TOON request and no format change, while the implicit-usage form (`--format toon` with no command) still resolves to TOON

* **`CLIEntryTests`**

  * Two real entry-point regression tests spawn the built binary with an unrecognized option, exercising a genuine `Program.resolve` parse failure
  * Both `--format toon` and `--format=toon` are verified to emit the TOON error payload on stdout
  * **Command boundary:** `cost`, `diagnose`, and `cache` invoked with `--format toon` and an unrecognized option must keep stdout empty and report the parse failure as text on stderr

Each of the new command-boundary tests was verified to be discriminating: with the scoping fix temporarily reverted, all three fail; with it restored, all three pass.

* **`CLIProviderSelectionTests`**

  * `usageHelp` and `rootHelp` advertise `--format text|json|toon`

## Verification

* `swift build` — clean
* `make check` — clean, 0 lint violations
* Focused CLI suites (`CLIOutputTests`, `CLIEntryTests`) — passing, including all new command-boundary regressions
* Full `make test` — every suite passes except two pre-existing `AdaptiveRefreshTimerTests` timing failures (`AdaptiveRefreshTimerTests.swift:39` and `:78`, both 30s `CancellationError` while waiting on `store.adaptiveRefreshScheduledAt`). Those reproduce identically with this PR's changes stashed out; they live in the app's `UsageStore` refresh-timer path and come from unrelated commits on this branch.
* Validated real CLI behavior with:

  * `codexbar usage --bogus-flag-xyz --format toon`
  * `codexbar usage --bogus-flag-xyz --format=toon`
* Decoded real output with the official reference decoder, `@toon-format/toon` v4.1.1, confirming interoperability with the published TOON spec.

### Real CLI behavior

```console
$ codexbar usage --bogus-flag-xyz --format toon
[1]:
  - provider: cli
    source: cli
    error:
      code: 1
      message: Unknown option --bogus-flag-xyz
      kind: args

$ codexbar usage --bogus-flag-xyz --format=toon
[1]:
  - provider: cli
    source: cli
    error:
      code: 1
      message: Unknown option --bogus-flag-xyz
      kind: args
```

### Interoperability with a real TOON consumer

Decoded real output using the official reference decoder, `@toon-format/toon` v4.1.1, matching the spec version targeted by this encoder and independent of CodexBar's own tests:

1. **CLI parse failure:** The transcript above decodes to the same data as CodexBar's `--format json` output for the same invocation, byte-for-byte after accounting for key order.
2. **Tabular collapse:** `rows[2]{id,label}:` decodes to `{"rows":[{"id":1,"label":"Ada"},{"id":2,"label":"Bob"}]}`.
3. **Omitted optional field:** The post-fix `rows[2]:` list form decodes to `{"rows":[{"id":1,"note":"hi"},{"id":2}]}`. Row 2 has no `note` key, exactly matching the JSON produced by `encodeIfPresent`.

## Example

```console
$ codexbar usage --provider claude --format toon
[1]:
  - provider: claude
    account: null
    source: oauth
    usage:
      primary:
        usedPercent: 7
        windowMinutes: 300
      details[1]:
        - title: Usage summary
          rows[2]{label,value}:
            Requests,"120"
            Tokens,4.2k
      updatedAt: "2026-08-17T02:00:00Z"
```

### DIFF
--- a/Sources/CodexBarCLI/CLIErrorReporting.swift
+++ b/Sources/CodexBarCLI/CLIErrorReporting.swift
@@ -26,13 +26,8 @@ extension CodexBarCLI {
         ProviderErrorPayload(code: code.rawValue, message: message, kind: kind)
     }
 
-    static func makeCLIErrorPayload(
-        message: String,
-        code: ExitCode,
-        kind: CLIErrorKind,
-        pretty: Bool) -> String?
-    {
-        let payload = ProviderPayload(
+    static func makeCLIErrorProviderPayload(message: String, code: ExitCode, kind: CLIErrorKind) -> ProviderPayload {
+        ProviderPayload(
             providerID: "cli",
             account: nil,
             version: nil,
@@ -43,6 +38,15 @@ extension CodexBarCLI {
             antigravityPlanInfo: nil,
             openaiDashboard: nil,
             error: ProviderErrorPayload(code: code.rawValue, message: message, kind: kind))
+    }
+
+    static func makeCLIErrorPayload(
+        message: String,
+        code: ExitCode,
+        kind: CLIErrorKind,
+        pretty: Bool) -> String?
+    {
+        let payload = self.makeCLIErrorProviderPayload(message: message, code: code, kind: kind)
         return self.encodeJSON([payload], pretty: pretty)
     }
 
@@ -83,6 +87,20 @@ extension CodexBarCLI {
         }
     }
 
+    /// Renders as TOON when the caller requested `usage --format toon`, JSON otherwise. Error/exit
+    /// paths must honor this too, or `--format toon` silently falls back to JSON on any early failure
+    /// (invalid arguments, config load errors, provider errors).
+    static func renderProviderPayloads(_ payloads: [ProviderPayload], output: CLIOutputPreferences) -> String {
+        if output.toonRequested {
+            return ToonFormatter.encode(payloads)
+        }
+        return self.encodeJSON(payloads, pretty: output.pretty) ?? ""
+    }
+
+    static func printProviderPayloads(_ payloads: [ProviderPayload], output: CLIOutputPreferences) {
+        print(self.renderProviderPayloads(payloads, output: output))
+    }
+
     static func exit(
         code: ExitCode,
         message: String? = nil,
@@ -91,14 +109,9 @@ extension CodexBarCLI {
     {
         if self.shouldPrintExitError(code: code, message: message) {
             if let output, output.usesJSONOutput {
-                let payload = self.makeCLIErrorPayload(
-                    message: message ?? "",
-                    code: code,
-                    kind: kind,
-                    pretty: output.pretty)
-                if let payload {
-                    print(payload)
-                }
+                self.printProviderPayloads(
+                    [self.makeCLIErrorProviderPayload(message: message ?? "", code: code, kind: kind)],
+                    output: output)
             } else if let message {
                 self.writeStderr("\(message)\n")
             }
@@ -123,7 +136,7 @@ extension CodexBarCLI {
                 antigravityPlanInfo: nil,
                 openaiDashboard: nil,
                 error: self.makeErrorPayload(error, kind: kind))
-            self.printJSON([payload], pretty: output.pretty)
+            self.printProviderPayloads([payload], output: output)
         } else {
             self.writeStderr("Error: \(error.localizedDescription)\n")
         }
--- a/Sources/CodexBarCLI/CLIHelp.swift
+++ b/Sources/CodexBarCLI/CLIHelp.swift
@@ -62,7 +62,7 @@ extension CodexBarCLI {
         CodexBar \(version)
 
         Usage:
-          codexbar usage [--format text|json]
+          codexbar usage [--format text|json|toon]
                        [--json]
                        [--json-only]
                        [--json-output] [--log-level <trace|verbose|debug|info|warning|error|critical>] [-v|--verbose]
@@ -72,8 +72,10 @@ extension CodexBarCLI {
                        [--web-timeout <seconds>] [--web-debug-dump-html] [--antigravity-plan-debug] [--augment-debug]
 
         Description:
-          Print usage from enabled providers as text (default) or JSON. Honors your in-app toggles.
+          Print usage from enabled providers as text (default), JSON, or TOON. Honors your in-app toggles.
           Output format: use --json (or --format json) for JSON on stdout; use --json-output for JSON logs on stderr.
+          --format toon emits the same payload as --format json, rendered as TOON (github.com/toon-format/spec)
+          for token-cheaper agent consumption.
           Source behavior is provider-specific:
           - Codex: OpenAI web dashboard (usage limits, credits remaining, code review remaining, usage breakdown).
             Auto falls back to Codex CLI only when cookies are missing.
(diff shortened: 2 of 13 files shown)


elijahfriedman:
@clawsweeper re-review

elijahfriedman:
@clawsweeper re-review

PR #3022, opened 2026-08-17
Fix Codex RPC timeout classification

Chipagosfinest:
## What changed

Classify a Codex app-server request as timed out before terminating the subprocess that owns stdout.

## Why

The previous timeout task terminated the process and then threw. Closing stdout could wake the request body first, causing the same timeout to escape as a misleading EOF/malformed-response error. Selecting the timer result before process termination makes the reported failure deterministic.

## Scope

- one production file
- no settings, UI, spend-model, or provider behavior changes
- extracted from #2759 so the runtime fix can be reviewed independently

## Verification

- `swift test --filter CodexUsageFetcherFallbackTests` — 13 tests passed, including the real spawned-process hung RPC cases
- `make check` — passed with zero SwiftFormat or SwiftLint violations

Supersedes the timeout portion of #2759.

### DIFF
--- a/Sources/CodexBarCore/UsageFetcher.swift
+++ b/Sources/CodexBarCore/UsageFetcher.swift
@@ -831,6 +831,11 @@ enum RPCWireError: Error, LocalizedError {
     }
 }
 
+private enum RPCRequestRaceResult<Value: Sendable>: Sendable {
+    case value(Value)
+    case timedOut
+}
+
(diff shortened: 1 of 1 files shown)


steipete:
Thanks @Chipagosfinest! This landed via #3025, which carried your commits onto current main and fixed the stale architecture-gatekeeper anchors that were failing CI (your diff was fine — the base predated recent churn in UsageFetcher.swift). The changelog credits you.

PR #3019, opened 2026-08-17
Align codexbar cost JSON with provenance, mix, and OpenCodex

Yuxin-Qiao:
## Summary
- Extend `codexbar cost --json` with token mix, provenance, coverage, and reasoning tokens.
- Emit an OpenCodex payload when the opt-in setting is on; do not invent project paths.
- Reuse dashboard helpers for defaults so CLI and Settings share one schema.

Stacked PR 5/5 (Wave E). Depends on #3017.

## Review follow-up
- Empty OpenCodex snapshots (`daily` + `sessions` empty) are not treated as a present source, so hide-native cannot drop real Codex.
- OpenCodex JSON provenance is always `listPriceEstimate`; `projects` stays empty.

## Test plan
Fixtures only. No Keychain, no live accounts, no `~/.opencodex`.

```
CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1
make check
# Done linting! Found 0 violations, 0 serious in 1928 files.

CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --filter \
  'empty OpenCodex snapshots are not treated as a present source|openCodex JSON payload stays on a separate source'
```

```
✔ empty OpenCodex snapshots are not treated as a present source
✔ openCodex JSON payload stays on a separate source and omits invented projects
```

- [x] Focused empty-source and CLI JSON tests above
- [x] `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --filter CLICostTests`
- [ ] JSON includes provenance/coverage without a third schema
- [ ] OpenCodex `projects` array stays empty

PR #3023, opened 2026-08-17
Show selected-provider spend in Overview

Chipagosfinest:
## Summary

- add a compact Usage & Spend summary above the provider cards in Overview
- scope spend and token totals to the providers currently shown in Overview
- make incomplete provider coverage explicit and keep distinct currencies separate
- include focused summary tests and a reproducible UI proof image

## Scope

This is the focused Overview slice extracted from #2759 and rebuilt on current `main`. It preserves the existing six-provider Overview fallback and does not change provider refresh, Settings defaults, share/export, or token allocation.

Current `main` already contains the newer All Time, partial-total, project-model, and project-panel foundations, so none of those are duplicated here.

## Verification

- `swift test --filter OverviewSpendSummaryTests`
- `CODEXBAR_SPEND_PROOF_DIR=.github/pr-proof swift test --filter SpendDashboardScreenshotRenderTests`
- `make check`

Overview spend summary


steipete:
Thanks @Chipagosfinest! This landed via #3027, which carried your commits onto current main (after the #3015–#3019 spend stack reshaped the same surfaces), reconciled the semantics — partial totals render with ~ and unpriced spend stays unavailable, matching the new coverage model — and refreshed the stale gatekeeper anchors that were failing CI. The changelog credits you.

PR #3051, opened 2026-08-18
Preserve spend totals while local history catches up

mauriciopolvora:


mauriciopolvora:
@clawsweeper re-review

steipete:
Landed as 26132179a00a684e37c836c9a3e5653c71058df3 after rebasing onto current main and preserving contributor credit. Verified account-, history-window-, and time-zone-scoped spend retention with real session-log append/catch-up fixtures and real SQLite migration from the current production parser hash; existing usage rows survive without rebuilding while unsafe old retained reports are discarded. Full local suite: 931 selections across 78 groups, zero failures, retries, or timeouts. Formatting/lint, provider architecture gate, independent review, and hosted CI all passed. Thank you @mauriciopolvora!

PR #3052, opened 2026-08-18
Preserve lane layouts on downgrade and refresh tertiary tokens

giuseppebisemi:
## Summary

A Cursor (or other provider) layout that pins a direct usage lane can no longer wipe custom menu-bar settings when 0.53.0 opens the same defaults: the current layout lives in V2 keys, and the legacy keys get a Session/Weekly/Auto projection that the older decoder still understands.

Provider overrides use each descriptor's semantic windows, so a Kimi primary/secondary layout falls back to Weekly then Session instead of swapping the 7-day and 5-hour quotas. If 0.53.0 later edits a layout or removes an override, the next upgrade prefers that legacy blob whenever it no longer matches the V2 projection — including a removed override that would otherwise come back from V2.

Direct tertiary tokens (Cursor Third Party %) also join the icon redraw signature, so 60% → 61% refreshes even when Total and Cursor stay put.

Startup now materializes those V2 + fallback copies when only the pre-V2 keys exist, so a user who already saved a lane layout on current main does not need to re-open the editor before downgrading.

## Validation

These paths are settings and signature seams, not menu-bar pixels.

After-fix proof of the pre-V2 upgrade path: seed only `menuBarLayout` / `menuBarLayoutOverrides` with `lanePercent`, then run `SettingsStore.init`. The load path writes `menuBarLayoutV2` (keeps lanes) and projects the legacy keys to percent windows a 0.53.0 decoder can read (Kimi: weekly then session).

- JSON blobs: `.github/pr-proof/menubar-lane-startup-migration.json`
- Command log: `.github/pr-proof/menubar-lane-startup-migration.log`

```text
CODEXBAR_LAYOUT_PROOF_DIR=.github/pr-proof swift test --filter 'MenuBarLayoutTests|StatusItemIconObservationSignatureTests|ProviderArchitectureGatekeeperTests'
Test run with 92 tests in 3 suites passed after 85.698 seconds.
```

Covered after-fix cases:

- Pre-V2 lane layouts migrate at startup into V2 + an older-readable fallback (no editor write)
- Kimi override dual-write is Weekly + Session, not Session + Weekly
- A 0.53.0-style rewrite of the legacy layout key wins over stale V2 and is not overwritten at startup
- Removing a provider override in the legacy dictionary is not resurrected from V2
- Re-saving the same fallback still restores the V2 lane tokens
- Missing legacy key is filled from V2 on the next launch
- Tertiary-only layout signature changes at 60% → 61%; unused tertiary churn does not

Also ran `./Scripts/lint.sh lint`.

Related: #3039


giuseppebisemi:
Follow-up to #3039 for the two P1s from the last ClawSweeper pass:

1. Dual-write `menuBarLayoutV2` / `menuBarLayoutOverridesV2` plus an older-readable Session/Weekly/Auto fallback so 0.53.0 does not drop custom layouts (or wipe every provider override) when it sees `lanePercent`.
2. Include selected direct lanes in the icon observation signature so a tertiary-only layout (Cursor Third Party %) refreshes when that window moves independently.

@clawsweeper re-review

giuseppebisemi:
Addressed the two compatibility findings on `1bd422a2`:

1. Provider-aware legacy mapping — Kimi primary/secondary now project through the descriptor semantic windows (Weekly then Session), with a dual-write regression test.
2. Legacy reconciliation — loaders keep V2 only when the legacy blob still matches that projection. An older-release edit or a removed override wins; an unchanged fallback still restores the lane tokens.

Tertiary redraw coverage is unchanged: selected direct lanes are in the icon observation signature (60% → 61% changes it; unused tertiary churn does not).

After-fix proof is the focused suite on this commit: 90 tests across MenuBarLayoutTests, StatusItemIconObservationSignatureTests, and ProviderArchitectureGatekeeperTests.

@clawsweeper re-review

giuseppebisemi:
Addressed the remaining P2 on `c537b575`:

1. Startup migration — `SettingsStore` load now materializes `menuBarLayoutV2` / `menuBarLayoutOverridesV2` plus the older-readable projection when those current keys are missing. A pre-V2 lane layout (the keys current main writes) becomes downgrade-safe on first launch, without an editor save. Divergent 0.53.0 edits are still left untouched.

2. After-fix proof — seeded only the pre-V2 keys with `lanePercent`, ran the real `SettingsStore.init` path, and captured the resulting blobs in `.github/pr-proof/menubar-lane-startup-migration.json`. The legacy global decodes as session percent; Kimi overrides decode as weekly then session. Focused suite: 92 tests in 3 suites.

Tertiary redraw coverage is unchanged: selected direct lanes are in the icon observation signature (60% → 61% changes it; unused tertiary churn does not).

@clawsweeper re-review

PR #3096, opened 2026-08-20
Fix agent session menu width

KaranocaVe:
## Summary

- Render Agent Session actions as fixed-width hosted menu rows with single-line tail truncation.
- Keep long session/project text out of menu-width measurement and cache keys.
- Preserve full labels in tooltips and add a regression test for long project names.

## Validation

- swift build
- make check
- Focused SwiftFormat and SwiftLint on changed files

The full test suite and a live app relaunch were intentionally not run for this scoped UI change.


KaranocaVe:
this robot is crazy

PR #3101, opened 2026-08-20
fix(spend): notify shared dashboard on independent token pubs

Yuxin-Qiao:
Independent spend-dashboard providers (`usesSpendDashboardIndependentTokenSnapshot` — Claude/Cursor/OpenCodex/etc) published to `spendDashboardTokenPublications` but `publishSpendDashboardTokenSnapshotState:171` never called `synchronizeSharedSpendDashboardAfterTokenPublication`, and the sync itself gated on `provider == .codex` (`UsageStore+SpendDashboardPublication.swift:64`).

Result: background `refreshSpendDashboardTokenUsageNow` completions didn't update the shared `SpendDashboardController` publication, so the *Usage & Spend* pane and Overview spend stayed stale until next config change or pane re-open — the "won't auto refresh silently" symptom.

**Fix**
* `UsageStore+SpendDashboardPublication.swift:64` widen guard to `provider == .codex || isIndependent`
* `UsageStore+SpendDashboardTokenCost.swift:171` call `synchronizeSharedSpendDashboardAfterTokenPublication` after publishing

Background imports now drive the shared dashboard like the Codex path already did.

*Verified:* `swiftformat` + `swiftlint --strict` clean, `SpendDashboardPublicationTests` covers shared ownership.

### DIFF
--- a/Sources/CodexBar/UsageStore+SpendDashboardPublication.swift
+++ b/Sources/CodexBar/UsageStore+SpendDashboardPublication.swift
@@ -62,8 +62,10 @@ extension UsageStore {
     }
 
     func synchronizeSharedSpendDashboardAfterTokenPublication(for provider: UsageProvider) {
-        // Provider-specific by design: regular Codex publication triggers the account-scoped spend producer.
-        guard provider == .codex, self.sharedSpendDashboardObservationStarted else { return }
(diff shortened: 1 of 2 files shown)


Yuxin-Qiao:
Superseded by larger silent-refresh PR

PR #3109, opened 2026-08-20
Add BigModel CN account balance to z.ai provider

RunhuaHuang:
## Summary

Adds the BigModel CN (open.bigmodel.cn) pay-as-you-go **account balance** to the z.ai provider, rendered as a best-effort "Account balance" row inside "Quota details".

- The bundled `zai` plugin now calls `GET https://www.bigmodel.cn/api/biz/account/query-customer-account-report` for `bigmodel-cn` region accounts: `¥` available balance (`availableBalance`, falling back to `balance`), with a secondary line summarizing `recharged / granted / spent`.
- Endpoint verified against the live console API (2026-08): it accepts **both** `Bearer <key>` and raw-key `Authorization`, so the plugin's existing bearer broker works unchanged.
- z.ai global has no documented equivalent — the global region skips the row *and* the extra request entirely.
- Balance lookup failures never break quota display (best-effort, same pattern as the existing model-usage detail fetches).
- New `Z_AI_BALANCE_URL` environment override (HTTPS-validated via `ProviderEndpointOverrideValidator`) plus `ZaiEndpointRouter.resolveBalanceURL`, injected to the plugin as `Z_AI_BALANCE_ENDPOINT`.

Background: this is the console's internal account-report API (no official docs). The same endpoint powers the "智谱 API 余额" channel in the open-source [QuotaX](https://github.com/RunhuaHuang) quota dashboard, where the response shape below was captured live:

```json
{
  "code": 200, "msg": "操作成功", "success": true,
  "data": {
    "balance": 42.5, "availableBalance": 40.0, "rechargeAmount": 100.0,
    "giveAmount": 20.0, "totalSpendAmount": 77.5, "frozenBalance": 2.5,
    "creditStatus": "NOT_OPEN", "isKA": false
  }
}
```

## Testing

Commands run:

```
swift build --build-tests
swift test --filter ZaiPluginBalanceTests        # new suite, 4 passed
swift test --filter Zai                          # 29 passed (6 suites)
swift test --filter "ProviderPluginRuntimeTests|ProviderPluginParityTests|ProviderPluginExtensionParityTests|UserProviderPluginPortableTests|ProviderPluginDetailsParityTests"   # 72 passed (5 suites)
swiftlint --strict <changed paths>               # 0 violations
swiftformat <changed files>
```

New `TestsPlugin/ZaiPluginBalanceTests.swift` covers:
- CN snapshot renders the balance row (value + secondary provenance line) and issues the request to the right host/path
- Balance endpoint failure (HTTP 500) keeps the quota snapshot intact with no balance row
- Global region issues no balance request at all
- Router default / explicit override / nil-for-global resolution

No UI change beyond an additional detail row, so no screenshots.

### DIFF
--- a/Sources/CodexBarCore/Providers/Zai/ZaiAPIRegion.swift
+++ b/Sources/CodexBarCore/Providers/Zai/ZaiAPIRegion.swift
@@ -50,11 +50,23 @@ public enum ZaiAPIRegion: String, CaseIterable, Sendable {
             URL(string: "https://bigmodel.cn/coding-plan/team/usage-stats")!
         }
     }
+
+    /// BigModel CN pay-as-you-go account balance. The endpoint lives on the
+    /// `www.bigmodel.cn` console host (not the `open.` API host) and accepts both
+    /// `Bearer <key>` and raw-key Authorization (verified 2026-08). z.ai global has
+    /// no documented equivalent.
+    public var balanceURL: URL? {
+        switch self {
+        case .global: nil
+        case .bigmodelCN: URL(string: "https://www.bigmodel.cn/api/biz/account/query-customer-account-report")!
+        }
+    }
 }
 
 public enum ZaiEndpointRouter {
     private static let quotaPath = "api/monitor/usage/quota/limit"
     private static let modelUsagePath = "api/monitor/usage/model-usage"
+    public static let balancePath = "api/biz/account/query-customer-account-report"
 
     public static func resolveQuotaURL(
         region: ZaiAPIRegion,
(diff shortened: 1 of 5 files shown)


RunhuaHuang:
Thanks for the review — all three findings addressed in 72d762a:

- **P1 (timeout):** the balance request now passes an explicit `timeoutSeconds: 5`, well below the fetch deadline, so a stalling balance service can no longer delay the model-usage requests or discard the fetched quota snapshot. Covered by a new assertion on the recorded request's `timeoutInterval`.
- **P2 (region-aware validation):** extracted `validateBalanceEndpointOverride` and wired it into **both** `validateEndpointOverrides` overloads, so the region-aware path used by the provider fetch pipeline now rejects malformed / non-HTTPS `Z_AI_BALANCE_URL` like the other endpoint overrides. New test exercises both overloads with an `http://` override.
- **P2 (null amounts):** numeric conversion now goes through a null/undefined guard, so `availableBalance: null` correctly falls back to `balance` (previously rendered ¥0.00 because `Number(null)` is 0), and null secondary fields no longer produce misleading zero-valued rows. New fixture test covers both.

Verified: `swift test --filter Zai` (31 passed), `ZaiPluginBalanceTests` (6 passed), plugin infra suites (72 passed), `swiftlint --strict` 0 violations, `swiftformat` applied.

PR #3130, opened 2026-08-21
Persist Codex priority-turn scan cursor across relaunches

olddonkey:
## Summary

After every relaunch, CodexBar's first Codex refresh re-scans the whole codex CLI trace database. This PR makes the priority-turn scan cursor durable across launches so a relaunch resumes incrementally instead of re-scanning.

**Root cause (profiled).** `CostUsageScanner.codexPriorityTurns` runs three leading-wildcard `feedback_log_body LIKE '%…%'` predicates over `<codexHome>/logs_2.sqlite`. Leading `%` defeats every index, so it is a full body-column scan — 1.4 GB / 130k rows on the reference machine (release `CodexBarCLI`, `sample(1)`: `pread` + `strcspn`/`patternCompare` under `codexPriorityTurns`). With a cold OS page cache that is ~2 minutes of I/O-bound wall time (~20 s CPU); with a warm page cache it is ~2.5–3 s of CPU. The function already scans incrementally (`rowid > lastRowID`) through a process-global memo, but that memo was never persisted, and app refreshes always inspect priority turns (`bypassScannerDebounce` → `refreshMinIntervalSeconds = 0`), so **every process launch paid the scan again**.

**Measured** (isolated cache root, same real data, same machine, `codexbar cost --provider codex --days 30`; M1/M2 = a fresh process after the 60 s refresh interval, i.e. the relaunch path):

| binary | cold scan | within interval (cached) | relaunch M1 | relaunch M2 |
|---|---|---|---|---|
| `main` (f74117aeb) | 8.73 s / 7.46 s CPU | 0.82 s | **5.86 s / 5.46 s CPU** | **5.37 s / 5.00 s CPU** |
| this PR | 8.61 s / 7.71 s CPU | 0.84 s | **3.33 s / 3.08 s CPU** | **2.85 s / 2.63 s CPU** |

Token and cost totals are identical between the two binaries at every step; the cached fast path is unchanged. The saved ~2.5 s CPU per relaunch is the LIKE scan with a warm page cache; the cold-page-cache case (~2 min I/O wait) is removed for the same reason but was not re-measured (no `purge` without sudo).

## What changes

- Persist the memo cursor (`lastRowID`, sqlite file identity, coverage epoch, accumulated turns) as an optional `turnsCursor` inside the existing `StoredPriorityState` payload in the codex cost cache, and seed the in-memory memo from it before the cold path runs.
- All existing invalidation rules apply unchanged to a seeded cursor: file identity (inode) change, rowid regression, and a requested window that expands earlier than the cursor's coverage each force a full rescan. `forceRescan` drops the memo instead of seeding it.
- The cursor is **excluded from the identical-content save comparison** (its `lastRowID` advances on nearly every refresh because the codex CLI appends trace rows continuously) and is written as a **metadata-only update on the skip path**, so the cheap no-op save stays cheap.
- `StoredPriorityState` decodes the cursor **leniently** (`try? decodeIfPresent`): a malformed cursor yields `nil` and one cold trace scan, never a dropped `turnKeys` (which would trigger cache-wide reprocessing).
- The persisted cursor is validated with a **content anchor** (rowid the accumulation ended on + SHA-256 of that row's `ts\nbody`), captured with one primary-key lookup and checked before resuming; a missing row or digest mismatch forces a full rescan, so a recreated `logs_2.sqlite` that reuses an inode cannot resume from a stale cursor (review feedback). The inode check remains as a cheap pre-check.
- `CodexParserHash` is regenerated (hashed scanner sources changed). The previous hash `2d17f4981b78d07f` is added to `compatiblePredecessorParserHashes` — parsing and the persisted row shape are unchanged — so existing `cost-usage.sqlite` stores are **adopted on upgrade, not rebuilt**; an old priority payload without a cursor loads with `codexPriorityTurnsCursor == nil` and intact `turnKeys`.

Out of scope (deliberately unchanged): the LIKE predicates / query plans / parsers, pricing, the JSONL corpus scan and its 2 s budget, catch-up policy. No new dependencies.

## Tests

New `CostUsageScannerCodexPriorityCursorTests` (9 cases): relaunch reuses the persisted cursor and scans only appended rows (a mutated old row must not reappear); inode change still full-scans (with `maxRowID == lastRowID` so only identity can trigger it); window expanded earlier still full-scans; old payload without a cursor cold-scans; live memo wins over a stale seed; stale-cursor re-accumulation is idempotent (turns, source maps, insertion order, no duplicates); identical-content skip still persists an advanced cursor with zero file writes; malformed cursor still loads turn keys; force rescan drops the cursor and cold-scans; replaced database with a reused inode (anchor mismatch → full rescan, matching anchor → incremental); deleted anchor row → full rescan; payload without anchor fields → cold scan. Plus `CostUsageStoreTests`: predecessor-hash adoption of a cursorless payload without rebuild.

Note: the skip-path test uses the repo's process-global `saveCycleCheckpointForTesting` seam like other store tests; it is safe under the repo runner (`swift test --no-parallel`) and can interfere only in ad-hoc parallel `swift test` runs.

## Verification

- `make check`: parser hash current (`3c984b655688593f`), `0/1974 files require formatting`, swiftlint `0 violations, 0 serious in 1973 files`.
- Full suite `make test`: green on `c529ac35c` (tree `cc21c5df`; exit 0, 1116 s, ~9.2k tests) and again on `51845ab7d` after the content-anchor follow-up (exit 0, 852 s, 9231 tests; all cursor-suite cases incl. the reused-inode / deleted-anchor / missing-anchor-fields cases executed inside the gate) — exit 0, 1116.6 s, ~9.2k tests passed across all shards; all 9 `CostUsageScannerCodexPriorityCursorTests` cases and the new predecessor-adoption case executed inside the gate.
- Baseline `make test` on `main` (f74117aeb): green (1174 s).

## Memory note

The ~400 MB steady RSS some users see is a separate story (GUI framework residency + transient multi-provider scan buffers + malloc fragmentation; the CLI runs this same scan in ~107 MB). This PR targets the CPU burst, not the resident-memory number.

---
Implemented by grok-4.6 (xhigh) via implementation-loop; reviewed hunk by hunk plus an independent deep review; one iterate round (write-amplification and lenient-decode fixes) and two follow-ups (parser hash regeneration, predecessor adoption).


olddonkey:
Addressed the P2 from the Codex review (and ClawSweeper's merge-risk item): the persisted cursor no longer trusts the inode as its only identity.

**What changed (commit on this branch):** both the in-memory memo and the persisted cursor now carry a **content anchor** — the rowid the accumulation ended on plus the SHA-256 of that row's `"<ts>\n<feedback_log_body>"`. It is captured with one primary-key lookup after each successful accumulation (if that lookup fails nothing is memoized/persisted), and validated before resuming: a missing anchor row or a digest mismatch forces a full rescan. The inode check stays as a cheap pre-check. This also closes the same (shorter) window the in-process memo had. A missing anchor row can also be Codex pruning old rows in place; the conservative full rescan is intended there. SHA-256 uses the repo's CryptoKit/swift-crypto pattern so the Linux CLI still builds.

**Tests added:** replaced database with a reused inode (stale anchor → full rescan; matching anchor → incremental control, with a mutated old row that must not reappear), deleted anchor row → full rescan, payload without anchor fields → nil cursor / cold scan; the relaunch test now also asserts the anchor survives and advances. `CodexParserHash` regenerated accordingly (the unreleased intermediate hash is not listed as a predecessor; `2d17f4981b78d07f` from `main` still is).

The "Live Verification FAIL" in the ClawSweeper report is its own environment failing at `pnpm install` before running anything; not related to this change.

PR #3131, opened 2026-08-21
Prune stale fleet snapshot records after account key changes

steipete:
## Summary

PR #3082 changed the claude-swap accountID used by `AccountSnapshotSyncPayload.accountKey` from email-derived to slot-derived, which changes the CloudKit record name (`snap-<provider>-<accountKey>-<deviceID>`). `CloudSyncEngine.pushPendingSnapshots` only ever queued `.saveRecord`, so devices that synced before the upgrade kept the old email-keyed records forever — showing up as stale duplicate fleet accounts in the fleet menu until manually cleaned.

## Changes

- New pure planner `CloudSyncSnapshotPruner.recordNamesToDelete(persisted:desired:)`: groups the desired push by `(provider, deviceID)` read from payload fields (never parsed from record names — accountKey hashes and device IDs contain dashes) and returns persisted record names whose scope participates in the push but whose name is no longer produced. Other devices' records and providers absent from the push cycle are structurally untouchable.
- `pushPendingSnapshots` queues `.deleteRecord` for stale names via the existing `pendingRecordZoneChanges` API, drops them from `lastSnapshotHashes` (so a re-appearing key isn't save-suppressed), and clears envelope bookkeeping (`fleetSnapshots`, `encodedSystemFields`, `recordMetadata`) plus observable `state.fleetSnapshots` so the local menu updates immediately.
- `processEvent(.sentRecordZoneChanges)` previously did zero bookkeeping for `deletedRecordIDs`; it now runs the same idempotent cleanup, and the fetched-deletions path reuses the shared helper.

## Testing

- 5 new Swift Testing cases on the planner seam (no live CloudKit): upgrade scenario with dashed device IDs, device isolation, provider isolation, multi-provider independence, steady-state no-op.
- `make check` clean (0 SwiftLint violations, SwiftFormat 0/1970).
- `swift test --filter SyncModelTests` 15/15, `--filter CloudSync` 18/18, full `make test` 76/76 groups green on first pass.


steipete:
Closing as superseded by #3111, which landed on main first and fixes the same root cause (stale email-keyed fleet snapshot records left behind by the #3082 claude-swap account-key change) with a more conservative design: email-keyed predecessors are only deleted after CloudKit confirms the replacement slot-keyed save, with persisted retry state, account-switch isolation, and shared-predecessor reference counting.

This PR's approach (prune-on-push via a pure `(provider, deviceID)`-scoped planner) deletes eagerly before the replacement save is confirmed, which #3111's history shows is unsafe against transient save failures — a failed replacement would orphan the account from the fleet view entirely.

One capability from this PR that #3111 deliberately does not cover: generic pruning of records for accounts that are *removed* (not migrated) — #3111's cleanup is scoped to Claude Swap email→slot predecessors only. If removed-account records linger as a real problem, that should be built as a follow-up on top of #3111's confirmed-save machinery rather than this branch.

PR #3120, opened 2026-08-21
fix(codex): tokscale parity for token counts - max cached, clamped, reasoning split, stale

Yuxin-Qiao:
## Summary

- rebuild the parity branch on current main; replace the prior corrupted remote history
- derive cached usage as max(cached_input_tokens, cache_read_input_tokens), matching tokscale before its clamp to input
- add field-level stale-regression detection for out-of-order token_count events and run it before watermark latching so stale snapshots cannot enter interleaved mode
- keep reasoning additive metadata; do not subtract it from output because tokscale treats reasoning as a display subset of output, while CodexBar stores both fields alongside output
- compare reasoning in stale detection only when both snapshots provide it; an omitted reasoning field is unknown, not zero (ClawSweeper P2)
- apply the maximum cache-field selection in the JSONSerialization fallback path, not just the fast byte parser (ClawSweeper P2)
- parse bare usage rows from non-event rollout lines (one-shot codex exec / headless output): canonical usage envelopes, OpenAI and completion-style aliases, cached-input subtraction, timestamp fallback to the last accepted row
- regenerate the Codex parser hash after scanner changes
- fix `codex json fallback applies maximum cache selection` fixture: Swift multiline literal had inserted newlines inside the JSON (invalid JSONL) and the huge integer alone no longer forced the fallback; now built via joined string parts with an escaped cache-read key so the fast parser declines and the fallback path is exercised

## Evidence

- Head: `fix/codex-tokscale-parity-p0` @ ec305bc7713a1b661f8a4a3d3ed3af7990de52d6
- `swift test --filter CostUsageScannerTests`: 22/22 passed, including `codex json fallback applies maximum cache selection`
- `make check`: clean (SwiftFormat, SwiftLint strict 0 violations, parser hash current, locale/docs/package gates)
- CI: https://github.com/steipete/CodexBar/actions/runs/32625155825 — success (lint, changes, build-linux-cli x64/arm64, build-linux-musl-cli, swift-test-macos 0/2 and 1/2, lint-build-test)

Reference: tokscale `crates/tokscale-core/src/sessions/codex.rs`

PR #3136, opened 2026-08-22
Price OpenCodex usage once per entry and stop per-entry catalog/overlay reloads

olddonkey:
## Summary

The OpenCodex spend source (`~/.opencodex/usage.jsonl` → `OpenCodexUsageFanOut` → `OpenCodexUsageAggregator.snapshot`) re-resolved its pricing context **per entry** on every refresh:

- `listPriceUSD` called `CostUsagePricing.codexCostUSD` without a pre-resolved models.dev catalog, so each call went through `ModelsDevCache.load` → `FileManager.attributesOfItem` — a `stat` **plus an extended-attribute read** — and without a pre-resolved custom-pricing overlay, so each call also re-resolved the overlay file location.
- Every windowed entry was priced **three times** (day, session and hour accumulators), and day keys / hour buckets went through `Calendar` per entry.

On a 35k-entry log (all inside the 30-day window) that is ~100k stat+xattr syscalls and ~70k Calendar interval computations per refresh. In the running app this was the 25–35 s CPU spike on every adaptive refresh (sampled on 0.54.0: `snapshotsBySubscription` → `attributesOfItem` → `getxattr`/`listxattr`, plus Calendar/ICU and String hashing).

### Changes (snapshot output is byte-identical)

- **Resolve the models.dev catalog and the custom-pricing overlay once** per fan-out / snapshot and pass them down; **price each windowed entry once** and reuse the value for the day/session/hour/model merges. A missing catalog is substituted with an empty catalog so the degraded path never falls back to per-call loads. (`OpenCodexUsageAggregator`, `OpenCodexUsageFanOut`)
- **Memoize the local-day key and hour-bucket start per calendar interval**, using the calendar's own `[start, end)` intervals (DST-correct; no 86400/3600 arithmetic). Day keys still come from `CostUsageLocalDay.key`, which derives y-m-d from the same Gregorian-in-timezone calendar whose day interval the memo caches, so the memo is exact.
- **`ModelsDevCache.load` reads (mtime, size) via POSIX `stat`** instead of `attributesOfItem` (which also reads xattrs); memo/invalidation semantics unchanged. This helps every caller repo-wide. (`ModelsDevPricing.swift`)
- `CodexParserHash` is regenerated because `ModelsDevPricing.swift` is in the hashed set; the previous hash (`3c984b655688593f`) is added to `compatiblePredecessorParserHashes` since parsing and the persisted row shape are unchanged, so existing `cost-usage.sqlite` stores are adopted on upgrade instead of rebuilt (same pattern as #3130).

Not in this PR (follow-up): `OpenCodexUsageStore` keys its sqlite cache on `path|size|mtime`, so the append-only log is fully re-parsed (and DELETE+re-inserted) whenever it grew. That is where the remaining transient memory on this path lives; making the store incremental is a separate change.

### Measured

Release `CodexBarCLI`, isolated cache root, same machine, real 41.7 MB / ~35k-entry `usage.jsonl` (all inside the window), `cost --provider codex --format json --days 30`.

| Scenario (OpenCodex path only — empty codex home, identical frozen inputs) | main `27c7f334e` | this PR |
|---|---|---|
| cold (no store cache) | 14.3 s real · 9.1 s user · 4.9 s sys · 193 G instr · 440 MB peak | **2.6 s · 2.4 s · 0.1 s · 40 G · 412 MB** |
| warm (store cache hit) | 13.8 s · 8.3 s · 5.3 s · 166 G · 144 MB | **1.2 s · 1.1 s · 0.04 s · 13 G · 115 MB** |

CLI JSON output is identical in both scenarios (diffed with only `updatedAt` stripped). Full `cost --provider codex` run on live data in the app's steady-state case (log grew since the last refresh): ~11 s → ~3.5 s real (7.3–9.2 s → 3.2 s user); cold 26 s → 14 s. Peak footprint is unchanged — that is the re-parse path noted above.

### Tests

- `OpenCodexUsageFanOutTests`: equivalence of the new `snapshot` against an independent reference implementation (mixed providers, reported/estimated/unreported/unsupported entries, custom-pricing overlay, duplicate request IDs, entries outside the window, DST transitions in America/Los_Angeles, pricing through a fixture models.dev catalog so the catalog path is actually exercised); metadata-read counting proving one catalog load per snapshot for 60 entries and zero with an injected catalog; day/hour memo boundary cases (spring-forward gap, fall-back repeated hour, exact hour/midnight boundaries, and America/Santiago midnight DST transitions) compared against `CostUsageLocalDay.key` / `calendar.dateInterval(of: .hour)` computed per entry.
- `ModelsDevPricingTests`: memo still invalidates on cache-file size change and on mtime change after the `stat` switch; exactly one metadata read per `ModelsDevCache.load` (task-local recorder).
- `CostUsageStoreTests`: predecessor-hash set includes `3c984b655688593f`.

### Process

Implemented by grok-4.6 (xhigh) via implementation-loop; reviewed hunk by hunk plus an independent deep review (no correctness defect; its findings — overlay-once, empty-catalog substitution, task-local recorder, deterministic catalog-path test, Santiago DST cases — were applied in one iterate round). Full `make check` and `make test` run locally on the final tree.


olddonkey:
Re the ClawSweeper merge-risk item (parser-hash adoption of existing cost caches):

- The adoption coverage is part of this PR: `CostUsageStoreTests` `compatible predecessor parser hash adopts without rebuilding` asserts the full `compatiblePredecessorParserHashes` set (now including `3c984b655688593f`) and exercises adoption of a predecessor store without a rebuild (`rebuildCount == 0`, `userVersion` upgraded, `parser_hash` meta rewritten to the current value). The mechanism is the same as #3130's predecessor entry; nothing in this PR changes parsing, pricing math, or the persisted row shape — only how often the models.dev catalog / custom-pricing overlay are resolved and how day/hour keys are computed.
- `Scripts/regenerate-codex-parser-hash.sh --check` is current (`fc4cae716e9066c5`) and the full local `make check` + `make test` (77/77 sharded groups) ran on the submitted head, in addition to CI.
- The "Live Verification" failure in the bot run is its own sandbox (`corepack` trying to download pnpm before `swift run`), not this change — `CodexBarCLI cost --provider codex` is exactly what the before/after numbers in the description were produced with.

Also measured in the running app (dev build as my menu-bar app, real data): the OpenCodex merge that used to dominate every adaptive-refresh spike (~10 s × up to 3 merges) is now ~2 s of a spike; peak memory on that path is unchanged, which is the follow-up noted in the description (incremental `OpenCodexUsageStore`).

olddonkey:
Pushed a comments-only commit (653057d47): doc comments on the once-per-snapshot pricing context (why a missing catalog becomes an empty one, the two overlay levels in `listPriceUSD`, the day-key memo invariant, the task-local test recorder). No code change; `CodexParserHash` is regenerated to `5f8507161b23757c` because `ModelsDevPricing.swift` is in the hashed set — the predecessor list is unchanged (`3c984b655688593f` is still the only shipped hash being adopted). `make check` and the focused store/OpenCodex/ModelsDev/gatekeeper tests (187) pass on this head.

steipete:
#3141 and #3150 are merged — thanks. This PR (and #3140/#3135 stacked above it) now conflicts with the landed parser-hash/scanner changes from #3150 and #3120. Please rebase onto current main: regenerate the parser hash on the merged tree (`./Scripts/regenerate-codex-parser-hash.sh`) and re-check whether your predecessor-hash entry intent still holds against main's current `compatiblePredecessorParserHashes` list. I'll merge each as it greens — #3136 first, then #3140, then #3135.

PR #1931, opened 2026-07-06
Guide Gemini users to Antigravity on consumer-tier shutdown

Yuxin-Qiao:
## Summary
- Detect Google's consumer-tier shutdown signals (`UNSUPPORTED_CLIENT`, `IneligibleTierError`, and explicit Gemini Code Assist / Antigravity migration responses) in Gemini quota, `loadCodeAssist`, and token-refresh paths.
- Surface a typed `consumerTierDeprecated` error immediately, clearing stale quota so the migration signal cannot be hidden by cached usage.
- Show an explicit Settings action to enable and refresh Antigravity only after CodexBar observes that typed signal. Ordinary login, `401 UNAUTHENTICATED`, and Antigravity-not-running errors remain unchanged; there is no automatic enablement or fallback.
- Keep migration state scoped to each `UsageStore`, clear it after a successful Gemini refresh or provider reset, and document the June 2026 provider boundary.

Fixes #1929

## Proof
- [x] 33 focused tests across `GeminiConsumerTierMigration`, `GeminiProviderMigrationSettings`, `GeminiLoginAlert`, and `GeminiStatusProbeAPI`
- [x] `make check`
- [x] `make test` — all 48 shards passed
- [x] Autoreview — no accepted/actionable findings

## Live verification boundary
No real OAuth or Keychain flow was run. Live verification would require a consumer Google account currently receiving Google's shutdown response; stubbed API/status-probe tests cover all three observable response paths. Terminal-only Gemini failures cannot activate the Settings action because CodexBar does not capture that process output.

PR #3139, opened 2026-08-22
Detect Google's live Gemini consumer-tier shutdown response

betive37:
## Summary

On a Google AI Pro (consumer) account the Gemini provider reports a bare **`Gemini API error: HTTP 403`** instead of the Antigravity migration guidance, and the login action deletes `~/.gemini/oauth_creds.json` to launch a `gemini` sign-in that Google rejects with *"This client is no longer supported for Gemini Code Assist for individuals. To continue using Gemini, please migrate to the Antigravity suite of products"* — so the user loses their stored credentials and gets no path forward.

Root cause: Google's live shutdown response does not have the shape `GeminiStatusProbe` was matching against. Captured today against a real consumer token:

**`POST cloudcode-pa.googleapis.com/v1internal:loadCodeAssist` → HTTP 200**
```json
{
  "allowedTiers": [{ "id": "standard-tier", "name": "Gemini Code Assist", "isDefault": true }],
  "ineligibleTiers": [{
    "reasonCode": "UNSUPPORTED_CLIENT",
    "reasonMessage": "This client is no longer supported for Gemini Code Assist for individuals. To continue using Gemini, please migrate to the Antigravity suite of products: https://antigravity.google",
    "tierId": "free-tier",
    "tierName": "Gemini Code Assist for individuals"
  }]
}
```
No `currentTier`. `loadCodeAssistStatus` only ran `throwIfConsumerTierDeprecated` on **non-200** bodies, so the signal was ignored and the call returned `.empty`.

**`POST …/v1internal:retrieveUserQuota` → HTTP 403**
```json
{"error": {"code": 403, "status": "PERMISSION_DENIED",
  "message": "You do not have a valid license of this product… (#3501)",
  "details": [{"@type": "…ErrorInfo", "reason": "SUBSCRIPTION_REQUIRED", "domain": "cloudaicompanion.googleapis.com"}]}}
```
No migration wording at all, so text matching cannot classify it — it fell through to `apiError("HTTP 403")`.

Net effect: `geminiObservedConsumerTierDeprecation` never became true, so the **Enable Antigravity provider** settings action added in #1931 never appeared for the users it was written for.

## Changes

**`GeminiStatusProbe`**
- `loadCodeAssistStatus` reads `ineligibleTiers[].reasonCode` / `reasonMessage` on the 200 body through the existing `isConsumerTierDeprecationSignal` — no new text heuristics. No `currentTier` + unsupported-client → `.consumerTierDeprecated`, thrown before the quota call.
- Two signals suppress the flag, because the `loadCodeAssist` branch runs *before* `resolveAccountPlan` and would otherwise pre-empt its mappings: a named `paidTier.name` (authoritative even without `currentTier`, pinned by `paid tier name survives missing current tier`) and an `hd` claim (Workspace/education, which the resolver reads as Workspace when paired with `free-tier`, and which the June 2026 shutdown does not cover). Google's captured consumer response carries neither. The claim is already extracted before `loadCodeAssistStatus` runs, so it only needed threading through.
- The quota 403 maps to `.consumerTierDeprecated` only when the same fetch saw that flag **and** the account is not on `standard-tier`. Standard/Enterprise subscriptions are outside the June 2026 shutdown, so a licensed account's unrelated 403 still surfaces as `HTTP 403` even if Google lists the consumer tier as ineligible for every CLI caller.
- `parsePaidTierName`, the new `isConsumerClientUnsupported(in:paidTierName:hostedDomain:)`, and the `loadCodeAssist` response parsing sit in a file-scoped extension to stay under `type_body_length`.

**`UsageStore`** — one stored `geminiMigrationObservation` (`none` / `localAntigravityHandoff` / `googleConsumerTierShutdown`) replaces the single Bool; a later local-tooling failure never downgrades a shutdown already seen. `geminiObservedConsumerTierDeprecation` becomes a derived accessor, so the settings action behaves exactly as before.

**Login flow** — when Google's shutdown was observed, `GeminiLoginRunner.run` returns a new `.consumerTierDeprecated` outcome *before* `clearCredentials()` and the Terminal launch, and the alert shows the Antigravity guidance. It **warns rather than blocks**: the alert offers **Switch Account…** beside Cancel (Cancel is the default button, since confirming clears stored credentials), and confirming re-runs the ordinary login without the guard — that is how a user moves off a shut-down consumer account onto a Workspace, education, or Code Assist Standard/Enterprise one. Confirming does not clear the observation, so the settings action stays put and the next attempt warns again. The narrower flag matters here: the local `oauthCredentialsUnavailableWithAntigravity` handoff deliberately does **not** guard login, because reinstalling or relaunching Gemini CLI is the fix for that case and Workspace accounts must keep that path. Ordinary login is unchanged (default parameter, no other call sites).

Gatekeeper anchors are untouched — `UsageStore.swift` keeps its line count and the two Gemini anchors (`GeminiStatusProbe.swift:146`, `GeminiLoginRunner.swift:7`) are byte-identical.

## Tests

Fixtures mirror the captured bodies above: `loadCodeAssistUnsupportedClientResponse(currentTierId:paidTierName:)` and `quotaSubscriptionRequiredResponse()`.

- `GeminiConsumerTierMigrationTests` (+9): 200 + `ineligibleTiers` → deprecated · consumer tier + flag + quota 403 → deprecated · `standard-tier` + flag + quota 200 → normal snapshot · `standard-tier` + flag + quota 403 → `HTTP 403` · no flag + quota 403 → `HTTP 403` · `paidTier.name` without `currentTier` + entry + quota 200 → plan stays `Plus` · same shape + quota 403 → `HTTP 403` · `hd` claim without `currentTier` + entry + quota 200 → normal snapshot · Workspace (`free-tier` + `hd`) + entry + quota 403 → `HTTP 403`.
- `GeminiProviderMigrationSettingsTests` (+4): shutdown arms the login guard · local handoff does not · local handoff after a shutdown does not disarm it · clearing disarms.
- `GeminiLoginAlertTests` (+3) and new `GeminiLoginRunnerTests`: the deprecated outcome maps to the Antigravity alert, carries the `Switch Account…` recovery and its explanation, ordinary login failures carry no recovery button, and the runner never arms the credentials watcher (i.e. never reaches the deletion path).

## Commands run

- `swift test --filter "Gemini|ProviderArchitectureGatekeeperTests"` — 143 tests, all passing
- `make check` — 0 violations
- `make test` — full sharded suite, 8276 tests. The only failure is `StatusMenuSwitcherRefreshTests/"merged provider switch updates live tab rows in place"`, which fails identically on a clean tree at `27c7f334e` with these changes stashed (verified) — a pre-existing headless-AppKit menu-item-identity flake, unrelated to this change.

## Real behavior proof

Reproduced end to end on the reporting account (Google AI Pro, personal Google account). Same machine, same `~/.gemini/oauth_creds.json`, back-to-back runs — only the binary differs.

**Before — released 0.54.0 CLI**
```console
$ codexbar usage --provider gemini --format json
[{"provider":"gemini","source":"auto",
  "error":{"kind":"provider","code":1,"message":"Gemini API error: HTTP 403"}}]
```

**After — this branch**
```console
$ .build/arm64-apple-macosx/debug/CodexBarCLI usage --provider gemini --format json
[{"provider":"gemini","source":"auto",
  "error":{"kind":"provider","code":3,
  "message":"Google no longer supports Gemini CLI OAuth for individual, AI Pro, or Ultra accounts. Enable CodexBar's Antigravity provider, sign in to Antigravity or run `agy`, then refresh."}}]
```

The error kind changes from the generic `apiError` (code 1) to `consumerTierDeprecated` (code 3), which is the sentinel that activates the **Enable Antigravity provider** settings action and now also guards the login action.

**And the path the message points at works on the same machine**
```console
$ .build/arm64-apple-macosx/debug/CodexBarCLI usage --provider antigravity --format json
{"provider":"antigravity","source":"cli","loginMethod":"Google AI Pro",
 "primary":{"windowMinutes":10080,"usedPercent":3.79,"resetsAt":"2026-08-29T04:35:08Z",
   "resetDescription":"You have used some of your weekly limit, it will fully refresh in 6 days, 18 hours."},
 "secondary":{"windowMinutes":300,"usedPercent":0,"resetsAt":"2026-08-22T15:26:36Z"}}
```

The underlying HTTP exchange behind the "before" run, captured with `curl` on the same token, is the `loadCodeAssist` 200 / `retrieveUserQuota` 403 pair quoted at the top of this description.


betive37:
Both asks are addressed in `8f2d5d4`.

**[P2] Honor paidTier before declaring a consumer shutdown** — confirmed and fixed. `resolveAccountPlan` treats `paidTier.name` as authoritative even without `currentTier`, pinned by the existing `paid tier name survives missing current tier`, and my branch threw before that resolver ran.

The gate is folded into the flag itself rather than added at the throw site, because the quota-403 mapping had the same hole — a paid response with no `currentTier` would have fallen through the `guard let tierId` branch and then been reclassified at the 403 instead:

```swift
let isConsumerClientUnsupported = Self.isConsumerClientUnsupported(in: json, paidTierName: paidTierName)
// paidTierName == nil && hasUnsupportedClientIneligibleTier(in: json)
```

Two regression tests cover the combined response shape (`paidTier.name` + no `currentTier` + the `UNSUPPORTED_CLIENT` entry): with quota 200 the plan stays `Plus`, and with quota 403 the error stays `HTTP 403`. Both failed before the fix — the 403 one returned `.consumerTierDeprecated`, which is the second hole. Google's captured shutdown response carries no `paidTier`, so the consumer path is unchanged. `CHANGELOG.md` untouched in this commit.

**Real behavior proof** — added to the PR description. Same machine, same `~/.gemini/oauth_creds.json`, back-to-back runs, only the binary differs:

```console
$ codexbar usage --provider gemini --format json          # released 0.54.0
[{"provider":"gemini","source":"auto",
  "error":{"kind":"provider","code":1,"message":"Gemini API error: HTTP 403"}}]

$ .build/arm64-apple-macosx/debug/CodexBarCLI usage --provider gemini --format json   # this branch
[{"provider":"gemini","source":"auto",
  "error":{"kind":"provider","code":3,
  "message":"Google no longer supports Gemini CLI OAuth for individual, AI Pro, or Ultra accounts. Enable CodexBar's Antigravity provider, sign in to Antigravity or run `agy`, then refresh."}}]
```

Error kind goes from generic `apiError` (code 1) to `consumerTierDeprecated` (code 3) — the sentinel that activates the settings action and now guards login. The account is a personal Google AI Pro one; `--provider antigravity` returns a full snapshot on the same machine, so the guidance points somewhere that works.

`swift test --filter "Gemini|ProviderArchitectureGatekeeperTests"` 139 passing · `make check` 0 violations · `make test` 8272 tests, only the pre-existing `StatusMenuSwitcherRefreshTests` AppKit flake (verified to fail identically on a clean `27c7f334e`).

@clawsweeper re-review

betive37:
Fixed in `59858e2`.

**[P1] Preserve the Workspace and education login path** — the concern is real, but the exposure is at the quota-403 site rather than the branch cited.

`GeminiStatusProbe.swift:528-530` is inside `guard let tierId else`, so it only runs when `currentTier` is **absent**; a `free-tier` response never reaches it. The actual hole was the 403 mapping, whose gate was `caStatus.tier != .standard` — and Workspace is `free-tier` + `hd`, so it passed straight through. Both shapes did fail, and both now have regressions:

```
keeps a workspace account without current tier out of the shutdown path  → was .consumerTierDeprecated
keeps http 403 for a workspace account on the free tier                  → was .consumerTierDeprecated
```

Rather than gate each site, the hosted domain is folded into the flag next to the paid-tier signal, so every consumer of it inherits the exemption:

```swift
paidTierName == nil && hostedDomain == nil && self.hasUnsupportedClientIneligibleTier(in: json)
```

`claims.hostedDomain` is already extracted before `loadCodeAssistStatus` is called, so it only needed threading through the signature. The `loadCodeAssist` response parsing moved into the file's parsing extension to stay under `type_body_length`; no behavior change there. `docs/gemini.md` records why both signals outrank the ineligible-tier listing. `CHANGELOG.md` untouched since `c48fff7`.

`swift test --filter "Gemini|ProviderArchitectureGatekeeperTests"` 141 passing · `make check` 0 violations · `make test` 8274 tests, only the pre-existing `StatusMenuSwitcherRefreshTests` AppKit flake (verified to fail identically on a clean `27c7f334e`). The live consumer-account run still returns `consumerTierDeprecated` (code 3) after the refactor.

One note on the automated live check: `swift run CodexBarCLI usage --help` failed before reaching `swift` — `sh -lc pnpm install --frozen-lockfile` could not download pnpm through corepack in the sandbox. That is environmental, not a regression from this branch.

@clawsweeper re-review

betive37:
Fixed in `a0f3626`.

**[P1] Restore a supported account-switch recovery path** — confirmed. The guard returned before the credential-clearing switch, and the observation only clears on a successful Gemini refresh or a provider reset, neither reachable from the blocked action. So a user moving off an affected consumer account was stuck for the session, and the alert did not say so.

On the **decision needed**, I implemented the recommended option — keep the safe default, add a warned path — using the repo's existing two-button alert pattern:

- The alert now offers **Switch Account…** beside **Cancel**, with **Cancel as the default button**, deliberately inverting the `Open Browser`/`Cancel` ordering used elsewhere: confirming deletes `~/.gemini/oauth_creds.json`, so a stray Return must not destroy credentials.
- Confirming re-runs the ordinary login with the guard off — the same flow as before this PR.
- Confirming does **not** clear `geminiMigrationObservation`. The settings action stays available, and the next attempt warns again; nothing depends on hidden session state.
- The message explains the one case where continuing helps: `"Signing in again only helps if you are switching to a Workspace, education, or Code Assist Standard/Enterprise account. Continuing clears the stored Gemini credentials and opens Gemini CLI."`

If you would rather keep the action fully blocked, that is dropping `confirmButtonTitle` from the `.consumerTierDeprecated` case in `geminiLoginAlertInfo` — the presenter falls back to the single-button alert on its own. Happy to switch it either way.

Covered at the model seam rather than through live AppKit, per AGENTS.md: the deprecated outcome carries the recovery title and its explanation, and ordinary login failures carry none.

`swift test --filter "Gemini|ProviderArchitectureGatekeeperTests"` 143 passing · `make check` 0 violations · `make test` 8276 tests, only the pre-existing `StatusMenuSwitcherRefreshTests` AppKit flake (verified to fail identically on a clean `27c7f334e`).

The automated live check failed again before reaching `swift` — `sh -lc pnpm install --frozen-lockfile` could not download pnpm through corepack in the sandbox. Environmental, not from this branch.

@clawsweeper re-review

PR #3105, opened 2026-08-20
perf(spend): parallelize loads and memoize model build

Yuxin-Qiao:
Consolidates previous #3099 + remaining perf work, now one reviewable PR.

**Parallelize**
* `SpendDashboardController.swift:237` provider baselines: `for await` → `withTaskGroup` (was 400ms-6s additive for Claude/Cursor/OpenCodex 365d).
* `SpendDashboardController.swift:433` Codex multi-account: sequential `for account` → `withThrowingTaskGroup` (was 2s×N).

**Memoize model**
* `SpendDashboardModel.swift:300` conversionCache per `(source→target)` (was `CurrencyExchange.convert` per input, 8× per build).
* Hoist `bounds` once per `build` and pass to `buildCurrencyGroup` (was recomputed per currency group).
* `SpendDashboardModel.swift:1069` reuse static `utcCalendar` for Mistral/OpenRouter bucket (was `gregorianCalendar(timeZone: GMT)` per `day()` 2920×).

**Debounce & throttle (74ab7d7d)**
* `UsageStore+SpendDashboardPublication.swift` 250ms debounce for `withObservationTracking` and token publication bursts (was scan storm on every provider publish).
* `SpendDashboardController.swift:1450` 30s throttle for `refreshDateWindow` same-day revisits (was `build` + `publish` on every `didBecomeActive`).
* `SpendDashboardController.swift:1110` display-only fast path for `hiddenSourceIDs / preferredCurrencyCode / hideNative` (was full Codex scan for filter changes).

**Evidence**
* `SpendDashboardController.swift:237,433` parallel
* `SpendDashboardModel.swift:300,1069` memo
* `swiftformat` + `swiftlint --strict` clean, `swift build --target CodexBar` ok
* Before: `全部` 3-10s empty → 2-3s, model build 40ms → 8ms
* After: wall-time = slowest provider

**Real behavior proof (74ab7d7d)**
```
$ swift test --filter SpendDashboardControllerTests
✔ Suite SpendDashboardControllerTests passed after 3.054 seconds. (24 tests)
$ swift test --filter SpendDashboardClockRolloverTests
✔ Suite SpendDashboardClockRolloverTests passed after 0.011 seconds. (4 tests)
$ swift test --filter SpendDashboardSourceConcurrencyTests
✔ Suite SpendDashboardSourceConcurrencyTests passed after 0.129 seconds. (11 tests)
$ swift build
Build complete! (11.48s)
$ swiftformat --lint
0/3 files require formatting
[...]

PR #3140, opened 2026-08-22
Parse the OpenCodex usage log incrementally instead of re-reading it every refresh

olddonkey:
> Builds on #3136 (same OpenCodex path). Until that merges, this PR's diff includes its commits; the new work is the last commit.

## Summary

`~/.opencodex/usage.jsonl` is append-only, but `OpenCodexUsageStore` keyed its sqlite cache on `path|size|mtime` — so the identity changed on every append and the cache never hit in the steady state. Every refresh re-read the whole log (42 MB / ~35k entries on my machine), rebuilt it as one 42 MB `String`, ran a `JSONSerialization` per line, then did `DELETE FROM entries` plus ~35k inserts. Even a cache *hit* re-decoded a stored JSON payload for every row.

#3136 removed the per-entry pricing work on this path; this removes the per-refresh re-read.

### Changes

- **Parse cursor.** The store persists the log path, the file identity (`st_dev`/`st_ino`, not mtime), the byte offset just past the last consumed record, and a SHA256 of the first 64 KiB. On load it either serves the cached rows unchanged (`size == parsedOffset`), or reads only `[parsedOffset, size)` and merges those rows with `INSERT OR REPLACE` — which keeps the existing "last occurrence in file order wins" dedupe. A changed file identity, a shrunken file, or a prefix-digest mismatch falls back to the full re-parse, so rotation, truncation and in-place rewrites are still handled.
- **Trailing records.** A complete record with no terminating newline is returned (a full parse of the same bytes returns it too) but is *not* committed and does not advance the cursor, so a later append that glues bytes onto it cannot desync the cache from a full parse.
- **Typed columns (schema v2).** Token fields are nullable INTEGER columns instead of a JSON payload, so reading the cache no longer decodes JSON per row. A schema bump rebuilds cleanly; no code path ever reads a v1 table.
- **Byte-sliced parsing.** The parser slices lines out of the file's bytes (`mappedIfSafe` + `memchr`) instead of materializing the log as a `String`. `parseLines` now uses the same splitter, so there is one line-splitting rule instead of two (`\n` only; a lone `\r` or a form feed is no longer a record separator, and a raw U+2028/U+0085 inside a JSON string no longer destroys the record).

Ordering is still produced by Swift comparisons, not SQLite collation, and the returned array is byte-identical to a full re-parse.

### Measured

Release `CodexBarCLI`, isolated cache root, real 42 MB / 35k-entry log, `cost --provider codex --format json --days 30`. The OpenCodex path is isolated by pointing the run at an empty Codex home and giving both binaries an identical frozen copy of the log; CLI JSON output is identical apart from `updatedAt`.

| Scenario | parent (#3136) | this PR |
|---|---|---|
| cold rebuild, no cache (runs once per upgrade) | 3.71 s · 452 MB peak | **2.68 s · 290 MB** |
| warm cache hit | 1.25 s · 120 MB | **0.58 s · 82 MB** |
| after the log grew (the app's per-refresh case) | 4.05 s · 480 MB | **1.95 s · 118 MB** |
| cache database | 18 MB | **5.1 MB** |

The cold figure is higher than an earlier revision of this PR (187 MB) because the full parse no longer memory-maps the log — see the SIGBUS note under "Side-effect review" below. That path runs on a cold cache or a schema rebuild; the per-refresh path reads only the appended tail and is unaffected.

For scale: before #3136 the same cold case was 19–21 s / ~450 MB.

### Tests

`OpenCodexUsageStoreIncrementalTests` (new): incremental result equals a full re-parse for append, partial trailing line, a newline-less record later glued to more bytes, duplicate request IDs, truncation, rotation with a matching 64 KiB prefix and in-place rewrite; tail-only reads verified with a byte/line recorder (zero bytes read when nothing changed); schema-v2 round trip for a missing `usage`, a single zero-valued field and mismatched totals; a v1 database rebuild; and a busy-writer case proving a failed `BEGIN IMMEDIATE` leaves both the rows and the cursor untouched. `OpenCodexUsageParserTests` gained cases for the separators whose handling changed.

Known limitation, deliberately accepted and commented in the code: the prefix digest covers only the first 64 KiB, so an in-place rewrite *past* 64 KiB that preserves both size and inode is not detected. The log is append-only; rotation changes the inode and truncation trips the size check.

### Side-effect review

An independent review focused purely on runtime side effects (hangs, crashes, data loss) found three things worth fixing, all now addressed:

- **Two OpenCodex homes could be mixed in one cache.** Every home shares one cache database while the log path is per-home. This PR had replaced the per-read identity check with a cursor check made from a *separate* connection, so a loader could read its own cursor, have another home's CLI replace the cache, and then use that home's rows as its baseline — the write-lock check only rejected same-file older offsets, so it kept those rows, reset the cursor and appended its own tail. The cursor and the rows are now read in **one transaction**, the write carries the exact base cursor the parse was derived from, and an incremental write is rejected whenever the durable cursor differs from that base **in any field**.
- **A crash class.** The full parse mapped the whole log and then walked it; a truncation between mapping and touching those pages raises SIGBUS, which Swift cannot catch. The parser now opens one descriptor, snapshots its size with `fstat` and reads exactly the range it intends to parse, treating a short read as "changed under us". That also bounds a read that would otherwise chase a growing file, and gives the full-reload path the same post-read identity check the incremental path had. The cost is the cold-path memory noted above.
- **Transient I/O failures were published as "no spend".** Any `stat` failure returned an empty array, which the dashboard turns into a confirmed-empty state that removes the OpenCodex source. Only `ENOENT`/`ENOTDIR` now mean "absent"; every other errno throws so the source is marked unavailable instead.

Schema v2 also moved to its own database filename, so downgrading to an older build leaves it a usable v1 cache rather than one it can read but never write.

The same review confirmed what it could not break: the retry state machine is bounded at two iterations, there is no cyclic SQLite lock acquisition, the cursor cannot point past EOF, and no file descriptor, statement or connection leaks on any error path.

**Disclosed, not fixed here** (pre-existing, and outside this PR's files): the forced-refresh reconciliation tail calls the synchronous OpenCodex merge from a `@MainActor` method (`SpendDashboardController.merge(outcome:capture:)` → `SpendDashboardSource+OpenCodex.swift`), so that read happens on the main thread. It is reached only from `SpendDashboardController.refresh()`, i.e. the refresh button in the spend dashboard preferences pane — not the menu bar and not the periodic refresh. This PR makes that path roughly an order of magnitude cheaper (a cache hit or a tail read instead of a full 42 MB re-parse), but the synchronous call remains. Happy to move it off the main actor in a separate PR if you want it.

### Process

Implemented by grok-4.6 (xhigh) via implementation-loop; reviewed hunk by hunk plus an independent deep review. That review found a merge blocker — `writeEntries` ignored the result of `BEGIN IMMEDIATE`, so under a concurrent writer (the CLI opens the same database as the app) the `DELETE` and the cursor write would land in autocommit and the rollback would be a no-op, leaving an empty cache with a valid cursor — plus seven smaller findings; all were fixed in one iterate round. Full `make check` and `make test` (77/77 sharded groups) pass locally on this head.


olddonkey:
Both items are addressed on the pushed head (`7b57ab10b`).

## P2 — stale cursor commits after acquiring the write lock

The race is real and I reproduced the mechanism as described: the write decided what to persist from a cursor snapshot taken *before* the lock, so a delayed loader could commit an older `parsedOffset` over a newer one, re-insert rows the other loader had already stored, and — after a later truncation — leave the `size == parsedOffset` cache-hit path serving rows for content no longer in the log.

`writeEntries` now re-reads the durable cursor **after `BEGIN IMMEDIATE` succeeds and before any mutation**, on the same write connection:

- For an incremental append (`replaceAll == false`), a durable cursor with the same `path` and `fileIdentity` whose `parsedOffset` is `>=` the proposed one means this work is stale → `ROLLBACK`, no cursor write, no row inserts.
- A full reload (`replaceAll == true`) still replaces even a newer cursor, because it re-derived the whole file — that is what truncation, rotation and a schema rebuild need. This is stated in a comment at the check.
- A missing cursor, a different file identity, or an older durable offset writes exactly as before.

`loadEntries` re-runs its cursor path **once** against the freshly committed durable state and falls back to a full reload if that write is stale too, so a caller never returns a partial result and never loops.

Regression test `stale incremental write does not regress a newer durable cursor` is ordered rather than racy — no threads, no sleeps: load once to capture a cursor, append and load again to commit a newer one, then drive the write path with the first snapshot through a test-only seam, and assert the newer cursor survives, `SELECT COUNT(*)` matches the deduped parse, and a subsequent `loadEntries` still equals a full re-parse. The existing busy-writer test (`failed begin immediate leaves rows and cursor untouched`) is unchanged.

## P3 — release-owned changelog

Removed. `CHANGELOG.md` on this branch is now byte-identical to `main`. For whoever writes the release notes:

> - OpenCodex: parse `usage.jsonl` incrementally. The cache keyed on `path|size|mtime`, so every refresh re-read, re-parsed and re-inserted the whole append-only log (~42 MB / 35k entries on a heavy machine); the store now keeps a parse cursor (file identity + byte offset + prefix digest) and reads only the appended tail, and token fields are typed columns instead of a JSON payload decoded per row — a refresh after the log grew drops from ~4.6 s / 422 MB to ~2.1 s / 110 MB with identical output, and the cache database shrinks from 18 MB to 5 MB (#3140).

Note that the stacked predecessor #3136 still carries its own changelog commit; say the word and I will drop it there too rather than force-push a PR that is otherwise sitting quietly.

## Checks

`make check` and the full `make test` (77/77 sharded groups, zero failures, zero timeouts) pass on `7b57ab10b`. The CLI A/B in the description was re-run after the fix and is unchanged: OpenCodex path cold 2.74 s / 187 MB, warm 0.52 s / 75 MB, with byte-identical JSON output against both `main` and the parent branch.

olddonkey:
Addressed on `79baa7694`.

## P2 — revalidate the file after tail parsing

Correct, and thank you for pushing on it: the identity check ran *before* the tail read, so a rotation in that window made the parse start at the old `parsedOffset` inside the replacement file, and those bytes were then merged with cached rows from the old file and persisted under the old identity.

`incrementalReload` now re-stats the log immediately after `parseLog` returns and before anything is inserted or returned. It requires the same `path` and `st_dev`/`st_ino`, and re-runs the cursor validation against the **post-read** stat, so the size check and the prefix digest are re-evaluated too. On mismatch it discards the parsed tail, writes neither rows nor cursor, and runs a single `fullReload` against the file as it now exists — the same bounded fallback the stale-cursor path already uses, no new loop. If the log disappeared entirely, the load returns empty, matching `loadEntries`'s existing behaviour for a missing file.

Still undetectable, and commented at the check: a replacement that preserves path, device, inode, size **and** the first 64 KiB. That is the pre-existing digest threat model documented at `canReuseCursor`.

**Deterministic coverage** — no sleeps, no real threads. A task-local post-parse hook (unset and free in production; one test asserts it is unset by default) lets a test replace the log while the parse result is in hand:

- replacement **longer** than the old `parsedOffset`, new inode
- replacement **shorter** than the old `parsedOffset`, new inode

Both assert that the returned entries equal a full re-parse of the replacement, that no request ID from the old file survives, that the persisted cursor describes the replacement, that no duplicate rows exist, and that the next load is a cache hit reading zero bytes.

## Merge order with #3136

Noted — this branch is stacked on #3136 deliberately (both touch the same OpenCodex path and #3140 builds on the price-once work). Whenever #3136 lands I will rebase this onto `main`; if you would rather review them as one change or in the other order, say so and I will restructure.

## Checks

`make check` and the full `make test` (77/77 sharded groups, zero failures, zero timeouts) pass on `79baa7694`. The CLI A/B in the description was re-measured after both concurrency fixes and is unchanged — OpenCodex path cold 2.74 s / 187 MB, warm 0.52 s / 75 MB, byte-identical JSON output against `main` and against the parent branch.

olddonkey:
Pushed `e331eab0d`. In addition to the review items above, I had this branch audited by a second reviewer looking **only** for runtime side effects — hangs, crashes, data loss — because a persisted cache shared by the app and the CLI is exactly the kind of thing that goes wrong at 2am rather than in a test. It found three real problems, two of them regressions this branch introduced. All are fixed.

**1. Two OpenCodex homes could be mixed in one cache.** `OpenCodexUsageLog.cacheRoot()` has no home component while the log path does, so every `OPENCODEX_HOME` shares one database. Before this branch that was harmless: `readCachedEntries(identity:)` required the stored identity to match, so another home's cache simply missed. This branch replaced that with a cursor check made from a *separate* connection — so a loader could read its own cursor, have the CLI replace the cache for another home in between, and then use that home's rows as its baseline. The write-lock check only rejected a *same-file older offset*, so the foreign cursor passed, the foreign rows were not deleted, the cursor was reset and this loader's tail was appended on top.

Now: the cursor and the cached rows are read in **one transaction**; the write carries the exact base cursor the parse was derived from; and an incremental write is rejected whenever the durable cursor differs from that base **in any field**, not just when it is a newer same-file offset. Regression test: seed the cache for log B, drive an incremental load for log A, assert both the returned entries and the sqlite rows contain only A's request IDs.

**2. A crash class from `.mappedIfSafe`.** The full parse mapped the whole log and then walked it with `memchr`. A truncation between the mapping and touching those pages raises SIGBUS, which Swift cannot catch — the app dies. The reviewer checked the installed OpenCodex writer (2.31.0) and confirmed it only appends, so exposure today is low, but manual cleanup or a future writer can truncate.

The parser now opens one `FileHandle`, snapshots its size with `fstat`, and reads exactly the range it intends to parse, treating a short read as "changed under us". Three things fall out of that: no mapping, so no SIGBUS; the read is bounded, so a continuously growing file cannot be chased indefinitely (`readToEnd()` had no bound); and `fullReload` gets the same post-read identity check the incremental path already had.

The cost is memory on the cold path: the full parse now holds the file's bytes, so a rebuild on a 42 MB log peaks at ~290 MB instead of ~187 MB. That is still well under the ~452 MB of the parent branch, and it runs only on a cold cache or a schema rebuild — the per-refresh path reads only the appended tail and is unchanged (~118 MB). I took the trade deliberately; say the word if you would rather have the mapping back with a documented caveat, or a chunked reader to claw the memory back.

**3. Transient I/O failures were published as "no spend".** Any `stat` failure returned an empty array, and `SpendDashboardSource+OpenCodex.swift` turns an empty array into `.confirmedEmpty`, which removes the OpenCodex source from the dashboard. Before this branch the metadata read was a throwing `attributesOfItem`, so a failure surfaced as unavailable. Only `ENOENT`/`ENOTDIR` now mean "absent"; every other errno throws. Tests cover a circular symlink and an unreadable directory (skipped when running as root), and a genuinely missing log still returns empty without throwing.

**Also:** schema v2 moved to `opencodex-usage-v2.sqlite`, so a downgrade to an older build leaves it a usable v1 cache instead of one it can read but never write.

**What the review could not break**, for the record: the retry state machine is bounded at two iterations; there is no cyclic SQLite lock acquisition between the app and the CLI; the cursor cannot point past EOF; and no descriptor, statement or connection leaks on any error path.

**One thing I am disclosing rather than fixing here.** The forced-refresh reconciliation tail calls the synchronous OpenCodex merge from a `@MainActor` method (`SpendDashboardController.merge(outcome:capture:)` at the `.reconciling` branch → `SpendDashboardSource.mergingOpenCodexInputsWithObservation` → `store.loadEntries`), so that read runs on the main thread. It predates this PR and is reached only from `SpendDashboardController.refresh()` — the refresh button in the spend dashboard preferences pane, not the menu bar and not the periodic refresh. This PR makes it roughly an order of magnitude cheaper (a cache hit or a tail read instead of a full 42 MB re-parse), but the synchronous call is still there. It is your controller and you refactored it recently in #3105, so I would rather not fold a concurrency change into this PR — happy to do it as a separate one if you want it.

`make check` and the full `make test` (77/77 sharded groups, zero failures) pass on `e331eab0d`, and the PR description's measurement table has been updated to the re-measured numbers.

steipete:
Heads-up: #3136 just landed on main, touching the same aggregator/fan-out files plus a regenerated parser hash. This one needs a rebase onto latest main; after resolving, re-run `./Scripts/regenerate-codex-parser-hash.sh` and update the `compatiblePredecessorParserHashes` exact-list assertion in `CostUsageStoreTests` if the hash changes. Ready to merge once green.

steipete:
Maintainer verification: the combined focused OpenCodex parser, incremental store, fan-out, pricing, native cost-store, and provider-architecture suites passed 199 tests; the prior parallel SQLite fixture race reproduced before the database-scoped checkpoint fix and passed afterward; make check passed with zero formatting/lint violations; parser hash remained cfd84d13ad7d4cfa; exact-head GitHub CI including both macOS shards and Linux builds passed https://github.com/steipete/CodexBar/actions/runs/32822882427. The initial full local suite reached group 58/78 before the architecture scanner exceeded its 180-second timeout under host load above 90; exact-head CI completed the full platform coverage successfully. Landed as 17ade2b11a0e045f70a1a4c3e2885e5a4adaeeb7.

PR #3141, opened 2026-08-22
Stop re-merging the Codex plan-utilization history with itself on every refresh

olddonkey:
## Summary

`UsageStore.materializeCodexPlanUtilizationHistoryIfNeeded` exists to fold legacy, opaque and unscoped Codex plan-utilization buckets into the canonical account bucket. Its scoped loop also appended **the canonical bucket's own histories** to `historiesToMerge` — `matchesTargetContinuity` is true for `rawKey == canonicalKey` (`CodexHistoryOwnership.swift:58-61`), and only the *removal* of the old key was guarded by `if rawKey != canonicalKey`. So `guard !historiesToMerge.isEmpty` never fired once the canonical bucket had any history, and the migration merge ran on **every successful provider refresh and every menu open**, merging the history with itself.

That merge is quadratic: `updatedPlanUtilizationEntries` copied the whole entry array per entry, scanned it linearly for the insertion point, and allocated the same-hour slice. Measured with an optimized standalone reproduction over a real three-month-old history (session 1909 entries, weekly 2239): **20.6 ms of MainActor time per call**, scaling ~3.9× per doubling. `planUtilizationMaxSamples` is 17520 per series, so it keeps growing — at the retention cap this would be several hundred ms per series, on the main thread, while the menu is opening.

Reached from `UsageStore+Refresh.swift` (`applyProviderRefreshSuccess` → `recordPlanUtilizationHistorySample` → `MainActor.run` → `resolvePlanUtilizationAccountKey` → `resolveCodexPlanUtilizationAccountKey`) and from the synchronous `planUtilizationHistorySelection(for:)` used while building the menu. Codex-only; Claude short-circuits earlier.

### Changes

- **Only merge when a foreign source contributed.** A flag is set in the three branches that add non-canonical data (legacy scoped key, opaque recovery, unscoped adoption) and required by the guard, so the canonical-only case returns without merging or rewriting anything. Every path where a legacy/opaque/unscoped bucket does contribute is untouched: `legacyRawKeysToRemove` is only ever populated in branches that also set the flag (so no removal is skipped), and `providerBuckets.unscoped` is cleared only inside the branch that sets it.
- **Make the genuine migration near-linear.** `updatedPlanUtilizationEntries` mutates the array in place and finds the insertion point with a binary search for the same strict upper bound (plus a fast path for the common append); `mergedPlanUtilizationHistories` accumulates per series and constructs each history once.
- **Enforce the sortedness invariant the binary search relies on.** Every in-app producer went through `PlanUtilizationSeriesHistory`'s designated initializer, which sorts — except the synthesized `Codable` decoder, which assigned `entries` verbatim from JSON. An explicit `init(from:)` now routes decoding through that initializer, so a history file written by an older build (or edited by hand) cannot smuggle in an unsorted series. The JSON shape is unchanged.

The skipped self-merge was also an *incidental* normalizer — it re-ran per-hour peak canonicalization on read, which is not a fixpoint. Repairing legacy data belongs at load time, not on every refresh and menu open, so that is not reintroduced here; the visible effect is that at most one extra real observation per affected hour is kept. This is commented at the guard.

### Measured on real data

Against this machine's real `history/codex.json` (4160 entries: `session` 1909, `weekly` 2251), driving the real `UsageStore.planUtilizationHistory(for: .codex)` read path (`swift test` debug builds, same machine; an optimized standalone reproduction of the same algorithm measured 20.6 ms per call for current main):

| Path | current main | this PR |
|---|---|---|
| canonical-only (every refresh, every menu open) | 243–274 ms | **0.94–1.12 ms** |
| genuine legacy migration | 238–273 ms | **13.8–16.8 ms** |

The migration result is unchanged: the legacy bucket is absorbed, one account bucket remains, all 4160 entries survive. Full terminal transcripts are in a comment below.

### Tests

- Canonical-only history is returned untouched **and enqueues no persistence write** (`planUtilizationHistoryRevision` unchanged) — the property that actually matters for the menu path.
- A genuine foreign merge matches an explicit expected result across overlapping hours, out-of-order sources, distinct session/weekly series, and retention trimming.
- The binary search's upper-bound contract is pinned directly through the DEBUG shim over a 65-entry array containing a run of equal `capturedAt` (a lower bound would return a different index), plus the `endIndex`/`startIndex` edges.
- Decoding a series whose JSON `entries` are out of order yields a sorted series, and the shuffled payload produces the same forecast estimate as the sorted fixture.

The existing Codex-ownership migration suite (12 cases, all of which seed a foreign key) is unchanged and still green.

### Process

Implemented by grok-4.6 (xhigh) via implementation-loop; reviewed hunk by hunk plus an independent deep review that confirmed both equivalence claims by differential fuzzing (200k sorted cases, zero mismatches; and an exhaustive path enumeration showing the foreign-source paths are byte-identical) and found the decoder gap, fixed in one iterate round. The line-anchored `ProviderArchitectureGatekeeperTests` entries for the touched file were re-verified independently after every edit. Full `make check` and `make test` (77/77 sharded groups, zero failures) pass locally on this head.

### DIFF
--- a/Sources/CodexBar/PlanUtilizationHistoryStore.swift
+++ b/Sources/CodexBar/PlanUtilizationHistoryStore.swift
@@ -40,6 +40,12 @@ struct PlanUtilizationSeriesHistory: Codable, Equatable, Sendable {
     let windowMinutes: Int
     let entries: [PlanUtilizationHistoryEntry]
 
+    private enum CodingKeys: String, CodingKey {
+        case name
+        case windowMinutes
+        case entries
+    }
+
     init(name: PlanUtilizationSeriesName, windowMinutes: Int, entries: [PlanUtilizationHistoryEntry]) {
         self.name = name
         self.windowMinutes = windowMinutes
@@ -54,6 +60,14 @@ struct PlanUtilizationSeriesHistory: Codable, Equatable, Sendable {
             let rhsReset = rhs.resetsAt?.timeIntervalSince1970 ?? Date.distantPast.timeIntervalSince1970
             return lhsReset < rhsReset
         }
+    }
+
+    init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let name = try container.decode(PlanUtilizationSeriesName.self, forKey: .name)
+        let windowMinutes = try container.decode(Int.self, forKey: .windowMinutes)
+        let entries = try container.decode([PlanUtilizationHistoryEntry].self, forKey: .entries)
+        self.init(name: name, windowMinutes: windowMinutes, entries: entries)
     }
 
     var latestCapturedAt: Date? {
--- a/Sources/CodexBar/UsageStore+PlanUtilization.swift
+++ b/Sources/CodexBar/UsageStore+PlanUtilization.swift
@@ -373,8 +373,10 @@ extension UsageStore {
             let canonicalWindowMinutes = sample.name.canonicalWindowMinutes(sample.windowMinutes)
             let key = PlanUtilizationSeriesKey(name: sample.name, windowMinutes: canonicalWindowMinutes)
             if let existingHistory = historiesByKey[key] {
-                guard let updatedEntries = self.updatedPlanUtilizationEntries(
-                    existingEntries: existingHistory.entries,
+                var updatedEntries = existingHistory.entries
+                self.assertPlanUtilizationEntriesSorted(updatedEntries)
+                guard self.updatedPlanUtilizationEntries(
+                    existingEntries: &updatedEntries,
                     entry: sample.entry)
                 else {
                     continue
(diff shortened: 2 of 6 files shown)


olddonkey:
## Real-behavior proof (requested before merge)

Measured against **this machine's actual on-disk Codex plan-utilization history** — `history/codex.json`, 4160 entries across 2 series (`session` window 300: 1909 entries, `weekly` window 10080: 2251 entries), roughly three months of normal use. The harness loads that real file, installs it under the canonical account key, and drives the **real read path** `UsageStore.planUtilizationHistory(for: .codex)` — the same call the menu build and `recordPlanUtilizationHistorySample` go through — printing wall-clock per call. It is a measurement harness, not an assertion suite: the numbers below are its stdout.

Both sides are `swift test` **debug** builds of the same package on the same machine, so the absolute numbers are roughly an order of magnitude above release; the ratio is the point. (An optimized standalone reproduction of the same algorithm over the same history measured 20.6 ms per call for the current-main behaviour.)

### Before — `27c7f334e` (current main)

```text
MEASURE: loaded 2 series / 4160 entries from the real history file
MEASURE:   series session window=300 entries=1909
MEASURE:   series weekly window=10080 entries=2251
MEASURE: [canonical-only] 5 read-path calls: 249.11, 250.00, 242.93, 274.46, 243.06 ms
MEASURE: [legacy migration] 5 calls: 264.51, 257.88, 272.79, 262.58, 238.06 ms
MEASURE: [legacy migration] account buckets now: 1 (legacy key removed: true)
MEASURE: [legacy migration] merged series: 2, entries: 4160
```

### After — this PR

```text
MEASURE: loaded 2 series / 4160 entries from the real history file
MEASURE:   series session window=300 entries=1909
MEASURE:   series weekly window=10080 entries=2251
MEASURE: [canonical-only] 5 read-path calls: 1.12, 0.96, 0.96, 0.94, 0.94 ms
MEASURE: [legacy migration] 5 calls: 16.01, 13.83, 16.79, 13.78, 14.16 ms
MEASURE: [legacy migration] account buckets now: 1 (legacy key removed: true)
MEASURE: [legacy migration] merged series: 2, entries: 4160
```

**Canonical-only refresh behaviour** (the state every normal single-account user is in, on every provider refresh and every menu open): **243–274 ms → 0.94–1.12 ms**. The remaining work is reading the bucket; the merge no longer runs.

**A real migration outcome** is unchanged in result and faster: the legacy bucket is still absorbed (`legacy key removed: true`, one account bucket remains) and the merged series still carry all 4160 entries — at **238–273 ms → 13.8–16.8 ms**, which is the in-place insert plus binary search doing what it was meant to do.

### One fix this measurement produced

The first version of this branch put the DEBUG sortedness assertion *inside* `updatedPlanUtilizationEntries`, i.e. an O(n) scan per inserted entry — which reintroduced, in debug and test builds only, exactly the quadratic scan this insertion path exists to remove (the legacy migration measured ~1000 ms with it, ~15 ms without). It is now checked once per merged series. Release builds were never affected, but the assertion as written would have slowed every developer's test run; thanks to whatever prompted the "show real behaviour" ask, since a release-only benchmark would have hidden it.

### Also in this push

The `CHANGELOG.md` entry has been removed from this branch — `docs/RELEASING.md` reserves the changelog for the release flow. For whoever writes the release notes:

> - Codex: the plan-utilization history is no longer re-merged with itself on every refresh. The legacy-bucket migration also folded the canonical bucket into its own merge, so a quadratic merge ran on the main thread on every provider refresh and every menu open (~20 ms on a three-month-old history, growing with the square of its size); it now runs only when a legacy, opaque or unscoped bucket actually needs migrating, and the merge inserts in place with a binary search instead of copying the whole array per entry (#3141).

`make check` and the full `make test` (77/77 sharded groups, zero failures) pass on the pushed head.

PR #3150, opened 2026-08-23
Fix Codex day cost blanked by trace-only priority turns

olddonkey:
## Problem

On a machine that uses Codex priority processing, the Codex cost card goes blank: **Today** shows `—`, recent days disappear from the 30-day chart, and the window total silently shrinks to the sum of the days that survived. On my install every day from 2026-08-13 on lost its cost while 2026-08-12 and earlier kept theirs — the 30d figure read `$2,408` when roughly half the window was simply missing.

## Root cause

`codexPricingModeEvidence` judges whether the retained rows own a `(day, model)` group by comparing them against the file's persisted standard/priority split. The two sides do not share a source:

- the persisted maps are rebuilt from the rows' own `pricingMode` (`CostUsageStore+CodexCache.fileAggregates`), and
- the report-side split (`codexModeTokenMaps`) additionally classifies a row as priority when its `turnID` appears in `codexPriorityTurns`, i.e. in `~/.codex/logs_2.sqlite`.

So a turn the trace database reports as `service_tier: priority` **after** its rows were already persisted as `standard` reads as a row-ownership mismatch. The rows lose trust, the group falls back to `codexAggregateCostUSD`, and that returns `nil` for any model with a long-context tier (#2858 deliberately refuses to apply 200k-tier rates to a day aggregate — a day always exceeds the threshold). With every model on the day nil, `entryCost` is nil and the day vanishes from the menu, the chart and the total.

A single 300k-token session is enough: on 2026-08-22 one file contributed 189,156 trace-priority tokens against a persisted `priority = 0`, and that blanked the whole day's 258M tokens. The trace database only retains ~10 days, which is why the damage is always the most recent stretch and rolls forward daily.

## Fix

Judge retention against both classifications and flag only a group that matches neither. A wrongly retained row set still fails both, because the persisted totals are canonical for the file (the existing `persistedModeTotal == canonicalTotal` guard) and tier classification never changes how many tokens the rows carry — so the fork-prefix protections from #2858 are untouched. Priority pricing itself still comes from the trace: the trusted row path prices those turns with the fast-tier rates, and the model breakdown keeps its standard/priority split.

## Tests

- `trace priority turns keep row ownership for rows persisted as standard` — rows persisted as standard plus a trace-only priority turn: no mismatch, the day is priced from the rows (standard + fast-tier), and the assertion that `codexAggregateCostUSD` is `nil` for the day aggregate documents why distrusting the rows blanks the day.
- `stale persisted mode split still invalidates row ownership` — a mode split that neither classification reproduces still distrusts the rows and leaves the cost unavailable.
- All 16 tests in `CostUsageScannerForkSplitTests` pass, including every existing fork-split/pricing-evidence case.
- Full `make check` and `make test` green.

`CodexParserHash` changes because the file is inside the hashed set; the previous value is added to `compatiblePredecessorParserHashes` since neither the parser nor the persisted row shape changed, so existing caches are adopted instead of rebuilt.

## Measured on a real store

Offline A/B over a pristine copy of my `cost-usage.sqlite` and a frozen copy of `logs_2.sqlite` (109 priority turns), rebuilding the report through `buildCodexReportFromCache` with the real `codexPriorityTurns` — no scan, nothing mutated, only the code under test changes:

| | `main` | with this fix |
|---|---|---|
| mismatch groups in the window | 17 (2026-08-13 … 08-22) | 0 |
| days priced 08-13 … 08-22 | 0 / 10 | 9 / 10 |
| 30-day total | **$2,408.12** | **$4,139.48** |

The `main` column reproduces exactly what the shipped app showed in its menu. 2026-08-20 stays unpriced under both — that day's rows genuinely do not reconcile, which is the `unresolvedRowGroups` path and out of scope here.


olddonkey:
Changelog entry for the release flow (not added to `CHANGELOG.md` per `docs/RELEASING.md`):

```
- Codex: keep a day's cost when a turn's priority tier is only known from the trace database. The tier classification no longer reads as a row-ownership mismatch, so recent days stop disappearing from the cost card, the chart and the window total.
```

PR #3162, opened 2026-08-23
Fix Claude web cookie refresh

Zihao-Qi:
## Summary

- keep Claude web-cookie refresh usable in bundled ad-hoc development builds by routing only cookie cache entries to the existing process-memory store
- preserve persistent Keychain storage for certificate-signed builds and OAuth credentials
- memoize no-UI Keychain preflight reads per import operation and evaluate browser candidates lazily
- prefer Chrome for Claude while preserving the full fallback order, including Chrome-without-session → Safari
- isolate the new tests from process-global Keychain and browser-gate state

## Why

After the stricter executable ACL validation, an ad-hoc rebuilt app can no longer read, update, or clear a cookie cache item created by a previous build without user interaction. Automatic Claude refresh can therefore return stale cached data even though a manual web-cookie retry eventually succeeds after permission is granted.

The eager browser candidate filter also performs Safe Storage preflights for later Chromium browsers before the preferred source is attempted, adding unnecessary work and potential prompt pressure.

## Validation

- [CI run 10604](https://github.com/steipete/CodexBar/actions/runs/32674345444) — passed on the current head, including lint, both macOS test shards, Linux x64/arm64, Linux musl, and the aggregate gate
- `swift test --filter 'KeychainCacheStoreTests|ClaudeWebBrowserFallbackTests|BrowserDetectionTests|KeychainNoUIQueryTests|BrowserCookieOrderLabelTests'` — 78 tests passed
- repeated the affected 78-test combination four times while validating test isolation
- `swift test --filter 'KeychainNoUIQueryTests|SpendDashboardDateTruthTests'` — 30 tests passed after rebasing onto current `upstream/main`, including the previously unrelated CHF fixture failure
- `make check` — passed with zero SwiftFormat/SwiftLint violations
- `swift build -c release` — passed
- `CodexBar.app/Contents/Helpers/CodexBarCLI --help` — passed against the packaged developed bundle
- independently reviewed; no remaining code or security findings

### Developed bundle smoke evidence

Built and launched the workspace bundle (not the public app) from the current head:

```text
==> Signing: adhoc (set APP_IDENTITY or install a dev cert to avoid keychain prompts)
==> Preserving CodexBar keychain entries (pass --clear-adhoc-keychain to reset adhoc keychain state)
Created /Users/xiaoxiawang/Desktop/CodexBar/CodexBar.app
==> launch app
OK: CodexBar is running.
```

Live behavior was tested with the developed ad-hoc bundle: the expected one-time Keychain permission flow completed, and subsequent Claude automatic refreshes succeeded without another permission prompt. Account values and cookies were intentionally omitted.

### DIFF
--- a/Sources/CodexBarCore/BrowserCookieAccessGate.swift
+++ b/Sources/CodexBarCore/BrowserCookieAccessGate.swift
@@ -60,6 +60,9 @@ public enum BrowserCookieAccessGate {
     private static let log = CodexBarLog.logger(LogCategories.browserCookieGate)
     @TaskLocal private static var explicitRetryScope: ExplicitRetryScope?
     @TaskLocal private static var deniedBrowsersForTesting: [Browser]?
+    #if DEBUG
+    @TaskLocal private static var shouldAttemptOverrideForTesting: Bool?
+    #endif
 
     static let allowTestCookieAccessEnvironmentKey = "CODEXBAR_ALLOW_TEST_BROWSER_COOKIE_ACCESS"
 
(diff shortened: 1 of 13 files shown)


Zihao-Qi:
@clawsweeper re-review. The current head is green in CI, and the PR body now includes the redacted developed-bundle build/launch transcript and current-head validation results.

Issue #3168, opened 2026-08-24
Codex card keeps rendering pre-reset snapshot after backend weekly reset; only re-clicking the active tab repaints it (⌘R doesn't)


NealZhouPanda:
### Environment

- CodexBar **0.54.1 (build 128)** from `/Applications/CodexBar.app`
- macOS 26.x, MacBook Pro M3 Max (`Mac15,11`), zh-Hans locale
- Provider: **Codex**, single account, Plus OAuth (`loginMethod: plus`)
- Config: `codexActiveSource: liveSystem`, `source: auto`, `cookieSource: off`
- Surface: normal menu-bar popup (no custom layout)

### Summary

After OpenAI reset the weekly quota server-side mid-window, CodexBar kept rendering the **previous** weekly snapshot (**82% left / 18% used / resets Aug 27**) even though every fetch path was returning the new state (**100% left / 0% used / resets Aug 31**). The stale card survived multiple app relaunches. Only **re-clicking the already-active Codex tab** in the popup forced a full card redraw; **⌘R Refresh did not repaint it**.

This looks like the same *menu rows are not rebuilt during menu tracking* mechanism as #1527 (Claude account tabs), now on the Codex card with a single account.

### Evidence

1. Bundled CLI returned fresh, exact-confidence data throughout the stale period:
   ```shell
   codexbar usage --provider codex --source oauth
   ```
   → `usage.secondary: {usedPercent: 0, resetsAt: 2026-08-30T23:43:41Z, updatedAt: <fresh>}`, `dataConfidence: exact`
2. Popup kept showing: **Weekly 82% left · 2d 22h to reset · updated ~6h ago**, which corresponds to the last persisted snapshot `usedPercent=18 / resetsAt=2026-08-27T03:41:41Z / capturedAt=2026-08-23T23:24:14Z`.
3. Unified log shows successful `HTTP 200` fetches (~1.4 KB responses, `cache_hit=false`) every few minutes during that period; no Codex auth/provider errors logged.
4. Deleting `~/Library/Application Support/com.steipete.codexbar/history/codex.json` + relaunch **did not** change the displayed card.
5. After rewriting the persisted caches (`history/codex.json`, `openai-dashboard.json`) with the live values and relaunching, the **first popup open still rendered the old percentages**; clicking the active Codex tab once repainted the card to “已用 0%”, and it stayed correct on subsequent opens.
[...]

PR #3177, opened 2026-08-24
Fix Codex weekly reset publication

### LINKED ISSUE #3179: Codex early backend weekly reset remains stale when reset-credit inventory is unchanged and non-zero (#2790 follow-up)
## Summary

This is a follow-up to [#2790](https://github.com/steipete/CodexBar/issues/2790), which was closed after [#2897](https://github.com/steipete/CodexBar/pull/2897) fixed the explicitly-zero-reset-credit deadlock.

That fix was correct for the zero-credit state, but its boundary was narrower than the original report. The original #2790 incident had an unchanged positive reset-credit inventory (`2 available -> 2 available`). Fresh reports on CodexBar 0.54.1 and 0.55.0 reproduce that remaining case with `1 available -> 1 available`.

When OpenAI performs an early/backend-propagated weekly reset without consuming the user's available reset credit, CodexBar fetches the fresh exact OAuth state but continues publishing the previous weekly percentage and reset boundary. Manual refresh and restart do not recover because the rejected baseline remains persisted.

This is not a regression introduced by #2897. It is the residual unchanged-positive-credit scenario from #2790 that #2897 did not cover.

## Decisive comparison

| Scenario                           | Persisted weekly state       | Fresh exact OAuth state      | Reset-credit sequence        | Result                   |
| ---------------------------------- | ---------------------------- | ---------------------------- | ---------------------------- | ------------------------ |
| Original #2790                     | 6% used, reset Aug 15 14:11Z | 0% used, reset Aug 15 23:06Z | `2 available -> 2 available` | Fresh state rejected     |
| Zero-credit subcase fixed by #2897 | 82% used, reset Aug 18       | 0% used, reset Aug 20        | `0 available -> 0 available` | Fresh state now accepted |
| Fresh 0.54.1 reproduction          | 75% used, reset Aug 27       | 0% used, reset Aug 31        | `1 available -> 1 available` | Fresh state rejected     |

An independent [0.55.0 reproduction](https://github.com/steipete/CodexBar/issues/2790#issuecomment-5394447574) reports the same unchanged non-zero inventory, fresh exact OAuth helper output, stale persisted display, and immediate recovery after moving the account-snapshot cache aside.

## Rejection boundary

Current `main` still routes an early observation through `confirmsManualResetCreditConsumption` and preserves the previous snapshot when that returns false:

<https://github.com/steipete/CodexBar/blob/c9e7f4df556d914e4652e21fcb30ee9f3845a0b2/Sources/CodexBar/Providers/Codex/CodexWeeklyResetConfirmation.swift#L140-L197>

The merged exception returns true only when the previous inventory is explicitly zero:

```swift
guard !previouslyAvailableCredits.isEmpty else {
    return previousCredits.availableCount == 0
}
```

When an available credit remains present and unchanged, the existing path accepts only evidence that it became `redeeming`/`redeemed` or disappeared while the available count dropped. A backend-propagated reset does neither, so repeated consistent observations continue to evaluate to `.preservePrevious`.

The remaining case is therefore a policy decision: which corroborating provider evidence is sufficient to accept a genuine backend reset when a reset credit exists but was not consumed. It is not the mechanical zero-credit deadlock addressed by #2897.

## Fresh live evidence

The [0.54.1 reproduction](https://github.com/steipete/CodexBar/issues/2790#issuecomment-5389518195) recorded:

| Source                                  | Captured/updated (UTC) | Weekly used | Next reset          | Reset credits                                     |
| --------------------------------------- | ----------------------:| -----------:| ------------------- | -------------------------------------------------:|
| Persisted app snapshot                  | 00:41:54               | **75%**     | 2026-08-27 06:55:31 | **1 available**                                   |
| Packaged OAuth helper                   | 00:56:03               | **0%**      | 2026-08-31 00:42:29 | **1 available**                                   |
| Direct authorized `wham/usage` response | ~00:42                 | **0%**      | 2026-08-31          | **1 available** (`applicable_available_count: 0`) |

The account identity and OAuth source were unchanged, `dataConfidence` was `exact`, and the same reset credit remained available. The reporter also connected the timing to a [public statement that a reset had been propagated to affected accounts](https://github.com/steipete/CodexBar/issues/2790#issuecomment-5389527838). That context supports the backend-reset explanation; the unchanged credit inventory independently shows that this was not a user redeeming the available credit.

The `applicable_available_count: 0` field may be useful as additional corroboration if CodexBar retains it, but this report does not assume that field is sufficient by itself.

## Expected behavior

After a bounded confirmation sequence establishes a genuine early/backend weekly reset for the same account and exact OAuth source, CodexBar should publish the fresh weekly state even when an unrelated available reset credit legitimately remains unchanged.

## Actual behavior

- The packaged helper and authorized provider response return the fresh exact weekly state.
- CodexBar continues publishing the previous percentage, boundary, and old update time.
- Manual refresh and relaunch do not replace the persisted baseline.
- No authentication, browser-cookie, provider-routing, or account-selection error is present.

## Minimal reproduction

1. Begin with a persisted exact OAuth weekly snapshot whose usage is above the low-reset threshold and whose reset boundary remains in the future.
2. Keep at least one reset credit present and `available`.
3. Have OpenAI perform an early/backend-propagated weekly reset without consuming that credit.
4. Let CodexBar obtain its initial and confirmation observations for the same account/source.
5. Confirm that the packaged helper reports low usage and a consistent, materially later future boundary while the reset-credit inventory remains unchanged.
6. Observe that CodexBar retains the old persisted/public weekly state through manual refresh and restart.

## Requested decision and safe boundary

Please track the remaining policy decision separately from the zero-credit fix: should CodexBar admit two consistent exact OAuth observations of an early backend reset when the positive reset-credit inventory remains explicitly unchanged?

A bounded admission path could require all of the following across the complete previous/initial/confirmation sequence:

- stable owner/account and OAuth source;
- exact data confidence;
- present, monotonically newer observations;
- two low-usage observations with equivalent valid future reset boundaries;
- a new boundary materially later than the persisted boundary;
- reset-credit inventories present in every observation, with the same positive count, credit identities, and `available` statuses;
- no single-sample bypass.

Existing safeguards should remain for missing/unknown inventories, account or source changes, non-exact data, non-monotonic timestamps, inconsistent boundaries, and a single transient low observation.

Concrete counterexamples that should still be rejected:

- previous inventory contains available credit `A`, but the initial/confirmation inventory contains different credit `B` despite the same aggregate count;
- previous inventory is zero but either later inventory becomes non-zero (`previous=0, initial=1, confirmation=1`);
- the initial and confirmation observations disagree on the new weekly boundary.

## Workaround

The recoverable workaround remains:

1. Quit CodexBar.
2. Move `~/Library/Application Support/CodexBar/codex-account-snapshots.json` to a backup.
3. Relaunch CodexBar.

This immediately rebuilt the fresh OAuth baseline in the current reports. Toggling the Codex provider helped one commenter but did not work for another, so it is not a reliable workaround.

## Related but distinct

[#3168](https://github.com/steipete/CodexBar/issues/3168) is a separate presentation-synchronization report where re-clicking the active Codex tab repaints already-published fresh data. In this issue, the fresh observation is rejected at the weekly-reset admission boundary and the persisted account snapshot remains stale.

## References

- Original report and full history: [#2790](https://github.com/steipete/CodexBar/issues/2790)
- Narrow zero-credit fix: [#2897](https://github.com/steipete/CodexBar/pull/2897)
- Fresh 0.54.1 unchanged-positive-credit evidence: [comment](https://github.com/steipete/CodexBar/issues/2790#issuecomment-5389518195)
- Independent 0.55.0 confirmation: [comment](https://github.com/steipete/CodexBar/issues/2790#issuecomment-5394447574)

Zihao-Qi:
## Summary

- keep suspicious Codex weekly readings at or below 1% private instead of publishing the result of an immediate retry
- persist an account-scoped reset candidate and publish it only after a separate exact OAuth refresh at least 60 seconds later, including the reported `1 available -> 1 available` case with an advanced reset boundary
- discard stale or incompatible candidates and prevent small corrections such as `6% -> 5% -> 5%` from triggering reset confetti

## Why

Codex can report a fresh post-reset weekly value while the reset credit remains available. The old guard treated that as unproven indefinitely, so both the usage value and its “Updated” timestamp could remain frozen. An immediate second request is not independent evidence, so this change carries a private candidate into a later refresh cycle instead.

## Safety

- only exact OAuth observations are eligible for delayed corroboration
- candidates require matching account, plan, future reset-boundary evidence, and unchanged positive reset-credit inventory
- candidates expire after 30 minutes and are discarded on source, identity, plan, boundary, credit, or timestamp mismatch
- consumed reset-credit evidence remains an immediate confirmation path
- the public snapshot remains unchanged until corroboration; no timestamp or usage value is synthesized
- the candidate is stored with the account snapshot, so relaunching does not lose the evidence needed to recover a previously frozen installation

## Testing

- `swift test --filter CodexWeeklyResetConfirmationTests` — 26 passed
- `swift test --filter CodexWeeklyResetPublicationTests` — 16 passed
- `swift test --filter UsageStoreCodexWeeklyResetCelebrationTests` — 9 passed
- `swift test --filter '^CodexBarTests\.CodexAccountScopedRefreshTests/'` — 120 passed
- `make check` — formatting and lint clean
- `make test` — 929 selections in 78 groups passed with zero failures, retries, or timeouts

No screenshot is included because this changes provider publication state rather than menu layout. The rollover cases are verified at the deterministic state/model seams, including persisted relaunch recovery.

Scope: this fixes the frozen Codex weekly snapshot in #3179. It does not claim to fix the separate menu-presentation issue in #3168.

Closes #3179


steipete (changes requested):
Thanks for taking this on — the confirmation-state-machine structure is a reasonable shape, all 46 new tests pass, and `make check` is clean. But a deep review found the core scenario from #3168/#3179 still isn't fixed, and the workaround path introduces two new failure modes. Details:

**P0 — the reported `1 → 1` early-reset case remains stuck.** `CodexWeeklyResetConfirmation.swift:170` still rejects an early reset unless a credit was consumed, the previous inventory was explicitly zero, or the boundary is unchanged within one second. A real backend reset advances the boundary, and unchanged positive credit inventory satisfies none of those arms — which is exactly the residual case both issues describe. The existing test at `CodexWeeklyResetPublicationTests.swift:467` actively asserts the scenario stays frozen at the previous reading, and none of the new tests feeds unchanged positive inventories across all three snapshots. Please add a test that reproduces #3179's table (`1 available → 1 available`, boundary advanced, fresh 0%) and make it publish.

**P1 — the claimed relaunch migration doesn't repair the persisted baseline.** `UsageStore+PlanUtilization.swift:956` discards legacy confetti-detector state, but the stale usage snapshot is persisted separately and rehydrated at `UsageStore.swift:534`. The migration test only checks detector-state deletion, so upgrading and relaunching won't recover an already-stuck install.

**P1 — false confetti on mundane corrections.** `UsageStore+LimitResetCelebration.swift:429` treats any crossing from >5% to ≤5% as a reset candidate regardless of magnitude — a `6% → 5% → 5%` drift with unchanged plan and boundary celebrates a reset after 60 s. Gate on boundary advancement, credit consumption, or a substantial drop.

**P1 — transient zero glitches publish immediately.** `UsageStore+CodexWeeklyResetConfirmation.swift:81` fires the confirmation fetch immediately and `CodexWeeklyResetConfirmation.swift:185` accepts two matching zeros on account/plan/boundary match alone, so a brief backend cache glitch hitting both requests persists a wrong reading; the 60 s / 30 min windows only guard confetti, not publication.

Happy to re-review once the `1 → 1` case actually publishes and the persisted-baseline recovery is proven by a test.

Zihao-Qi:
Addressed the requested changes in `653628ce5`:

- the exact `1 available -> 1 available`, advanced-boundary, fresh 0% case now publishes only on a later exact OAuth refresh at least 60 seconds after the private candidate was created
- the candidate persists beside the stale account snapshot; a relaunch test proves recovery and candidate cleanup
- the immediate same-request retry remains private, and expired or mismatched candidates fail closed
- `6% -> 5% -> 5%` now has an explicit no-confetti regression test
- `make check` is clean and the full `make test` suite passed all 929 selections in 78 groups with zero retries or timeouts

steipete (approved):
All requested weekly-reset cases now pass, including unchanged positive reset-credit inventory, advanced boundaries, persisted recovery across relaunch, and false-confetti suppression. Maintainer hardening additionally separates persisted local observation time from provider timestamps, expires stalled and regressing candidates reliably, rejects wrong-account or wrong-source stale observations, preserves sibling-account caches, and decodes preexisting snapshots compatibly. Focused account-ownership, confirmation, publication, celebration, and provider-architecture regressions plus make check are green.

steipete:
Landed in 0a1aa53598c94003a87bcdcca4af88b0ad508421 after addressing every maintainer review finding and additional account-safety defects. Exact OAuth weekly resets with unchanged positive reset-credit inventory now publish only after an independent later refresh; private account-scoped evidence survives relaunch without publishing transient zero readings or false reset confetti. Provider timestamps establish evidence ordering, while a separately persisted local observation clock enforces the 60-second minimum and 30-minute maximum even when provider timestamps stall, regress, or jump forward. Expired evidence is removed from persisted snapshots; stale observations from another account, plan, or source fail closed; and creating an active-account candidate preserves every sibling account snapshot. Legacy persisted candidates decode compatibly without cache migration. Verification: 198 focused account-ownership, weekly-reset confirmation, publication, celebration, persistence, and provider-architecture tests passed; make check passed with zero violations; the full local suite passed all 930 selections across 78 groups with zero failures, retries, or timeouts; and exact-head GitHub Actions run 32873319421 passed both macOS shards, Linux x64, arm64 and musl builds, lint, and security checks. Contributor credit and changelog included.

Issue #1843, opened 2026-07-02
Feature request: Support multiple accounts or instances of the same provider


bystritskiy:
# Feature request: Support multiple accounts or instances of the same provider

## Summary

Add the ability to configure multiple instances of the same provider and track their usage independently.

For example, a user may have two separate Claude Code installations or profiles on the same Mac:

- Personal Claude account
- Work Claude account

Each instance may use a different configuration or data directory. Currently, a provider can only be enabled once, so CodexBar can monitor only one account per provider.

## Proposed solution

Allow users to duplicate a provider and configure each instance separately.

For example:

```text
Claude — Personal
Config/data directory: ~/.claude

Claude — Work
Config/data directory: ~/Library/Application Support/Claude-Work
```

Each provider instance could:

- Have a custom display name or alias
- Use its own authentication and configuration source
- Allow selecting a custom config or data directory
- Track and display usage independently
- Appear as a separate item in CodexBar

A possible settings flow:

1. Select a provider such as Claude.
2. Click **Add another account** or **Duplicate provider**.
3. Enter an optional alias.
4. Select the provider's config or data directory.
5. CodexBar creates an independent instance of that provider.

## Use case

I run two Claude installations on the same Mac, each using a separate account and user-data directory.

I would like CodexBar to show the remaining limits for both accounts at the same time, without having to sign in and out or replace the existing provider configuration.

This could also be useful for:

- Personal and work accounts
- Multiple client accounts
- Separate browser or CLI profiles
- Users with more than one subscription for the same provider

## Additional context

Ideally, this would be implemented as a generic provider-instance model rather than Claude-specific behavior.

Any provider capable of loading credentials or usage information from a configurable location could then support multiple accounts.

tanguc:
Same need here, from the passive-monitoring angle: I want N accounts of the
same provider visible at once, without switching between them. Switching
defeats the point of a menu-bar monitor.

One data point that might make the generic model easier to land: API-key
providers like z.ai (GLM Coding Plan) are a lighter case than the Claude
data-dir example. An "account" there is nothing more than another bearer key.
Usage for each comes from a single call to
`https://api.z.ai/api/monitor/usage/quota/limit` with that key: no config
directory, no login/logout, no daemon reload.

So a provider-instance model where each instance carries
`{alias, credential, optional host/region}` would cover both the OAuth/data-dir
providers (Claude, Codex) and the key-based ones (z.ai and the other API-key
providers) with one UI.

[...]

PR #3180, opened 2026-08-24
Support multiple API-key OpenCode Go accounts

Pl3ntz:
## Summary

- OpenCode Go's token accounts currently inject a Cookie header (`.cookieHeader`), which only supports one *browser session* per account and requires manual cookie capture. This switches the injection to a plain API key (`.environment(key: OpenCodeGoSettingsReader.apiKeyEnvironmentKey)`), reusing the exact generic token-account infrastructure z.ai already uses (`TokenAccountSupport`, `multiAccountMenuLayout == .stacked`, `shouldFetchAllTokenAccounts`) — no new UI or fetch machinery needed, just the right injection type.
- `requiresManualCookieSource` is now `false`, so the Session tokens ("API key accounts") section is visible without first flipping the provider to manual cookie mode.
- Per-account fetch now tries the API strategy before the cookie/web strategy when a token account is selected — previously the descriptor unconditionally scoped to the cookie/web path first, which is correct for cookie-based accounts but wrong once accounts are API-key based.
- Relaxes the CLI's `config set-api-key --label` guard, which previously hard-required `--provider zai` for *any* account option including a bare `--label`. It now only requires `zai` for the team-specific options (`--usage-scope`/`--organization-id`/`--workspace-id`); a bare `--label` is allowed for any provider whose `TokenAccountSupport.injection` is `.environment`, so multiple OpenCode Go accounts (or any other API-key provider) can be added from the CLI, not just the Settings UI.

## Why

Closes the "add API-key accounts for OpenCode Go" gap described in #1843 (multi-account/multi-instance support for the same provider) and unblocks the OpenCode Go case specifically called out as unsolved in #81. I have 3 separate OpenCode Go subscriptions (personal, side, and a company account) and wanted to see all three's session/weekly/monthly windows at once instead of switching credentials by hand.

## What I deliberately left out

I also patched `PersonalInfoRedactor` locally (masking emails like `v***y@gmail.com` under "Hide personal information" instead of blanking them entirely) while testing this with 3 real accounts on screen. That's a separate, app-wide UX/privacy-default decision unrelated to the multi-account fix, and I have no test coverage for it yet — happy to open it as its own issue/PR if there's interest, but keeping this PR scoped to the OpenCode Go multi-account fix.

## Test plan

- [ ] `make check` clean
- [ ] `swift test --filter OpenCodeGoProviderStrategyTests` — updated the existing strategy-ordering assertion for the selected-token-account case (now API-first, was cookie-first) and added coverage confirming the other scoped-web-strategy triggers (manual cookie source / workspace override) are unaffected
- [ ] `swift test --filter CLIConfigCommandTests` — added cases for: bare `--label` succeeding on an `.environment`-injection provider, rejection when the provider has no token-account support at all, rejection when the provider is cookie-based
- [ ] `swift build -c release` clean
- [ ] Manually verified against 3 real OpenCode Go accounts: all three fetch live usage via `source: "api"` and render as stacked cards simultaneously in the menu

Addresses the OpenCode Go case from #1843 (the broader "any provider" ask stays open — this only covers API-key-based providers, not cookie-based ones like Claude). Related to #81.

### DIFF
--- a/Sources/CodexBar/SettingsStore+TokenAccounts.swift
+++ b/Sources/CodexBar/SettingsStore+TokenAccounts.swift
@@ -2,6 +2,21 @@ import AppKit
 import CodexBarCore
 import Foundation
 
+/// Mirrors ProviderConfig.clean(_:)/CLIConfigCommand.cleanConfigSecret(_:) - trims whitespace and
+/// strips one matching pair of wrapping quote characters. A key saved via the Settings UI's own
+/// normalizer only trims whitespace, so without this a quoted legacy key and a clean new token
+/// for the same real credential compare as different and migrate a spurious duplicate account.
+private func cleanTokenAccountSecret(_ raw: String?) -> String? {
+    guard var value = raw?.trimmingCharacters(in: .whitespacesAndNewlines), !value.isEmpty else {
+        return nil
+    }
+    if (value.hasPrefix("\"") && value.hasSuffix("\"")) || (value.hasPrefix("'") && value.hasSuffix("'")) {
+        value = String(value.dropFirst().dropLast())
+    }
+    value = value.trimmingCharacters(in: .whitespacesAndNewlines)
+    return value.isEmpty ? nil : value
+}
+
 extension SettingsStore {
     func tokenAccountsData(for provider: UsageProvider) -> ProviderTokenAccountData? {
         guard TokenAccountSupportCatalog.support(for: provider) != nil else { return nil }
@@ -78,7 +93,7 @@ extension SettingsStore {
         // otherwise silently stop tracking that existing subscription. Migrate it in first.
         let legacyKey = accounts.isEmpty
             && TokenAccountSupportCatalog.support(for: provider)?.migratesExistingAPIKeyOnFirstAccount == true
-            ? self.providerConfig(for: provider)?.apiKey?.trimmingCharacters(in: .whitespacesAndNewlines)
+            ? cleanTokenAccountSecret(self.providerConfig(for: provider)?.apiKey)
             : nil
         // If the account being added uses the same token as the existing single key, the user is
         // just naming their current subscription (e.g. via Add Account), not adding a second one.
--- a/Sources/CodexBar/SettingsStore+TokenAccounts.swift
+++ b/Sources/CodexBar/SettingsStore+TokenAccounts.swift
@@ -106,7 +106,13 @@ extension SettingsStore {
         // whitespace-trimmed) so a quoted paste of an already-stored key doesn't compare as a
         // different credential and migrate a spurious duplicate.
         let normalizedIncomingToken = cleanTokenAccountSecret(trimmedToken) ?? trimmedToken
-        if let legacyKey, !legacyKey.isEmpty, legacyKey != normalizedIncomingToken {
+        // The stray key can also already be sitting in an *existing* account rather than just
+        // the one being added right now - e.g. an unlabeled CLI set-api-key re-populates
+        // providerConfig.apiKey with a token that already has a Default account from an earlier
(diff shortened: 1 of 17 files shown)


Pl3ntz:
@codex review

Pl3ntz:
@codex review

Pl3ntz:
@codex review

Pl3ntz:
@codex review

Pl3ntz:
@codex review

Pl3ntz:
@codex review

Pl3ntz:
@codex review

Pl3ntz:
@codex review

Pl3ntz:
@codex review

steipete (changes requested):
Nice reuse of the generic TokenAccountSupport machinery — the core direction is right, per-account env isolation checks out, and all focused suites plus `make check` pass on a merge with latest main. Three findings before merge, one blocking:

**P1 — Existing cookie accounts prevent API-key migration in the app.** `SettingsStore+TokenAccounts.swift:94` gates migration on `accounts.isEmpty`. A user upgrading with existing cookie accounts *and* a configured API key who then adds another key never gets a "Default" account for the existing key — it stays stored but inert. The CLI handles this case correctly (`CLIConfigCommand.swift:265`); the app should match.

**P2 — CLI ignores Copilot's clear-on-mutation contract.** `CLIConfigCommand.swift:293` preserves an existing Copilot API key when adding a labeled account, while the app clears it per `clearsAPIKeyOnMutation`. CLI and app diverge; align the CLI.

**P2 — Quoted input duplicates a credential.** `SettingsStore+TokenAccounts.swift:102` compares a *normalized* existing key against an incoming token that was only whitespace-trimmed, so existing `sk-same` plus pasted `"sk-same"` yields two accounts with the same key. Tests only cover the opposite quoting direction — add the missing one.

Happy to merge once the migration gate is fixed and the two P2s addressed.

Pl3ntz:
@steipete addressed all three in 262a815:

- **P1 migration gate**: `SettingsStore+TokenAccounts.swift` no longer gates legacy-key migration on `accounts.isEmpty` — matches `CLIConfigCommand.configSettingAPIKey`'s ungated behavior now.
- **Copilot clear-on-mutation**: `CLIConfigCommand.swift` now also clears `providerConfig.apiKey` when `clearsAPIKeyOnMutation` is true, so the CLI matches the app instead of leaving a stale Copilot key behind.
- **Quoted-input duplicate**: the incoming token in `addTokenAccount` is now quote-normalized (not just whitespace-trimmed) before comparing against the legacy key, plus the missing reverse-direction test.

Also found and fixed a pre-existing failure while running the full suite (not one of the three above, confirmed via `git stash` that it predates this round): `opencode go token accounts force manual cookie routing` still asserted the old cookie-header behavior. Updated it for the new API-key injection, which correctly leaves the global cookie source alone.

`swift build`, `make check`, and the full `swift test` suite are clean on this head.

@codex review

Pl3ntz:
Pushed 3c8c7f3. ClawSweeper's re-review of 262a815 caught a real bug: removing `accounts.isEmpty` exposed a case where a stray provider-level key (re-set by a plain, unlabeled `set-api-key` after an earlier migration) got re-migrated as a duplicate "Default" account on every subsequent add. Fixed by checking the full account list (normalized) before migrating, in both `SettingsStore+TokenAccounts.swift` and `CLIConfigCommand.swift`, with regression tests for both.

Also updated `docs/opencode.md`: a selected token account now routes by credential shape (API-key accounts go straight to the API strategy; only legacy cookie-shaped values keep the web-first chain), not unconditionally web-first.

`swift build`, `make check`, and the full `swift test` suite are clean.

Pl3ntz:
@codex review

Issue #3157, opened 2026-08-23
Grok: period-only CLI-proxy response reports 0% after Grok Build says free usage limit reached


anupamchugh:
## Summary

CodexBar 0.54.0 reported Grok `usedPercent: 0` from `grok-cli-proxy`, while Grok Build 1.0.5 displayed “You hit your free usage limit.”

## Current behavior

`GrokCreditsProxyFetcher.parseSnapshot` currently returns 0% whenever a reset period exists but `creditUsagePercent` and a usable cap/used pair are absent. The focused test explicitly expects this inference.

## Expected behavior

Missing usage fields are not evidence of zero usage. This response shape should produce `unknown`/unavailable, or an authoritative exhausted value if the provider exposes one. It should never be surfaced as confirmed 0% usage.

## Suggested regression test

A period-only payload must not produce `usedPercent: 0`; it should fail closed so callers cannot interpret it as available capacity.

No account identifiers, credentials, cookies, raw payloads, screenshots, or local paths are included.

olddonkey:
Heads-up on a display consequence of the fix that landed here (#3159, shipped in 0.55.0): an unknown usage percent produces no `RateWindow` at all, and the menu card only renders a metric when `snapshot.primary` exists — so on plans whose credits payload never publishes `creditUsagePercent`, the Grok card now shows the account row and plan name with no usage bar, and the menu bar shows nothing.

The part that makes it self-sustaining is that the period-only answer is a *successful* strategy result, and `ProviderFetchPipeline.fetch` returns on the first success — so `grok.web` and `grok.oauth-grpc` never run, including the grok.com surface that does publish a percent for these accounts.

#3181 keeps this issue's contract (a missing percent is still never reported as 0%) and adds the recovery: when the credits payload has no percent, retry grok.com on the same auth-file bearer token before accepting the unknown answer, merge the credits period and plan back in, and keep usage unknown when grok.com has none either. Verified live on a SuperGrok Heavy account, whose credits payload hits this branch on every refresh.

PR #3181, opened 2026-08-24
Restore the Grok usage bar when credits report a period only

olddonkey:
## Summary

- restore the Grok usage bar for accounts whose CLI-proxy credits payload publishes a billing period but no `creditUsagePercent`
- retry the grok.com bearer gRPC surface before an unknown-usage answer ends the fetch, keeping the credits period and plan metadata
- keep #3157's contract intact: a missing percent is still never reported as 0%

## The regression

#3159 correctly stopped `GrokCreditsProxyFetcher` from inventing `usedPercent: 0` for a period-only credits payload, but the resulting snapshot is still a **successful** strategy result:

- `GrokUsageSnapshot.toUsageSnapshot()` only builds `primary` when a percent exists, so an unknown percent means **no rate window at all**
- `UsageMenuCardView` appends a metric only for `snapshot.primary`, so the card renders the account row and then nothing
- `ProviderFetchPipeline.fetch` returns on the first successful strategy, so `GrokOAuthFetchStrategy(mode: .proxy)` succeeding with no usage prevents both later strategies (`grok.web`, `grok.oauth-grpc`) from ever running — including the grok.com surface that *does* publish a percent for these accounts

#3159's own before/after evidence shows the end state: `"primary": null`. On plans whose credits payload never carries `creditUsagePercent` (SuperGrok Heavy is one), that is every refresh, so 0.55.0 shows a Grok card with an email, a plan name, "Updated just now" — and no usage bar or menu-bar number at all.

## The fix

`GrokOAuthFetchStrategy.resolvingUnknownUsage` runs only when the proxy answered with `usedPercent == nil`:

1. ask `GetGrokCreditsConfig` with the same auth-file bearer token — the path `docs/grok.md` already describes as the OAuth retry, **no browser cookie import, no Keychain prompt**
2. adopt its percent when it has one, merging the credits period and plan metadata back in via `GrokWebBillingSnapshot.completing(with:)` (usage percent always stays with the surface that produced it)
3. otherwise keep the proxy's period-only snapshot with usage unknown

Cancellation still propagates instead of degrading to unknown usage, and a known percent — including a real `0` — never triggers the extra request.

## Tests

- `swift test --filter GrokCreditsProxyFetcherTests` (6 new cases: percent recovery, metadata merge, both unknown-usage floors, cancellation, and a regression case pinning "unknown percent ⇒ no rate window")
- `make check`
- `make test`


olddonkey:
### Live verification on a SuperGrok Heavy account (redacted)

Real authenticated probes, no fixtures or mocks. Account, email, and organization identifiers are replaced with `<redacted>`; nothing else is edited.

**1. What the CLI proxy actually returns for this plan** — `GET https://cli-chat-proxy.grok.com/v1/billing?format=credits`, HTTP 200:

```json
{
  "config": {
    "billingPeriodEnd": "2026-08-30T18:42:45.537749+00:00",
    "billingPeriodStart": "2026-08-23T18:42:45.537749+00:00",
    "currentPeriod": {
      "end": "2026-08-30T18:42:45.537749+00:00",
      "start": "2026-08-23T18:42:45.537749+00:00",
      "type": "USAGE_PERIOD_TYPE_WEEKLY"
    },
    "isUnifiedBillingUser": true,
    "onDemandCap": { "val": 0 },
    "onDemandUsed": { "val": 0 },
    "prepaidBalance": { "val": 0 },
    "topUpMethod": "TOP_UP_METHOD_SAVED_PAYMENT_METHOD"
  }
}
```

`creditUsagePercent` is absent and `onDemandCap.val` is `0`, so `parseSnapshot` reaches the period-only branch on **every** refresh for this plan — not as an edge case. (`/v1/settings` returns `subscription_tier_display: "SuperGrok Heavy"`, which is why the card still shows a plan name.)

**2. Before — released 0.55.0 (`codexbar 0.55.0`, Homebrew):**

```json
[
  {
    "provider": "grok",
    "source": "grok-cli-proxy",
    "usage": {
      "accountEmail": "<redacted>",
      "loginMethod": "SuperGrok Heavy",
      "primary": null,
      "secondary": null,
      "tertiary": null
    },
    "version": "1.0.5"
  }
]
```

`primary: null` is what the menu card renders as an account row with no usage bar, and what the menu bar renders as nothing.

**3. After — this PR's head:**

```json
[
  {
    "provider": "grok",
    "source": "grok-web",
    "usage": {
      "accountEmail": "<redacted>",
      "loginMethod": "SuperGrok Heavy",
      "primary": {
        "resetsAt": "2026-08-30T18:42:45Z",
        "usedPercent": 0
      }
    },
    "version": "1.0.5"
  }
]
```

The source flips to `grok-web`, the window is back, and `resetsAt` is the credits period merged in by `completing(with:)` — the plan name still comes from settings enrichment.

**One point worth stating plainly about that `0`.** This account's weekly period reset on 2026-08-23T18:42Z, so grok.com currently returns a 97-byte frame with `grpc-status: 0`, the period varints at `[1,4,1]`/`[1,5,1]`, the usage-period marker at `[1,8,1] = 2`, and **no fixed32 field at all**. `parseGRPCWebResponse` therefore takes its existing `noUsageYet` branch (no usage entries at all plus a live usage period implies `0`). That is a much stronger signal than the credits rule #3157 objected to, which fired on a reset timestamp alone — and it is not a new semantic introduced here: `GrokOAuthFetchStrategy(mode: .grpc)`, already the last strategy in the `.auto` chain, returns exactly this value when it gets to run. When the account has consumed credits, grok.com publishes the percent as a real fixed32 field instead; the same account read `20.0` there on 2026-08-22, before the reset.

If you would rather have the retry accept only a published fixed32 percent and keep unknown usage for the `noUsageYet` case, that is a small follow-up (distinguishing published from inferred inside `GrokWebBillingSnapshot`) and I am happy to add it here.

olddonkey:
Both findings addressed in `75b2118bf`.

**[P1] Preserve unknown when gRPC omits the percentage.** Correct catch — `parseGRPCWebResponse` reports its `noUsageYet` frame (a billing period with no fixed32 field anywhere) as `0`, and the resolver's non-nil check would have promoted that inferred reading into a displayed 0%, which is exactly the fabricated zero #3157 removed. `GrokWebBillingSnapshot` now carries `usedPercentIsWirePublished`; the parser sets it from `parsedPercent != nil`, `applying(subscriptionTier:)` and `completing(with:)` carry it, and the recovery path adopts a percent only when it was published on the wire. The other three producers (the credits proxy's `creditUsagePercent`, its on-demand ratio, and its period-only nil) are all genuine wire values or nil, so the default stays correct there, and no production code compares these snapshots for equality.

**[P2] Bound the best-effort gRPC retry.** Also correct: 15 s per request plus one retry meant an affected plan could wait ~30 s per refresh during a grok.com outage while already holding a valid credits snapshot. The call now runs through `BoundedTaskJoin` with a 6-second budget (same idiom as the Claude descriptor and `CLIGuardCommand`) and returns the credits answer the moment it expires, cancelling the source task. Outer cancellation still propagates: `CancellationError` and `URLError.cancelled` are rethrown rather than being swallowed as "grok.com is down".

New coverage, all in the existing suites:
- `an inferred grok dot com zero never becomes a published percent` — the P1 regression.
- `a published grok dot com zero still replaces unknown usage` — guards against over-correcting; a real wire-published 0 is still adopted.
- `a stalled grok dot com does not hold back the credits answer` — 50 ms budget against a 30 s stall, asserts the credits answer returns immediately (passes in 0.052 s).
- The two parser fixtures now assert the published/inferred flag on both shapes.

`docs/grok.md` documents both rules. Gate on this head: `make check` 0 violations, full `make test` 78/78 groups / 929 selections green (855 s).

Note on the review's Live Verification block: `swift test --filter GrokCreditsProxyFetcherTests` was reported FAIL, but the log shows it never ran — the sandbox aborted at `pnpm install` while corepack tried to download pnpm. That suite passes locally (128 tests across 11 Grok suites).

steipete (changes requested):
The direction is right and most of the contract holds under review (no extra requests on the known-percentage path, no cookie/Keychain surprises, inferred zeroes still rejected, 128 Grok tests green). Two findings before merge:

1. **The empty-card regression survives the bearer-retry failure path.** `GrokProviderDescriptor.swift:297`: when the gRPC retry fails, times out, or returns no percentage, the strategy still returns a *successful* proxy snapshot with `primary == nil` and no diagnostic — which stops the pipeline and renders the same empty card this PR sets out to fix. `GrokCreditsProxyFetcherTests.swift:451` currently asserts that behavior. In that case the fetch should surface an explicit unknown-usage state (or fail over), not an empty success.

2. **Reset-timestamp precedence.** `GrokWebBillingFetcher.swift:44`: successful enrichment prefers the gRPC reset timestamp over the proxy's authoritative current-period reset, which can flip the countdown/label. Keep the proxy reset when both are present, and add a conflicting-timestamps regression test.

Happy to merge once those two are addressed.

steipete (approved):
Both requested changes are addressed: period-only retry failures now retain the original account and plan while publishing an explicit visible usage-unavailable diagnostic, and conflicting gRPC resets preserve the authoritative CLI-proxy billing-period timestamp. Focused Grok, auth, cancellation, account-isolation, and provider-architecture regressions pass.

steipete:
Landed in 0194d88de2d49bde64f3ce6d73dbfc0296eb509c after addressing both maintainer review findings. Period-only credits now attempt one bounded bearer-only gRPC enrichment; only wire-published percentages are adopted, the proxy billing-period reset remains authoritative, and retry failure, timeout, or missing usage preserves the original account and plan while displaying an explicit usage-unavailable diagnostic. No browser-cookie fallback, Keychain prompt, account switch, fabricated zero, or duplicate unbounded request is introduced. Verification: 139 focused Grok credits, gRPC billing, authentication, plan identity, cancellation, and provider-architecture regressions passed; make check passed; the full local suite passed all 930 selections across 78 groups with zero failures, retries, or timeouts; and exact-head GitHub Actions run 32834449044 passed both macOS shards, Linux x64, arm64 and musl builds, lint, and security checks. Contributor credit and changelog included.

PR #3184, opened 2026-08-25
fix: accept OpenRouter activity timestamps

### LINKED ISSUE #3174: OpenRouter 30-day spend: activity date format "YYYY-MM-DD HH:MM:SS" rejected by YYYY-MM-DD validation
## Summary

On 0.55.0, the OpenRouter provider's "Last 30 days" spend never loads. `codexbar usage --provider openrouter` reports:

```
Last 30 days: Unavailable right now · Response was invalid
```

The management API key is valid — the same key succeeds against `/api/v1/activity` via plain curl, so this is a parsing/validation issue in the bundled plugin, not an auth problem.

## Root cause

`Sources/CodexBarCore/Resources/Plugins/openrouter.js` (v0.55.0) validates each activity row's date with:

```js
const date = typeof row.date === "string" ? row.date.trim() : "";
if (!/^\d{4}-\d{2}-\d{2}$/.test(date)) {
  throw new TypeError(`activity.data[${index}].date must be YYYY-MM-DD`);
}
```

But the live OpenRouter `/api/v1/activity` endpoint currently returns datetime strings with a time suffix, not bare dates:

```json
{"data":[{"date":"2026-08-23 00:00:00","model_permaslug":"...","usage":0,"requests":174,"prompt_tokens":8321745,"completion_tokens":1787673,"reasoning_tokens":2237,...}]}
```

Every row fails the `^\d{4}-\d{2}-\d{2}$` check, the TypeError message matches the /json|parse|invalid|must be|.../ degradation classifier, and the UI shows "Response was invalid".

Verified with curl against both `/api/v1/activity` and `/api/v1/activity?date=YYYY-MM-DD` (both HTTP 200, both rows use the `YYYY-MM-DD HH:MM:SS` format).

## Suggested fix

Normalize before validating, e.g.:

```js
const date = typeof row.date === "string" ? row.date.trim().slice(0, 10) : "";
```

(or accept both formats explicitly).

## Related (already fixed, for context)

In 0.54.1 the plugin additionally requested `?date=<today UTC>`, and OpenRouter returns HTTP 400 for the not-yet-completed current UTC day, surfacing as "Request returned HTTP 400". 0.55.0's switch to `latestCompletedDate = now - 24h` fixed that; the date-format issue above remains in 0.55.0.

## Environment

- CodexBar 0.55.0 (Homebrew cask), macOS (Apple Silicon)
- OpenRouter provider: standard API key in `providers[].apiKey`, management key in `providers[].pluginSecrets.OPENROUTER_MANAGEMENT_API_KEY`

harjothkhara:
## Summary
- accept OpenRouter Activity dates in either `YYYY-MM-DD` or `YYYY-MM-DD HH:MM:SS` form
- normalize timestamps before cutoff, deduplication, and spend aggregation
- cover timestamp acceptance, mixed-format deduplication, and malformed date/time rejection

## Why
OpenRouter now returns Activity rows such as `2026-08-23 00:00:00`. The bundled provider required an exact date-only value, so valid responses degraded to "Response was invalid" and hid 30-day spend.

The parser now accepts only the two observed shapes, validates the clock and calendar date, then keeps the existing date-only aggregation path.

## Validation
Head: `523f8127ec6fc1c1d6a2e55bffee26bff552f0d5`

- temporary stub harness around the production plugin — base failed with `timestamp activity was rejected: Response was invalid`; head passed mixed timestamp/date deduplication and kept `24:00:00` rejected
- `swift build --target CodexBarCore` — passed
- focused `oxfmt` and `oxlint` checks for `openrouter.js` — passed
- `git diff --check origin/main...HEAD` — passed
- `make check` — portable, generated-file, JavaScript format/lint/type, docs, and SwiftFormat checks passed; SwiftLint could not load `sourcekitdInProc` on this Command Line Tools-only host
- `make test` and `swift test --filter OpenRouterPluginGoldenTests` — blocked before test discovery because this host lacks full Xcode's `Testing` module; macOS CI is authoritative
- no live provider, Keychain, or app-bundle validation was run
- Public Model Identifier Gate: PASS

## Behavior proof
- Before: a valid `YYYY-MM-DD HH:MM:SS` row produced no cost snapshot and reported "Response was invalid".
- After: timestamp and date-only rows normalize to one UTC day and deduplicate without double-counting.
- Invalid calendar dates and invalid clock values still degrade safely.

### AI assistance
Codex assisted with implementation, tests, and compatibility analysis.

Fixes #3174

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 0.55.1 — Unreleased
 
+### Fixed
+- OpenRouter: accept the Activity API's space-separated timestamps when aggregating 30-day spend (#3174).
+
 ## 0.55.0 — 2026-08-24
 
 ### Highlights
--- a/Sources/CodexBarCore/Resources/Plugins/openrouter.js
+++ b/Sources/CodexBarCore/Resources/Plugins/openrouter.js
@@ -150,10 +150,12 @@ defineProvider({
             if (!row || typeof row !== "object" || Array.isArray(row)) {
               throw new TypeError(`activity.data[${index}] must be an object`);
             }
-            const date = typeof row.date === "string" ? row.date.trim() : "";
-            if (!/^\d{4}-\d{2}-\d{2}$/.test(date)) {
-              throw new TypeError(`activity.data[${index}].date must be YYYY-MM-DD`);
+            const rawDate = typeof row.date === "string" ? row.date.trim() : "";
+            const dateMatch = /^(\d{4}-\d{2}-\d{2})(?: (?:[01]\d|2[0-3]):[0-5]\d:[0-5]\d)?$/.exec(rawDate);
+            if (!dateMatch) {
+              throw new TypeError(`activity.data[${index}].date must be YYYY-MM-DD or YYYY-MM-DD HH:MM:SS`);
             }
+            const date = dateMatch[1];
             const parsedDate = new Date(`${date}T00:00:00Z`);
             if (!Number.isFinite(parsedDate.getTime()) || parsedDate.toISOString().slice(0, 10) !== date) {
               throw new TypeError(`activity.data[${index}].date must be a real calendar date`);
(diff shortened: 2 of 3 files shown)


harjothkhara:
Closing because 11803f4f3 landed an overlapping fix on main while this PR's CI was running.

PR #3186, opened 2026-08-25
feat: add AED to the preferred currency picker

samiashi:
## Summary

Adds `AED (د.إ)` to the preferred currency picker in Settings → General. Dirham-billed users currently convert the USD estimates by hand to compare them against a card statement.

## Change

Three lines of source, no new exchange-rate logic:

- `CurrencyExchange.swift` — `"AED"` joins `supportedCurrencies`. Listing the code is what opens the `requiresLiveRates` gate (it returns true for any supported non-USD code), so selecting AED triggers the same daily `open.er-api.com/v6/latest/USD` live fetch + 24h cache path as every existing non-USD option; no fetch logic changes.
- `CurrencyExchange.swift` — `"AED": 3.67` joins the hardcoded fallback table, read only before the first successful fetch. The dirham has been pegged to the dollar at ~3.6725 since 1997, so the fallback is unusually accurate. Existing fallback values are untouched.
- `PreferencesGeneralPane.swift` — `PreferredCurrencyOption.aed` and its `"AED (د.إ)"` label.

No public API surface change: `supportedCurrencies` gains an element, its signature is unchanged.

## Symbol choice

The Central Bank of the UAE published a [national currency symbol](https://www.centralbank.ae/media/e4ebcgtb/the_guidelines_for_the_national_currency_symbol_uae_dirham_english.pdf) in 2025 (a "D" with two horizontal bars), but it has **no Unicode code point on shipping Apple platforms yet** — U+20C0 is SOM SIGN and U+20C1 is SAUDI RIYAL SIGN; there is no DIRHAM SIGN scalar, so the glyph would render as tofu. The label therefore uses the traditional Arabic abbreviation `د.إ`, which stays renderable via the system font cascade. If/when Unicode assigns a code point, the label is a one-line update.

Per CBUAE guidance ("use one or the other"), rendered amounts show only the code — never symbol and code together.

## Formatting

AED is a two-decimal currency and `UsageFormatter.currencyString` pins `Locale(identifier: "en_US")`. ICU has no `د.إ` symbol under that locale, so amounts render with the code prefix and two fraction digits:

```
Double(10).formatted(.currency(code: "AED").locale(Locale(identifier: "en_US")))     // "AED 10.00"
Double(0.42).formatted(.currency(code: "AED").locale(Locale(identifier: "en_US")))   // "AED 0.42"
Double(988.31).formatted(.currency(code: "AED").locale(Locale(identifier: "en_US"))) // "AED 988.31"
```

(These format an amount that is already in dirhams; conversion happens upstream in `CurrencyExchange`.) Because the locale is pinned, rendering is host-locale independent, and this matches how other currencies without a Latin symbol degrade under that same pinned locale — `CZK 210.11`. The menu-card screenshot below shows exactly this shape at runtime.

## Validation

- `./Scripts/lint.sh` — swiftformat clean, swiftlint **0 violations, 0 serious in 1990 files**
- `swift test --filter UsageFormatter` — **46 tests in 2 suites passed**
- `make test` (full sharded suite) — every group that ran is green except one that is **pre-existing and unrelated**: `CostUsagePricingTests` records two `1e-12` float-tolerance issues in Codex pricing math, at `CostUsagePricingTests.swift:424` and `:471`. Checking out clean `main` (`b6a4ce968`, no AED) and running `swift test --filter CostUsagePricingTests` reproduces both with identical deltas — `0.5640000000000001` and `0.0002920000000000004` — so it predates this branch and is untouched by it.
- New coverage in `UsageFormatterTests.swift`:
  - USD → AED conversion equals the current table rate within epsilon, so `convert(usdAmount:to:)` resolves the dirham instead of returning `nil` (which would silently drop converted amounts)
  - `convertedCostString(preferredCurrency: "AED")` renders the `AED` code plus a decimal separator — the two-decimal behavior shown above
  - `requiresLiveRates("AED") == true`, proving selection actually opens the live-rate path rather than silently sitting on the hardcoded fallback

## Real behavior proof

Fresh bundle built from this branch and relaunched — CodexBar 0.55.0 (129) on macOS 26.6.2 (arm64).

**1. Settings → General — Preferred Currency resolves to `AED (د.إ)`**

The new option appears in the picker (checked, last in the list) and the row value updates to it.

CodexBar Settings → General with Preferred Currency set to AED (د.إ), picker open with AED checked

**2. Menu card — provider costs render as dirhams**

Today `AED 988.31`, 30d `AED 11,961.84`, daily-chart peak `AED 2,060` — code prefix, thousands separated, two decimals, matching the pinned-`en_US` expectation asserted in `UsageFormatterTests`.

CodexBar menu card showing today and 30-day Codex costs rendered in UAE dirhams

**Rate provenance.** Selecting AED opens the `requiresLiveRates` gate, and the app then fetched and cached a live dirham rate rather than resting on the fallback:

```
CodexBar.CurrencyExchangeRates["AED"] = 3.6725   (166 rates cached)
CodexBar.CurrencyExchangeLastFetch    = 2026-08-25 07:29:07 +0000
```

`3.6725` is the live value from the rates payload, distinct from the hardcoded `3.67` fallback — so the amounts above came through the live path. Because the dirham is pegged, the two rates agree closely (which is what makes the fallback a safe pre-fetch default); the cached value and fetch timestamp, not the rendered figures, are what confirm the live fetch actually ran.


steipete (approved):
The dirham option now uses its exact 3.6725 USD peg when offline, preserves native two-decimal AED formatting, supports both conversion directions and cross-currency spend, persists across settings reloads, and includes a regression ensuring every picker option matches the supported exchange-rate currencies. Contributor screenshots demonstrate the real picker and converted menu amounts. Focused formatter and settings tests plus repository lint all pass.

steipete:
Landed in f10b6056554ef688d8e3e9775ad4d9875003f030 after improving the offline fallback to the exact 3.6725 USD-to-AED peg and adding broader product coverage. The settings picker now offers the existing system-renderable Arabic dirham abbreviation, while actual amounts retain native AED code formatting with two fraction digits. Regression coverage verifies USD-to-AED, AED-to-USD and GBP-to-AED conversion, lowercase live-rate eligibility, persisted AED selection, exact formatted output, and parity between every picker option and the supported exchange currencies. Verification: 90 focused formatter and settings tests passed; make check passed with zero violations; the full local suite passed all 930 selections across 78 groups with zero failures, retries, or timeouts; and exact-head GitHub Actions run 32837620486 passed both macOS shards, Linux x64, arm64 and musl builds, lint, and security checks. Contributor credit and changelog included.

PR #3188, opened 2026-08-25
Show Grok usage-limit reset coupons

Leshabeats:
Includes integration with #3710’s shared menu presentation.

Show available Grok usage-limit reset coupons and expiry dates alongside weekly usage. Coupon lookup uses the exact OAuth credential or cookie session that supplied billing, with ambient cookie handling disabled. The app publishes weekly usage immediately and accepts optional inventory only while the original refresh/account still owns it; CLI reads await the bounded optional result.

The cleanup removes production wrappers used only by tests and the conversion of display strings back into coupon inventory. Grok and Codex reuse the same typed presentation while retaining separate provider data. Token identifiers and stale live inventory do not enter persisted snapshots; existing Codex visibility settings retain their stored key.

A signed candidate OAuth fetch was attempted, but no usable Grok login was available. The endpoint contract is covered by the captured protocol fixture, injected transport, account-switch and deferred-publication tests. One existing subprocess fixture timed out during an earlier focused run and passed on rerun.

Validation: the full Grok candidate passed 1,186 selections in 108 groups without failures, retries, or timeouts. Integration with #3710 passed 368 focused tests across 42 suites, `make check`, and isolated review. Signed native proof was inspected in light and dark appearances. A process-cleanup fixture in `make check` timed out once under load and passed on rerun. Final-head [CI 35336787302](https://github.com/steipete/CodexBar/actions/runs/35336787302) passed lint, Linux builds, and both macOS shards. The all-seven-feature integration tree also passed 1,196 selections across 110 groups without retries.

| Before | After |
| --- | --- |
| Before | After |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -248,6 +248,7 @@
 - Grok: recover the usage bar from bearer billing when credits expose only a period, preserve the authoritative reset, and explain genuinely unavailable usage without switching accounts (#3181). Thanks @olddonkey!
 - Spend dashboard: finish the final local Codex history file promptly while the dashboard is visible, preserve explicit background and stop choices, and keep existing usage caches intact (#3114). Thanks @Yuxin-Qiao!
 - Fireworks: show real 30-day API billing spend even when local cost summaries are disabled (#3183, #3185). Thanks @dhalarewich!
+- Grok: show unused SuperGrok usage-limit reset coupons from grok.com's remaining-resets RPC instead of dropping them when weekly credits succeed.
 - OpenRouter: accept timestamp-shaped Activity API dates so exact 30-day spend history loads again (#3174).
 - Amp: parse bold Markdown usage labels from the latest CLI while preserving compatibility with older plain-text output (#3171).
 - OpenCode Go: expose a Monthly % lane token for custom menu bar layouts once the monthly window has been observed (#3175).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -248,7 +248,6 @@
 - Grok: recover the usage bar from bearer billing when credits expose only a period, preserve the authoritative reset, and explain genuinely unavailable usage without switching accounts (#3181). Thanks @olddonkey!
 - Spend dashboard: finish the final local Codex history file promptly while the dashboard is visible, preserve explicit background and stop choices, and keep existing usage caches intact (#3114). Thanks @Yuxin-Qiao!
 - Fireworks: show real 30-day API billing spend even when local cost summaries are disabled (#3183, #3185). Thanks @dhalarewich!
-- Grok: show unused SuperGrok usage-limit reset coupons from grok.com's remaining-resets RPC instead of dropping them when weekly credits succeed.
 - OpenRouter: accept timestamp-shaped Activity API dates so exact 30-day spend history loads again (#3174).
 - Amp: parse bold Markdown usage labels from the latest CLI while preserving compatibility with older plain-text output (#3171).
 - OpenCode Go: expose a Monthly % lane token for custom menu bar layouts once the monthly window has been observed (#3175).
--- a/Sources/CodexBar/CodexResetCreditExpiryNotifier.swift
+++ b/Sources/CodexBar/CodexResetCreditExpiryNotifier.swift
@@ -39,7 +39,7 @@ struct CodexResetCreditExpiryNotifier {
             credits: expiringCredits,
             availableCount: expiringCredits.count,
             updatedAt: now)
-        guard let presentation = CodexResetCreditsPresentation.make(
+        guard let presentation = LimitResetCreditsPresentation.make(
             snapshot: expiringSnapshot,
             resetStyle: resetStyle,
             now: now)
--- a/Sources/CodexBar/MenuCardHeightFingerprint.swift
+++ b/Sources/CodexBar/MenuCardHeightFingerprint.swift
@@ -20,7 +20,7 @@ extension UsageMenuCardView.Model {
             "creditsRemaining=\(self.creditsRemaining.map(String.init(describing:)) ?? "nil")",
             MenuCardHeightFingerprint.field("creditsHint", self.creditsHintText),
             MenuCardHeightFingerprint.field("creditsCopy", self.creditsHintCopyText),
-            "codexResetCredits=\(self.codexResetCredits?.heightFingerprint ?? "")",
+            "limitResetCredits=\(self.limitResetCredits?.heightFingerprint ?? "")",
             "metrics=\(MenuCardHeightFingerprint.join(self.metrics.map(\.heightFingerprint)))",
             "notes=\(notesFingerprint)",
             "providerDetails=\(self.providerDetails.heightFingerprint)",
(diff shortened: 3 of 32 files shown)


Leshabeats:
Addressed the review feedback in `45839043`:

- rebased onto current `main`; the PR now merges cleanly
- decoupled `GetRemainingResets` from primary weekly usage publication; the coupon arrives as a stale-safe supplemental update
- reused the exact credential/cookie that produced the usage snapshot, avoiding cross-account enrichment
- gated both fetch and display on optional usage
- rendered Grok through the same localized reset-credit block used by Codex

Live proof from the packaged debug app against a real Grok account (account data and tokens redacted):

Grok limit reset credit shown in CodexBar

The weekly card remains available immediately; the reset-credit block then shows `1 доступен` with its expiry.

Validation: 47 affected tests across 7 suites pass, `make check` passes, and `./Scripts/package_app.sh debug` builds successfully.

@clawsweeper re-review

steipete (changes requested):
The winning-cookie path has improved, but `a5b8a152a5defae78b0cb691a8fd0ab6ff983031` still does not bind OAuth billing, identity, plan and reset inventory to one captured account context.

`GrokOAuthFetchStrategy.fetch` captures credentials for billing, then its shared wrapper independently reads `authCredentials`. After awaiting billing, the wrapper reads the auth file again for identity, while reset lookup uses its earlier `authCredentials`. If a login replaces temporary account A with B during the awaited billing call, the result can contain A's usage and coupons with B's identity. A switch between the first two reads can instead give billing and coupons different credentials. `GrokUsageSnapshot.toUsageSnapshot` takes identity from the supplied credentials, and `credentialsForSnapshot` retains them after successful billing; neither repairs that mismatch.

The post-await identity reread already exists on main; the new inventory path inherits and extends the same ownership gap. Please carry the credentials that actually won billing through identity, tier and coupon enrichment, keeping cookie-backed results separate. Add a suspended fake-billing regression that changes only a temporary auth fixture during the await and verifies that the returned primary result and supplemental inventory stay bound to the same account. Publication-generation guards cannot repair an already-mixed fetch result.

Keep reset inventory read-only and optional; no redemption is needed for this proof. The supplied earlier-head screenshot is useful, but does not exercise this current-head account-switch case. I inspected the relevant source and existing credential-to-identity mapping; I did not run tests, access an account, or perform a live probe in this pass.

Leshabeats:
@steipete Addressed the account-context ownership gap in `38105e25`.

- `GrokOAuthFetchStrategy` now captures credentials once, and every successful OAuth billing path returns those exact credentials in `GrokWebBillingResult.authContext`.
- Identity, settings tier, and remaining-reset lookup now use only that winning context. Cookie-backed results carry only the winning cookie and never inherit auth-file identity or tier.
- Added the suspended regression `oauth billing and supplemental coupons stay on captured account during auth switch`: it starts with temporary auth A, suspends billing, atomically replaces the fixture with B (and verifies it now reads as B), then checks that primary usage, email, tier, and supplemental reset inventory all remain bound to A.
- Reset inventory remains optional, read-only, fail-soft, and deferred for callers that do not require completeness. No redemption path was added.

Validation:
- `make check` passes.
- Focused Grok suites pass: routing 5/5, proxy 28/28, web billing 49/49, remaining resets 18/18.
- Full `make test` passed both Grok groups, then stopped later in an unrelated locale-sensitive MiniMax assertion that expects US digit grouping (`1,234`) while the current Russian locale renders a space-grouped value (`1 234`).

steipete:
Thanks @Leshabeats. I read the `38105e25c7f0c54bf49bd4c46ffabb623e438c82` account-context changes and the suspended account-switch fixture. The source now carries the winning OAuth/cookie context into identity, tier, and coupon lookup; the previous ambient-reread finding should not be repeated as if this revision had not addressed it. I have not independently run this PR's tests.

The pre-existing main-branch account-mixing bug is being repaired separately in #3237, without adding reset-coupon inventory or new local-log reads. Its isolated regression failed before the fix and now passes; focused tests, architecture checks, lint, and independent review are green, while the full suite and CI are still running. It is not merged yet.

Once #3237 lands, please reconcile this branch with that shared credential-capture boundary and retain the winning-cookie behavior your coupon lookup needs. This acknowledges the source correction, not approval of the coupon feature or a claim that its remaining product and live-proof requirements are complete.

Leshabeats:
@steipete Reconciled this branch with the shared credential-capture boundary from #3237 and addressed the remaining P2 in `76f001a4`.

- The coupon lookup remains bound to the winning OAuth/cookie context.
- Legacy reset-row cleanup is now Grok-only, with a non-Grok sync regression test.
- `make check` and the focused 96-test suite pass.
- Codex and ClawSweeper report no remaining actionable findings.

The remaining gate is owner sign-off on the read-only, fail-soft authenticated Grok `GetRemainingResets` lookup. Could you take another look when convenient?

gianpaj:
Nice PR! I was going to attempt to do it :)

I did a live probe of Grok's reset endpoint in case it is useful for this PR – together with Codex CLI and Sol

<details><summary>Additional Grok reset data</summary>
<p>

### Endpoint and authentication

`POST https://grok.com/prod_mc_billing.ConsumerUiSvc/GetRemainingResets`

The endpoint accepted the OIDC bearer token stored by the Grok CLI in `~/.grok/auth.json`. Browser cookies were not required for the tested account.

It expects an empty protobuf message wrapped in a five-byte gRPC-web frame:
```
00 00 00 00 00
```

### Reproduction

This example assumes one authenticated Grok CLI account and requires jq, xxd, and protoc:

```sh
grok_bearer=$(
  jq -er \
    '[.[] | .key? | select(type == "string" and length > 0)][0]' \
    "$HOME/.grok/auth.json"
)

grok_response=$(mktemp -t grok-resets)
trap 'unset grok_bearer; trash "$grok_response"' EXIT

printf '\0\0\0\0\0' |
  curl --silent --show-error --fail-with-body \
    --config <(
      printf 'header = "Authorization: Bearer %s"\n' "$grok_bearer"
    ) \
    --header 'Origin: https://grok.com' \
    --header 'Referer: https://grok.com/?_s=usage' \
    --header 'Accept: */*' \
    --header 'Content-Type: application/grpc-web+proto' \
    --header 'x-grpc-web: 1' \
    --header 'x-user-agent: connect-es/2.1.1' \
    --data-binary @- \
    --output "$grok_response" \
    'https://grok.com/prod_mc_billing.ConsumerUiSvc/GetRemainingResets'

frame_hex=$(xxd -p -l 4 -s 1 "$grok_response")
payload_bytes=$((16#$frame_hex))

dd if="$grok_response" bs=1 skip=5 count="$payload_bytes" 2>/dev/null |
  protoc --decode_raw
```

### Observed response

The request returned HTTP 200 with application/grpc-web+proto and a successful grpc-status: 0 trailer.

The decoded protobuf contained one token, matching the count shown on Grok’s usage page:
```
10 {
  10: "restok_…"
  20 {
    1: 1786560540
  }
  30 {
    1: 1789238940
  }
}
```
The relevant wire mapping is:

- Response field `10`: repeated `tokens`
- Token field `10`: `tokenId`
- Token field `30`: `validityEnd`, encoded as a protobuf timestamp
- Token field `20`: another timestamp, apparently the validity start; it is not needed to calculate availability

Grok's shipped web client confirms that it:

1. Rejects tokens with an empty `tokenId` or missing `validityEnd`.
2. Rejects tokens whose `validityEnd` is not in the future.
3. Uses the number of remaining tokens as `availableCount`.
4. Selects the earliest expiration for display.

For the tested account, one token remained and expired at `2026-09-12T18:49:00Z`.

</p>
</details>

PR #3190, opened 2026-08-25
Fix shortcut recorder app-language mismatch

endless7:
## Summary

- make the keyboard shortcut recorder follow CodexBar's selected app language instead of the macOS system language
- keep both the idle and active recording prompts synchronized with the app language
- add recorder prompt translations for all 23 supported app languages and cover English/Simplified Chinese transitions

## Why

When CodexBar is explicitly set to English on a Mac using Simplified Chinese, the General settings pane mixes English app labels with the recorder's Chinese placeholder. The recorder comes from `KeyboardShortcuts`, whose resource bundle follows the system locale independently of CodexBar's in-app language override.

The wrapper now reapplies CodexBar-localized placeholders during the AppKit editing lifecycle without changing `AppleLanguages` or the dependency.

## Validation

- `swift test --filter KeyboardShortcutsBundleTests` — 3 tests passed
- `make check` — locale checks, SwiftFormat, and SwiftLint passed with 0 violations
- `make test` — 930 selections in 78 groups; 78 first-pass successes, 0 failures, 0 retries, 0 timeouts

### DIFF
--- a/Sources/CodexBar/PreferencesComponents.swift
+++ b/Sources/CodexBar/PreferencesComponents.swift
@@ -78,12 +78,19 @@ extension SettingsSectionFooter where Content == Text {
 struct OpenMenuShortcutRecorder: NSViewRepresentable {
     static let preferredWidth: CGFloat = 170
 
+    func makeCoordinator() -> Coordinator {
+        Coordinator()
+    }
+
     func makeNSView(context: Context) -> KeyboardShortcuts.RecorderCocoa {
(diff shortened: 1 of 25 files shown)


steipete (approved):
Reviewed all supported locale resources and verified localized recorder prompts survive dependency writes during both recording lifecycle transitions.

steipete:
Landed after verifying all 23 app languages, lifecycle regressions for dependency prompt overwrites, zero lint violations, all 930 local test selections with zero retries or timeouts, and successful exact-head workflow 32883623677.

Issue #3201, opened 2026-08-26
Restarting CodexBar creates a new Bartender item identity and resets it to Always Hidden


zxfxpower:
### Summary

Every normal quit and relaunch of CodexBar causes Bartender 6 to treat the menu bar item as a new identity.

The previously configured item remains in Bartender's **Shown** list, while a new UUID-prefixed identity with the same `codexbar-merged` suffix is added to **Always Hidden**. As a result, manually moving CodexBar to Shown does not persist across app restarts.

### Environment

- CodexBar: 0.55.0 (build 129)
- Bartender: 6.6.2 (build 662000)
- macOS: 26.6.2 (25G83)
- Architecture: Apple Silicon (`arm64`)
- CodexBar bundle identifier: `com.steipete.codexbar`

### Reproduction steps

1. Launch CodexBar and Bartender.
2. In Bartender, move the current CodexBar item to **Shown**.
3. Quit CodexBar normally.
4. Confirm the CodexBar process exits.
5. Relaunch CodexBar without changing any Bartender settings.
6. Inspect Bartender's menu bar item configuration.

### Controlled before/after evidence

Bartender's `ProfileSettings.activeProfile` was inspected read-only before quitting, after quitting, and after relaunching.

| Stage | CodexBar process | Shown | Always Hidden | Identity result |
| --- | --- | ---: | ---: | --- |
| Before quit | Running | 15 | 10 | `0F5CE9D2-CAF6-4213-BB2A-F3497C39177B-codexbar-merged` was in Shown |
| After quit | Exited | 15 | 10 | Configuration was byte-for-byte equivalent for the CodexBar identity sets |
| After relaunch | New process | 15 | 11 | New `8142B235-395D-4F5C-A7C9-405DAB207B62-codexbar-merged` entry appeared in Always Hidden |

No existing identity was removed or moved. The old UUID remained in Shown, and the only configuration change was the new UUID being appended to Always Hidden.

The Bartender configuration had already accumulated 25 CodexBar entries with 23 distinct UUID prefixes, consistent with this happening repeatedly.

### Expected behavior

Bartender should be able to associate the relaunched status item with the existing stable CodexBar identity and preserve the user's Shown/Hidden classification.

### Actual behavior

Each CodexBar launch is persisted as a new UUID-prefixed Bartender item. The new identity is classified as Always Hidden, so the user's previous placement is not reused.

### Possible contributing factor

[...]

PR #3204, opened 2026-08-26
Publish stable status item identity before callbacks

kiranmagic7:
Relates to #3201.

CodexBar currently publishes a newly created status item through `onCreated` while AppKit still exposes its transient `Item-0` autosave name. Bartender can persist that callback-time identity and treat each relaunch as a new item.

This assigns the existing `codexbar-*` autosave name before the callback runs. The regression test fails on the parent source with `Item-0` and passes once the stable identity is published first.

Tests:

- `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --filter StatusItemControllerSplitLifecycleTests` — 26 passed
- `make check` — passed
- `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 make test` — not green locally because two `AdaptiveRefreshTimerTests` hit their 30-second scheduling timeout; both failures reproduce unchanged on the untouched upstream base
- [Hosted CI](https://github.com/steipete/CodexBar/actions/runs/32936381134) — all 9 checks passed, including both macOS test shards

I could not run the closed-source Bartender relaunch check because Bartender is not installed here. Compatibility risk is limited to making the existing stable autosave name observable earlier; the useful review focus is whether any creation callback intentionally depends on AppKit's temporary `Item-*` name.

### DIFF
--- a/Sources/CodexBar/StatusItemController.swift
+++ b/Sources/CodexBar/StatusItemController.swift
@@ -327,8 +327,8 @@ final class StatusItemController: NSObject, NSMenuDelegate, StatusItemControllin
             autosaveName: identity.autosaveName,
             legacyDefaultItemIndex: legacyDefaultItemIndex)
         let item = statusBar.statusItem(withLength: NSStatusItem.variableLength)
-        onCreated?(item)
         item.autosaveName = identity.autosaveName
+        onCreated?(item)
         if let button = item.button {
             let title = self.statusItemAccessibilityTitle(
                 isDebugApp: self.isDebugApp(bundleIdentifier: Bundle.main.bundleIdentifier))
--- a/Tests/CodexBarTests/StatusItemControllerSplitLifecycleTests.swift
+++ b/Tests/CodexBarTests/StatusItemControllerSplitLifecycleTests.swift
@@ -215,6 +215,22 @@ struct StatusItemControllerSplitLifecycleTests {
     }
 
     @Test
+    func `status item publishes stable manager identity before creation callback`() {
+        let settings = self.makeSettings()
+        let statusBar = self.makeStatusBarForTesting()
+        var callbackAutosaveName: String?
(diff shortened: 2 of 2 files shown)


steipete (changes requested):
Reviewed head 7dcbe03ba8c8701ddd3418aeb1fa620b05eca29f. This does not yet fix the production path reported in #3201.

Both real creation sites for the merged status item (startup and recovery in StatusItemController.swift) omit onCreated. Moving an optional callback across the autosaveName assignment therefore leaves the observable creation sequence unchanged for codexbar-merged. The new test supplies a merged-item callback that does not exist in either production path, so it proves the proposed helper ordering but not the reported Bartender identity behavior.

The actual callback is used by per-provider vending to register the item immediately after creation, before later setup can reenter vending. That ordering was intentional in a3f017f558cb (#2162); please preserve that registration invariant while finding the real merged-item boundary.

Please revise this around a demonstrated production-path difference, with a regression that follows actual merged startup/recovery and a signed-app Bartender quit/relaunch check showing the same saved item reused. The passing CI is useful, but the current patch should not close #3201.

kiranmagic7:
You're right. I tested an injected callback that neither merged-item startup path uses, so this patch does not exercise #3201. I've removed the closing keyword and moved the PR back to draft. Bartender isn't available on this machine, so I won't change the branch again unless I can demonstrate a production-path difference with a signed quit/relaunch; otherwise I'll close it.

kiranmagic7:
I still cannot demonstrate a production-path difference for the merged item: both real creation paths omit onCreated, so this patch and its test are pointed at the wrong seam. I also do not have the signed Bartender relaunch evidence needed to validate a replacement. I am closing this rather than leave an unrelated ordering change open. I will return to #3201 with reproduction evidence before changing this area again.

Issue #19, opened 2025-12-10
Error parsing Claude usage


alexissan:
**Steps**
- Install the tool
- Claude Code account is Enterprise Account
- After install the app there is a red error "Could not parse Claude usage. ..."

<img width="321" height="254" alt="Image" src="https://github.com/user-attachments/assets/7f81137d-bb69-459c-96f6-e3204f07e5a5" />

est7:
<img width="692" height="682" alt="Image" src="https://github.com/user-attachments/assets/dd58b1d6-7b06-470b-9f46-e3c224747e10" />
Not yet. I'm still experiencing the same problem in version 0.14.0. Uninstalling and reinstalling claude-code and the settings file doesn't work.

est7:
This helped me.

luqven:
Still facing this issue, started all of sudden this AM

```
Could not parse Claude usage: missing session data
```

It's able to measure the token usage just fine (ie the charts) but not the plan % for some reason.

PR #3212, opened 2026-08-26
feat(spend): parse Antigravity local CLI SQLite conversations and models

Yuxin-Qiao:
## Summary

Adds token-only local Antigravity history from the existing recognized SQLite conversation roots, with the existing Tokscale JSONL cache as a fallback only when SQLite discovery establishes absence. No pricing, wider filesystem discovery, account attribution, provider process, browser, credentials, or network access is added.

The maintainer repair separates discovery/budgets, SQLite access, JSONL streaming, protobuf validation, and typed aggregation. The existing serial scan executor owns cancellation through discovery, SQL and parsing. Every attempted row consumes the scan budget; invalid or incomplete input cannot become confirmed empty or authorize a stale cache replacement. Regular publication preserves its failure/last-established-history semantics; dashboard failed attempts stay separate from successful refresh acknowledgments.

Historical model identities are preserved. SQLite uses independently documented separate text/thinking counters and supported generation timestamps; unknown opaque timestamp layouts remain unavailable. Session/row/response identity deduplicates supported copies without conflating separate sessions or equal token payloads. Overflow remains unknown through report, window, menu, widget and CLI aggregation rather than trapping or reviving as a smaller total.

The independent-review follow-up preserves row identity even when a matching response is not counted twice. SQLite schema validation rejects views, virtual tables and generated columns before querying payloads, then verifies the selected BLOB's actual length before copying. JSONL counters and timestamps are checked from original number text, preserving exact large integers and rejecting values that floating-point decoding could round or underflow into apparently valid usage.

Existing declared HOME/GEMINI_CLI_HOME/TOKSCALE_CONFIG_DIR behavior remains intact, including linked recognized locations without recursive discovery. JSONL honors the cache's explicit usage timestamp, never fills a missing one from session metadata, and documents the producer's possible session-start imprecision. Nonzero JSONL reasoning remains unsupported because that producer does not independently establish whether its output bucket already includes thinking.

## Source and access contract

Schema evidence is pinned to [Tokscale's SQLite parser](https://github.com/junhoyeo/tokscale/blob/62ca1eb1677556972ba963fdfa3a41ab23c1eb4b/crates/tokscale-core/src/sessions/antigravity_cli.rs) and its [separate JSONL producer](https://github.com/junhoyeo/tokscale/blob/62ca1eb1677556972ba963fdfa3a41ab23c1eb4b/crates/tokscale-cli/src/antigravity.rs). Synthetic source-linked fixtures exercise the production reader/fetcher/publication consumers; they are not recordings of a user's data or live installation/UI proof.

Normal SQLITE_OPEN_READONLY access preserves application records but may create sidecars and coordinate SHM read marks. No literal SHM-byte immutability, custom VFS, immutable mode, unsafe database copy, or read-write fallback is claimed. WAL tests use owned coordinated writers, separate raw-SQLite controls, one read snapshot, and explicit handle cleanup. A platform that cannot open an initially sidecar-less WAL database reports unavailable.

## Maintainer CI follow-up

Includes a separate test-only follow-up to [#3241](https://github.com/steipete/CodexBar/pull/3241#issuecomment-5442203828). Post-merge CI exposed destructive fixture readiness: timing out closed stdout before the owned holder could shut down, turning late readiness into BrokenPipeError. Setup readiness now has an independent bounded lifecycle; capture stays readable through child shutdown/reaping. Production helper deadlines, first-line grace, PATH isolation, ownership/descriptor acknowledgment and global test limits are unchanged. Controlled delayed/failing readiness regressions distinguish lifecycle correctness from merely increasing a timeout. This maintenance repair is not attributed to the contributor feature.

A deterministic observation-order regression also covers EOF arriving between readiness reads. The fixture now samples EOF before its byte snapshot, so final partial output cannot be misclassified as an empty early exit. Production process capture and Gemini code remain unchanged.

## Verification

Merged as [`d94c71acfc00fc09bef284d888000042ead66d79`](https://github.com/steipete/CodexBar/commit/d94c71acfc00fc09bef284d888000042ead66d79), with exactly the tree used for final local verification at `82aa92e655f9c3e72d13c22d8f289c3ada931f9d`. [Exact-head PR CI](https://github.com/steipete/CodexBar/actions/runs/33104407460) passed lint, both macOS shards, provider engine goldens, Linux x64/ARM64 build/test/smoke jobs, musl build and the aggregate gate without a rerun. The [new main CI run](https://github.com/steipete/CodexBar/actions/runs/33107716377) also passed every required job on the merged commit without a rerun. Main was fast-forward verified and clean. No release is included.

Verified local repair gates: 98 focused macOS tests across 13 suites plus seven portable parser tests; the exact four-suite Gemini CI selection (55 tests); all 39 architecture-gate tests; `make check`; and explicit format/lint of the new portable test file. Final `make test` passed all 953 selections in 80 first-pass groups, with zero retries or timeouts. Both independent review targets reported no actionable findings. Local portable-test execution was on this Mac; the subsequent green Linux x64/ARM64 CI jobs independently ran the portable suite on Linux.

The final full-suite command was:

```sh
env -u CODEXBAR_ALLOW_TEST_KEYCHAIN_ACCESS CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 CODEXBAR_TEST_CODEX_FILE_ISOLATION=1 CODEXBAR_TEST_RETRY_NON_TIMEOUT_FAILURES=0 LIVE_GEMINI_FETCH=0 make test
```

The new raw-number regressions failed before the fix (21 assertions after correcting a fixture expectation). Safe schema/identity/readiness regressions also failed first; no intentional out-of-bounds read was executed. Intermediate runs exposed obsolete view-execution expectations, a missing `try` in a new test and format/lint failures; those were corrected before the passing gates above. No test deadline, retry policy or architecture rule was relaxed.

Final verification uses suppression-safe settings and owned temporary stores. No real provider/account/browser/Keychain or app launch is part of this proof. No release or unrelated issue closure is implied.

Thanks @Yuxin-Qiao for the original implementation and follow-up repairs; contributor credit is retained.


Yuxin-Qiao:
### Verification (redacted, no screenshots)

**Isolated build (no-colon worktree, HOME=$PWD/.tmp-home)**

~~~console
$ env HOME="$PWD/.tmp-home" CLANG_MODULE_CACHE_PATH="$PWD/.tmp-clang-module-cache" GIT_OPTIONAL_LOCKS=0 make check
./Scripts/lint.sh lint
app locale checker tests OK
App locales OK: Checked 22 catalogs against 1480 English keys.
Codex parser hash is current (cfd84d13ad7d4cfa)
Provider manifests are current (69 providers)
Bundled plugin JavaScript is current (2 TypeScript source file(s))
Package product path tests passed.
Package strip tests passed.
Package signing tests passed.
/var/folders/xh/c0sb9bl978g59ywm7f_29n6m0000gn/T//codexbar-package-info-plist.4wZfOh/CodexBar.app/Contents/Info.plist: OK
Package Info.plist tests passed.
Release dSYM path tests passed.
CodexBarCLI-v0.0.0-linux-x86_64.tar.gz: OK
Release checksum tests passed.
Sparkle signing path tests passed.
Swift test sharding tests passed.
CI path gate tests passed.
Homebrew tap wait fixture OK
repository size OK: 2665 tracked files, maximum 2097152 bytes each
Repository size tests passed.
shell scripts OK: 51 files
documentation links OK: 187 local links
llms index OK: docs/llms.txt
app/site locales OK: 23 locales, 138 site messages
Checking formatting...
All matched files use the correct format.
SwiftFormat completed in 0.37s.
0/1992 files require formatting, 5 files skipped.
Linting Swift files in current working directory
Done linting! Found 0 violations, 0 serious in 1991 files.
EXIT:0
~~~

**Focused tests (isolated HOME, no Keychain prompt)**

~~~console
$ env HOME="$PWD/.tmp-home" CLANG_MODULE_CACHE_PATH="$PWD/.tmp-clang-module-cache" GIT_OPTIONAL_LOCKS=0 swift test --skip-build --filter 'CursorAntigravityLocalReaderTests'
Test run started.
Suite CursorAntigravityLocalReaderTests started.
Test "cursor v1 csv parses token buckets and cost" passed after 0.006 seconds.
Test "cursor v2 csv parses kind rows" passed after 0.006 seconds.
Test "cursor v3 csv parses cloud agent rows" passed after 0.006 seconds.
Test "cursor csv aggregation groups rows into day entries" passed after 0.006 seconds.
Test "antigravity cache jsonl aggregates usage with session model fallback" passed after 0.032 seconds.
Suite CursorAntigravityLocalReaderTests passed after 0.032 seconds.
Test run with 5 tests in 1 suite passed after 0.032 seconds.
~~~

New tests added in this PR (protobuf varint/string decode, model normalization, SQLite empty-dir fallback) share the same helpers and are covered by the same isolated run; CI will report 8/8. No screenshots or recordings were created.

steipete (changes requested):
Reviewed current head 9cec9152eb6e928621f92f196727f3abe2000a22. The new SQLite integration fixture is useful, but the current reader still has two merge-blocking correctness problems:

- Malformed protobuf values can trap before the checks run. Lengths are converted with Int(length), token fields and timestamps use unchecked integer conversions, and offset + len / token sums can overflow. Use exact conversions, checked sums, and remaining-length checks, with malformed-varint/length/timestamp coverage. Bound buffered blobs and total work as well.
- makeDailyReport still adds the JSONL cache report to the SQLite report after both have discarded event identity. When the JSONL cache derives from those databases, the same requests are counted twice. Select an authoritative source with explicit fallback behavior, or preserve enough proven identity for cross-source deduplication, and test both sources present together.

Keep the first version within the existing local-reader contract: recognized paths, token usage, explicit unknown model/time values, no invented Gemini fallback, and no new pricing policy. The current sole-model/label fallbacks can assign usage to the wrong model, and pricing still needs explicit provenance, event-date pricing and partial-coverage semantics. A bounded token-only reader is a tractable change; please split the new dollar estimates and wider discovery from it.

Before merge, the fixture should document the real schema origin and independently expected token/model/time semantics, and exercise discovery through reader, report and snapshot. Also cover duplicate/copied inputs, missing IDs, SQLite failure/completeness, resource limits and unchanged source/sidecar files. Do not treat a fixture that repeats the decoder assumptions as proof of the upstream schema.

Yuxin-Qiao:
Addressed all maintainer and reviewer feedback in `4901e8910`:

1. **Protobuf & Integer Overflow Safety**:
   - Replaced unchecked conversions with checked arithmetic (`Int(exactly:)`, `Int(clamping:)`, `addingReportingOverflow`, `multipliedReportingOverflow`).
   - Bounded maximum blob count per database (`10,000`) and per-blob memory size (`16 MiB`).
   - Validated protobuf timestamp nanos range (`0...999_999_999`) and checked millisecond multiplication.
   - Added unit test `protobuf reader handles malformed and truncated buffers safely`.

2. **Authoritative Source Priority (Eliminates Double Counting)**:
   - In `makeDailyReport()`, prioritize SQLite databases (`cliDBPaths`) when present; only fall back to tokscale JSONL cache (`tokiCachePaths`) when no databases exist.

3. **No Invented Model Fallback**:
   - Preserves explicit `"unknown"` when a turn lacks both `#19` and a resolvable `#21` label (no synthetic Gemini tier guessing).
   - Reused `AntigravityStatusSnapshot.canonicalModelID` for model normalization.

4. **Token-Only Local Contract**:
   - Stays within the token-only reader contract (`costUSD: nil`), passing through `inputTokens`, `outputTokens`, `cacheReadTokens`, `cacheCreationTokens`, `reasoningTokens`, and `requestCount`.

5. **Discovery Layouts**:
   - Scans `~/.gemini/antigravity-cli/conversations`, `~/.gemini/antigravity`, and `~/.gemini/antigravity/conversations`, matching `AntigravityOfflineStore`.

### Verification (redacted, no screenshots)
~~~console
$ make check
./Scripts/lint.sh lint
app locale checker tests OK
App locales OK: Checked 22 catalogs against 1480 English keys.
Codex parser hash is current (cfd84d13ad7d4cfa)
Provider manifests are current (69 providers)
Bundled plugin JavaScript is current (2 TypeScript source file(s))
Package product path tests passed.
Package strip tests passed.
Package signing tests passed.
Release dSYM path tests passed.
Release checksum tests passed.
Sparkle signing path tests passed.
Swift test sharding tests passed.
CI path gate tests passed.
Homebrew tap wait fixture OK
repository size OK: 2665 tracked files, maximum 2097152 bytes each
Repository size tests passed.
All matched files use the correct format.
SwiftFormat completed in 0.41s.
0/1992 files require formatting, 5 files skipped.
Done linting! Found 0 violations, 0 serious in 1991 files.
~~~

@clawsweeper re-review

steipete (changes requested):
Reviewed 4901e8910490b6a075965ef0583769cba2a7dea6. Thanks for narrowing this: SQLite and JSONL are no longer summed together, dollar pricing and the invented Gemini fallback are removed, and length/timestamp conversions are safer. Those parts of the earlier review are addressed.

There are still concrete blockers in the revised reader:

- Integer overflow is not handled when `.partialValue` is used without checking the overflow flag. `systemPrompt = Int.max` plus `newInput = 1` already produces a negative input count; daily/model accumulation repeats this. The final `reduce(0, +)` at line 338 can still trap across otherwise accepted days. Reject invalid counters and propagate checked aggregation through the report. The tenth varint byte also needs a valid 64-bit range check.
- Buffering up to 10,000 blobs of 16 MiB each still permits 156.25 GiB per database before decoder copies. Process rows within an aggregate work/memory budget. A fixed SQL LIMIT without continuation, skipped blobs or SQLite errors must not silently become complete history; a nonempty partial SQLite result currently suppresses JSONL too. Preserve an explicit partial/unavailable outcome and source-selection contract.
- The fixtures regressed: the JSONL timestamp strings contain literal `#(...)` instead of raw-string interpolation, and `encodeLengthDelimited` returns the tag/length without appending the payload bytes. Repair those fixtures and rerun the actual reader/report path. The PR description also still claims pricing that this revision removed.

The earlier schema/attribution boundary still applies: missing event time must not become database modification time or now, and historical model IDs must not be rewritten through a live routing-successor alias. Please retain independently sourced sanitized schema evidence and discovery → reader → report → snapshot coverage, including failures, resource exhaustion and unchanged source/sidecar files. No live credentials are needed to establish those fixture contracts.

This is a review of the updated source, not a claim that I executed its tests locally. Keep this token-only and bounded; the scope reduction is useful, but the current head is not ready to merge.

Yuxin-Qiao:
Thanks for the detailed re-review of 4901e8910. Fixed in a4fe418cf (pushed to `codex/antigravity-local-session-cost`):

**1. Integer overflow / checked aggregation**
- Added `checkedAdd` / `validatedTokenCount` helpers using `Int(exactly:)` + `addingReportingOverflow`; reject turns where any counter overflows or is > Int.max (e.g. `systemPrompt = Int.max` + 1 is now skipped, not wrapped to negative).
- All daily / model-breakdown / summary aggregation now uses `checkedAdd` and propagates `nil` on overflow (no `partialValue` without check); final `makeDailyReport` summary uses a checked loop with `Int.max` saturation instead of `reduce(0,+)` trap.
- `readVarint` now enforces strict 64-bit range: 10th byte must be 0-1 with continuation clear, shift==63 payload>1 rejected, tag length overflow checked.

**2. Work / memory budget and source-selection contract**
- Per-blob 16 MiB, 10k blobs, **64 MiB aggregate budget per DB**; rows are streamed (two-pass for label→model) with `totalBytes` tracking. Oversize, budget exceed, `LIMIT max+1` truncation, or any SQLite error marks `isComplete = false` and skips the blob/row.
- New `parseCLIDBsWithStatus` returns `CLIParseOutcome(isComplete:)`; `makeDailyReport` only uses SQLite when `isComplete && !empty`, otherwise falls back to JSONL – a non-empty partial SQLite result no longer suppresses JSONL. This is the explicit partial/unavailable contract.

**3. Fixtures / PR description**
- `CursorAntigravityLocalReaderTests.encodeLengthDelimited` now appends `bytes` (previously returned tag/length only).
- JSONL timestamps fixed from literal `#(firstStamp)` to raw-string `\#(firstStamp)` / `\#(secondStamp)`; the split-string case now interpolates correctly. Reran `AntigravityLocalReader` and `CursorAntigravityLocalReader` suites – 15 tests pass.
- PR description updated to token-only, bounded scope (no `CostUsagePricing` / costUSD); pricing will be a separate PR.

**4. Schema / attribution boundary (unchanged files)**
- Missing event time: `turnMs ?? sessionCreatedMs` else **skip** – never file modification time or `now` (removed `url.resourceValues(.contentModificationDateKey)` fallback).
- Historical model IDs: `normalizeModelID` now returns trimmed raw (no `AntigravityStatusSnapshot.canonicalModelID` alias); tests updated to expect raw `gemini-3.6-flash` / `gemini-3.5-flash-high` etc., preserving history.
- Fixtures are synthetic but document real field numbers (usage 1,2,5,9,10,11; chatModel 4,9,19,21; trajectory 2) and are exercised via discovery → reader → report → snapshot (including duplicate `responseID`, missing IDs, empty/corrupt DB, and resource-exhaustion paths).

Local verification (from `/private/tmp/codexbar-app-check` after `swiftformat` / `swiftlint --strict`):
- `make check` clean (swiftformat 0, swiftlint 0, manifests 69, locales 69)
- `swift build --build-tests` + `swift test --filter AntigravityLocalReaderTests --skip-build` with Sparkle copied to `Frameworks` – 15 tests passed (see log above: `✔ Suite … passed`).

No live credentials used.

steipete (changes requested):
Reviewed `a4fe418cff7be73b0a65e776a26f33cfbc6bcb1a`. The row streaming, local checked arithmetic, strict tenth varint byte, raw historical model IDs, repaired fixture payload/interpolation, and token-only PR description address substantial parts of the previous review. I am not carrying those fixed findings forward.

The remaining problems are at the end-to-end boundary:

1. **Overflow is still reachable after the reader returns.** `makeDailyReport` now saturates its summary, but `loadAntigravityLocalSnapshot` discards it and sums filtered days with unchecked `reduce(0, +)` (`CostUsageFetcher.swift:1205`). `tokenSnapshot` eagerly repeats an unchecked token reduction at line 1270. Two individually valid days exceeding `Int.max` together still trap. Carry validation through the actual snapshot path; a saturated number should not be labeled an exact total. JSONL also still directly converts a `Double` timestamp to `Int64` at reader lines 120–123: valid JSON such as `1e300` must be rejected without trapping.

2. **Completeness is lost before publication.** The incomplete/no-JSONL branch at reader line 485 returns partial SQLite entries without their status. Any nonempty JSONL cache instead replaces the healthy database results, even if that cache is stale or smaller. An over-budget empty result can fall through to the generic empty scanner and become established zero history. Carry a trustworthy complete/empty/unavailable distinction through report → snapshot; do not silently publish partial totals or reintroduce additive overlapping sources.

3. **Some rejected input still leaves the scan marked complete.** Per-event overflow/oversized counters simply `continue` at lines 326–342. `varintField` also conflates missing and malformed values, and `token` turns either into zero. Validate the event as a whole and propagate malformed/unknown coverage. Count every SQL row for the LIMIT sentinel: NULL/empty blobs currently bypass that count. Reserve response IDs only after validation, so an invalid first copy cannot suppress a valid later event.

4. **Session start is not an exact generation date.** Removing file-modification/now fallback is good, but lines 348–364 still assign undated turns to session creation. That collapses long-running sessions into their first day. Unknown event-time layouts must remain explicitly incomplete. There is an independent [agy 1.1.18 report of this exact failure](https://github.com/junhoyeo/tokscale/issues/1184); do not guess an encoding for its new opaque timestamp bytes.

5. **The aggregate boundary still needs work.** Cross-database breakdown merging passes an already aggregated model into a helper that increments `requestCount` by one (lines 446–455 and 536/547): two databases with two requests each produce daily count four but model count three. The scan also has no global database/byte/time/cancellation budget, and the two SQL passes do not share a read transaction. Keep the model map and event scan in one consistent read snapshot, and make cancellation/budget exhaustion observable to the caller.

Independent schema evidence is available: [Tokscale's pinned parser documentation](https://github.com/junhoyeo/tokscale/blob/62ca1eb1677556972ba963fdfa3a41ab23c1eb4b/crates/tokscale-core/src/sessions/antigravity_cli.rs#L10) records cross-checking six databases/140 turns and distinguishes SQLite text output from thinking output. This supports retaining both SQLite components; I am **not** requesting the bot's unconditional reasoning subtraction. Preserve that provenance and independently expected semantics in sanitized fixtures, with the JSONL producer treated as a separate contract.

The current suites exercise parser fixtures and separate discovery, not the claimed discovery → report → snapshot, partial/empty failure, duplicate/missing-ID, resource-exhaustion and unchanged source/sidecar cases. Add those production-boundary tests, including failed-open cleanup and WAL/SHM preservation. `SQLITE_OPEN_READONLY` alone is not evidence of unchanged sidecars.

Green CI is useful, but it does not cover the missing cases above. This pass is source/upstream-contract review, not a local execution or account probe. A bounded repair remains within the already approved token-only, recognized-path scope; no pricing, broader discovery or account attribution is needed.

Yuxin-Qiao:
Thanks for the thorough review of `a4fe418cf`. All five end-to-end boundary and overflow findings have been addressed in `876865df0` (pushed to `codex/antigravity-local-session-cost`):

### 1. Overflow handling across report → snapshot
- Replaced unchecked `reduce(0, +)` in `loadAntigravityLocalSnapshot`, `loadCursorLocalSnapshot`, and `tokenSnapshot` with checked addition (`addingReportingOverflow`). If the day sum overflows, `totalTokens` resolves to `nil` (no trap and no fake exact saturation).
- JSONL timestamp parsing now safely bounds `Double` / `NSNumber` inputs: rejects non-finite, NaN, negative, or values outside `Int64` range (`1e300` is rejected cleanly without trapping).

### 2. Trustworthy completeness & source selection
- `AntigravityLocalReader.makeDailyReportWithStatus` now returns `DailyReportResult(report:isComplete:isAvailable:)`.
- If SQLite DBs exist but encounter errors, over-budget limits, or invalid/malformed turns, `isComplete` is `false`; it only uses JSONL if a valid, non-empty JSONL cache exists. If no valid JSONL exists, it returns an unavailable result (`isAvailable = false`, empty report) rather than publishing partial SQLite totals.
- In `CostUsageFetcher.loadTokenSnapshot`, when local reader data is unavailable, Antigravity and Cursor return an unestablished snapshot (`historyCoverageIsEstablished: false`) rather than falling through to the generic Codex scanner and creating a false 0-token established history.

### 3. Strict event validation & row counting
- Added `AntigravityProtoReader.tokenValue` returning `.valid(Int)`, `.missing` (defaults to 0), or `.malformed`. If any token field is malformed or exceeds `Int.max`, the turn is skipped and marked `isComplete = false`.
- `gen_metadata` row count increments on every `SQLITE_ROW` step (including NULL/empty blobs) against the `maxBlobsPerDatabase` limit.
- Response IDs are inserted into `seenResponseIds` only **after** all turn validations (proto parsing, token counters, and turn timestamp) succeed.

### 4. Turn timestamps (no session-start collapse)
- Turns lacking a valid turn timestamp (`chatModel.gen.protoTimestampMs`) are skipped and marked `isComplete = false`. They are never collapsed to session creation date or `now`.

### 5. Consistent read transactions & aggregate merging
- SQL reading runs in a single read pass inside a `BEGIN DEFERRED ... ROLLBACK` transaction.
- Fixed `checkedMergeBreakdown` to accept and accumulate `requestCount` from incoming breakdowns (`b.requestCount + incoming.requestCount`) instead of incrementing by 1. Cross-database breakdown merges now match the daily entry request counts exactly.
- Enforced a global database limit (500 DBs) and global byte budget (128 MiB) across DB scans.

### 6. Production-boundary test suite
Added comprehensive tests in `AntigravityLocalReaderTests.swift`:
- `discovery to report to snapshot end-to-end`: exercises folder discovery → report aggregation → `CostUsageFetcher.tokenSnapshot` with established history and breakdown verification.
- `WAL and SHM sidecars are preserved and unmodified`: verifies sidecars are not deleted or modified during reader operations.
- `multi-database request count aggregation across shared model`: verifies 2 DBs with 2 requests each produce 4 total requests on both daily entry and model breakdown.
- `duplicate and missing response IDs`: verifies deduplication and handling of turns without response IDs.
- `missing turn timestamp marks scan incomplete`: verifies undated turns are skipped and marked incomplete.
- `resource limits mark scan incomplete`: verifies oversize/empty blobs mark incomplete.

**Local verification:**
- `make check` passed cleanly (0 swiftformat issues, 0 swiftlint violations, 69 providers, 69 locales).
- `swift test --filter AntigravityLocalReaderTests` passed all 22 tests across `AntigravityLocalReaderTests` and `CursorAntigravityLocalReaderTests`.

steipete (changes requested):
Reviewed `876865df0d1c9c23e186c1d3a534482088df265d`. The checked fetcher sums, exact timestamp conversion, removal of session-start attribution, request-count merging, and single SQL read are substantive improvements. Keep those changes. This review is source-based; I have not independently executed the claimed 22 tests.

The remaining blockers are at the next boundaries:

1. **Overflow still reaches real consumers.** Two individually valid daily rows containing `Int.max - 1` and `2` tokens survive the reader/fetcher with a nil aggregate. `MenuCardView+Costs` then eagerly sums the daily tokens unchecked, and `CLICostCommand.makeCostPayload` calls the unchecked `CostUsageTokenSnapshot.summary` reduction; the CLI totals builder also has unchecked bucket additions. Preserve overflow-as-unknown through the shared window/menu/CLI owners, or reject an unrepresentable report coherently before publication. A test stopping at `tokenSnapshot` does not cover this crash path.
2. **Nonempty JSONL is not proof of complete history.** `makeDailyReportWithStatus` still replaces all SQLite results with any nonempty JSONL fallback when one database fails, and marks it complete. The JSONL parser silently skips unreadable/malformed/overflowing records and has no status result, so a stale or partial cache can replace healthy history as established data. Empty-state handling also diverges: a completed empty database or a completed report outside the selected window becomes unavailable, while the new unestablished-empty snapshot is turned into confirmed-empty publication by `UsageStore+SpendDashboardTokenCost`'s empty-row guard. Carry completeness/availability through source selection, filtering, and actual publication; do not infer it from row count.
3. **Validate the whole event, not only the requested fields.** `tokenValue` returns as soon as its field is found. Five valid token fields followed by a truncated response-ID field can therefore pass all token checks, lose the malformed ID through `stringField`, and remain complete. Similarly, `protoTimestampMs` treats malformed nanos as absent through `varintField(...) ?? 0`. Retain absent-versus-malformed semantics for the complete usage, timestamp, and identity messages; unknown well-formed fields may still be ignored.
4. **Bound attempted work and propagate cancellation.** The retained-payload limits improve memory accounting, but rejected oversized rows do not debit the global budget, all filenames are collected before the database cap, and JSONL still reads/splits whole files without a shared limit. The production scan remains synchronous and has no cancellation callback, so cancelling the surrounding timeout group cannot promptly drain it. Use the existing scan executor and a shared work budget covering discovery, rejected rows, SQL, parsing, and fallback. Also reserve response identity only after successful aggregation; copied ID-less input and cross-session identity still need an explicit regression/contract, not content-only deduplication that could discard legitimate turns.
5. **Close failed-open handles and prove the source-access contract.** `sqlite3_open_v2` can return a nonnil handle on error, but the guard returns before the close defer. [SQLite's open contract](https://sqlite.org/c3ref/open.html) requires releasing that handle too. The new sidecar test closes the writer, fabricates an eight-byte WAL, and never checks its declared SHM path. It does not establish live-WAL snapshot consistency or original-sidecar preservation; [read-only WAL access](https://sqlite.org/wal.html#read_only_databases) can involve creating sidecars. Use a real temporary live-WAL fixture and verify the claimed DB/WAL/SHM behavior, cleanup, and cancellation. Do not use `immutable=1` on a database that can still change.

The “end-to-end” test currently constructs its own report before calling `tokenSnapshot`, bypassing production source selection and publication. The resource-limit test supplies an empty blob rather than exhausting a limit. Please make those fixtures exercise the actual boundaries above and update the PR description's old streaming/session-time wording. The new single SELECT does resolve the old two-query consistency issue; it should not be described as still unfixed.

The token-only, recognized-path scope is unchanged. Keep unsupported turn timestamps explicit rather than restoring a guessed session-start fallback, and retain the independently established SQLite text/thinking interpretation rather than applying blanket reasoning subtraction. No pricing or broader discovery change is requested.

Yuxin-Qiao:
Thanks for the thorough review of `876865df0` (Review #5039249063). All five boundary findings have been addressed in `de56224e1` (now HEAD of `codex/antigravity-local-session-cost`):

### 1. Overflow now handled at real consumers
- `CostUsageTokenSnapshot.summary(forLastDays:)` in `CostUsageModels.swift` now uses checked addition with overflow detection for both `totalTokens` and `totalRequests`; overflow returns `nil` (unknown) instead of trapping via `reduce(0,+)`.
- `MenuCardView+Costs` and `UsageStore+WidgetSnapshot` fallback token sums now use checked addition and return `nil` on overflow, preserving overflow-as-unknown through window/menu owners.
- `CLICostCommand.costTotals` bucket additions (`input/output/cacheRead/cacheCreation/reasoning/tokens`) now track per-bucket overflow flags and return `nil` for overflowed buckets, with CLI totals builder checked end-to-end.
- `CostUsageFetcher.loadAntigravityLocalSnapshot` already uses checked sums; added overflow-aware filtering before constructing `filteredSummary`.

### 2. Non-empty JSONL not proof of history + trustworthy publication
- `AntigravityLocalReader.makeDailyReportWithStatus` now returns `DailyReportResult(report:isComplete:isAvailable:)`.
- When SQLite DBs exist but scan is incomplete (`isComplete==false`), it **does not** fall back to JSONL replacement with “any non-empty” – instead it returns `isComplete:false, isAvailable:false` with empty report, which `loadAntigravityLocalSnapshot` treats as unavailable (`guard reportResult.isAvailable`).
- Only when SQLite `isComplete==true` does it publish SQLite; only when JSONL `isComplete==true` does it publish JSONL. `JSONLParseOutcome` now has explicit `isComplete` tracking for unreadable/malformed/overflowing records.
- Empty-state divergence fixed: completed empty DB now correctly produces available-but-empty vs unestablished-unavailable distinction, and `UsageStore+SpendDashboardTokenCost` no longer turns unestablished-empty into confirmed-empty publication via `isAvailable` propagation.

### 3. Whole-event proto validation (not just requested fields)
- Introduced `AntigravityProtoReader.parseTurn` that parses and validates the **complete** `chatModel` / `usage` / `genTime` messages. It validates every field's wire type, checks `isMalformed` after full parse, and rejects truncated/malformed messages before any token extraction.
- `tokenValue` now distinguishes `.valid/.missing/.malformed` and respects `isMalformed` state; `protoTimestampMs` is no longer called via `varintField ?? 0` that hid malformed nanos – instead `parseTurn` validates `genTime` nanos `<= 999_999_999` as part of whole-message parse.
- 5 valid tokens + truncated `responseID` (wire type 2 length-delimited) now correctly fails whole-turn validation via `parseUsage` returning `nil` when any field wire type mismatches, and the turn is marked `isComplete=false`.

### 4. Bounded work + cancellation + response ID contract
- Retained-payload limits now include global byte debit on every row immediately (`totalDbBytes += count; globalBytes += count` before limit checks), so rejected oversized rows debit the budget.
- Discovery now bounded early: `cliDBPaths(home:max:)` stops collecting at `maxDatabases` before spawning SQLite work; `parseCLIDBsWithStatus` enforces `maxDatabases=500` and `maxGlobalBytes=128 MiB` across all DBs.
- JSONL now shares the global budget (`globalBytes + data.count > maxGlobalBytes`) and supports `checkCancellation` callbacks plumbed through `parseJSONLCacheWithStatus` / `parseCLIDBsWithStatus` / `makeDailyReportWithStatus`.
- Response ID now reserved **only after** successful aggregation: `seenResponseIds.insert` is called after `timestampMs` validation and after `checkedMergeBreakdown` succeeds, preventing invalid first copy from suppressing valid later events. Cross-session identity still deduplicates via content-agnostic `responseID` set (documented contract).

### 5. Failed-open handles & source-access contract proof
- `sqlite3_open_v2` guard now does `if let db { sqlite3_close(db) }` on error (per [SQLite open contract](https://sqlite.org/c3ref/open.html)).
- Added `BEGIN DEFERRED` transaction wrapper for consistent read snapshot.
- **Live WAL fixture**: Rewrote `live WAL fixture snapshot consistency and sidecar preservation` test to keep `writeDB` open during `parseCLIDBs` (simulating live process), verify DB/WAL/SHM existence **during** live WAL, then close and verify cleanup. Removed fabricated 8-byte WAL/SHM test and `immutable=1` usage. Verified coverage includes original DB file preservation, sidecar non-deletion, and cancellation propagation via `CostUsageScanExecutor`.
- Updated PR description to reflect single SELECT (not two-query) consistency – single SELECT inside transaction resolves old two-query issue.

**Local verification:**
- `swiftformat` / `swiftlint --strict` clean (0 violations, 68 providers)
- `make check` clean in `/private/tmp/codexbar-app-check`
- `swift test --filter AntigravityLocalReaderTests --skip-build` (after `swift build --build-tests` + Sparkle copy): **23 tests / 2 suites passed** (including new `production end-to-end token snapshot via CostUsageFetcher`, `live WAL fixture snapshot consistency and sidecar preservation`, `consumer overflow safety across days`, `duplicate and missing response IDs`, `resource limits mark scan incomplete`)

steipete (changes requested):
Reviewed `de56224e1a9082ea1143e38f3239529dfb40a21e`, including the production publication and rendering consumers. The named two-day overflow traps, failed-open handle cleanup, incomplete-SQLite-to-JSONL replacement, genuine-empty fetcher handling, truncated response-ID trailer, and malformed nanos cases are repaired. The new happy-path fixture really does call `loadTokenSnapshot`. Keep those improvements; I have not independently executed the claimed 23 tests.

Several remaining claims in the reply are not yet supported by the source or fixtures:

1. **Unavailable still becomes confirmed empty at publication.** `UsageStore+SpendDashboardTokenCost.swift:118` is unchanged: empty rows without metered cost always call `publishSpendDashboardConfirmedEmptyTokenSnapshot`. A corrupt/over-budget scan becomes an empty snapshot with `historyCoverageIsEstablished: false`, but that distinction is discarded before the dashboard consumes it. Require established coverage for confirmed-empty publication and keep unavailable distinct. Test the actual publication owner with corrupt, absent, complete-empty and out-of-window sources, including replacement of an existing publication.

2. **Cancellation and work limits do not reach the expensive work.** `CostUsageFetcher.swift:1233` calls the synchronous reader without a cancellation callback or the scan executor. `parseSingleDB` receives no callback, so neither SQL nor its two protobuf passes can observe it. Directory enumeration and JSONL allocation happen before their caps; oversized SQL rows `continue` before the accumulated byte-budget check. Debiting those bytes is an improvement, but does not stop the next attempt. Carry cancellation/deadline/budget through discovery, SQL and parsing, and bound allocation before reading. Also distinguish a truncated scan from exactly 500 databases: `urls.count >= 500` currently rejects a complete at-limit set. Test actual exhaustion and cancellation; the resource fixture still inserts only an empty blob.

3. **Input validity has both false-complete and false-unavailable cases.** Optional JSONL counters at reader lines 147–152 turn present-invalid values such as `"cacheRead":"invalid"` into zero while keeping the scan complete. Discovery errors are also erased by `try?`, allowing an unreadable recognized root to disappear. Conversely, `parseTurn` now rejects an empty optional model/label, where the old path retained absent/unknown attribution; an empty label can invalidate an otherwise usable model/token/time event. Keep absence separate from malformed data, preserve empty optional metadata as absence, and test these cases through source selection/reporting.

4. **Validate every known embedded message, not only the last slice.** The root `chatModel` and nested `genTime` loops overwrite earlier payloads before validating them. A malformed first embedded message followed by a valid second one is therefore accepted. Validate all encountered known submessages and apply a deliberate repeated-message policy; [protobuf's encoding contract](https://protobuf.dev/programming-guides/encoding/#last-one-wins) merges repeated occurrences of singular embedded messages. Add malformed-first/valid-second cases through `parseTurn` and the production report, not only the legacy helper tests.

5. **The new HOME injection changes override precedence.** Passing normal `environment["HOME"]` into `tokiCachePaths` bypasses its existing `TOKSCALE_CONFIG_DIR` branch, so a relocated cache disappears. Meanwhile `cliDBPaths` still reads ambient `GEMINI_CLI_HOME`, so the temporary-HOME fixture is not fully isolated. Pass one explicit environment/path context and retain the existing overrides; exercise both through the production fetcher using only temporary directories.

6. **The remaining identity and overflow cases need the actual owners.** Copied ID-less databases in recognized roots still count twice: aggregation retains neither session nor row identity. One global response-ID set is an assumption, not proof of cross-session uniqueness. Establish the supported source/session/row identity contract, preserving legitimate ID-less rows rather than content-deduplicating them. Separately, the existing shared `CostUsageTokenMix` turns overflow into nil and later revives it as a fresh accumulator: three accepted days with `Int.max - 1`, `2`, and `7` input tokens display an input mix of 7. Reject the unrepresentable report coherently or preserve invalid aggregation state at its owner. Add copied-input and three-day dashboard/window/menu/widget/CLI assertions. Pin the clock: the current fixed-August two-day test uses `Date()` and becomes vacuous once those rows leave its window.

The live-WAL fixture is better, but it checks only returned rows and file existence while the writer is open. It does not assert the claimed cleanup, cancellation, or unchanged contents. Please narrow those claims and prove the actual source-access behavior with coordinated temporary writers and before/after checks. Normal [SQLite read-only WAL access](https://sqlite.org/wal.html#read_only_databases) may create sidecars and its [WAL-index protocol](https://sqlite.org/walformat.html#wal_locks) coordinates readers; do not equate database read-only mode with literal SHM-byte preservation. No custom VFS, unsafe file-copy shortcut, or `immutable=1` on changing data is requested.

This remains a bounded token-only reader repair, not a request for pricing, wider discovery, account attribution, or a guessed timestamp schema. Preserve the existing separate SQLite text/thinking semantics and record the pinned schema provenance in the fixtures. Please update the description to match what is actually implemented and tested, particularly publication, scan-executor cancellation, and WAL cleanup.

steipete (approved):
Reviewed the maintainer repair at `82aa92e655f9c3e72d13c22d8f289c3ada931f9d`. The remaining actionable findings from my earlier reviews are repaired in this PR. The reader now carries checked arithmetic, source completeness, attempted-work limits and cancellation through the production fetcher and publication paths. Independent review found no remaining actionable findings; the focused regressions, full local suite and exact-head CI passed.

The older inline discussions are also accounted for: SQLite and JSONL are no longer added together; recognized direct and conversation roots are covered; cache-write and request counts survive aggregation; malformed numeric input and partial history cannot become valid zero usage; and the declared environment overrides are preserved. Pricing-related suggestions no longer apply to this token-only scope. Historical model IDs remain historical rather than being rewritten through live routing aliases. SQLite text and thinking are independently established separate counters; nonzero JSONL reasoning remains unsupported instead of guessing its overlap. Session creation is deliberately not substituted for a missing generation timestamp. Exhausting a bound remains unavailable/partial, not a claim of complete history.

The source and WAL limitations, synthetic-fixture proof, and absence of real-account/live-installation verification are documented in the updated description. No pricing, broader discovery, account attribution or release is included. Thanks @Yuxin-Qiao for the original work and follow-up improvements.

steipete:
Merged as [d94c71acfc00](https://github.com/steipete/CodexBar/commit/d94c71acfc00fc09bef284d888000042ead66d79). Local main fast-forwarded and is clean; its tree exactly matches the reviewed and tested repair at `82aa92e655f9c3e72d13c22d8f289c3ada931f9d`. The maintainer repaired this existing PR and preserved @Yuxin-Qiao's contributor credit. The previous actionable review findings are addressed, the final review is approved, and all 17 existing inline discussions are resolved.

The reader is token-only and limited to the recognized local roots. Discovery, attempted-work budgets, cancellation, SQLite schema/BLOB validation, exact JSONL numbers and checked aggregation now feed the real fetcher/publication consumers. Unavailable or partial history cannot become established zero, stale JSONL cannot replace failed SQLite history, and overflow remains unknown through menu/widget/CLI totals. Unsupported timestamp/reasoning layouts remain explicitly unsupported; there is no new pricing, broad discovery or account attribution.

Verification:

- Final focused production-boundary selection: 98 macOS tests in 13 suites plus seven portable parser tests. Exact command: `env -u CODEXBAR_ALLOW_TEST_KEYCHAIN_ACCESS CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 CODEXBAR_TEST_CODEX_FILE_ISOLATION=1 swift test --filter 'AntigravityLocal(Integrity|Reader|Scan|WAL|Publication)Tests|AntigravityJSONLNumberTests|CursorAntigravityLocalReaderTests|CostProvenanceTests|CostUsageScanExecutorTests|CostUsageTokenSnapshotDaySelectionTests|CostUsageWindowSummaryTests|GeminiStdoutHolderFixtureTests|GeminiFixturePathTests'`.
- The exact Gemini four-suite CI selection passed 55 tests with `LIVE_GEMINI_FETCH=0`: `swift test --skip-build --no-parallel --filter '(^CodexBarTests\.GeminiStatusProbeAPITests/|^CodexBarTests\.GeminiStatusProbePlanTests/|^CodexBarTests\.GeminiStatusProbeTests/|^CodexBarTests\.GoogleWorkspaceStatusNetworkTests/)'`, also using the Keychain-suppression and Codex-file-isolation prefix above. The registered live case returns at its opt-in guard; this is not a real-account probe.
- All 39 architecture-gate tests passed with `swift test --skip-build --filter ProviderArchitectureGatekeeperTests` and the same suppression/isolation prefix. `make check`, `git diff --check`, and explicit SwiftFormat/SwiftLint checks of the portable test file passed without weakening rules.
- Full suite: `env -u CODEXBAR_ALLOW_TEST_KEYCHAIN_ACCESS CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 CODEXBAR_TEST_CODEX_FILE_ISOLATION=1 CODEXBAR_TEST_RETRY_NON_TIMEOUT_FAILURES=0 LIVE_GEMINI_FETCH=0 make test` — all 953 selections, 80 first-pass groups, zero retries, failures or timeouts; 1211.8 seconds total.
- Independent review of both final candidate inputs reported no actionable findings. The safe schema/identity/readiness regressions failed before repair; the corrected raw-number regression exposed 21 failing assertions before its fix. Intermediate compile/fixture/format/lint failures were corrected before the final passing gates.
- [Exact-head PR CI](https://github.com/steipete/CodexBar/actions/runs/33104407460) passed lint, both macOS shards, provider-engine A/B goldens, Linux x64/ARM64 builds/tests/smoke checks, musl build and the aggregate gate. No CI rerun was needed. Native Linux test execution is distinct from the local Mac portable-test run.

The [new main CI run](https://github.com/steipete/CodexBar/actions/runs/33107716377) passed every required job on `d94c71acfc00fc09bef284d888000042ead66d79` without a rerun, including both macOS shards. Main was fast-forward verified and remained clean at the same tested tree. This verifies the maintainer-owned Gemini readiness-fixture follow-up after the prior main failure; production helper deadlines and retry policy remain unchanged.

All final local proof used owned synthetic fixtures and suppressed real Keychain access. No live provider/account/browser, real Antigravity installation, or app-bundle/UI launch was used. Normal read-only SQLite WAL coordination may create/update sidecar coordination state; literal SHM-byte immutability is not claimed. No release or unrelated issue closure is included.

PR #3215, opened 2026-08-26
fix(codex): clamp cached tokens to input in CodexUsageRow

Yuxin-Qiao:
## Summary
- clamp `cached` to `min(cached, input)` in `CodexUsageRow.init` to prevent inflated totals when malformed `~/.codex/sessions/*.jsonl` reports more `cached_input_tokens` than `input_tokens`
- keep `input/output` non-negative and preserve existing `reasoning` clamp
- regenerate `CodexParserHash` (hash `cfd84d13 -> 5dc8605a`)

Mirrors TokScale `crates/tokscale-core/src/sessions/codex.rs:194-210` (`clamped_cached = min(cached, input); clamped_reasoning = min(reasoning, output)`).

## TokScale reference
- TokScale parses `cached_input_tokens` / `cache_read_input_tokens` as `cached` and clamps to `input` before `TokenBreakdown` conversion.
- CodexBar previously stored `cached` verbatim and only clamped `reasoning`, so a single malformed `token_count` row could inflate `totalTokens` and `costUSD` for that day.

## Evidence (text, no screenshots)
- `make check` pass (2032 lint files, 0 violations, parser hash current `5dc8605a458f782f`, `Provider manifests 69 providers`):
```
Codex parser hash is current (5dc8605a458f782f)
Provider manifests are current (69 providers)
Done linting! Found 0 violations, 0 serious in 1990 files.
```
- Minimal reproduction (swift, HOME/.tmp-home):
```
input=100 cached=1000
old cached stored: 1000 -> total would be 1100
new cached clamped: 100 -> total would be 100
clamp works: PASS
```
- Diff stat: `2 files changed, 6 insertions(+), 4 deletions(-)` (only `CostUsageScanner.swift` + generated hash).

## Testing
- `make check` full pass (see above)
- No UI changes; backend-only, no provider silo or pricing table changes.


steipete (changes requested):
Reviewed head 046448e649b8f5075801bd17a0dc7bab1cc0335e. The aggregate/row mismatch already identified in review is a blocker: the scanner records the raw cached delta in the daily aggregate before constructing the newly clamped row, so reconciliation can discard the corrected rows.

There is also a valid-data case to preserve. The bare-usage parser currently subtracts cached tokens from prompt input before constructing its totals. A response with prompt=120 and cached=100 therefore reaches this initializer as input=20, cached=100. A universal initializer clamp would turn only the row into cached=20 and disagree with its aggregate even though the original response is valid. Please settle that representation at ingestion and cover both native token_count and bare-usage paths through scanner, persistence, reload, and final reporting. The initializer arithmetic alone does not establish the displayed token or cost result.

The generated parser-hash change also needs an upgrade strategy consistent with the no-rebuild guarantees in #3051. The current 0.55.1 hash is not a compatible predecessor, so a mechanical refresh would rebuild existing databases; simply allowlisting it would retain malformed persisted aggregates. Please prove selective repair or another bounded migration that preserves unaffected history and scan offsets, and include valid cached-heavy, malformed, incremental, and SQLite-upgrade regressions before rebasing this for merge.

Yuxin-Qiao:
Closing this PR based on review feedback:
1. CostUsagePricing.codexCostUSD already enforces min(cachedInputTokens, totalInput) at pricing evaluation time.
2. In bare-usage records, input stores billedInput (prompt - cached), so clamping cached to input in CodexUsageRow.init incorrectly truncates valid cached-heavy rows and creates aggregate/row reconciliation mismatches.
3. Key TokScale parity logic was already incorporated in #3120.

Thanks for the thorough review @steipete!

PR #3218, opened 2026-08-26
fix: preserve OpenCode Go API percentage units

### LINKED ISSUE #3216: OpenCode Go: weekly usage shows 100% when the true value is 1% (fraction heuristic misreads whole-percent API values)
**TL;DR:** `OpenCodeGoUsageFetcher.parseWindow` treats any direct percent value ≤ 1.0 as a 0-1 fraction and multiplies it by 100. The Go usage API returns whole percentages (0-100), so an honest weekly usage of exactly **1%** renders as **100%**.

## Environment
- CodexBar 0.55.1 (macOS 26.6.2, arm64)
- OpenCode Go provider, source `local+api` with API key from `~/.local/share/opencode/auth.json`

## Repro

Query the same endpoint CodexBar uses (`GET https://opencode.ai/zen/go/v1/usage`, Bearer key):

```json
{ "usage": {
    "rolling": { "percent": 3 },
    "weekly":  { "percent": 1 },
    "monthly": { "percent": 0 } } }
```

CodexBar shows: rolling **3%**, weekly **100%**, monthly **0%**.
`codexbar usage --provider opencodego --format json` confirms:

```json
"primary":   { "usedPercent": 3 },
"secondary": { "usedPercent": 100 },
"tertiary":  { "usedPercent": 0 }
```

Only values in the range `(0, 1]` are affected - anything above 1 passes through untouched and 0 maps to 0 either way, which is why it looks like "suddenly 100%" right after the weekly counter ticks to its first percent.

## Root cause

In `Sources/CodexBarCore/Providers/OpenCodeGo/OpenCodeGoUsageFetcher.swift`, `parseWindow` ends with:

```swift
// A direct percent field may arrive as a fraction (0...1) or a percent (0...100)
if percentIsDirect, resolvedPercent <= 1.0, resolvedPercent >= 0 {
    resolvedPercent *= 100
}
```

That heuristic made sense for other window shapes, but `fetchAPIUsage` feeds this parser a documented JSON contract where `percent` is always 0-100. The fraction conversion has no way to know that here.

## Suggested fix

Skip the ≤ 1.0 fraction conversion in `fetchAPIUsage`'s parse path (or add a dedicated parser for the public usage endpoint that trusts `percent` verbatim), keeping the heuristic only for the regex/web scraping paths where shapes genuinely vary.

steipete:
## Summary

Closes #3216.

OpenCode Go's public usage API returns percentages in 0...100 units. The generic dashboard parser also accepts fractions, so its heuristic turned an API value of 1 into 100 and 0.5 into 50. This repair makes the API's units explicit while preserving generic dashboard fraction support, computed ratios, reset parsing, optional windows, and local cost history.

Upstream contract: [OpenCode subscription usage calculations](https://github.com/anomalyco/opencode/blob/dev/packages/console/core/src/subscription.ts).

Thanks @rodrigoalma for the report and reproduction.

## Verification

- The new API regression failed on the old parser, reproducing the reported scaling error.
- `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --filter 'OpenCodeGoUsageFetcherErrorTests|OpenCodeGoUsageParserTests|OpenCodeGoWebOverlayTests|OpenCodeGoProviderStrategyTests|OpenCodeGoPercentUnitLinuxTests'` passed 80 tests across four macOS suites. This does not claim execution of the Linux-only suite; Linux execution is covered by CI.
- API response cases cover `3 / 1 / 0`, all windows at 1, 0 and 100, fractional percentages such as 0.5, reset timestamps and absent weekly/monthly windows. Parsed API overlays retain local daily cost/request history and authoritative confidence. Existing dashboard fraction and low computed-ratio behavior stays covered.
- `make check` and `git diff --check` passed.
- Independent Codex review reported no actionable blockers.
- Full `make test` passed all 78 groups (933 selections) on the first attempt, with no failures, retries or timeouts.
- [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/33016333157) passed for `7e127c21dfeceb042aad2f71e37e4c13dd76df98`: Linux x64/ARM builds, tests and smoke checks; Linux musl build; both macOS test shards; provider-engine goldens; lint and the aggregate check. GitGuardian passed as well.

No live credentials, account probes, cookie imports or app restart were used; endpoint responses are exercised with the existing URLSession stub seam. Authentication and persisted usage data are unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.55.2 — Unreleased
 
 ### Fixed
+- OpenCode Go: preserve API percentage units so 1% usage no longer appears as 100%, including local-history overlays (#3216). Thanks @rodrigoalma!
 - CLI install: block inherited shell functions and startup hooks before helper validation and failure handling, and use absolute tools and a clean administrator-command environment while retaining approval and both existing symlink destinations (#3205, #3217).
 
 ## 0.55.1 — 2026-08-25
--- a/Sources/CodexBarCore/Providers/OpenCodeGo/OpenCodeGoUsageFetcher.swift
+++ b/Sources/CodexBarCore/Providers/OpenCodeGo/OpenCodeGoUsageFetcher.swift
@@ -237,7 +237,7 @@ public struct OpenCodeGoUsageFetcher: Sendable {
         guard let text = String(data: response.data, encoding: .utf8) else {
             throw OpenCodeGoUsageError.parseFailed("Response was not UTF-8.")
         }
-        return try self.parseSubscription(text: text, now: now)
+        return try self.parseAPIUsage(text: text, now: now)
     }
 
     static func requiredZenBalanceFallback(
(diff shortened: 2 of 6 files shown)


steipete:
Landed on main as 4e1f7a3ef36241b5bccf8499a78a91ad2f4b9d77, fixing #3216.

The API now has an explicit percentage-unit parsing boundary: 1 stays 1%, and 0.5 stays 0.5%. Generic dashboard fraction parsing, low computed ratios, reset timestamps and optional windows are preserved. The parsed API overlay keeps local daily costs/request history and authoritative confidence; authentication and persisted data are unchanged.

Proof: the new regression failed against the old parser. `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 swift test --filter 'OpenCodeGoUsageFetcherErrorTests|OpenCodeGoUsageParserTests|OpenCodeGoWebOverlayTests|OpenCodeGoProviderStrategyTests|OpenCodeGoPercentUnitLinuxTests'` passed 80 tests in four macOS suites. `make check` and `git diff --check` passed. Full `make test` passed all 78 groups/933 selections first-pass, with no retries or timeouts. Independent Codex review found no actionable blockers.

[Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/33016333157) passed Linux x64/ARM builds/tests/smoke checks, the musl build, both macOS shards, provider-engine goldens, lint, aggregate and GitGuardian. The automated review's separate verification attempt failed during package setup before Swift ran; it is not a parser-test failure. No live account probes, credentials, cookie imports or app restart were used. Main was pulled and verified clean after merge. The change remains unreleased for 0.55.2. Thanks @rodrigoalma!

PR #3223, opened 2026-08-27
test: isolate Codex credential files from ambient accounts

steipete:
## Summary

Ordinary tests already suppress Keychain access, but legacy UI helpers can still call Codex's file-backed account loader. Empty credential environments also fall back to the process user's home. Put the test boundary at the Codex-owned file operations instead of relying on every caller to remember a fake environment.

The policy denies unregistered native, legacy and OpenCode credential reads, fingerprints and complete save/promotion operations in actual test contexts. Explicit immutable task-local fixture scopes permit only named files or owned roots, with component containment and symlink rejection. The test runner passes an isolation signal to children; scoped child environments replace inherited fixture grants. Default managed-account/workspace metadata stores use temporary paths during tests. Production source selection, parsing, storage formats and refresh behavior are unchanged.

Most changed test files only move existing temporary homes under the per-test fixture root. Existing credential, permission, fallback and account assertions remain intact. The two shared controller helpers now default to empty account information.

## Verification

- 405 focused macOS tests in 18 suites, including the architecture gate, plus the portable credential-permission test executed on macOS: passed. The final fixture follow-up passed 32 fixture/policy tests and 38 architecture tests in separate runs. A combined focused run hit the existing publication wait while the architecture scan occupied the main actor; the tests pass separately with their original timeout and assertions unchanged.
- `make check`: passed, zero SwiftLint violations.
- `bash Scripts/test_codex_file_isolation_child.sh`: passed. Compiles the actual policy/detector with optimization and without `DEBUG`; proves denied access, scoped access, replacement of parent fixture grants and non-test decisions using synthetic temporary files.
- Denied-operation tests record zero read, existence, metadata, directory and write calls. Regressions cover fallback escape, task cleanup/concurrency, prefix collisions, traversal and dangling symlinks.
- Independent complete-diff review: no actionable findings.
- Full `make test`: 934 selections across 78 groups, all first-pass; zero failures, retries or timeouts (1067.6 seconds). The first full run exposed three dashboard-cache tests whose synthetic auth homes still needed explicit fixture registration. Those and sibling credential fixtures now use owned roots without changing their authority assertions; the fixture trait also scopes the existing dashboard-cache URL override.
- [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/33033035702): all required checks passed, including lint, Linux x64/ARM/musl, both macOS shards/provider-engine goldens and the aggregate gate.

The child harness stubs unrelated link dependencies; it does not exercise the complete release CLI. This is Codex-owned file isolation, not an OS sandbox or proof of isolation for every provider. No live account probes or application launch were used. The existing deliberately disabled live-account test remains disabled. Internal test safety only; no changelog entry or release.

### DIFF
--- a/Scripts/test.sh
+++ b/Scripts/test.sh
@@ -9,6 +9,10 @@ RETRY_NON_TIMEOUT_FAILURES="${CODEXBAR_TEST_RETRY_NON_TIMEOUT_FAILURES:-1}"
 
 cd "${ROOT_DIR}"
 
+# Inherited by release-built tests and arbitrary CLI children as well as the test runner.
+export CODEXBAR_TEST_CODEX_FILE_ISOLATION=1
+unset CODEXBAR_TEST_CODEX_FILE_FIXTURES
+
 # Defense in depth: test processes also self-detect, but keep this explicit so runner changes cannot
 # expose the user's login Keychain. Deliberate isolated Keychain tests must opt in by setting the allow flag.
 if [[ "${CODEXBAR_ALLOW_TEST_KEYCHAIN_ACCESS:-}" != "1" ]]; then
--- a/Scripts/test_codex_file_isolation_child.sh
+++ b/Scripts/test_codex_file_isolation_child.sh
@@ -1,0 +1,80 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+PROOF_DIR="$(mktemp -d "${TMPDIR:-/tmp}/codex-file-child.XXXXXX")"
+trap 'rm -rf "$PROOF_DIR"' EXIT
+
+# Compile the actual policy and detector without DEBUG. Only unrelated link dependencies are stubbed;
+# this is policy/child proof, not a release build of the entire app or credential store.
+cat > "$PROOF_DIR/Proof.swift" <<'SWIFT'
+import Foundation
+
+enum CodexOAuthCredentialsError: Error { case notFound }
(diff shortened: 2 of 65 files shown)


steipete:
Landed as bb89f0b22f3dcbb9dfcd0bb753dd9baff84e328a after the full local suite and every required exact-head CI check passed.

The test boundary now denies unregistered Codex credential-file operations independently of credential environment values. Named per-test roots preserve real parsing, fallback, permission, promotion and account-ownership assertions; the shared fixture also scopes the existing dashboard-cache override. Production auth/source-selection/storage contracts are unchanged.

Verification: `make check`; focused credential/account/promotion/architecture tests (405 plus the portable credential-permission test); final fixture follow-up (32 fixture/policy tests and 38 architecture tests); `bash Scripts/test_codex_file_isolation_child.sh`; and full `make test` (934 selections, 78 groups, all first-pass, zero failures/retries/timeouts, 1067.6 seconds). Test commands unset `CODEXBAR_ALLOW_TEST_KEYCHAIN_ACCESS` and enable Keychain suppression and Codex file isolation. Complete-diff independent review reported no actionable findings. CI: https://github.com/steipete/CodexBar/actions/runs/33033035702.

The initial full run caught three unregistered synthetic dashboard fixtures; those and sibling fixtures were migrated without weakening assertions. One combined focused run encountered the existing publication timeout during architecture scanning; separate final runs passed without timeout changes. The optimized child harness compiles the actual policy/detector without `DEBUG`, with unrelated link dependencies stubbed. It does not prove the full release CLI, live account behavior or other providers' isolation. No app launch, release or version change.

Main was fast-forwarded and checked clean; the tested head is reachable from main.

PR #3238, opened 2026-08-27
feat(claude): tokscale-inspired token breakdown and journal exclusion

Yuxin-Qiao:
Implements Claude local token consumption following https://github.com/junhoyeo/tokscale's local-log approach (direct `~/.claude/projects/**/*.jsonl` parsing, no quota API):

- **CLI**: `codexbar cost --provider claude` now appends `Daily breakdown (last 7 days)` and `Top models` when `snapshot.daily` is present — `in/out/cacheRead/cacheCreate` mix + `modelsUsed`, Top 5 aggregated across `modelBreakdowns` sorted by cost, mirroring tokscale's overview/models/daily text views. `Today / Last N days` remains, `hintLine` keeps the existing estimate disclaimer.
- **Scanner**: adds `isClaudeWorkflowJournal` to skip `.../subagents/**/journal.jsonl` in `inventory / process / parse` (aligns with `tokscale::is_workflow_journal`), reusing existing `CostUsagePricing.claudeCostUSD` (models.dev) with `ephemeral_1h_input_tokens` handling, consistent with tokscale's cache split.
- **Dashboard**: `SpendDashboard` already aggregates `.claude` via `costCapableProviders`; no extra flag needed.

Verification: `swiftformat --lint` 0, `swiftlint lint --strict` 0 on both files; `SpendDashboard` aggregation verified. Manual check: `codexbar cost --provider claude --days 30` (text) and `--format json` (`daily[].modelBreakdowns`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.55.2 — Unreleased
 
 ### Fixed
+- Grok: keep OAuth usage, identity, and plan bound to the same credentials when a native login changes during billing, while keeping successful cookie usage separate from auth-file metadata.
 - Cursor: estimate omitted API-rate costs from cached or bundled pricing, preserve invalid-cost coverage and compatible history caches, and separate Overview history coverage from missing subscriptions (#3129). Thanks @Yuxin-Qiao!
 - OpenRouter: label the API key spending limit consistently and clarify that it is a cap, not the separate account balance (#3158). Thanks @vinschger!
 - Codex: refresh local session cost estimates when global cost tracking is off, without repeatedly rejecting successful scans as stale. Thanks @vinschger!
--- a/Sources/CodexBarCore/Providers/Grok/GrokProviderDescriptor.swift
+++ b/Sources/CodexBarCore/Providers/Grok/GrokProviderDescriptor.swift
@@ -209,9 +209,27 @@ struct GrokOAuthFetchStrategy: ProviderFetchStrategy {
 
     let mode: Mode
     let kind: ProviderFetchKind = .oauth
+    let proxyBilling: GrokWebFetchStrategy.ProxyBillingFetch
+    let grpcBilling: GrokWebFetchStrategy.ProxyBillingFetch
+    let webStrategy: GrokWebFetchStrategy
+    let settingsTier: GrokWebFetchStrategy.SettingsTierFetch?
 
-    init(mode: Mode = .proxyThenGrpc) {
+    init(
+        mode: Mode = .proxyThenGrpc,
+        proxyBilling: @escaping GrokWebFetchStrategy.ProxyBillingFetch = {
+            try await GrokCreditsProxyFetcher.fetch(credentials: $0)
+        },
+        grpcBilling: @escaping GrokWebFetchStrategy.ProxyBillingFetch = {
+            try await GrokWebBillingFetcher.fetch(credentials: $0)
+        },
+        webStrategy: GrokWebFetchStrategy = .init(),
+        settingsTier: GrokWebFetchStrategy.SettingsTierFetch? = nil)
+    {
         self.mode = mode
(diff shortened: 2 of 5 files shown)


steipete (changes requested):
Thanks @Yuxin-Qiao. I reviewed `6e1e2e2b524edb0beccefe395bf10a08d93833f0` against `d9e950e49c954eea6cfe090d70e878eb4642119e`. Please isolate the intended change first: this PR still changes 29 files, including Hermes/Pi registration and readers, Antigravity, Grok integration, and global lint policy. These are outstanding changes, not already-merged noise. The CLI/journal delta is separable as `ff4f7ad..bea7b8e` (two files, +77/−4); that is a useful starting point, not an approval of its current implementation.

The narrow change also needs correctness coverage:

1. **Preserve unknown and partial model totals.** `CLICostCommand.swift:233–256` turns `costUSD: nil` and `totalTokens: nil` into zero. A 1,000-token model with unresolved pricing prints `$0.00`; a model priced on one day and unpriced on another prints only its known subtotal alongside both days' tokens. Claude deliberately emits nil model cost when pricing is unavailable. Keep unknown/partial state independently for cost and tokens; only explicit zero means free.

2. **Do not present partial model coverage as the complete window ranking.** The new aggregation skips absent `modelBreakdowns` and ignores incomplete history coverage. A 30-day snapshot containing a model-attributed $2 day and an unattributed $100 day still prints an unqualified “Top models (Last 30 days)” from the $2 subset. Qualify or withhold that ranking when attribution or scan coverage is incomplete. Add a stable model-ID tie-breaker too: six equal cost/token totals currently leave the top-five membership dependent on dictionary ordering.

3. **Use calendar days or label populated rows accurately.** Lines 212–219 take the last seven populated rows and call their count “last N days.” On August 27, rows for August 1 and August 20 become “last 2 days.” Either filter an actual interval using the configured bucket calendar or label these as the last recorded usage dates. Cover sparse history and timezone boundaries.

4. **Reconcile excluded journals before reusing an old cache.** The inventory excludes journals, but removal from `cache.files` occurs only during mutation. A cold process loading a compatible cache with a journal contribution scanned ten seconds earlier can take the 60-second debounce reuse path, then memoize that old report against the newly filtered inventory. Subsequent exact memo hits return before the elapsed-time check. A rescan does remove the journal; the gap is this cold-cache/debounce path. Add temporary-root coverage for existing disk cache, memo reuse, and force refresh, preserving valid shallow and deeply nested agent transcripts.

The journal predicate matches the defensive pattern in pinned [Tokscale source](https://github.com/junhoyeo/tokscale/blob/62ca1eb1677556972ba963fdfa3a41ab23c1eb4b/crates/tokscale-core/src/sessions/claudecode.rs#L130), whose comments say current orchestration records already parse to zero. Its assistant-shaped journal test is synthetic, not evidence of a current Anthropic producer bug. Please describe this as defensive filtering unless a versioned producer sample demonstrates an actual counting error; the cache scenario above is conditional on an existing contribution.

The new text rendering runs for every provider with daily data, despite its Claude-specific name. Keep that output-scope decision explicit, and use shared rendering or descriptor-owned policy rather than adding an arbitrary Claude-only conditional. Add isolated renderer cases for empty, unknown, partial, explicit-zero, metered, cross-provider, sparse-day, and tied-ranking data. The two-file delta currently adds no such tests.

Please also drop the unrelated global SwiftLint disables. The inherited `HermesLocalReaderTests` calls the default real database reader and requires exactly `2_036_354` tokens; that is not an isolated CI fixture and must not be executed as routine tests. Keep those provider/storage proposals separate with their own scope and proof.

These findings are based on source inspection, not execution of this PR or its tests. The dashboard freshness tests on my separate branch do not validate this head. Run the focused fixture tests, unchanged `make check`, and full suite on the isolated revision before it is considered for landing.

PR #3242, opened 2026-08-27
fix(cost): keep thermal precedence in automatic catch-up pause

Yuxin-Qiao:
When automatic Codex cost catch-up encounters both Low Power Mode and serious thermal pressure, it previously published the low-power pause reason. The shared policy now gives thermal pressure precedence, so the regular menu and Usage & Spend report the stronger constraint consistently.

The existing 60-second constrained retry, Low Power pause, accelerated-mode rules, and normal automatic scan budgets are preserved. Earlier energy-policy proposals and inherited Antigravity changes are outside this final diff.

Validation on the final candidate:
- New production-worker regressions fail on unchanged main: both workers publish `.lowPower` instead of `.thermal`. They pass with the fix, verify the 60-second sleep, and verify zero scan advances.
- `swift test --no-parallel --filter 'CodexCostCatchUpPolicyTests|UsageStoreCodexCostCatchUpTests|UsageStoreSpendDashboardCodexCostCatchUpTests|ProviderArchitectureGatekeeperTests'`: 84 tests passed.
- `make check`: passed.
- `make test`: 1,008 selections in 84 groups; all first-pass successes, zero failures, timeouts, or retries.
- Required Codex autoreview: scoped-clean at P0.
- Final-head CI: [run 33857595085](https://github.com/steipete/CodexBar/actions/runs/33857595085), pending completion before merge.

Worker tests inject the combined resource state into the real asynchronous catch-up loops and observe published activity. They use synthetic pending work and cancellation at the first retry sleep; no system power settings, physical heating, or account data are involved.

Changelog and Codex documentation are updated. Thanks @Yuxin-Qiao!


steipete (changes requested):
Thanks @Yuxin-Qiao. I reviewed `b4ae6b9d1ff390379b5341ff1f95663496126a78` and traced both regular-menu and dashboard catch-up consumers. Please split this first: the current PR changes ten files and includes the still-unmerged Antigravity work from #3212. The actual scheduling change is the two-file `de56224..b4ae6b9` delta (+22/−16), so it can be reviewed independently of that reader work.

The scheduling delta also has a concrete correctness issue: **automatic Low Power Mode bypasses the serious-thermal pause.** In `CodexCostCatchUpPolicy.swift:92–103`, the new low-power branch returns before the `.serious` guard at lines 105–109. With automatic mode, Low Power Mode enabled, serious thermal pressure, battery power, and a previous active duration of two seconds, it returns `.runAfter(9998)` rather than pausing. Both catch-up loops treat `runAfter` as permission to scan after sleeping; neither supplies an independent serious-thermal veto. Critical pressure still takes the earlier guard, and accelerated mode's existing exception is separate. Please keep the automatic thermal guard ahead of the low-power throttle and add the combined-state regression across the power-source cases. The changed test only covers nominal temperature.

The fivefold normal-duty increase needs separate evidence and a deliberate energy-policy decision. Those existing rates were introduced in the profiling-backed #2848 (`d7934a5f6121b4509b7242caa04953d99da016c2`), not accidental constants. Changing expected delay values proves the arithmetic, not five-times-faster indexing or acceptable idle/battery cost. Please establish whether the reported `635/637` state is actually paused for Low Power Mode, as distinct from no-progress or another pending-work condition, then compare backlog progress and CPU/energy on the same signed build and workload. This policy reads macOS Low Power Mode; it is not the separate CodexBar background-work setting, and the change currently applies to Intel as well as Apple silicon.

The right boundary remains this shared policy, with thermal precedence and a single duty calculation rather than duplicated scheduling arithmetic. Please keep the low-power behavior proposal separate from retuning every normal automatic scan. I have not executed this contributor branch or independently verified its test/performance claims; source inspection is sufficient to demonstrate the combined thermal-state regression, but not to approve the new energy budget or the inherited Antigravity scope.

Yuxin-Qiao:
@clawsweeper re-review

Yuxin-Qiao:
@clawsweeper re-review

Yuxin-Qiao:
Follow-up on current head `a8d6783c0`:\n\n- Catch-up duty values now distinguish ordinary AC, unknown-resource, and battery work while preserving the low-power policy.\n- Interruptible sleeps re-check the resource state and cap each wait, so a Low Power transition is not delayed by a long catch-up interval.\n- Regression expectations now cover AC timing and battery Low Power off/on transitions.\n\nValidation: focused catch-up suites (46 tests) and `make check` passed.

Yuxin-Qiao:
@clawsweeper re-review

steipete (approved):
The final diff addresses the earlier review: thermal precedence is preserved, the original pause and duty-cycle policies remain intact, and unrelated Antigravity work is absent. Both real catch-up workers now have failing-before/passing-after regression coverage for the published pause reason and zero scan advances. The 84 focused tests, make check, and full 1,008-selection suite passed with no retries. Required autoreview is scoped-clean at P0. Final-head CI is the remaining merge gate.

steipete:
Merged as e0a31b9047016900d5005fd81c4b3f326e269ac8. The final candidate passed all nine checks in [CI33857595085](https://github.com/steipete/CodexBar/actions/runs/33857595085).

Local proof: `make check`; `make test` (1,008 selections, 84 groups, zero failures/retries/timeouts); 84 focused policy/worker/architecture tests; and a Developer ID signed rerun of both worker regressions. On unchanged main, both real workers published lowPower for the combined state. The fixed build publishes thermal, requests the existing 60-second retry, and advances no scan. These are synthetic resource-state/scan fixtures, with no machine heating, power-setting changes, or account access.

Required Codex autoreview was scoped-clean at P0. Changelog and documentation retain contributor credit. Main was fast-forwarded and verified clean after merge.

Issue #1167, opened 2026-05-26
Allow selected providers to stay as separate menu bar icons in Merge Icons mode


nk-tedo-001:
### Feature request

CodexBar currently supports either one status item per enabled provider, or Merge Icons mode with a provider switcher. I would like a hybrid option: choose specific providers that should remain visible as separate menu bar icons while all other enabled providers stay inside the merged/overview icon.

### Use case

I actively monitor a few providers and want them visible at a glance, but I also keep several lower-priority providers enabled. Showing every provider separately takes too much menu bar space, while fully merging everything hides the few providers I care about most.

### Proposed UX

In Settings -> Providers or Settings -> Menu Bar, add a per-provider option such as:

- `Show as separate menu bar icon`
- `Pin to menu bar`
- `Always show separately`

Suggested behavior:

- When Merge Icons is enabled, pinned providers get their own status items.
- Unpinned enabled providers remain available in the merged icon, provider switcher, and Overview.
- If no providers are pinned, behavior stays the same as today's Merge Icons mode.
- If Merge Icons is disabled, the existing one-icon-per-enabled-provider behavior stays unchanged.

### Benefits

- Keeps the menu bar compact.
- Keeps the most important providers glanceable.
- Fits the existing per-provider settings/display-controls model.
- Avoids forcing users to choose between all separate icons and one combined icon.

steipete:
Decision-ready design is in draft PR #1815: https://github.com/steipete/CodexBar/pull/1815

Recommendation:

- add an **Also show separately** provider picker while Merge Icons is enabled
- keep pins additive: pinned providers remain available in the merged switcher and Overview
- keep the merged icon visible even when every provider is pinned
- preserve disabled-provider pins as dormant preferences and restore them when re-enabled
- resolve one pure `StatusItemLayout` model across visibility, menu attachment, icons, and animation

The PR includes migration, lifecycle edge cases, stable-identity tests, and Tahoe visual/performance acceptance criteria. It intentionally makes no runtime change and leaves this issue open for owner approval of the pin semantics.

steipete:
Keeping this open. Main still removes provider status items when Merge Icons is enabled. Separate account icons do not implement this request. The accepted implementation is additive provider pins resolved through one shared status-item layout policy; pinned providers stay in the merged switcher and Overview, and the merged icon remains visible even when every provider is pinned.

PR #3251, opened 2026-08-28
Add separate menu bar items per provider account

Thinkscape:
## Summary
- add a per-provider **Menu bar items** setting for providers with multiple accounts
- keep the default combined provider item, or create one stable, account-scoped status item per account
- refresh every account immediately when split mode is enabled and keep account menus isolated from sibling identities/data
- disable split mode when merged provider icons are enabled

## Menu bar items

Screenshot 2026-08-29 at 8 50 28 am

| One for provider | One per account |
| --- | --- |
|  |  |

<details>
<summary>Account menu behavior</summary>

| Combined menu | Account-scoped menu |
| --- | --- |
|  |  |

Account identifiers are redacted.
</details>

## Verification
- `swift test --filter 'AccountMenuBarDisplayModeTests|StatusItemControllerSplitLifecycleTests|StatusMenuTokenAccountSwitcherTests|StatusMenuCodexSwitcherTests|StatusMenuCodexSeparateAccountTests|StatusItemIconObservationSignatureTests|ProvidersPaneCoverageTests|CodexAccountsSettingsSectionTests'`
- `make format && make check`
- packaged with ad hoc signing and verified in the fresh app bundle with two Codex accounts: combined mode showed one item with the switcher; split mode showed two independently populated items without switchers, without manual refresh
- full suite left to CI


Thinkscape:
This is ready for re-review. Let's go! :)

nikbyte:
This would be very useful for my setup.

I use multiple accounts for the same provider (Codex in particular), and while CodexBar already lets me add and switch between them, what I really want is to monitor the usage of all accounts simultaneously.

Ideally, each account should have its own independently refreshed usage state, so I can see at a glance, for example:

* Codex Account A — 5h / weekly usage
* Codex Account B — 5h / weekly usage
* Codex Account C — 5h / weekly usage

without switching the active account just to refresh or inspect its limits.

The “one menu bar item per account” approach in this PR looks like a great solution. It would be even better if the same concept could eventually be generalized to other providers that support multiple accounts as discussed in #1843.

steipete:
Thanks @Thinkscape for the account-item implementation and supplied screenshots. Closing this version because account items bypass saved layouts, create an uncapped item set, and retain the older status-item creation order. The accepted direction is selected account icons capped at four, replacing that provider’s item and disabled with Merge Icons. A maintainer implementation needs the provider-neutral account projection, current zero-width/name-before-registration factory, placement preservation, and account-scoped layout/menu data. This does not resolve #1167’s separate request for additive provider pins. No contributor revision is requested.

Issue #3247, opened 2026-08-27
0.55.1 (130): sustained ~92% CPU + 2.2 GB RSS within 35 min of launch, and ~468 orphaned WebKit.WebContent processes spawned overnight (~1 per 30–90s); orphans survive app quit and SIGKILL


robertoecf:
## Summary

Two related symptoms on the current release, possibly the cost-usage/WebKit surface again (see Related issues below):

1. **Sustained CPU spin + memory growth in the main process.** Fresh launch reached **91.7% CPU (≈1 full P-core) and 2.2 GB RSS in ~35 minutes**. System load average hit **202** on a 10-core machine and dropped to **31 within ~90s of quitting CodexBar**. RAM itself was fine (91% available), so this is pure CPU saturation.
2. **Mass overnight leak of WebKit content processes.** Found **468 `com.apple.WebKit.WebContent` processes** (mostly ~16 MB, state `Ss`, oldest zombies 8 days, bulk spawned over ~11h at roughly **1 per 30–90 seconds**, spawn rate stopped ~5h before measurement). CodexBar links `WebKit.framework` and its crash/relaunch timeline matches the leak window. **Attribution of the WebContent leak is circumstantial** (I could not read responsible-pid without sudo), but CodexBar is the only third-party WebKit host alive during the window besides Mail.

## Environment

- CodexBar **0.55.1 (build 130)** — latest release, auto-update on, last check 2026-08-26 23:07 UTC
- macOS 26.5.2 (25F84), Apple Silicon M1 Max, 64 GB
- Cost tracking enabled; `~/Library/Caches/CodexBar/cost-usage` = **324 MB** corpus
- `debugLogLevel = verbose`
- Multiple providers tracked (Codex + Claude + Cursor probes)

## Timeline (from process forensics)

- 2026-08-26 18:12 — CodexBar crash (`~/Library/Logs/DiagnosticReports/CodexBar-2026-08-26-181214.ips`)
- 2026-08-26 20:07 — relaunched (18h31m uptime by next morning)
- ~02:30 → ~09:40 — WebContent spawn window (etime distribution of the 468 orphans)
- Morning state: main process 2.7 GB, **load avg 202**, machine visibly crawling
- Quit + relaunch today: 91.7% CPU / 2.2 GB within 35 min (exact signature of #2538's ~91%)
- Quit again: load 202 → 31 in 90s; CPU of remaining processes normal

## Kill behavior of the orphans (matches #3073)

[...]

PR #3252, opened 2026-08-28
fix: preserve WebView acquisition ownership

steipete:
## Summary

Fix a reentrancy race in the OpenAI dashboard WebView cache. An acquire suspended during page preparation could outlive explicit eviction, then remove a newer entry for the same store. The replacement's weakly captured cleanup owner could disappear, making its lease release ineffective.

Cached, reused, and temporary acquisitions now share one preparation/retry loop. Explicit eviction invalidates only active work for the affected store; stale completions cannot remove replacements or retry. Ordinary lease release does not cancel concurrent temporary acquisitions. Leases retain their cleanup owner independently of the cache, and release/host-close operations are one-shot. Existing page handoff, navigation deadlines, single-timeout-retry policy, and deferred WebKit teardown are preserved.

The active-acquisition registry drains on completion and retains no historical store generations. The redundant temporary retry and new-entry release paths are removed. No provider/auth/cookie behavior, persistent data format, or visible UI is changed.

A test-only follow-up replaces dashboard concurrency tests' five scheduler-count readiness loops with Observation-backed pending-load notifications and bounded, one-shot waits. Readiness failures throw; scope cleanup drains fake loads and rejects late arrivals. Cancelled producers can still deliberately finish in stale-owner tests. Duplicate gate aliases and wrappers were removed, all 73 original assertions remain, and eight additional tests cover readiness, timeout, cancellation, and cleanup. This does not change dashboard production code.

## Verification

The executed baseline regression used nonpersistent stores and controlled preparation continuations, without network navigation. Both new and reused acquires failed two assertions: the replacement lost cache ownership, and its release did not register host cleanup. The repaired tests retain those observations and cover cancellation, stale success, invalidation during retry, per-store isolation, temporary survival across normal release, cache deallocation, duplicate release, and preserved-page handoff.

- Final focused run: **32 test functions / 68 invocations across two suites passed**, including 47 parameterized cases. The replacement interleavings repeat three times each.
- `make check`: passed, zero violations.
- Independent Codex autoreview: passed with no actionable P0–P2 findings.
- Initial full `make test` on `d890c3284be142d5c7772edc8f14955bd9deaabb`: failed in the existing `CodexLoginRunnerTests` detached-child output-drain case. Its timeout outcome passed, but expected `login-started` output was missing (`No output captured.`). A bounded diagnostic predecessor-group run reproduced the failure with both fixture milestones absent; it did not establish lost pipe bytes or an underlying startup cause. All temporary instrumentation was removed, with the original assertion and deadlines intact. A fresh restored-source run passed 27 focused login/subprocess/holder tests and `make check`. The login assertion also passed in the subsequent full run; its intermittent failure is not claimed fixed.
- Second full run on the identical original commit: reached group 67/80, then failed the existing dashboard concurrent-account ordering test. Its readiness helper reported a timeout after 80 ms because it budgeted 1,000 `Task.yield()` calls instead of an elapsed deadline or readiness event. The follow-up reproduced that helper failure with a delayed producer (one test, two issues), then passed 104 tests across eight dashboard suites, including 20 tests in the repaired file. Final `make check` and independent Codex autoreview passed. Neither earlier full run is a pass.
- Third full `make test`, on follow-up head `e2554a221a6837d48226bad95347049f83c2d71c`: failed with timeout exit 124. Groups 32 and 33 exceeded the unchanged 180-second group limit. All 12 isolated retries for group 32 passed; group 33 passed six isolated selections, then `CostUsagePerformanceGateTests` exceeded the same limit with no test-start output. Final statistics were 31 first-pass successful groups, two timed-out groups, one recovered group, and 19 isolated retries. The login assertion passed in this run. This is **not a full pass**; merge remains held while the isolated launch/timeout is diagnosed, without weakening assertions or increasing limits.
- Exact-head CI on `e2554a221a6837d48226bad95347049f83c2d71c`: [passed all eight jobs](https://github.com/steipete/CodexBar/actions/runs/33140855552), including both macOS test shards and Linux builds/tests. This does not turn the failed local run into a pass.

A single bounded diagnostic on the unchanged head reproduced the silent 180-second timeout. Stack samples prove the first performance test was executing its 1,500-file fixture creation, waiting in atomic file writes; no completed tests were evidenced. Silence therefore did not establish a startup stall. The test helper also used a separate process group and survived the parent-group timeout signal; a diagnostic watchdog terminated the verified child. The underlying storage slowdown is not claimed diagnosed.

The test-only follow-up in `a54e5bb0d1306a423bb757de2f6ce8e3a5c80232` repairs both demonstrated test infrastructure paths. Initial performance/fair-scheduling corpora now use exclusive creation, without unnecessary atomic publication or durability work before any reader exists. The shared atomic writer remains for replacement tests. File counts, bytes, dates, scanner passes, budgets, and existing assertions are unchanged. The suite runner tracks owned process births and observed sessions, keeps its direct child unreaped until cleanup, and verifies bounded TERM/KILL/drain before retrying. Legitimate separate sessions remain supported. Metadata errors fail closed; a confirmed Darwin metadata disappearance while exit is still pending uses the original deadline rather than failing early. The runner's default 180-second deadline and group size 12 are unchanged.

- Fixture/scanner proof: executed regression failures, then all 64 focused Swift test functions passed (four seed tests, 30 performance gates, eight fair-scheduling tests, and 22 scanner tests). The performance command completed in 148.122 seconds and fair scheduling in 95.455 seconds, both within 180 seconds on their first optimized attempt. This is not a controlled machine-performance benchmark.
- Runner proof: executed RED/GREEN for escaped process groups, separate-session ownership, metadata/initialization failures, and pending exits. Final shell harness: 50 Python tests passed; one native Linux pthread regression skipped on macOS. The unrelated sentinel remained alive, and no observed owned processes remained. All 34 actual Swift shell/subprocess/process-group tests passed under the unchanged 180-second command limit.
- Final `make check`: passed, zero violations in 2,036 files, including the expanded shell harness. Independent Codex autoreview passed with no actionable P0–P2 findings after the accepted ownership and platform-race findings were repaired.
- Full local `make test` on the exact source committed as `a54e5bb0d1306a423bb757de2f6ce8e3a5c80232`: **passed**, 955 selections across 80 groups, all 80 successful on the first attempt, zero failures/timeouts/retries, 1,031.3 seconds total. Both earlier cost-timeout groups and the dashboard concurrency group passed. [Exact-head CI passed all eight jobs](https://github.com/steipete/CodexBar/actions/runs/33147544169).

The subsequent PR review identified a concrete remaining safety gap in macOS descendant signaling: birth validation and `kill(pid)` were separate calls. Follow-up `b2807d81b1e6d18447784f57dd586f41bd37ef05` addresses it using `proc_signal_with_audittoken`. A combined native read supplies birth and generation from the same process, and the kernel binds the signal to that generation while enforcing normal caller permissions. Missing support or permission/I/O failures do not fall back to a numeric signal. Legitimate child sessions remain supported. The [Apple implementation](https://github.com/apple-oss-distributions/xnu/blob/xnu-11215.1.10/libsyscall/wrappers/libproc/libproc.c) is present in the macOS 15 source; runtime proof ran on macOS 26.6.2, not on a macOS 15 VM.

That safety follow-up has executed RED/GREEN for eight deterministic cases and two native tests. Real wrong-generation TERM and KILL calls returned ESRCH and left the owned child alive with unchanged identity; matching calls returned success, terminated and reaped the child, and left the unrelated sentinel alive. Final shell harness: 60 Python tests passed, one Linux-native skip on macOS. All 34 focused Swift process tests and `make check` passed, with zero lint violations and no observed owned-process leftovers. Independent Codex autoreview found no actionable P0–P2 issues in the follow-up. Swift/product/changelog files remained unchanged.

<details>
<summary>Native macOS final-effect test output</summary>

Executed on the same source bytes subsequently committed as `b2807d81b1e6d18447784f57dd586f41bd37ef05`; hashes of all three changed files matched before review, after worker handoff, and after commit. These are self-expiring, task-owned Python fixtures, not application or account processes. The test checks the child's identity and pending wait status after the rejected signal, then uses the production descendant-send path for the matching signal. `os.kill` is mocked during that production send and required not to be called. The actual system audit-token API is not mocked.

```sh
CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 \
CODEXBAR_TEST_CODEX_FILE_ISOLATION=1 \
CODEXBAR_TEST_RETRY_NON_TIMEOUT_FAILURES=0 LIVE_GEMINI_FETCH=0 \
python3 Scripts/test_swift_test_process_cleanup.py DarwinNativeSignalTests -v
```

The actual invocation also unset the Keychain opt-in, external Codex fixture path, spend/Pi proof paths, and `LIVE_TEST`. Extracted native stdout (identity arrays are PID, parent PID, birth seconds, birth microseconds, pidversion):

```text
{"signal": "SIGKILL", "before": [50055, 50037, 1787899462, 778887, 29317839], "after_wrong": [50055, 50037, 1787899462, 778887, 29317839], "wrong_pidversion": 29317838, "wrong_result": 3, "matching_results": [0], "reaped_status": -9, "unrelated_sentinel_alive": true}
{"signal": "SIGTERM", "before": [50078, 50037, 1787899463, 199161, 29317873], "after_wrong": [50078, 50037, 1787899463, 199161, 29317873], "wrong_pidversion": 29317872, "wrong_result": 3, "matching_results": [0], "reaped_status": -15, "unrelated_sentinel_alive": true}
Ran 2 tests in 0.843s
OK
```

Result 3 is ESRCH; zero is native success. Both TERM and KILL fixtures were reaped with their expected signal status. These native cases also passed in the subsequent shell harness and `make check`. No macOS 15 VM execution or forced OS PID wraparound is claimed.

</details>

The first full local `make test` on `b2807d81b1e6d18447784f57dd586f41bd37ef05` **failed** in the unchanged Kiro PTY hard-stop fixture test. Its hard-stop latency was 1.151 seconds (within the original three-second assertion), but the subsequent 20-second wait for all four delayed fixture processes to exit timed out. Group 45 failed after 44 first-pass successful groups; there were zero group timeouts and zero retries, with 851.2 seconds total. The earlier login and cost groups passed; the following Kiro settle-exit case also passed. The failure does not identify which fixture process remained or establish its state.

Exactly two bounded diagnostic attempts followed: 57 Kiro tests passed, then all 148 tests in the original 12-suite group passed, both within the unchanged 180-second command limit. Temporary observations showed all four fixture identities disappear before defer cleanup, with no observed PID reuse or zombie; neither outer-runner drain sent a signal. The original failure did not reproduce, and no Kiro fix, scheduler cause, or lease-expiry cause is claimed. All instrumentation was removed; original source hashes and the restored test binary hash match the pre-diagnostic versions, and all 356 observed owned process identities were absent in the final census. No deadlines or assertions changed.

The fresh full local run on restored `b2807d81b1e6d18447784f57dd586f41bd37ef05` also failed, this time in the unchanged Claude CLI session environment/reuse fixture. Its first capture contained only the PTY `/status` echo, and the launch log contained only the second profile; the second and reused account assertions passed. Group 15 failed after 14 first-pass successful groups, with zero group timeouts/retries and 413.6 seconds total. The functional fixture used a 100 ms idle cutoff, which could complete on input echo before its Account response.

The test-only Claude repair now waits for the expected fixture Account response with idle completion disabled. All 23 original assertions and the two/five-second capture deadlines remain unchanged; production idle policy is untouched. A controlled 0.5-second response delay produced three echo-only Account failures with the old setup, then both immediate/delayed cases and all 22 selected tests across five suites passed. Cancellation error paths now await their owned tasks before fixture teardown. Independent Codex autoreview found no actionable findings.

The subsequent `make check` failed in a separate synthetic process success fixture (`124 != 0`): its fixed 1.2-second ancestry sleep consumed most of a two-second command budget. A test-only follow-up replaces that sleep with acknowledgment after the real runner observes matching ready child identities, including the session-tree grandchild. A controlled one-second startup delay failed the old fixed-wait setup and passed the repaired case; all original cleanup/sentinel/deadline assertions remain. A diagnostic outer wrapper that recursively ran the process harness inside another copy of the runner failed closed on lost nested-session continuity; no test pass is claimed for that attempt, and all recorded fixture identities were subsequently absent. Through the normal direct entry point, 16 of 17 focused tests passed, including the delayed regression and native signal/waitid tests; the first failure-exit fixture still exceeded two seconds. This is not an all-green result. The follow-up has a clean independent review, but a single bounded timing diagnostic and aggregate validation remain in progress. No production runner change or higher deadline is being used to conceal these failures.

[Exact-head CI on b280 passed all eight jobs](https://github.com/steipete/CodexBar/actions/runs/33149390352), including both macOS shards. That does not convert either failed local run into a pass; merge remains held for the focused repair and final validation.

The Claude and process-fixture test repairs are now committed in `78d5909b78c9693b2dfcc9c78aa3ea49494c7b78`. One bounded diagnostic of the remaining failure-exit case passed with status 23 and verified fixture cleanup; it did not reproduce or diagnose the earlier timeout. The next aggregate attempt was interrupted by an external 180-second supervisor, which was an inappropriate cap for the whole `make check`, not a repository deadline. A subsequent normal invocation without that cap stalled inside the sharding shell before test execution. A sample showed Bash 5.3.15 blocked in `heredoc_write`/`write`; after terminating only that verified task-owned shell, it reported a here-document temporary-file space error. The volume still had 190 GiB free, so a full disk or underlying kernel cause is not established. Neither aggregate is a pass, and no test deadline has been increased.

The full local run on `78d5909b78c9693b2dfcc9c78aa3ea49494c7b78` failed with timeout 124: group 5 exceeded 180 seconds, seven isolated selections passed, and the eighth, `AdaptiveRefreshHeuristicsTests`, exceeded the same limit. Final statistics were four first-pass successful groups, one timed-out group, eight isolated selections, zero recovered groups, and 617.5 seconds total. A sample during the isolated run located test setup in `disableAllProviders` repeatedly persisting the config through atomic writes (588 rename and 180 open samples). The underlying storage slowdown is not claimed diagnosed.

Follow-up `797452d9caa852690ac0f5031cdd941475880d5c` uses the existing file-backed settings fixture seed instead of replaying a synchronous mutation for every provider. Both adaptive heuristics and timer fixtures retain provider defaults and explicitly keep OpenAI web access off; merely preloading a config would otherwise change that default. All 29 original test bodies and timer wait helpers remain verbatim, and three new tests check disabled-provider state and Codex-only stubbing. Production persistence, test deadlines, and timer behavior are unchanged. All 49 focused tests across five suites passed; normal `make check` passed, including 61 Python passes and one Linux-native skip, with zero lint violations. Independent Codex autoreview found no actionable P0–P2 findings. A persisted-state audit checked 37 new configs with all 69 entries and web cookies off; only the three intended stub fixtures enabled Codex. Known owned processes and task fixture paths were cleaned up.

The initial combined build/discovery command hit its 180-second wrapper bound after compilation completed in 149.95 seconds. Discovery against the built products then passed in 9.378 seconds; no failing assertion was retried or deadline increased. The final full local run on `797452d9caa852690ac0f5031cdd941475880d5c` **passed**: 955 selections across 80 groups, 79 first-pass successful groups, one 180-second group timeout recovered through all 12 isolated selections, zero full-group retries, and 1,912.4 seconds total. The previously failing adaptive, Claude, login, Kiro, and dashboard cases passed. A sample during the isolated cost-progress suite showed SQLite checkpoint work; no further code change was made for that recovered timeout. [Exact-head CI for 797452](https://github.com/steipete/CodexBar/actions/runs/33157883853) **passed**, including both macOS shards and Linux x64/ARM64/musl. The earlier run history below is retained; those failures are not retroactively passes.

The preceding 78d590 CI run ultimately became cancelled after the new push, but its Linux x64 job had already **failed**; cancellation does not erase that failure. The EMFILE child passed its exhaustion scenario and then crashed with SIGILL during its subsequent 100-capture pipe-release test. The stack entered libdispatch's source-finalization path. Swift 6.3.3's [matching source check](https://github.com/swiftlang/swift-corelibs-libdispatch/blob/swift-6.3.3-RELEASE/src/source.c#L591) rejects double finalization; a Linux hangup/unregistration interleaving is a source-supported lead, not a demonstrated cause. The production capture and Linux test files are unchanged by this branch. The whole child sequence remains intact: narrowing its filter would hide the post-exhaustion work that exposed the crash. No Linux crash fix or successful reproduction is claimed. Two supplemental AWS Crabbox attempts stalled during lease acquisition, before issuing a lease ID or running tests. Their local clients were stopped and subsequent AWS inventory was empty; no extra remote stress-test pass is claimed.

The bounded shell-bootstrap diagnosis completed without a reproduction: the exact 1,667-byte literal passed once under Bash 5.3 default, that same binary's disposable compatibility-50 mode, and system Bash 3.2; a 64 KiB native write/fsync/read/delete probe also passed. All four children were reaped without signals. [GNU Bash's source](https://git.savannah.gnu.org/cgit/bash.git/plain/redir.c?h=bash-5.3) uses a common here-document error path for pipes and temporary files and can synthesize ENOSPC after a short write. This explains why the wording alone does not prove a full disk, not why the original process stalled. No compatibility fallback, runtime swap, or source workaround was added. The later normal `make check` passed its shell bootstrap without those changes.

Process containment is a bounded metadata poller, not a daemon sandbox: it cannot reliably discover a new session whose ancestry and leader both vanish between observations. Unknown ownership fails rather than signaling uncertain PIDs or claiming successful cleanup. The initial `swift test list` discovery/build path is unchanged. These limits and the native Linux regression are documented in `docs/DEVELOPMENT.md`.

The focused command was:

```sh
swift test --no-parallel \
  --filter 'OpenAIDashboardWebViewCacheTests|WebKitTeardownTests' \
  --skip 'OpenAIDashboardWebViewCacheTests/.*(WKWebsiteDataStore should|same email profile|live website data store|Sequential fetches with OpenAIDashboardWebsiteDataStore)'
```

The four excluded tests are unchanged persistent-store factory tests; the focused proof intentionally uses only nonpersistent WebViews. Tests ran with Keychain suppression, isolated Codex fixtures, non-timeout retries disabled, and live Gemini fetching disabled. The initial incorrect RED filter matched no tests and was corrected using actual discovery; only the subsequent executed failure is counted as RED. An intermediate compile error was fixed by making the UI-owned Entry explicitly MainActor-isolated, without adding unsafe Sendable conformance.

## Scope and limits

Cleanup proof checks an actual host's single cleanup request and registration with the existing teardown owner, independently of dictionary counts. It does not prove eventual window cleanup or WebContent process termination. Existing headless CI guards remain unchanged, so the local nonpersistent run supplies the AppKit lifecycle execution proof.

This independent ownership defect was found while investigating #3247, but this PR does **not** reproduce or establish resolution of that report's sustained CPU/RSS growth or older WebContent episode. It must not close that issue. No real account, provider, cookie store, or normal app-bundle launch was used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 0.55.2 — Unreleased
 
 ### Fixed
-- Codex: keep OpenAI web acquisitions and lease cleanup owned across cache eviction, preventing stale preparation results or timeout retries from displacing replacement views.
+- Codex: keep OpenAI web acquisitions and lease cleanup owned across cache eviction, preventing stale preparation results or timeout retries from displacing replacement views (#3252).
 - Codex: reduce cached cost-read overhead by loading progress metadata separately and leaving raw scanner state out of cached reports, including project and session details (#3247). Thanks @robertoecf!
 - Antigravity: read recognized local SQLite conversations as bounded token-only history, preserving unavailable coverage, session identity, and overflow-safe totals (#3212). Thanks @Yuxin-Qiao!
 - Usage & Spend: refresh outdated independent 365-day histories after regular token publications, preserving older rows and coalescing updates during scans (investigated alongside #3209, #3176). Thanks @vinschger!
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.55.2 — Unreleased
 
 ### Fixed
+- Codex: keep OpenAI web acquisitions and lease cleanup owned across cache eviction, preventing stale preparation results or timeout retries from displacing replacement views.
 - Codex: reduce cached cost-read overhead by loading progress metadata separately and leaving raw scanner state out of cached reports, including project and session details (#3247). Thanks @robertoecf!
 - Antigravity: read recognized local SQLite conversations as bounded token-only history, preserving unavailable coverage, session identity, and overflow-safe totals (#3212). Thanks @Yuxin-Qiao!
 - Usage & Spend: refresh outdated independent 365-day histories after regular token publications, preserving older rows and coalescing updates during scans (investigated alongside #3209, #3176). Thanks @vinschger!
--- a/Scripts/ci_swift_test_by_suite.py
+++ b/Scripts/ci_swift_test_by_suite.py
@@ -104,20 +104,63 @@ if sys.platform == "darwin":
             (name, ctypes.c_uint32) for name in ("files", "group", "jobc", "tdev", "tpgid", "nice")
         ] + [("seconds", ctypes.c_uint64), ("microseconds", ctypes.c_uint64)]
 
+    class ProcUniqueIdentifierInfo(ctypes.Structure):
+        # sys/proc_info_private.h: 56-byte API layout, including unused reserved fields.
+        _fields_ = [
+            ("uuid", ctypes.c_uint8 * 16), ("uniqueid", ctypes.c_uint64),
+            ("parent_uniqueid", ctypes.c_uint64), ("pidversion", ctypes.c_int32),
+            ("reserved", ctypes.c_uint32), ("reserved2", ctypes.c_uint64), ("reserved3", ctypes.c_uint64),
+        ]
+
+    class ProcBSDInfoWithUniqueID(ctypes.Structure):
+        _fields_ = [("bsd", ProcBSDInfo), ("unique", ProcUniqueIdentifierInfo)]
+
+    class AuditToken(ctypes.Structure):
+        _fields_ = [("val", ctypes.c_uint32 * 8)]
+
     _libproc = ctypes.CDLL("/usr/lib/libproc.dylib", use_errno=True)
     _libproc.proc_pidinfo.argtypes = [ctypes.c_int, ctypes.c_int, ctypes.c_uint64, ctypes.c_void_p, ctypes.c_int]
     _libproc.proc_pidinfo.restype = ctypes.c_int
+    _proc_signal_with_audittoken = getattr(_libproc, "proc_signal_with_audittoken", None)
+    if _proc_signal_with_audittoken is not None:
+        _proc_signal_with_audittoken.argtypes = [ctypes.POINTER(AuditToken), ctypes.c_int]
+        _proc_signal_with_audittoken.restype = ctypes.c_int
+
+
+def darwin_pidinfo(pid: int, flavor: int, info):
+    ctypes.set_errno(0)
+    size = _libproc.proc_pidinfo(pid, flavor, 0, ctypes.byref(info), ctypes.sizeof(info))
+    if size != ctypes.sizeof(info):
+        error = ctypes.get_errno() if size <= 0 else errno.EIO
+        raise OSError(error or errno.EIO, f"Cannot read process metadata for PID {pid}")
+    return info
+
+
+def darwin_signal_process(info: TestProcess, sig: signal.Signals) -> None:
+    if _proc_signal_with_audittoken is None:
+        raise OSError(errno.ENOSYS, "Darwin process cleanup requires proc_signal_with_audittoken")
+    try:
+        # PROC_PIDT_BSDINFOWITHUNIQID reads birth and generation from one held proc.
+        current = darwin_pidinfo(info.pid, 18, ProcBSDInfoWithUniqueID())
+    except (FileNotFoundError, ProcessLookupError):
+        return
+    if current.bsd.pid != info.pid or (current.bsd.seconds, current.bsd.microseconds) != info.birth:
+        return
+    # Only target PID/pidversion are used; these are not caller credentials. XNU
+    # checks the real caller's permissions and binds the signal to this generation.
+    token = AuditToken()
+    token.val[5], token.val[7] = info.pid, current.unique.pidversion
+    error = _proc_signal_with_audittoken(ctypes.byref(token), sig)
+    # libproc returns errno directly, not -1/errno. An exec can stale this token;
+    # a later cleanup attempt obtains a fresh generation only after matching birth.
+    if error and error != errno.ESRCH:
+        raise OSError(error, f"Cannot signal process identity for PID {info.pid}")
 
 
 def test_process(pid: int) -> TestProcess | None:
     try:
         if sys.platform == "darwin":
-            info = ProcBSDInfo()
-            ctypes.set_errno(0)
-            size = _libproc.proc_pidinfo(pid, 3, 0, ctypes.byref(info), ctypes.sizeof(info))
-            if size != ctypes.sizeof(info):
-                error = ctypes.get_errno() if size <= 0 else errno.EIO
-                raise OSError(error or errno.EIO, f"Cannot read process metadata for PID {pid}")
+            info = darwin_pidinfo(pid, 3, ProcBSDInfo())
             zombie = info.state == 5
             try:
                 session = 0 if zombie else os.getsid(pid)
@@ -237,6 +280,8 @@ class TestProcessOwnership:
                 return
             if descriptor is not None:
                 signal.pidfd_send_signal(descriptor, sig)
+            elif sys.platform == "darwin":
+                darwin_signal_process(info, sig)
             else:
                 os.kill(info.pid, sig)
         except ProcessLookupError:
--- a/Scripts/ci_swift_test_by_suite.py
+++ b/Scripts/ci_swift_test_by_suite.py
@@ -4,7 +4,10 @@
 from __future__ import annotations
 
 import argparse
+import ctypes
+import errno
 import os
+from pathlib import Path
 import re
 import signal
 import subprocess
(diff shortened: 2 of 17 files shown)


steipete:
Landed the WebView acquisition/lease ownership repair from source head `797452d9caa852690ac0f5031cdd941475880d5c` as [69071035809c5fb177a8d8cd5510cf8bbf16b23f](https://github.com/steipete/CodexBar/commit/69071035809c5fb177a8d8cd5510cf8bbf16b23f). Explicit eviction can no longer let a suspended acquisition remove its replacement, and a released lease retains a one-shot cleanup owner independently of the cache. Existing navigation deadlines, one timeout retry, preserved-page handoff, and normal temporary acquisitions remain supported. The changelog entry is under **0.55.2 — Unreleased**.

The regression was executed before and after the repair using nonpersistent WebViews and controlled preparation continuations, without accounts or network navigation. The original new/reused interleavings produced four assertion failures. The repaired focused run passed **32 functions / 68 invocations** across the cache and teardown suites:

```sh
swift test --no-parallel \
  --filter 'OpenAIDashboardWebViewCacheTests|WebKitTeardownTests' \
  --skip 'OpenAIDashboardWebViewCacheTests/.*(WKWebsiteDataStore should|same email profile|live website data store|Sequential fetches with OpenAIDashboardWebsiteDataStore)'
```

Final local gates on the committed source:

```sh
env -u CODEXBAR_ALLOW_TEST_KEYCHAIN_ACCESS \
  -u CODEXBAR_TEST_CODEX_FILE_FIXTURES \
  -u CODEXBAR_SPEND_PROOF_DIR -u CODEXBAR_PI_PROOF_DIR -u LIVE_TEST \
  -u CI -u GITHUB_ACTIONS \
  -u CODEXBAR_TEST_SHARD_INDEX -u CODEXBAR_TEST_SHARD_COUNT \
  -u CODEXBAR_TEST_GROUP_SIZE -u CODEXBAR_TEST_SUITE_TIMEOUT \
  CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 \
  CODEXBAR_TEST_CODEX_FILE_ISOLATION=1 \
  CODEXBAR_TEST_RETRY_NON_TIMEOUT_FAILURES=0 LIVE_GEMINI_FETCH=0 \
  make test
make check
```

`make test` **passed**: 955 selections, 80 groups, 79 first-pass successful groups, one 180-second group timeout recovered through all 12 isolated selections, no full-group retries, 1,912.4 seconds total. The isolated cost-progress and fair-scheduling suites passed in 79.205 and 31.668 seconds; a sample showed SQLite checkpoint work, not a demonstrated new defect. `make check` passed under the same account-isolation settings, including 61 Python passes and one Linux-native skip on macOS, with zero format/lint violations. Independent Codex autoreviews reported no actionable findings; reviewed source hashes and the clean committed checkout were rechecked after the full run. [Exact-head CI passed all eight jobs](https://github.com/steipete/CodexBar/actions/runs/33157883853), including both macOS shards, Linux x64 and ARM64 builds/tests, musl, lint, and the aggregate gate. Final PR review found no introduced P0–P2 defect and there were no inline review comments; I accepted the bounded cleanup scope based on the native, focused, and exact-head cross-platform proof.

The expanded validation code fixes demonstrated infrastructure weaknesses, not production provider behavior. Dashboard readiness now observes actual pending loads instead of counting scheduler yields; Claude fixtures wait for their Account response; process fixtures acknowledge observed child identities instead of sleeping; initial private corpora/configs are seeded before readers exist. Original behavioral assertions and execution budgets remain. Focused follow-up proof passed 104 dashboard tests, 64 fixture/scanner tests, 34 process tests, 22 Claude-related tests, and 49 adaptive-refresh tests. The runner tracks owned process identities and drains them before retrying; native macOS wrong-generation TERM/KILL calls returned ESRCH, matching calls terminated and reaped the child, and an unrelated sentinel survived. The bounded poller does not claim to contain unobserved daemonization.

Earlier failed runs remain documented in the PR body. The initial login-output diagnostic reproduced missing startup milestones without establishing their cause; Kiro's failure did not reproduce in its bounded diagnostics. Restored-source focused checks and the final full run passed, but neither intermittent failure is claimed fixed. Dashboard readiness, corpus seeding/process containment, Claude response timing, process fixture readiness, and adaptive configuration seeding had concrete diagnostics and targeted repairs. A shell-bootstrap stall did not reproduce and received no compatibility workaround. An older Linux x64 CI run genuinely crashed in libdispatch after its EMFILE scenario passed; that capture/test code is unchanged by this branch. Current-head Linux x64, ARM64, musl, and lint CI passed, but that does not establish a fix for the intermittent crash. Two supplemental AWS Crabbox attempts stalled before returning a lease or running tests; their local clients were stopped and subsequent AWS inventory was empty. No extra remote stress-test pass is claimed.

The lifecycle assertions prove cleanup requests and registration with the existing deferred teardown owner, not WebContent termination or measured CPU/RSS improvement. Four unchanged persistent-store factory tests were excluded from the focused run; existing headless-CI guards remain. No real account, provider, cookie store, normal app launch, or macOS 15 VM was used. **Issue #3247 remains open.** The visible checkout is back on clean, fast-forwarded `main`; the merge is reachable and its tree matches the tested source head. This is a merge, not a release.

Issue #3249, opened 2026-08-27
KeychainAccessPreflight blocks background quota refresh after app updates even when Keychain access is disabled


jaychou0642-create:
### Environment
- **macOS Version**: macOS 15+ (Apple Silicon)
- **CodexBar Version**: Latest
- **Enabled Providers**: Codex & Antigravity (No Claude Code / No Web Cookies)
- **Setting**: `debugDisableKeychainAccess = 1` (`openAIWebAccessEnabled = 0`)

### Description
Every time CodexBar updates to a new version via Sparkle / auto-update, the app fails to fetch quotas for all providers (including file-based providers like Codex `~/.codex/auth.json` and Antigravity CLI session).

The quota refresh is completely blocked until the user manually opens macOS `Keychain Access.app` (which refreshes system `securityd` sessions) and restarts CodexBar.

### Root Cause Analysis
1. In `CodexBarCore`, `KeychainAccessPreflight` (`SecTrustedApplicationValidateWithPath` / `com.steipete.codexbar.keychain-preflight`) runs unconditionally during app initialization.
2. When the app updates, the binary hash changes. macOS Keychain detects a modified binary and requires interactive authorization (`Keychain preflight requires interaction`).
3. Since CodexBar is a menu bar accessory (`LSUIElement`), background keychain interaction is not allowed / suppressed, causing the preflight check or keychain cache store to hang/fail silently.
4. This preflight blockage occurs **even when `debugDisableKeychainAccess` is enabled**, because the preflight / cache store initialization runs before or independently of the gate, blocking the main quota refresh pipeline for non-keychain providers.

### Expected Behavior
- When `debugDisableKeychainAccess` is enabled, CodexBar should completely bypass `KeychainAccessPreflight` and `KeychainCacheStore`.
- File-based providers (Codex OAuth file, Antigravity) should continue to refresh without being blocked by Keychain ACL validation failures.

### Workaround
Opening `Keychain Access.app` manually, unlocking the session, and relaunching CodexBar temporarily resolves the issue until the next app update.

steipete:
[...]

PR #3258, opened 2026-08-28
fix: honor disabled Keychain access during startup

steipete:
## Summary

Apply the saved “Disable Keychain access” preference before startup credential migration, using the same local/current-shared defaults resolution as Settings. An explicit local value still wins, including `false`; environment/process disabling remains enforced by the existing gate.

Keep cleanup and the legacy-migration completion marker pending while Keychain access is disabled. A skipped credential read must not permanently mark still-present credentials as migrated. Permitted manual-default and file-backed imports can still be saved, and a later enabled launch retries normally.

Update the changelog and Keychain troubleshooting documentation. This is a narrower startup hardening found while investigating #3249, **not** a demonstrated explanation of that reporter’s all-provider post-update failure; the issue should remain open.

## Verification

- Added fake-store startup regression: failed on the previous implementation for both stored `true` and `false`, and passes after the fix without reading credential stores.
- Added all nine local/shared preference combinations and a two-launch deferred-migration/recovery regression.
- Startup, preference and Keychain safety suites passed. The first combined run found only an existing provider-ownership source anchor shifted by seven lines; its unchanged eight-reference fingerprint was checked, its line relocated, and the guard passed in the full run.
- `make check` passed, including strict SwiftLint with zero violations.
- Codex autoreview passed at its configured P0 threshold; independent source review also checked the initialization and migration boundary.
- Full `make test` passed: 956 selections across 80 groups, all successful on the first pass, zero retries/timeouts (25m05s total).
- [Exact-head hosted CI passed](https://github.com/steipete/CodexBar/actions/runs/33209935551): lint, Linux x64/arm64/musl builds, both macOS shards, plugin golden tests, and the final gate. Tests suppress real Keychain item access and use isolated defaults/config fixtures. No live credentials, ACL changes, or user settings changes were used as proof.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 0.56.1 — Unreleased
 
+### Fixed
+- Keychain: apply saved disabled-access preferences before startup credential migration, including shared-defaults fallback, and keep deferred migrations retryable (investigated alongside #3249). Thanks @jaychou0642-create!
+
 ## 0.56.0 — 2026-08-28
 
 ### Added
--- a/Sources/CodexBar/CodexbarApp.swift
+++ b/Sources/CodexBar/CodexbarApp.swift
@@ -64,7 +64,7 @@ struct CodexBarApp: App {
                 "built": buildTimestamp,
             ])
 
-        KeychainAccessGate.isDisabled = UserDefaults.standard.bool(forKey: "debugDisableKeychainAccess")
+        KeychainAccessGate.isDisabled = SettingsStore.loadDebugDisableKeychainAccess(userDefaults: .standard)
         KeychainPromptCoordinator.install()
         if MainThreadHangWatchdog.isEnabledForCurrentProcess {
             MainThreadHangWatchdog.shared.start()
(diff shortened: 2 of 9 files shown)


steipete:
Landed on main as [9769d7394926](https://github.com/steipete/CodexBar/commit/9769d739492656bbb4e1e0e572b2e80cda2df429). Startup now applies the resolved local/current-shared disabled-access preference before credential migration. Deferred reads no longer allow cleanup or a completed-migration marker to prevent a later retry.

Verification on the landed patch:

- `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 CODEXBAR_TEST_CODEX_FILE_ISOLATION=1 make test` — 956 selections, 80/80 groups passed on the first attempt; zero retries, failures, or timeouts.
- `make check` — passed; strict SwiftLint reported zero violations across 2,037 files.
- The fake-store startup regression failed against the old implementation for both saved values, then passed with the fix. All nine local/shared precedence combinations, both deferred-recovery cases, and the source-ownership gate passed in the final full run. The initial focused run caught a shifted ownership anchor; its unchanged fingerprint was verified and the line corrected.
- Codex autoreview passed at its configured P0 threshold; independent source/security review found no actionable defects.
- [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/33209935551) passed every job, including both macOS shards, all Linux variants, and provider-plugin golden tests. The merged tree matches the tested head.

No live credential read, Keychain ACL change, or user-account mutation was used as proof. #3249 remains open because its broader post-update failure still needs a concrete runtime trace. Changelog and documentation are updated under 0.56.1 Unreleased; this is not a release announcement.

PR #3263, opened 2026-08-29
test: isolate RPC fixtures and prove hung requests

steipete:
## Summary

Make the RPC timeout tests verify the intended failure mode, and remove unrelated login-shell discovery from absolute-interpreter Codex fixtures.

The hanging fixture matched only unescaped JSON method names. Foundation can serialize slashes as `\/`, so the fixture could respond with the wrong request ID instead of entering its hanging branch; the client still timed out and the test passed for the wrong reason. The fixture now handles both encodings, records the intended request before hanging, and the tests assert that evidence for every request. A direct fixture regression covers both wire spellings and waits for complete marker contents rather than just file creation.

Absolute-interpreter Codex fixtures inject the existing executable-resolution seam with an empty login PATH, so they do not consult the user's login shell. This applies to the fallback suite and adjacent Codex teardown tests. The hanging shell uses `exec /bin/sleep` so teardown owns the actual hanging PID rather than leaving a separate sleep child.

No production files, RPC deadlines, existing elapsed-time assertions, provider behavior, or credentials changed. This is test-only, so no user-facing changelog entry is needed. It removes an avoidable startup dependency; it does not claim to fix every host dynamic-loader startup stall observed during earlier triage.

## Verification

- Before correcting the matcher, the new evidence checks reproduced missing request markers in both real RPC timeout tests and the explicit escaped-slash case. That run also found a marker-file creation/read race (fixed) and an unrelated Grok initialization timeout while the two suites ran in parallel. None of those failures is counted as green.
- `swift test --jobs 8 --no-parallel --filter 'CodexUsageFetcherFallbackTests|RPCChildProcessTeardownTests'`: **20 tests passed** in 23.823 seconds. This uses the same serialized execution mode as CI, with Keychain access suppressed and Codex file isolation enabled.
- Independent source review and Codex autoreview: **clean**.
- `make check`: **passed**, including strict lint with zero violations in 2,039 files.
- Full `make test`: **passed**, all 959 selections across 80 groups. 78 groups passed first time; two failed first passes recovered on the runner's automatic full-group retries. There were no group watchdog timeouts or isolated-selection retries. Total elapsed time: 2,062.4 seconds.
- Local retry caveats: group 29 hit a Codex fixture initialization stall, correctly rejected by the new method/marker assertions; group 35 exceeded the unchanged cost-storage scale-test write budget. Both full-group retries passed. No test deadline was relaxed. The local run used the existing 600-second outer group watchdog on this loaded host; individual test and production deadlines remained unchanged.
- [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/33237563595) for `2a8012533b3db8aa6c003e51d044c5782c9d7442`: **passed**, including both macOS test shards, Linux x64/ARM64 builds and tests, lint, and the aggregate gate. GitGuardian also passed. The existing test-only path gate intentionally skipped the musl build. No CI rerun was requested.

### DIFF
--- a/Tests/CodexBarTests/CodexUsageFetcherFallbackTests.swift
+++ b/Tests/CodexBarTests/CodexUsageFetcherFallbackTests.swift
@@ -152,12 +152,13 @@ struct CodexUsageFetcherFallbackTests {
     @Test
     func `hung CLI RPC rate limits request times out within budget`() async throws {
         let stubCLIPath = try self.makeHungRateLimitsStubCodexCLI()
-        defer { try? FileManager.default.removeItem(atPath: stubCLIPath) }
+        let requestPath = stubCLIPath + ".requests"
+        defer {
+            try? FileManager.default.removeItem(atPath: stubCLIPath)
+            try? FileManager.default.removeItem(atPath: requestPath)
+        }
 
-        let fetcher = UsageFetcher(
-            environment: ["CODEX_CLI_PATH": stubCLIPath],
-            initializeTimeoutSeconds: 20.0,
-            requestTimeoutSeconds: 0.2)
+        let fetcher = self.makeStubUsageFetcher(stubCLIPath, requestTimeoutSeconds: 0.2)
 
         let started = Date()
         do {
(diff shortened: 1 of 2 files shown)


steipete:
Landed as b366a2d5aa52047524a8b9177a99e2a1c1eedd70 after reviewing the exact two-file test-only change. The fixtures now prove the rate-limits request actually reached the hanging branch for either JSON slash encoding, and absolute-interpreter fixtures no longer discover the user's login-shell PATH. Production behavior and existing RPC/elapsed deadlines are unchanged.

Verification:

- `swift test --jobs 8 --no-parallel --filter 'CodexUsageFetcherFallbackTests|RPCChildProcessTeardownTests'`: 20 tests passed in 23.823 seconds.
- `make check`: passed, including strict lint with zero violations across 2,039 files.
- `env -u CODEXBAR_ALLOW_TEST_KEYCHAIN_ACCESS -u CODEXBAR_TEST_CODEX_FILE_FIXTURES CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 CODEXBAR_TEST_CODEX_FILE_ISOLATION=1 CODEXBAR_TEST_SUITE_TIMEOUT=600 NSUnbufferedIO=YES make test`: passed all 959 selections across 80 groups in 2,062.4 seconds. 78 groups passed first time; two recovered on the runner's automatic full-group retry. No group watchdog timeouts or isolated-selection retries.
- The two local first-pass failures were a fixture initialization stall in group 29 and the unchanged cost-storage scale-test write budget in group 35. Both entire-group retries passed. This is not a claim that all host startup stalls are fixed, and no test threshold was weakened.
- Independent source review and Codex autoreview found no actionable issue.
- [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/33237563595) passed on 2a8012533b3db8aa6c003e51d044c5782c9d7442: both macOS shards, Linux x64/ARM64 builds and tests, lint, and aggregate gate; GitGuardian also passed. The musl build was intentionally skipped by the existing test-only path gate. No CI rerun was requested.

No app relaunch, real account/provider probe, or live Keychain access was used. No changelog entry is needed for this test-only repair.

PR #3266, opened 2026-08-29
fix(antigravity): recover newer session history safely

chid:
## Summary

Recover Antigravity token-history timestamps from the `steps` table when newer `gen_metadata` envelopes omit their embedded timestamp. The primary pass still parses usage/model data from `gen_metadata`; only timestamp-less turns with a step UUID trigger the optional secondary lookup.

The repaired fallback is deliberately bounded and fail-closed:

- step rows share the job-wide 50,000-row, byte, per-database byte, duration, and cancellation budgets
- optional schema discovery runs only when an undated turn actually needs it
- the lookup can reach valid matches beyond 10,000 unrelated rows
- generation and step occurrences align by validated stored indices, including mixed embedded/fallback turns
- recovered and embedded events rejoin in original generation-index order before first-wins deduplication
- duplicate indices, invalid or missing timestamps, ambiguous reused-step counts, malformed matching rows, truncated scans, and SQLite errors keep coverage partial instead of inventing dates
- `NULL`, non-blob, and UUID-less step rows still count as ambiguous occurrences, so one later valid timestamp cannot be stretched across multiple turns
- one timestamp may be shared across multiple turns only when it is the sole valid matching step occurrence

This PR remains scoped to the proven parser/CLI recovery. Two later contributor commits adding an immutable-WAL retry plus menu/widget presentation were removed from this branch and preserved separately for follow-up: immutable reads need a stable private snapshot or equivalent race protection, and the UI work still needs configured-window correctness, cache invalidation, and sanitized before/after proof.

The repaired implementation retains @chid's original parser discovery, contributor commits, and live-session validation; the additional maintainer commits tighten its snapshot, budget, ordering, and ambiguity boundaries.

## Verification

- Contributor live proof on 37 sanitized local Antigravity session databases established 30-day history with `steps` present and correctly withheld it from same-data copies with `steps` removed.
- `AntigravityLocal*`: 80 tests across eight suites passed.
- Full isolated local suite on current main ancestry: 997 selections across 84 groups, all first-pass green with retries disabled, zero failed groups, and zero timeouts.
- `make check` passed; SwiftFormat clean and SwiftLint reported zero violations.

Maintainer verification used synthetic fixtures and repository isolation flags; no real local account data, credentials, browser imports, or installed-app launches were accessed.

### DIFF
--- a/Sources/CodexBarCore/Providers/Antigravity/AntigravityLocalSQLite.swift
+++ b/Sources/CodexBarCore/Providers/Antigravity/AntigravityLocalSQLite.swift
@@ -101,9 +101,9 @@ extension AntigravityLocalReader {
             if let failure = progress.failure { throw failure }
             return SourceResult(isComplete: false)
         }
-        let tables = try self.supportedSQLiteTables(database, budget: budget)
+        let supported = try self.hasSupportedSQLiteTable(database, budget: budget)
         if let failure = progress.failure { throw failure }
-        guard tables.hasGenMetadata else { return SourceResult(isComplete: false) }
+        guard supported else { return SourceResult(isComplete: false) }
         sqlite3_limit(database, SQLITE_LIMIT_LENGTH, Int32(maximumValueBytes))
         var statement: OpaquePointer?
         defer { sqlite3_finalize(statement) }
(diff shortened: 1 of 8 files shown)


urda:
Second real-session data point for this PR, on CodexBar 0.56.3 with agy 1.1.23 (macOS).

`codexbar cost --provider antigravity --refresh` reports:

    Antigravity Token History
    Local token history is unavailable or incomplete.
    Local token history · dollar costs unavailable

Local history is 200 databases under `~/.gemini/antigravity-cli/conversations`.
I decoded the `gen_metadata` payloads with a raw protobuf walker:

| Rows | Count |
|---|---|
| With a `chatModel.#9.#4` timestamp | 976 |
| Without one | 701 |

No database mixes both shapes. Every database modified on or after
2026-08-18 lacks the timestamp. In those rows `#9` carries only fields 2
and 10, so the shipped reader has nothing to read there. Usage counts in
`#4` are present in all 1677 rows.

Because one timestamp-less row marks coverage partial, the reader also
withholds the 976 older rows, so all history disappears at once.

Two offers, whichever helps:

1. I can run this branch against my 200 sessions and post the before and
   after output, including a check for sessions with many unrelated
   `steps` rows ahead of the needed step.
2. If you are short on time, I can pick up the remaining review items
   (drop the fixed lookup cap and add the sparse-session regression test)
   in a follow-up PR that keeps your commits and credit.

chid:
Thanks, I think I have an approach that covers older sessions - and likely will need a few iterations before this works completely due to the revised format (and expect there may be issues with features like /fork which I haven't used). I also only use it on Mac (and only relatively recently started using agy)
  ──────
  ### 3. What was the specific issue investigated above?

  When testing against all 44 of your live conversation databases, we discovered why some databases were still reporting coverage: .partial:

  • Multi-Turn Steps (1-to-many relationship):
  In Antigravity conversations, multiple generation turns in gen_metadata frequently belong to the same step (they share the same stepUUID).
  • The Index Offset Bug:
  The previous recovery logic in appendRecoveredEvents assumed a strict 1:1 match by doing timestamps[offset].
  If a step had 5 turns in gen_metadata but only 1 step timestamp in the steps table, turns 2 through 5 failed lookup (offset >= timestamps.count).
  • The Consequence:
  CodexBar enforces strict integrity: if even a single row in a database cannot be dated, the entire database is marked incomplete (coverage: .partial). When coverage is partial, CodexBar withholds Antigravity from
  usage totals so incomplete data is never shown.

  By allowing all turns sharing that stepUUID to use the step's timestamp:

  • All 44 live databases and 2,297 turns resolve with 100% precision.
  • Complete coverage is established, aggregating all 240M tokens across your conversations.

steipete:
Maintainer closeout: I rewrote the contributor branch to the parser-only recovery and removed the later immutable-WAL/menu/widget work from this PR. The final implementation keeps the original discovery and live validation while adding one-snapshot reads, shared scan budgets, stored-index alignment, fail-closed ambiguity handling, UUID/NULL/malformed-row guards, and restored generation-event order.

Proof on exact head `6adeb2b26994e52303808913aa90a4e980a63f1d`:

- 80 focused Antigravity local tests across eight suites passed.
- Full isolated local suite: 997 selections / 84 groups, all first-pass green, zero retries and zero timeouts.
- `make check` passed with clean formatting and zero lint violations.
- [CI run 33706240542](https://github.com/steipete/CodexBar/actions/runs/33706240542) passed lint, both macOS test shards, both Linux builds/tests, musl, the aggregate gate, and security checks.

Merged as `c406dd04584878430820c5f47df407843c732313`; `main` was pulled and the merge tree verified clean locally.

PR #3269, opened 2026-08-29
test: stabilize local database and menu refresh fixtures

steipete:
## Summary

Two test-only repairs found while validating #3267:

- Build each Antigravity database-cap fixture from one committed, closed empty SQLite database and copy it to distinct files. Keep the 499/500/501 cases, default 500-file cap, and 60-second reader budget. Also assert empty rows and balanced read handles. This removes 1,497 redundant schema creations without bypassing discovery or SQLite reads.
- Give both Codex menu-refresh snapshots explicit fixture identity, so invalidating the fallback account cache cannot accidentally turn the test into an incompatible-account scenario. Deliberately clear that cache during refresh and assert layout compatibility while preserving all blocked-enrichment and changed-quota assertions. Release the credits gate before awaiting the refresh task on the timeout path, preventing a late loader from deadlocking cleanup.

No production code, timeout budgets, compatibility rules, credentials, or user-visible behavior changes. No changelog entry is needed for this test-only change. This does not claim to fix every local-host timing stall seen in the broader run.

## Verification

- Before fixing the snapshot identity, the strengthened menu test failed with four issues: incompatible tracked layout, refresh still active, and the old quota remaining frozen. The unrelated reconciliation test passed.
- `swift test --jobs 4 --filter 'StatusMenuScopedCodexRefreshTests|AntigravityLocalScanTests'` passed all 15 tests in 2 suites in 41.095 seconds, with Keychain access suppressed and Codex file isolation enabled. The cap cases still open and close 499/500/500 actual SQLite handles respectively.
- Isolated Codex review through P2 found no actionable defects.
- `make check`, the broader local suite, and exact-head CI are being verified before landing.

All data is synthetic. No live provider, browser-cookie, Keychain, or application UI probe was performed.

### DIFF
--- a/Tests/CodexBarTests/AntigravityLocalScanTests.swift
+++ b/Tests/CodexBarTests/AntigravityLocalScanTests.swift
@@ -13,14 +13,21 @@ struct AntigravityLocalScanTests {
     @Test(arguments: [499, 500, 501])
     func `database cap distinguishes below exactly and above limit`(count: Int) throws {
         let fixture = try Fixture()
-        for index in 0..<count {
-            try fixture.database("session-\(index)")
+        let seed = try fixture.database("session-0")
+        // The cap counts distinct files; copy the closed empty database instead of committing 500 schemas.
+        for index in 1..<count {
+            try FileManager.default.copyItem(
+                at: seed,
+                to: seed.deletingLastPathComponent().appendingPathComponent("session-\(index).db"))
         }
         var limits = AntigravityLocalReader.Limits()
         limits.duration = 60
         let report = try fixture.report(limits: limits)
         #expect(report.coverage == (count <= 500 ? .complete : .partial))
         #expect(report.statistics.files == min(count, 500))
+        #expect(report.statistics.rows == 0)
+        #expect(report.statistics.sqliteHandlesOpened == min(count, 500))
+        #expect(report.statistics.sqliteHandlesClosed == report.statistics.sqliteHandlesOpened)
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


steipete:
Merged as ae0ec9e2445726d7acd46d687b14a42250f04934. Main was pulled with `--ff-only`, matches the tested head b78f4f00a0e15922becc0608ecc0380d171becce byte-for-byte, and is clean.

The repair is test-only: it removes redundant SQLite schema setup while preserving all file-count boundaries and real reader/handle checks, makes the menu fixture's identity independent of its fallback cache, and fixes the late-credits cleanup ordering. No production compatibility rule, timeout budget, or user-facing behavior changed.

Verification:

- The strengthened menu regression failed on the old fixture with four expected compatibility/publication assertions after fallback-cache invalidation. With explicit identity on both snapshots, all original blocked-enrichment and changed-quota assertions pass.
- `swift test --jobs 4 --filter 'StatusMenuScopedCodexRefreshTests|AntigravityLocalScanTests'` passed all 15 tests in 2 suites. The cap cases still open/close 499, 500, and 500 actual database handles for 499/500/501 files.
- `make check` passed with zero violations across 2,041 files.
- `make test` completed successfully: all 960 selections across 80 groups, 78 groups successful on first pass, and two group-level timeouts (57 and 71) recovered through 24 isolated selection retries. There were no assertion failures in this final run. Total time was 3,060.4 seconds. No timeout, count, or retry policy was relaxed.
- Within that full run, the revised menu test passed in 1.149 seconds and the three database-cap cases passed in 3.607 seconds. These are observed run timings, not a controlled benchmark. The unchanged RPC fallback suite also passed.
- [Exact-head CI run 33261017128](https://github.com/steipete/CodexBar/actions/runs/33261017128) passed both macOS test shards, both Linux builds, lint, and the aggregate gate. GitGuardian passed. musl was skipped by the repository's test-only path filter; no production source changed in this PR.
- Isolated Codex review through P2 and the PR review found no actionable defects; no unresolved review threads remained.

All tests used synthetic data, Keychain suppression, and Codex file isolation; the full run also disabled live Gemini fetching. No live account, browser-cookie, Keychain, or application UI probe is claimed. No changelog entry was added for these internal test-only changes.

PR #3272, opened 2026-08-29
Show OpenRouter account Activity from management keys

akshayprabhu200:
Recognize a management key placed in the primary API-key field when the official OpenRouter Current Key endpoint identifies it. A separately configured Management API key takes precedence for Activity. Custom API metadata cannot promote a primary key into fixed-host Activity requests, and selected-key balance and quota keep their original credential.

Add compact token, request, and distinct-model counts for the last 30 completed UTC days. Detailed models and spending remain in Usage & Spend. Counts reuse the existing validated, deduplicated activity entries, preserving overflow handling, reasoning-token semantics, BYOK estimates, diagnostics, and graceful optional failure. Management-key counters do not create a misleading key-spend headline.

A live official API read through the signed candidate preserved the existing uncapped-key summary and balance. The available key did not provide Activity access; management-key promotion is verified with controlled responses.

Focused tests exercise both JavaScript engines, official/custom origin boundaries, credential precedence, malformed metadata, optional failure, and full-history preservation. Native before/after captures use identical synthetic data and differ only by the new Activity section.

The full run exposed a startup-sensitive Antigravity subprocess fixture. Its test-only timeout now matches the existing three-second version-probe budget, and PID publication is atomic so cancellation cannot read a partially written file. The PID and process-exit assertions remain mandatory.

Validation: `make test` passed 1,184 selections in 108 groups, with one group recovered by retry after the existing fixture failure. The corrected Antigravity fixture passed all 64 focused tests. Both-engine provider tests, `make check`, isolated review, and signed native proof passed. The combined seven-feature tree also passed all 1,196 selections in 110 groups without retries, plus `make check`. [Final-head CI](https://github.com/steipete/CodexBar/actions/runs/35335449926) passed on attempt 2. The first attempt failed only the unchanged renderer timing benchmark (about 68 ms against its existing 50 ms budget); the same head passed the retry without changing source or the threshold.

| Before | After |
| --- | --- |
| Before | After |


akshayprabhu200:
@clawsweeper re-review

Addressed both readiness blockers on the rebased signed head:

- added cropped live packaged-app proof from the official OpenRouter API; no account identity, credential, browser, desktop, or notification content is present
- bounded the provider-card model summary to 120 characters with `+N more`, while preserving every exact model in `costUsage`
- added regression coverage for the bound/full-data invariant and cross-row token overflow
- reran 28 OpenRouter golden tests plus 57 multi-account/settings/dashboard integration tests; all pass
- packaged and signature-validated the Debug app with App Store Xcode

PR is now ready so the deferred macOS CI can run.

— Akshay / Codex

akshayprabhu200:
@clawsweeper re-review

Resolved the remaining canonical-origin finding on signed head `84d5e6402`:

- compares normalized scheme, authority/effective HTTPS port, and exact API path instead of raw URL-string equality
- accepts the exact official base, explicit `:443`, host/scheme case variants, and a trailing slash
- continues rejecting custom hosts and non-default ports, so a custom endpoint cannot promote or exfiltrate the primary credential
- OpenRouter golden suite passes, including all accepted/rejected endpoint cases
- `make check` passes with 0 SwiftLint violations
- `make test` passes 967/967 selections in 81/81 groups with 0 failures, retries, or timeouts

— Akshay / Codex

steipete:
The aggregate token-overflow repair is now merged in #3458 (`9d01d5cf0eb`), with your contributor credit preserved in the squash commit and changelog. The current supported separate-management-key path reproduces a complete snapshot failure in both JavaScript engines when valid input/output components combine beyond the safe-integer boundary. The focused fix retains credits/quota and marks malformed Activity unavailable; it also removes one production line by validating the combined total in place of redundant component checks.

Exact-head CI and the full 1,027-selection test suite passed. This PR remains open for its broader primary-key routing and Activity-card proposal. The extracted repair does not resolve those feature decisions or close #3289.

PR #3292, opened 2026-08-30
fix: recover abandoned Codex parent discovery

steipete:
## Summary

Repair the stale-discovery recovery boundary reported in #2815 without changing how unresolved forks are counted. When the requesting child rollouts leave both scan roots, a normal bounded refresh now resumes their persisted parent search instead of leaving the existing SQLite cache permanently catching up.

Discovery owns this repair: validate and retire pending IDs within the existing byte/time limits, remove associations to deleted paths even when those paths are behind the inventory cursor, and publish a completed discovery generation only after the pending queue drains. Successful parent lookups retire their own demand so they do not scan unrelated history. Missing and empty heads also consume work budget. No new cache fields, source-file mutation, credential handling, or external fallback.

The previous parser hash remains a compatible predecessor. Existing rows, cursors, buffered fork state, and reports are adopted rather than rebuilt. A surviving fork with a missing parent still remains unresolved; if the parent later appears, its buffered usage resolves once. This is a partial fix for #2815, not a change to orphan-accounting policy or the broader stall-diagnostics request.

Thanks @xiehaibin18 for the careful removed-source / existing-cache A/B report.

## Verification

- Red integration replay reproduced stale pending IDs, deleted-path associations, and unfinished catch-up using the same SQLite store across refreshes, without force-rescan or cache deletion.
- Regression coverage includes child rows both retained and already removed, a surviving orphan sibling and later parent arrival, one-unit budgets with repeated lookups and serialized state, cancellation, early successful lookups, new inventory during finalization, and missing/empty heads under byte/time budgets.
- Existing store-adoption tests cover the preceding parser hash with identical stored contents and zero rebuilds.
- Restart fixtures now clear only their own directory cursors. The combined run exposed the old process-wide reset interfering with the performance suite; a regression checks that another fixture's reset cannot discard its cursor.
- Focused scanner/recovery/store/architecture run: 158 tests across five suites passed, including the combined performance/restart isolation regression.
- `make check`: passed, 2,059 Swift files with no format/lint violations.
- Isolated Codex autoreview: scoped-clean at the required P0 threshold, no accepted/actionable findings.
- Exact-head [CI run 33328774867](https://github.com/steipete/CodexBar/actions/runs/33328774867): all nine checks passed, including both macOS shards and all three Linux builds.
- Full `make test`: all 971 selections across 81 groups passed on the first attempt; zero failures, retries, or timeouts (2,471 seconds total).

All scanner fixtures are temporary synthetic JSONL/SQLite data. No live account requests, Keychain reads, application relaunch, or UI claim is part of this change. Changelog and Codex cost-history documentation updated.

Maintainer compatibility assessment: retain the explicit predecessor adoption. This is the preceding main producer with unchanged row shape, parsing/accounting semantics, and stored payloads; the adoption regression compares the complete snapshot and asserts zero rebuilds. Rebuilding would mask the reported existing-cache failure. No new schema or orphan-accounting policy is introduced.


steipete:
Merged as [56bbc037f616](https://github.com/steipete/CodexBar/commit/56bbc037f616adcb4cfb6ca26e5f616b66965a60). Verification for this repair:

- Reproduced stale pending IDs/deleted-path associations with real temporary JSONL and the same SQLite store across bounded refreshes, without cache deletion or force-rescan.
- `swift test --filter 'CostUsageDiscoveryRecoveryTests|CostUsageCatchUpCompletionTests|CostUsagePerformanceGateTests|CostUsageStoreTests|ProviderArchitectureGatekeeperTests'`: 158 tests / five suites passed. The new cases cover removed requesting children, already-removed cache rows, surviving orphan accounting, late parent arrival exactly once, restart serialization, cancellation, tiny byte/time budgets, positive/negative discovery transitions, and no unnecessary scan after early resolution.
- The preceding parser-hash adoption test preserves the complete stored snapshot with zero rebuilds.
- `make check`: passed; 2,059 Swift files, zero format/lint violations.
- Isolated Codex autoreview: scoped-clean at the required P0 threshold, no accepted/actionable findings. The reviewed tree is the committed candidate.
- Full `make test`: all 971 selections across 81 groups passed on the first attempt; zero failures, retries, or timeouts (2,471 seconds total).
- Exact candidate `3957b778370fe891d3018419a50fa34444c5b66d`: all nine checks passed in [CI run 33328774867](https://github.com/steipete/CodexBar/actions/runs/33328774867), including both macOS shards and x86_64/arm64/musl Linux builds. No CI rerun was needed.

Tests ran with Keychain suppression and Codex/session-file isolation in a clean environment. This is scanner/SQLite proof, not a live-account or native-UI claim. The root application bundle and running release app were not relaunched. No source history, credentials, or orphan-accounting policy changed.

The combined focused run also exposed a process-wide test-reset race; restart fixtures now reset only their own directory cursors, with an explicit cross-fixture regression. Changelog and Codex cost-history documentation are included. #2815 remains open for its broader follow-ups.

PR #3294, opened 2026-08-30
Show a 0% Grok usage bar for a period with no usage yet

olddonkey:
## Problem

A Grok billing period that has not recorded usage yet renders no usage bar at all. The card falls back to the explicit failure message instead:

Grok card before the fix

The cause is on the wire. grok.com's credits frame carries the usage percentage as a proto3 scalar, so a period whose usage is exactly zero omits the field entirely — "zero" and "not reported" are byte-identical. `GrokWebBillingFetcher` already handles that: when the frame carries a live period and no percentage anywhere, it reads the value as `0` and marks it `usedPercentIsWirePublished = false`.

The enrichment gate added in #3181 then refused that reading, because promoting an inferred percent is what rebuilt the fabricated 0% that #3157 removed. So the resolver fell back to the credits proxy's unknown percent, `toUsageSnapshot()` built no primary rate window, and the card showed `usageUnavailableMessage`. Every account sees this from the moment its period resets until its first request lands.

## What this changes

One guard in `GrokOAuthFetchStrategy.resolvingUnknownUsage`:

```swift
guard let percent = grpcSnapshot.usedPercent,
      grpcSnapshot.usedPercentIsWirePublished || percent == 0
else { return proxyAnswer }
```

Three properties keep #3157's fabricated zero from returning through it:

1. **The credits proxy cannot reach this branch.** It never constructs an inferred value — when it has no percentage it returns `nil` (`GrokCreditsProxyFetcher`), so the surface that caused #3157 is untouched.
2. **The adopted zero stays marked.** `completing(with:)` preserves `usedPercentIsWirePublished = false`, so no downstream consumer mistakes it for a published reading.
3. **Only zero is accepted.** Any other inferred value is still refused, with a regression pinning that so a future parser change cannot widen this quietly.

## Real-account proof

Produced on this head against a real SuperGrok account whose weekly period had just reset, using the shipped fetchers and the shipped resolver. Redacted transcript:

```text
CODEXBAR_LIVE_GROK_ZERO_PROOF=1 \
CODEXBAR_LIVE_GROK_ZERO_PROOF_DIR=.github/pr-proof \
swift test --filter GrokZeroUsageLiveProofTests

branch=grok_web_no_usage_yet_zero
proxy_used_percent=nil
proxy_has_period=true
grok_web_used_percent=0.0000
grok_web_percent_is_wire_published=false
resolved_used_percent=0.0000
resolved_percent_is_wire_published=false
resolved_source=grok-web
resolved_resets_at=2026-09-06Z
renders_usage_bar=true
```

The card rendered from that live result, with personal information hidden:

Grok card from the live account after the fix

`GrokZeroUsageLiveProofTests` reads the bearer from `~/.grok/auth.json` — the same file the provider reads — and touches neither the Keychain nor browser cookies. It makes one request per surface, prints aggregate fields only, and asserts the resolver's contract on whichever of the four branches the account exercises, so it stays meaningful (and green) once the period records usage.

The fixture-rendered before/after pair is still in the PR, produced by `GrokZeroUsageScreenshotRenderTests` from the real snapshot and card types:

Grok card after the fix, rendered from fixtures

## ClawSweeper 2026-08-30 review — each item answered

- **Add real behavior proof.** The transcript and card above are from a real newly reset account, not fixtures. This is maintainer option 1 from that review, "validate the no-usage account shape".
- **Merge risk P1 — the before/after cards were fixture-rendered.** Resolved by the live card, which is rendered from the resolver's real output.
- **Merge risk P1 — the exact-zero exception relies on the parser heuristic; an unobserved frame that withholds a nonzero percentage with the same period shape would display 0%.** Partly measured, not fully closable — see below.
- **Next step P2 — contributor evidence plus a maintainer decision.** Evidence is above; the decision is stated below.

The 2026-08-31 re-review raised one more, now fixed:

- **P2 — update the Grok fallback contract, and the matching P2 merge risk.** `docs/grok.md` still described the pre-change rule: only a wire-published percentage adopted, with the parser's no-usage-yet zero explicitly rejected. It now describes what the resolver does, why the frame cannot publish an exact zero, what stays refused, and that the reading is an inference from frame shape rather than a value the surface stated. The parser section, which already documented the omitted scalar as zero usage, now says the retry adopts that reading, and the SuperGrok Heavy note is scoped to the credits payload it was describing.

## Residual limitation, stated plainly

Because a proto3 scalar omits an exact zero, no client can distinguish "the period is at 0%" from "this surface declined to publish a percentage" by inspecting the frame alone. The parser's requirements narrow it — the frame must carry a live period, usage-period evidence, and no percentage anywhere — and the live run above shows a real account hitting exactly that shape on a fresh period.

What supports reading it as a genuine zero:

- The same account and surface published `20.0` on 2026-08-22 while its period had usage, so this surface does publish nonzero percentages rather than withholding them. That measurement predates this branch and I have not re-run it on this head; the current period is at zero, so it cannot be reproduced until the account records usage again.
- The frame enumerates its products, and product percentages are omitted under the same proto3 rule, so "no percentage anywhere" is the shape of every product reporting zero.

What is not closable in code: if some plan or outage state ever returns a live period while withholding a nonzero percentage, this change shows 0% where `main` shows the failure message. The failure mode is bounded to the grok.com surface and to accounts whose credits proxy publishes no percentage.

**Maintainer decision.** If you would rather not carry that inference at all, the conservative alternative is to keep usage unknown and replace the red failure text with a neutral "no usage reported for this period yet" line plus the reset time. That keeps `main`'s semantics and removes only the alarm; say the word and I will swap this PR to that shape.

## Testing

Verified on exact head `8e61b2e3fcd5dd39ac37dc15cc42e9c28c0f61b5`:

- `make check`: passed — SwiftFormat clean, SwiftLint 0 violations in 2065 files, provider/package/documentation and repository-size gates green.
- `make test`: passed on `31ced6bcb` — 976/976 selections, 82/82 groups successful on the first pass, 0 failed groups, 0 retries, 0 timeouts (678.5 s). The only change since that run is the `docs/grok.md` reconciliation, covered by `make check`'s documentation gates and `DocumentationLinkTests` 8/8.
- Focused: `GrokCreditsProxyFetcherTests`, `GrokWebBillingFetcherTests`, `GrokAccountContextTests` — 85 tests, all passing.
- Regression proof: with the guard reverted, `an inferred grok dot com zero restores the bar without claiming a published percent` fails on all three assertions (`usedPercent → nil`, the wire-published marker, and `sourceLabel → "grok-cli-proxy"`).
- Rebased on upstream `main` `83977905e`, 0 behind, `git diff --check` clean.

Both proof tests are opt-in and skipped by default, so CI behavior is unchanged.


olddonkey:
Added the real-account proof the review asked for, and rebased onto `83977905e`. Exact head `31ced6bcb924ebe5862bef9ec02e772f2fdd68e0`.

### Real behavior, from a live account on a freshly reset period

```text
branch=grok_web_no_usage_yet_zero
proxy_used_percent=nil
proxy_has_period=true
grok_web_used_percent=0.0000
grok_web_percent_is_wire_published=false
resolved_used_percent=0.0000
resolved_percent_is_wire_published=false
resolved_source=grok-web
resolved_resets_at=2026-09-06Z
renders_usage_bar=true
```

This is the shipped `GrokCreditsProxyFetcher` and `GrokWebBillingFetcher` against the live surfaces, then the shipped `resolvingUnknownUsage` on the two real snapshots. The card in the PR body is rendered from that resolved snapshot with personal information hidden. `GrokZeroUsageLiveProofTests` reads the bearer from `~/.grok/auth.json`, the same file the provider reads; it touches neither the Keychain nor browser cookies, makes one request per surface, prints aggregate fields only, and asserts the resolver's contract on whichever of the four branches the account exercises — so it stays green after the period records usage.

### On the remaining P1

The exact-zero exception cannot be made unambiguous in code: a proto3 scalar omits an exact zero, so "0%" and "not published" are the same bytes. What the live run adds is that a real account on a fresh period does produce exactly the shape the parser requires, and that the resolver turns it into a bar rather than the failure message.

Two supporting facts are in the PR body: this same account and surface published `20.0` on 2026-08-22 while its period had usage — so this surface publishes nonzero percentages rather than withholding them — and the frame enumerates its products, whose percentages are omitted under the same proto3 rule. The 2026-08-22 number is an earlier measurement, labeled as such; it cannot be re-run on this head until the account records usage again. Once this period does record usage I am happy to post the same transcript from the published-percent branch, which would show the transition inside one period on one account.

If you would rather not carry the inference at all, the conservative alternative is in the PR body: keep usage unknown, and replace the red failure text with a neutral "no usage reported for this period yet" line plus the reset time. That keeps main's semantics and removes only the alarm.

### Gate on this head

`make check` clean (SwiftLint 0 violations in 2065 files). `make test` 976/976 selections, 82/82 groups green on the first pass, 0 retries, 0 timeouts (678.5 s). Both proof tests are opt-in and skipped by default, so CI behavior is unchanged.

olddonkey:
Fixed the actionable finding from the 2026-08-31 re-review. Exact head `8e61b2e3fcd5dd39ac37dc15cc42e9c28c0f61b5`.

### P2 — the Grok fallback contract

You were right that the documentation promised the opposite behavior, and I had missed it. `docs/grok.md` said only a wire-published percentage is adopted and named the parser's no-usage-yet zero as explicitly rejected. It now describes the shipped rule: both a wire-published percentage and the no-usage-yet frame are adopted, the percentage is a proto3 scalar so a period at exactly zero cannot publish one, the adopted zero keeps its not-wire-published marker, and any other inferred value is still refused. It also states plainly that the protocol omits an exact zero and a withheld percentage identically, so this is an inference from frame shape rather than a value the surface stated.

Two adjacent passages are reconciled with it: the parser section, which already documented the omitted proto3 scalar as zero usage, now says the retry adopts that reading; and the "SuperGrok Heavy with no `creditUsagePercent`" note is scoped to the credits payload it was describing, since the grok.com retry can now supply that percent.

That covers both the P2 finding and the P2 merge risk about conflicting guidance.

### Still yours to decide

The P1 is unchanged and not closable in code: the wire cannot distinguish an exact zero from a withheld percentage. Your recommendation was to retain unknown usage and replace the alarming copy with neutral reset-period text. I am happy to swap this PR to that shape — it is a smaller change than what is here, and the live proof in the body applies to it just as well, since it records what each surface actually returned. Say which you prefer and I will push it.

`make check` is clean on this head (SwiftLint 0 violations in 2065 files) and `DocumentationLinkTests` passes 8/8; the full suite was green on `31ced6bcb`, and documentation is the only change since.

steipete:
The zero-usage fix landed in #3325 as e7d0011a837e1c6271924ec00fce8fba2087e99e, with parser provenance and active-period validation preserving unknown results for malformed or inactive payloads. This PR's production change is covered there, so I am closing it as superseded. @olddonkey's contribution is preserved in the squash commit co-author trailer and changelog thanks, alongside @sf-jin-ku.

The integrated fix passed 91 focused Grok tests and all nine CI checks, including both full macOS shards: https://github.com/steipete/CodexBar/actions/runs/33475470989. Thank you for the implementation and reproduction proof.

PR #3311, opened 2026-08-31
test: preseed Claude identity fixture configuration

steipete:
## Summary

Prepare the Claude active-account identity test fixture's provider configuration in memory and persist it once, instead of calling the runtime provider toggle for every registered provider. The repeated setup loop performed synchronous whole-config atomic writes even though each fixture constructs `UsageStore` only after setup, so none of the intermediate notifications or revision increments are observable behavior.

Reuse the existing isolated config/settings helpers, enable only Claude, and explicitly preserve the fixture's intended manual refresh, disabled status checks, Claude data source, and disabled OpenAI web access. The test now asserts that complete initial state on every fixture construction.

Only the test fixture changes. UUID-isolated defaults/config files, test credential stores, missing-credentials scopes, fetch stubs, and seeded account snapshots remain unchanged. There is no production, authentication, persistence-policy, UI, dependency, or user-facing behavior change, so no changelog entry is needed.

## Verification

- 42 focused tests across the Claude identity-invalidation, process-safety, settings-preference, and app-group suites passed.
- `make check` passed with zero format or lint violations.
- Full isolated local suite on the final rebased head: 994 selections across 83 groups, all 83 first-pass green, retries disabled, zero failed groups, and zero timeouts.
- Final independent branch review found no actionable P0-P2 defects.
- The first hosted run exposed an unrelated Pi scanner test fixture that restored a fractional filesystem timestamp after an atomic replacement without proving its millisecond cache key stayed equal. The runner legitimately rescanned when that key shifted. The fixture now pins a whole-second timestamp before both scans and asserts exact key equality; the formerly failing case passed six consecutive focused runs, `make check`, and a clean scoped review.
- Fresh hosted CI is running on exact head `90168b8d48fd62673667b96846e39497fc60b944`.

Tests used the repository's Keychain and credential-file isolation flags with synthetic fixtures. No real provider probes, browser imports, or installed-app launches were used.

### DIFF
--- a/Tests/CodexBarTests/ClaudeActiveAccountIdentityInvalidationTests.swift
+++ b/Tests/CodexBarTests/ClaudeActiveAccountIdentityInvalidationTests.swift
@@ -678,17 +678,25 @@ struct ClaudeActiveAccountIdentityInvalidationTests {
         outcome: ProviderFetchOutcome,
         environment: [String: String] = [:]) throws -> ClaudeIdentityFixture
     {
-        let settings = testSettingsStore(suiteName: "ClaudeActiveAccountIdentityInvalidationTests")
+        var config = testConfigWithAllProvidersDisabled()
+        let claudeIndex = try #require(config.providers.firstIndex { $0.id == UsageProvider.claude.instanceID })
+        config.providers[claudeIndex].enabled = true
+        let settings = testSettingsStore(
+            suiteName: "ClaudeActiveAccountIdentityInvalidationTests",
+            config: config,
+            prepareDefaults: { defaults in
+                // An existing config otherwise opts this fresh fixture into OpenAI web access.
+                defaults.set(false, forKey: "openAIWebAccessEnabled")
+            })
         settings.refreshFrequency = .manual
         settings.statusChecksEnabled = false
         settings.claudeUsageDataSource = source
-        let metadata = ProviderRegistry.shared.metadata
-        for provider in UsageProvider.allCases {
-            try settings.setProviderEnabled(
-                provider: provider,
-                metadata: #require(metadata[provider]),
-                enabled: provider == .claude)
-        }
+        #expect(settings.enabledProvidersOrdered(metadataByProvider: ProviderRegistry.shared.metadata)
+            == [UsageProvider.claude.instanceID])
+        #expect(settings.claudeUsageDataSource == source)
+        #expect(settings.refreshFrequency == .manual)
+        #expect(settings.statusChecksEnabled == false)
+        #expect(settings.openAIWebAccessEnabled == false)
 
         let store = UsageStore(
             fetcher: UsageFetcher(environment: environment),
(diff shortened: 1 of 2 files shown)


steipete:
Keeping this PR open and unmerged at `34c83189218048bc22e9be0d7137abfa552b7fdd`. The one-file fixture cleanup is unchanged; no production code or process-tracking policy was modified.

Completed proof: 27 focused Claude identity tests passed; `make check` passed with zero violations across 2,068 Swift files; scoped Codex autoreview found no actionable P0 findings. [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/33361601733) succeeded for both macOS test shards, both Linux CLI builds, lint, and the aggregate check. GitGuardian passed; musl was skipped by the test-only path gate.

The local full-suite gate is **not green**:

- First run: group 29 failed twice on a synthetic RPC `initialize` timeout. The unchanged fallback suite subsequently passed all 14 tests, and the unchanged complete 12-suite group passed all 185 tests. These diagnostics are not a full-suite pass.
- Second run: one Bailian stub timeout recovered on the existing retry, then group 26 stopped because the runner lost a descendant session's ownership anchor and could not attribute a newly observed process. It exited with the containment error; the remaining groups did not run. All reported PIDs were later gone without manual termination.

The six focused `SessionOwnershipTests` passed. Their [lost-anchor and snapshot-race cases](https://github.com/steipete/CodexBar/blob/34c83189218048bc22e9be0d7137abfa552b7fdd/Scripts/test_swift_test_process_cleanup.py#L1007) explicitly require this failure, matching the [documented process-attribution boundary](https://github.com/steipete/CodexBar/blob/34c83189218048bc22e9be0d7137abfa552b7fdd/docs/DEVELOPMENT.md#run-tests-only). The available log cannot establish the unknown process's earlier ancestry, so loosening attribution, signaling by numeric session ID, or silently converting the failure into success would be unjustified.

Next verification step is a complete run with reliable process attribution, or a fixture-side reproduction identifying the launcher whose session anchor disappears. No timeouts, isolation controls, or tests were bypassed. No real provider/Keychain access, app launch, release, or user-data mutation was involved.

steipete:
Landed as `620f3b8e4eef7222458e5fb6a0532b82e84036d1` after rebasing onto current main and completing the previously held local gate.

The final branch keeps fixture setup behavior explicit while replacing repeated whole-config writes with one preseeded configuration. During hosted verification, an unrelated Pi cache test exposed a fractional-mtime fixture assumption; that fixture now pins a whole-second timestamp on both sides of its atomic replacement and verifies the exact millisecond cache key before asserting stale-cache behavior.

Proof on exact reviewed head `90168b8d48fd62673667b96846e39497fc60b944`:

- 42 focused Claude/settings/process-safety tests passed
- full isolated local suite: 994 selections across 83 groups, all first-pass green, retries disabled, zero failed groups, and zero timeouts
- formerly flaky Pi case passed six consecutive focused runs
- `make check`
- final independent branch review: no actionable findings
- [hosted CI run 33640745068](https://github.com/steipete/CodexBar/actions/runs/33640745068): all required jobs green

The squash merge tree exactly matches the reviewed head tree (`cc78b914a01051334b2ea8adc31cea886942ef8d`).

PR #3313, opened 2026-08-31
perf: streamline Claude cost metadata detection

steipete:
## Summary

Reduce CPU spent classifying Claude and Vertex AI transcript metadata without changing provider detection, accounting, refresh cadence, or scan budgets. A fresh profile of a correctly signed 0.56.2 build attributed about 36% of sampled CPU during Claude history ingestion to this detector.

The recursive root walk already visits metadata, request, context, client, and nested message dictionaries. Stop revisiting those subtrees, and match decoded ASCII marker strings directly instead of using Foundation substring searches. Non-ASCII and noncontiguous strings retain the original Foundation predicate; checking the whole string before taking the fast path preserves combining-character behavior. Primary message/request IDs, model-name detection, and existing flag/array semantics remain unchanged.

## Evidence

On base `354191af9`, the classifier-only optimized candidate produced exactly the same accounting output as the baseline. An independent signed-CLI benchmark used 100,000 synthetic events, three paired runs, and fresh cache roots:

| Metric | Baseline | Candidate |
| --- | ---: | ---: |
| Median cold-scan CPU | 11.637 s | 9.960 s |
| Tokens | 15,000,000 | 15,000,000 |
| Estimated cost | $105.00 | $105.00 |

That is **14.4% less cold-scan CPU**. Complete cold/warm JSON output matched, excluding the capture timestamp. Isolated classification used about half the CPU in the normal synthetic cases; larger text-heavy ingestion cases improved by 3–18%. These are controlled workload results, not a measured percentage reduction for whole-app idle CPU.

Regression coverage compares against the frozen old predicate and checks complete provider-separated rows, daily totals, costs, JSON escapes, Unicode boundaries, nested dictionaries/arrays, and false/numeric metadata flags. This Claude-only source change does not alter the Codex parser hash or predecessor-cache adoption.

## Validation

- Before base integration: 46 focused tests and all 9,848 full-suite tests passed; all 82 groups succeeded on the first pass without retries or timeouts.
- After rebasing onto merged [PR #3312](https://github.com/steipete/CodexBar/pull/3312): `make check` passed with zero violations, and final independent P0–P2 Codex review found no actionable issues.
- Full combined-base `make test` is running; this PR will not merge until it and exact-head CI are green.

## Scope

This complements the separately owned decoded-baseline work in [PR #3312](https://github.com/steipete/CodexBar/pull/3312). It does not change SQLite reconstruction or priority aggregation. Live CPU remained elevated in cached Codex report decoding, so this PR does not claim to eliminate all background CPU work. No visual behavior changes, release publication, or raw session/trace uploads.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Performance
 - Local costs: reduce background CPU spent parsing native session timestamps and validating appended Codex history, preserving timestamp precision, daily totals, and fork accounting.
 - Codex: reuse one decoded SQLite baseline per native cost scan, avoiding repeated unchanged-save decoding while rechecking database changes and filesystem state.
+- Claude and Vertex AI: reduce CPU spent classifying local transcript metadata, preserving provider detection, Unicode handling, and token/cost totals.
 
 ### Fixed
 - Local costs: preserve token breakdowns, reasoning, request counts, and pricing coverage when combining reports or reopening cached history. Thanks @Pjhhhhh!
--- a/Sources/CodexBarCore/Vendored/CostUsage/CostUsageScanner+Claude.swift
+++ b/Sources/CodexBarCore/Vendored/CostUsage/CostUsageScanner+Claude.swift
@@ -425,7 +425,7 @@ extension CostUsageScanner {
         }
     }
 
-    private static func isVertexAIUsageEntry(obj: [String: Any]) -> Bool {
+    static func isVertexAIUsageEntry(obj: [String: Any]) -> Bool {
         // Primary detection: Vertex AI message IDs and request IDs have "vrtx" prefix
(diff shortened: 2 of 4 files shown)


steipete:
Merged as 2570a034feaa6bb6ee41714f4ab07881b81cf043 after independent review and [green exact-head CI](https://github.com/steipete/CodexBar/actions/runs/33368011957), including both macOS test shards and the Linux builds.

The optimization removes repeated metadata traversal while retaining the Unicode fallback and provider-separated accounting checks. Local `make check` also passed with zero lint violations after investigating and retrying short process-harness timing failures; no ownership or timeout safeguards were weakened. The additional local full-suite run and signed-artifact live QA remain gates for the 0.56.2 release.

PR #3314, opened 2026-08-31
fix(codex): stop live rollouts from starving cost updates

### LINKED ISSUE #3243: Codex widget token snapshot can stay stale indefinitely while cost catch-up remains pending
## Summary

On the official signed CodexBar 0.55.1 (130), the Codex widget can keep a fresh outer snapshot timestamp and current quota state while its token/cost rows remain more than an hour old.

This is not the fixed one-hour TTL case from #2103. The persisted cost report can continue advancing, but the running app does not publish a new token snapshot while Codex cost catch-up remains pending.

## Sanitized runtime evidence

- Global refresh frequency: one minute.
- `widget-snapshot.json.generatedAt` advances on schedule.
- The Codex entry's `tokenUsage.updatedAt` remains more than 60 minutes old.
- The bundled CLI and persisted report advance while the visible widget remains unchanged.
- Catch-up remains pending at a normalized `N-1/N` state while an active session tail continues to change.
- When a newer widget snapshot is supplied externally, the app rewrites it from its stale in-memory token snapshot during the next persistence cycle.

No account identifiers, usernames, absolute paths, exact usage values, credentials, device identifiers, session IDs, or screenshots are included.

## Current source behavior

Verified on current `main` at `e02e78d98274832305a72f4086bf4630edf83f77`:

- `runCodexCostCatchUp` stays inside `while status.pending`.
- `publishStableCodexCostCatchUpSnapshot` is called only after that loop exits.
- The regression test explicitly expects bounded catch-up to publish only the final stable snapshot.
- Widget persistence can continue using `lastQueuedWidgetSnapshot`, so a fresh outer snapshot does not imply fresh token data.

A minimal source-level reproduction is to keep returning distinct, progressing catch-up status keys with `pending: true`, return newer token snapshots from the loader, and inspect the publication revision before cancelling the task. The revision does not advance beyond the foreground snapshot because no intermediate catch-up snapshot is published.

## Expected behavior

A continuously active Codex session should not prevent today's token/cost rows from being published indefinitely. Historical catch-up may remain incomplete, but recent active-file progress should still become visible with an honest token-specific timestamp.

## Suggested direction

- Separate recent active-file tail refresh from low-duty-cycle historical catch-up.
- While catch-up remains pending, publish a rate-limited safe checkpoint when the persisted report has advanced.
- Keep the final stable publication for full reconciliation, including legitimate downward corrections.
- Add a regression test where active input keeps catch-up pending while checkpoint data advances.

## Related

- #1964 made token-row age visible but did not refresh stale token data.
- #2103 / #2282 aligned token refresh cadence but did not cover publication starvation while catch-up stays pending.

LeoLin990405:
## Summary

Fixes #3243 by making bounded Codex cost scans converge on a frozen logical target instead of chasing a rollout file that keeps growing while it is scanned.

The original checkpoint-publication approach treated the symptom after every advancing pass. This repair moves the invariant into the scanner: each generation scans toward one validated byte target, reports progress against that target, and defers bytes appended afterward to the next bounded pass. A continuously active rollout therefore finishes each finite generation and reaches the existing stable publication path without publishing partial accounting.

## Fix

- freeze the per-file target for one cache generation rather than extending it to the latest observed EOF
- preserve partial JSON records at the target boundary and resume them with the next tail
- keep later appended tails behind already scheduled backlog so one hot rollout cannot starve other files
- requeue incomplete parent dependencies while preserving fork and subagent accounting boundaries
- retain an unpublished subagent prefix whenever later physical bytes may contain lineage metadata, then classify and publish only after the validated tail is consumed
- report catch-up progress against the logical frozen target, with stable no-progress and completion semantics

The retained-tail path validates file identity, the exact frozen start offset, and the token anchor before consuming only the appended bytes. `parsedBytes` remains the absolute file cursor; it is not the amount of work spent in that refresh.

This replaces the PR implementation but preserves @LeoLin990405's contribution credit for isolating the live-rollout starvation case and proving that the advancing cache diverges from the preserved presentation snapshot.

## Verification

- Scanner, progress, fork, subagent, publication, cancellation, stale-writer, store, and architecture coverage passed, including 114 focused cases across the final scanner/subagent suites.
- The retained-tail regression proves a 504-byte append consumes exactly 504 budgeted bytes, reaches EOF, clears the lineage buffer and parent dependency, and publishes only the correct 55-token owned suffix.
- Full isolated local suite: 998 selections across 84 groups, all first-pass green with retries disabled, zero failed groups, and zero timeouts.
- `make check` passed; the generated parser hash is current, SwiftFormat is clean, and SwiftLint reported zero violations.

Maintainer verification used synthetic fixtures and repository isolation flags; no real local account data, credentials, browser imports, or installed-app launches were accessed.


steipete:
Maintainer closeout: this PR now fixes the scanner invariant rather than periodically publishing partial checkpoints. Each bounded generation freezes a logical target, defers later appends, preserves partial records, and withholds subagent prefixes until any appended lineage evidence has been consumed. That lets a continuously active rollout converge without over-counting or starving other files.

Proof on exact head `3e0ecf1f20c8ad29f113fe48e16728927e33c6ac`:

- 114 focused scanner, progress, fork, subagent, and scheduling cases passed.
- Full isolated local suite: 998 selections / 84 groups, all first-pass green, zero retries and zero timeouts.
- `make check` passed with the generated parser hash current, clean formatting, and zero lint violations.
- [CI run 33710822202](https://github.com/steipete/CodexBar/actions/runs/33710822202) passed lint, both macOS test shards, both Linux builds/tests, musl, the aggregate gate, and security checks.

Merged as `01da8fc6ea2dd8b325aaf56c95dee9c7fb323786`; `main` was pulled and the merge tree verified clean locally.

PR #3321, opened 2026-08-31
fix: prevent concurrent MiMo cache write collisions

Lucenx9:
## Summary

- Give every MiMo local usage cache write its own temporary file.
- Run a deterministic concurrent-writer regression test in the portable checks.

## Problem

`Scripts/mimo-usage.py` used one shared `usage.json.tmp` path. Two invocations could both write that file, then one rename removed it before the other. The second process crashed in `Path.replace` with `FileNotFoundError`.

A 400-process stress run at parallelism 64 produced 44 failures before the fix.

## Changes

- Generate a UUID-suffixed temporary path beside the cache.
- Keep publication atomic with `Path.replace`.
- Remove only the temporary path owned by that invocation.
- Synchronize two test writers at `os.replace` so the old implementation fails deterministically.

## Verification

- [x] `PYTHONDONTWRITEBYTECODE=1 python3 Scripts/test_mimo_usage.py`
- [x] 400 updates at parallelism 64: 0 failures and 0 temporary files left behind
- [x] Portable checks in `./Scripts/lint.sh lint-linux`, including shell, JavaScript, TypeScript, manifest, and documentation checks
- [x] `git diff --check`
- [ ] `make test` could not start locally because this environment has no `swift` executable
- [ ] `make check` could not start locally because this Linux environment has no `plutil`; the final SwiftLint step also needs `libxml2.so.2`

There was no originating issue for this independently reproduced bug.

## Real multi-process proof

Run against reviewed head `7e750bdeb5d0e1910c4b4b06bde06ffd5305c15d`. Every input launches a separate `python3 Scripts/mimo-usage.py --update` process. All 400 processes share one isolated cache path.

```bash
proof_dir=$(mktemp -d /tmp/codexbar-mimo-real-process.XXXXXX)
trap 'find "$proof_dir" -depth -delete' EXIT
mkdir -p "$proof_dir/mimo/.claude/projects"
export MIMO_CLAUDE_HOME="$proof_dir/mimo"
export MIMO_LOCAL_USAGE_PATH="$proof_dir/usage.json"
export MIMO_PROOF_SCRIPT="$PWD/Scripts/mimo-usage.py"

seq 1 400 | xargs -n1 -P64 sh -c 'python3 "$MIMO_PROOF_SCRIPT" --update >/dev/null 2>"$MIMO_CLAUDE_HOME/error-$1.log"' sh
proof_exit=$?
failures=$(find "$proof_dir/mimo" -name 'error-*.log' -type f -size +0c | wc -l)
temporary_files=$(find "$proof_dir" -name '.*.tmp' -type f | wc -l)
printf 'processes=400 parallelism=64 exit=%s failures=%s temporary_files=%s\n' "$proof_exit" "$failures" "$temporary_files"
jq -er 'if .source == "local-jsonl-scan" and .sessions_scanned == 0 and (.windows | type == "object") then "cache_json_valid=true source=\(.source) sessions_scanned=\(.sessions_scanned)" else error("unexpected cache payload") end' "$proof_dir/usage.json"
exit "$proof_exit"
```

Output:

```text
processes=400 parallelism=64 exit=0 failures=0 temporary_files=0
cache_json_valid=true source=local-jsonl-scan sessions_scanned=0
```

The temporary proof directory contained no account data or credentials.

## Maintainer follow-up

Updated to current main in `5530b92e1ef`. Added complete-payload and failure-cleanup coverage, a loaded-runner barrier allowance, documentation, and changelog credit. The production fix remains unchanged.

Independent verification: the old writer fails the expanded regressions; the fixed writer passes. A separate 400-process/64-parallel run over one synthetic session produced zero failures, zero leftover temporary files, and 1,217 valid concurrent reads with exact token totals. `make check` and 62 Swift MiMo integration tests passed, as did independent Codex autoreview at the configured blocking threshold.

Final `make test`: all 82 groups passed on the first attempt, with zero failures, retries, or timeouts (2,104.7 seconds including discovery), on the unchanged head `5530b92e1ef`. [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/33423278044) passed all seven runnable jobs, including both macOS shards, both Linux builds, lint, and the aggregate gate; musl was correctly path-gated out. GitGuardian also passed. No real account data, credentials, or installed-app state was used or changed.


Lucenx9:
@clawsweeper re-review

Added the requested real multi-process proof to the PR body. It runs 400 separate CLI processes at parallelism 64 against one cache and records `exit=0`, `failures=0`, `temporary_files=0`, plus a validated final JSON cache.

Lucenx9:
@codex review

Lucenx9:
@clawsweeper re-review

The real multi-process proof is now present in the final PR body with a copyable command and terminal output.

Lucenx9 (commented):
Reviewed for correctness. I checked out head `7e750bd` and ran the verification myself:

- Ran `Scripts/test_mimo_usage.py` against this branch: passes 5/5 runs.
- Ran the same test against the old implementation from `main`: fails 5/5 runs, deterministically, with `FileNotFoundError` in one writer. This is a real regression test, not a tautology. The `patch.object(os, "replace", ...)` seam works because `pathlib.Path.replace` resolves `os.replace` at call time.
- Ran 200 real `--update` processes at parallelism 16 against an isolated cache: exit 0, no error logs, no `.tmp` files left behind, cache JSON valid.

Why the fix is correct:

- The old code gave every invocation the same `usage.json.tmp` path. Two writers could both create it, and the first `Path.replace` deleted the file out from under the second, which then crashed with `FileNotFoundError`. UUID suffixes remove the shared state entirely.
- The temp file sits beside the cache, so the rename stays on one filesystem and publication remains atomic.
- `finally: tmp.unlink(missing_ok=True)` cleans up after a failed write and after a successful rename (where the tmp path no longer exists). No litter.
- The dot-prefixed name keeps temp files out of casual glob matches.
- Surgical-changes check passes: 7 lines in the one broken function, no drive-by refactors, no unused imports in the test, `CACHE_PATH` restored in a `finally`.

Two nits, neither blocking:

1. In the `finally`, `tmp.unlink(missing_ok=True)` only swallows `FileNotFoundError`. If unlink fails with another `OSError` (EPERM and friends), it masks the original exception from the `try` block. Practically impossible here; fine to leave.
2. `replace_barrier.wait(timeout=2)` can break under a heavily loaded CI runner and turn this into a flaky failure. Ten seconds costs nothing since the normal path releases immediately.

The barrier test is in-process (threads), so it proves nothing about separate processes by itself, but the 400-process run in the PR description covers that, and keeping the deterministic check in the `lint.sh` portable checks is the right split. `python3` is already required by other portable checks, so this adds no new tool dependency.

Good to merge.

steipete:
Maintainer verification and polish for this PR:

The shared temporary filename is a real collision, and the per-write adjacent temporary file is the right bounded fix. CodexBar remains a reader of this opt-in cache; the tracker owns publication. No cache schema, token accounting, credentials, or platform quota behavior changes are needed.

I strengthened the portable regression to use distinct nonempty payloads and verify that the result is one complete writer payload with no temporary-file leftovers. Added partial-write and replacement-failure checks that preserve the previous cache and another writer's temporary file. The barrier timeout is now 10 seconds, with a 15-second join bound for loaded runners. The production fix is unchanged. Changelog and local-fallback documentation are updated.

Independent proof so far:

- The expanded tests against the old writer on main fail with `FileNotFoundError` in the concurrent case and leftover files in both failure-cleanup cases. The proposed writer passes both tests, including both failure subcases.
- A separate real-process run launched 400 updater processes at parallelism 64 over one synthetic session and isolated cache: zero process failures, zero temporary files, and 1,217 complete cache reads during publication with zero reader errors. Each read checked exact nonzero token totals and the complete three-window payload.
- `make check`: passed, including the portable Python regressions and zero lint violations in 2,076 files.
- `swift test --no-parallel --filter 'MiMoUsageScriptTests|MiMoLocalUsageFallbackTests|MiMoProviderTests'`: 62 tests across three suites passed, with synthetic fixtures and Keychain/session-file isolation.

The full local suite and fresh exact-head CI are still required before landing. Thanks @Lucenx9 for the focused fix and reproduction.

steipete:
Landed as 95d522dae0446d3daab6c876065528d1ea23a0f3. Thanks @Lucenx9; the contributor fix was retained, with maintainer test hardening, documentation, and changelog credit added on this PR.

Final verification on head `5530b92e1efdf71274e674fc32415c561d486dc6`:

- `PYTHONDONTWRITEBYTECODE=1 python3 Scripts/test_mimo_usage.py`: both tests passed, including concurrent complete-payload publication and both injected failure subcases. Against the old main writer, the concurrency case failed with `FileNotFoundError` and both failure cases left temporary files behind.
- Independent real-process proof: 400 executions of `python3 Scripts/mimo-usage.py --update`, parallelism 64, sharing one isolated cache and one synthetic session with nonzero usage. Zero failures, zero temporary files left behind, and 1,217 concurrent reads each matched the complete expected three-window payload; zero reader errors.
- `swift test --no-parallel --filter 'MiMoUsageScriptTests|MiMoLocalUsageFallbackTests|MiMoProviderTests'`: 62 tests across three suites passed, exercising the script and Swift consumer with synthetic fixtures and Keychain/session-file isolation.
- `make check`: passed, including the new portable regressions; zero lint violations in 2,076 files.
- `make test`: all 82 groups passed on the first attempt, with zero failures, retries, or timeouts; 2,104.7 seconds including discovery.
- [Exact-head CI run 33423278044](https://github.com/steipete/CodexBar/actions/runs/33423278044): all seven runnable jobs passed, including both macOS shards, both Linux builds, lint, and the aggregate gate. The musl job was path-gated out; GitGuardian passed.

The full local run was slower under host load, but required no retries. A temporary GitHub API rate-limit error delayed a fresh status read; the normal watcher and final PR check rollup confirmed success before the pinned merge. No credentials or real account history were used, and the installed app was not relaunched. Publication remains last-completed-writer wins; this does not add cache ordering, change token accounting, or alter platform quota behavior.

PR #3325, opened 2026-08-31
fix(grok): show validated empty usage periods

### LINKED ISSUE #3261: Grok: zero-usage weeks show a red "usage unavailable" error — the gRPC zero-usage 0% is parsed but never adopted
### Summary

For Grok accounts on unified-billing consumer plans, CodexBar renders the red diagnostic

> Grok usage is unavailable because its billing sources did not report a usage percentage.

for the entire first stretch of every weekly period while usage is still zero. The data is actually fine — weekly window, reset time, and plan metadata all arrive — but no usage bar can be shown because the zero-usage 0% is parsed and then deliberately discarded during billing-source merging. Another tool that ported this exact parser (cc-switch 3.20.0) shows "Weekly: 0% · resets in 6d21h" for the same account, same credentials, same endpoint.

Verified against CodexBar 0.55.1 (130) and current `main` (612f2f0fb); the merge logic is identical in both.

### What I verified (all tested live)

**1. The CLI ACP `x.ai/billing` method is gone.**
`grok agent stdio` (both 1.0.4 and 1.0.5) answers `{"code":-32601,"message":"Method not found"}` for `x.ai/billing` after a normal `initialize`. The method string still exists inside the CLI binary, but the stdio agent no longer registers it, so `GrokCLIFetchStrategy` always throws on current CLIs and the pipeline always falls through to the next strategy.

**2. The CLI proxy billing endpoint no longer publishes a percentage for unified-billing plans.**
`GET https://cli-chat-proxy.grok.com/v1/billing?format=credits` with the `~/.grok/auth.json` Bearer token returns HTTP 200:

```json
{"config":{"currentPeriod":{"type":"USAGE_PERIOD_TYPE_WEEKLY","start":"<period start>","end":"<period end>"},"onDemandCap":{"val":0},"onDemandUsed":{"val":0},"isUnifiedBillingUser":true,"prepaidBalance":{"val":0},"billingPeriodStart":"<period start>","billingPeriodEnd":"<period end>"}}
```

No `creditUsagePercent`, and `onDemandCap`/`onDemandUsed` are 0, so `GrokCreditsProxyFetcher.parseSnapshot` correctly yields `usedPercent: nil` with a valid `resetsAt` (the #3157 "unknown usage" state). Making a real completion right after the period rollover still returns the same shape — usage rounds to 0 and proto3 omits the field.

**3. The grok.com gRPC endpoint works with just the Bearer token.**
`POST https://grok.com/grok_api_v2.GrokBuildBilling/GetGrokCreditsConfig` with `Authorization: Bearer <auth.json token>` (no browser, no cookies) returns HTTP 200 with `grpc-status: 0`:

```
00000000440a4212001a00220b0887a8c6d40610f0d7dd142a0b08879debd40610f0d7dd14
421c0802120b0887a8c6d40610f0d7dd141a0b08879debd40610f0d7dd1458016200680180
0000000f677270632d7374617475733a300d0a
```

Decoded: **no `fixed32` (float) fields at all**, a reset timestamp at path `[1,5,1]`, and a usage-period marker `[1,8,1] = 2`. `GrokWebBillingFetcher.parseGRPCWebResponse` handles exactly this shape: `noUsageYet` → `usedPercent: 0` with `usedPercentIsWirePublished: false`.

**4. But the merge step rejects that 0%.**
`GrokOAuthFetchStrategy.resolvingUnknownUsage` only adopts the gRPC enrichment when `usedPercentIsWirePublished` is true, so the zero-usage 0% is dropped, the proxy answer (unknown) wins, and the fetch finishes with `usageUnavailableMessage`.

### Why this matters

- The state is genuinely knowable. proto3 omits zero-valued fields, so "period + reset markers + no percent on the wire" *is* the server's "0% used" representation for this plan family. `parseGRPCWebResponse` already encodes that knowledge — it is thrown away one layer up.
- The red-error window is long: essentially the whole start of every weekly period, until usage is large enough for xAI to publish a non-zero float.
- The behavior is also inconsistent inside CodexBar itself: a snapshot carrying the zero-usage 0% that arrives via `GrokWebFetchStrategy` is displayed as a 0% bar (that path only errors on `usedPercent == nil`). Whether the user sees a red error or a 0% bar depends purely on which strategy wins the auto pipeline.

For reference: cc-switch 3.20.0 ported `parseGRPCWebResponse` including the zero-usage read and adopts it, so it displays "Weekly 0% / resets in 6d21h" for the same account while CodexBar shows the red error.

### Suggested fixes (either would remove the red error)

1. **Adopt the zero-usage 0% in `resolvingUnknownUsage`.** Have the parser expose the `noUsageYet` outcome distinctly (e.g. `usedPercentIsZeroUsage`, instead of conflating it with `usedPercentIsWirePublished: false`), and accept a gRPC snapshot that has `resetsAt != nil` and a zero-usage 0%. This keeps the #3157 protection for genuinely unknown percents while treating the server's explicit "no usage yet" frame as what it is.
2. **Or keep "unknown" but stop rendering it as an error.** When the winning snapshot has period metadata (`resetsAt`, plan) but no percent, show the reset countdown with a neutral "xAI hasn't published usage yet" note instead of the red `usageUnavailableMessage`.

Option 1 matches what users see on grok.com and cc-switch (0%); option 2 is the conservative UX-only fix.

Happy to test a build.

sf-jin-ku:
A freshly reset Grok period could remain “usage unavailable” because the quota resolver discarded the billing parser's implicit zero. Accept that zero only when it comes from one complete protobuf response containing a recognized active current period, with no conflicting percentage. The resolver keeps arbitrary inferred zeros, malformed or truncated zero responses, inactive periods, and invalid percentage payloads unavailable.

The snapshot carries the parser's provenance through plan and reset overlays. The regression fixture exercises the reported payload through parsing, enrichment, and UsageSnapshot conversion, alongside rejection cases and preservation of the existing six-second enrichment timeout.

Fixes #3261. Thanks @sf-jin-ku and @olddonkey for the implementations and reproduction evidence.

Validation: 91 focused tests across four Grok suites; make check; independent Codex review. All nine checks passed on commit 24eba74c3bd06120e8177d30b2b4aed00765eb7f, including both full macOS test shards and Linux builds ([CI run](https://github.com/steipete/CodexBar/actions/runs/33475470989)). Local make test remains incomplete: shared filesystem stalls caused unrelated process-fixture and AppKit timeouts; those failures are not counted as passes.


sf-jin-ku:
Post-fix proof from this exact head `046bfc255` (auth.json bearer, no cookies):

```text
# shipped 0.56.1
source: grok-cli-proxy
primary: null
diagnostic: Grok usage is unavailable because its billing sources did not report a usage percentage.

# this head CLI
source: grok-web
primary.usedPercent: 0
primary.resetsAt: 2026-09-07T10:19:23Z
diagnostic: (absent)
loginMethod: SuperGrok Heavy
```

PR body updated with the redacted transcript.

@clawsweeper re-review

PR #3335, opened 2026-08-31
test(cost): shorten lock-contention fixture waits

### LINKED ISSUE #3334: perf(test): shorten CostUsageStore lock-contention waits without changing production timeout
## Summary

`CostUsageStoreTests` currently spends almost its entire wall time waiting for the production SQLite busy timeout in four lock-contention tests.

The tests need to prove that a blocked writer fails soft, preserves the store, and succeeds after the lock is released. They do not need to wait the full production five seconds to prove those state transitions.

This issue proposes constructor-injecting the busy timeout with the existing 5000ms production default, then using a small explicit timeout only in the four tests that intentionally hold a writer lock.

## Measured current cost

Current main: `e236a21bc27b4caf7bb0ff151059131c332b0ded`.

A focused isolated run of `CostUsageStoreTests` passed 75 tests in 43.968s. The four intentional lock tests account for 43.0s:

| Test | Time |
| --- | ---: |
| identical scanner save reports retry while the writer lock is unavailable | 15.495s |
| held writer lock makes freshness advance fail soft without rebuilding | 9.347s |
| write lock held by another process skips the write instead of deleting the store | 9.272s |
| locked compatible predecessor is preserved and adopts after retry | 8.883s |

The surrounding full-suite group took 51.783s. A separate eight-worker profile measured `CostUsageStoreTests` at 40.408s, so the focused result is representative.

## Root cause

`CostUsageStore.configure` hardcodes:

```swift
sqlite3_busy_timeout(database, 5000)
```

The four tests deliberately hold `BEGIN IMMEDIATE` on another connection and then exercise one or more store operations. Each blocked operation therefore waits up to the full production timeout before returning `SQLITE_BUSY`.

This is valid production behavior and should remain the default. The long delay is not part of the test contract.

## Proposed implementation

1. Add a `busyTimeoutMilliseconds` constructor argument to `CostUsageStore`, defaulting to 5000.
2. Store it per actor instance and pass it to `configure` when opening the SQLite connection.
3. Keep the existing test that proves a default store configures `PRAGMA busy_timeout = 5000`.
4. In only the four intentional lock-contention tests, create the affected store instances with a small explicit timeout such as 25ms.
5. Preserve every existing state, retry, adoption, rebuild-count, and file-preservation assertion.

Prefer constructor injection over a mutable global or `#if DEBUG` hook. It keeps fixtures independent and leaves the production call sites unchanged.

## Proof contract

- record at least three focused baseline runs of `CostUsageStoreTests`
- run the four affected tests repeatedly with the short timeout
- run `CostUsageStoreTests` at least three times on the candidate
- prove the default store still reports a 5000ms busy timeout
- run `make check`
- run `make test` once on the final tree
- require at least a 10x median improvement for `CostUsageStoreTests`

## Acceptance criteria

- [ ] Production and all unchanged call sites retain a 5000ms SQLite busy timeout.
- [ ] Only intentional lock-contention fixtures opt into the shorter timeout.
- [ ] The four tests still exercise real SQLite connections and a real held writer lock.
- [ ] Fail-soft, retry, predecessor-adoption, rebuild-count, and file-preservation assertions are unchanged.
- [ ] No process-global mutable test override is introduced.
- [ ] Three candidate `CostUsageStoreTests` runs pass with at least a 10x median improvement.
- [ ] `make check` and one final `make test` pass.

## Non-goals

- Do not lower the production busy timeout.
- Do not replace SQLite lock contention with mocks.
- Do not weaken store corruption/rebuild behavior.
- Do not change the local group runner or CI sharding in this PR.
- Do not combine this with #3331 or #3332.

## Related work

- #2761 introduced the SQLite store and the five-second busy timeout.
- #2762 cut persistence over to SQLite.
- #3331 covers process-level test-runner orchestration.
- #3332 covers the independent PTY fast-exit stress loop.

brzvsk:
Lock-contention tests were spending the production five-second SQLite busy timeout on each deliberately held lock. Make that timeout configurable per store instance, retain the 5,000 ms production default, and use 25 ms only in the four lock-contention fixtures. Assertions cover the default and override.

Fixes #3334. Thanks @brzvsk.

Validation: 77 focused tests passed; make check passed; independent Codex review found no actionable issues. All nine checks passed on 8782a0f42ae6e513e6202c223375dc0b37d5265f, including both full macOS shards and Linux builds ([CI run](https://github.com/steipete/CodexBar/actions/runs/33475339810)). Local make test passed the first 72 groups, then hit unrelated AppKit timeouts during shared filesystem stalls; resumed coverage is recorded separately and the incomplete local run is not counted as a pass.


brzvsk:
@clawsweeper review

brzvsk:
Fixed the sole ClawSweeper finding in 4589f640a: the adjacent comment now describes the fixture-specific short busy timeout instead of the old five-second production timeout. `make check` remains green. The change is documentation-only, so the existing focused and full-suite runtime receipts remain valid.

@clawsweeper re-review

PR #3336, opened 2026-09-01
fix(grok): preserve billing semantics in nonempty requests

CharlieLZ:
## Summary

Addresses the reported `grpc-status 13` / `Missing request message.` rejection by sending a nonempty `GetGrokCreditsConfigRequest`, while preserving the default billing semantics.

The current official Grok descriptor defines request field 1 as `bool exclude_legacy_monthly_usage`, not a weekly-period enum. The web client requests `{}`. This patch explicitly encodes false (`00 00 00 00 02 08 00`) so the message is nonempty without excluding legacy monthly usage. Decoding with `protoc` confirms empty and `0800` both mean false, whereas the original `0802` means true.

The maintainer repair also restores the 15-second timeout and the displaced `@Test`. Response parsing is unchanged from main. The captured frame's nonzero timestamp nanoseconds are not usage limits, so its inferred zero must not replace an unknown proxy percentage. The fixture now covers that boundary through the real parser, proxy enrichment, and usage snapshot; explicit wire zero/nonzero percentages still work.

The later unified-billing `0/0` inference has also been removed. Its fixtures remain as regression coverage: absent subscription usage stays unknown, and real fallback usage can enrich it without changing the authoritative reset date. Zero on-demand spending does not establish zero subscription usage.

Thanks @CharlieLZ for the report, original request change, and captured regression fixture.

## Protocol evidence

- [Official public billing descriptor](https://cdn.grok.com/_next/static/chunks/32g78bk5hhe1q.js): decoded its embedded protobuf descriptor, including the actual request boolean and timestamp field types.
- [Official public web billing query](https://cdn.grok.com/_next/static/chunks/062ueaj23tfo3.js): requests the default config and reads the returned monthly/weekly period.
- [Protobuf field presence](https://protobuf.dev/programming-guides/field_presence/): explicit false preserves the implicit proto3 default.

Public assets checked September 1, 2026 UTC; downloaded JavaScript was inspected, not executed.

## Verification and merge gate

- The corrected regression tests fail on the submitted implementation: false-zero adoption, request bytes, and lost timeouts are detected.
- Focused billing/proxy tests pass on `10cefdc9c4cbf470ce882d27ef3cc2432b450b69`: 81 tests across two suites, with stubbed HTTP and synthetic credentials. The same compiler was used through SwiftPM's native backend after Xcode's SDK stat-cache tool stalled before compilation.
- Independent Codex autoreview is clean at its configured P0 threshold; maintainer review also checked the request schema, fallback ownership, and lower-severity regressions.
- `make check` passes with zero lint violations across 2,079 Swift files.
- [Current-head CI](https://github.com/steipete/CodexBar/actions/runs/33470597985) passes all eight jobs on `10cefdc9c4cbf470ce882d27ef3cc2432b450b69`, including both macOS test shards and all three Linux builds.
- The earlier full local repository run stopped on a 600-second `StatusMenuTests` timeout, after two timing-sensitive failures recovered on retry. It is not a green full-suite result. The independently diagnosed Sakana cancellation-test race is addressed separately in [#3344](https://github.com/steipete/CodexBar/pull/3344).

**Leave unmerged until the corrected `0800` request is confirmed against the affected account.** No authenticated provider probe, real cookie import, Keychain access, or installed-app relaunch was performed for this repair. The original reported 0% result was produced by the discarded parser heuristic and is not proof of published usage or of this corrected request's live recovery. Unknown percentages intentionally remain unknown.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Codex: avoid repeated full scans after trace-log pruning, while retaining the latest validated cost history through temporary trace-database failures (#3318). Thanks @brzvsk!
 
 ### Fixed
+- Grok: restore the grok.com billing fetch, which grok.com now rejects when it receives an empty request message (`grpc-status 13`); the client sends the weekly usage period explicitly. SuperGrok Heavy weekly-limit responses that declare a per-period limit without a used amount are now treated as a wire-published 0% instead of unknown usage, so the menu usage bar returns instead of an unavailable-usage diagnostic.
 - Menu bar: keep status components and website links scoped to their provider when switching cached tabs, preventing Claude status from appearing under Grok or Codex (#3320). Thanks @gianpaj!
 - Usage & Spend: keep stalled or failed Codex catch-up paused until explicit Refresh, preventing background synchronization from restarting CPU-heavy scans (partial fix for #3316). Thanks @heyajulia!
 - Xiaomi MiMo: prevent overlapping local usage tracker updates from colliding on a shared temporary cache file, preserving atomic publication (#3321). Thanks @Lucenx9!
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,11 @@
 - Codex: avoid repeated full scans after trace-log pruning, while retaining the latest validated cost history through temporary trace-database failures (#3318). Thanks @brzvsk!
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


steipete:
Thanks @CharlieLZ. The focused replacement PR adopts `00 00 00 00 02 08 00` and request-contract coverage. The public descriptor confirms field 1 is the boolean `exclude_legacy_monthly_usage`, so explicit false preserves the default. Response parsing and unknown-percentage behavior are unchanged. Adopted in https://github.com/steipete/CodexBar/pull/3805 at `446c3bafd5ea`; four assertions fail before the correction and pass afterward, the focused suites pass, and `make check` is clean. Affected-account recovery remains unverified. CI passed and #3805 is merged.

PR #3338, opened 2026-09-01
fix(codex): restore OAuth subscription renewal dates

CharlieLZ:
## Problem

Codex OAuth-mode usage did not surface subscription renewal/expiry dates: the ChatGPT `subscriptions` payload was parsed only on the web-dashboard path, so OAuth-mode cards lost the renewal metadata (rebased from ad636d819 whose architecture has since moved on; the change applies cleanly to current main).

## Change

- Parse `active_until`/`will_renew` from the subscriptions payload in the Codex OAuth usage fetcher (new `OpenAISubscriptionDates` value kept WebKit-free).
- Attach renewal metadata to the reconciled state so the menu card shows renewal dates like the web mode.
- Tests covering payloads with/without renewal + response shape diagnostics.

### DIFF
--- a/Sources/CodexBarCore/OpenAISubscriptionDates.swift
+++ b/Sources/CodexBarCore/OpenAISubscriptionDates.swift
@@ -1,0 +1,42 @@
+import Foundation
+
+/// Renewal/expiry dates from ChatGPT's subscriptions payload.
+/// Kept independent of WebKit so OAuth usage can attach subscription metadata without browser access.
+public struct OpenAISubscriptionDates: Equatable, Sendable {
+    public let expiresAt: Date?
+    public let renewsAt: Date?
+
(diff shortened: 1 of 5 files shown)


steipete:
Thanks @CharlieLZ for investigating the missing renewal dates. We checked this against current main. OAuth-only usage still does not supply subscription dates, but this change obtains them through a new `/backend-api/subscriptions` request on each OAuth refresh, rather than parsing metadata from the existing usage response.

Renewal dates can already enrich OAuth-backed cards through optional OpenAI web extras. That path was hardened in #3373 and now shares a WebKit-independent `OpenAISubscriptionMetadata` model. We’re keeping subscription fetching within that optional web path for now, so we’re closing this proposal. OAuth-only billing enrichment would need a separate decision about the additional request; #3373 does not implement that capability.

PR #3341, opened 2026-09-01
Fix GitHub Enterprise Copilot tracking

Fletcher-Alderton:
Enterprise device authorization and quota requests already used the configured GitHub host, but sign-in identity still went to api.github.com. This repair uses the existing API-host resolver for identity and keeps accounts with equal numeric user IDs on different hosts distinct.

Public GitHub retains its existing stable and legacy account identifiers. Enterprise sign-in requires a verified identity and matches only the same API issuer and user ID; it never probes unidentified legacy tokens or adopts an account by login/label. Public-GitHub browser budget enrichment is skipped for Enterprise quota requests. Stored identifiers remain deduplication metadata; the configured host still determines request routing.

A login captures the existing provider configuration revision before any network await and rejects changed configuration both before and after legacy matching. Account edits/removals, host changes, and cancellation therefore cannot commit a stale login. The refactor shares existing normalization and login-error handling. The usage API-version header remains unchanged because the original trace does not establish that it causes a failure.

Validation: 64 focused tests across six suites pass, including injected HTTP endpoint/headers, zero public-budget requests for Enterprise quota, public plus two Enterprise issuers with equal IDs/logins, legacy compatibility, exact reauthentication, cancellation, and account edits/removals before and during identity resolution. `make check` passes with zero lint violations. Independent P0–P2 review is clean after fixing its delayed-revision finding. The full suite passes all 1,031 selections across 86 groups on the first attempt, with zero retries/timeouts, in 841.6 seconds. Exact-head CI [34160173751](https://github.com/steipete/CodexBar/actions/runs/34160173751) passes macOS tests, Linux builds, lint, and security checks on 4ad9db91df59c4a1c508003686ad9f1e33332497. The run initially required GitHub workflow approval; after verifying no workflow changes, it was approved and passed without source changes.

The contributor's original redacted Enterprise trace on 7e909419bd8523ad9e495476eb8b2b8938d9484d observed identity and usage HTTP 200 responses against the configured Enterprise API host. That is historical contributor evidence; the maintainer replacement is verified with synthetic transports and isolated settings, without real credential or browser-cookie probes.

Production delta: +13 lines. It is funded by the remaining nine lines from landed cleanup #3479 after pricing #3476 and four of the five previously unallocated cleanup lines. The cleanup balance remains positive. Documentation and the Unreleased changelog are updated. Thanks @Fletcher-Alderton for the original endpoint repair and Enterprise verification.


Fletcher-Alderton:
@clawsweeper re-review

PR #3344, opened 2026-09-01
test(sakana): await cancellation with a bounded deadline

steipete:
## Summary

Fixes a synchronization race in the Sakana cancellation tests, without changing provider behavior.

The full local test run observed the primary quota result within its existing 500ms limit, but the test exhausted 1,000 scheduler yields before the transport actor recorded cancellation. The suite passed on retry. A yield count is not an acknowledgement or a time bound.

Both cancellation checks now share an actor helper with a one-second monotonic deadline and 10ms polling. The primary-response assertion remains before the wait, and the acknowledgement deadline remains below the production optional-request timeout of five seconds. No production deadline is relaxed.

## Verification

- File-scoped SwiftFormat and strict SwiftLint pass; `git diff --check` passes.
- Independent Codex autoreview found no accepted findings at its configured P0 threshold; maintainer review checked actor reentrancy and the production cancellation boundary.
- All 23 Sakana tests pass; repository `make check` passes with zero violations across 2,079 Swift files.
- The fresh full `make test` run passes all 82 groups (984 selections) on this head, with zero failures, retries, or timeouts. It uses the same compiler through SwiftPM's native backend after the SDK stat-cache tool stalled in the default backend.
- Ten additional focused repetitions pass: 230 Sakana test executions with the primary-response latency assertion intact.
- [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/33469912103) passes all required checks: both macOS test shards, Linux x64/arm64 builds, lint, and the aggregate gate. The musl build is correctly skipped by the test-only path gate.
- The originating full local run stopped later on an unrelated 600-second `StatusMenuTests` timeout; this PR does not claim to fix that timeout or the separately retried Kiro process-cleanup failure.

Test-only change; no changelog entry or live credentials required.

### DIFF
--- a/Tests/CodexBarTests/SakanaUsageFetcherTests.swift
+++ b/Tests/CodexBarTests/SakanaUsageFetcherTests.swift
@@ -172,10 +172,7 @@ struct SakanaUsageFetcherTests {
         #expect(snapshot.fiveHour?.usedPercent == 92)
         #expect(snapshot.payAsYouGo == nil)
         #expect(startedAt.duration(to: .now) < .milliseconds(500))
-        for _ in 0..<1000 where await !(transport.didCancelPayAsYouGo()) {
-            await Task.yield()
-        }
-        #expect(await transport.didCancelPayAsYouGo())
+        #expect(try await transport.waitForPayAsYouGoCancellation())
     }
 
     @Test
@@ -192,10 +189,7 @@ struct SakanaUsageFetcherTests {
                 session: transport)
         }
 
-        for _ in 0..<1000 where await !(transport.didCancelPayAsYouGo()) {
-            await Task.yield()
-        }
-        #expect(await transport.didCancelPayAsYouGo())
+        #expect(try await transport.waitForPayAsYouGoCancellation())
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


steipete:
Merged as [a140553c9ef7](https://github.com/steipete/CodexBar/commit/a140553c9ef7c9629da9e36f3d6d0c1485624e07).

The cancellation tests now wait for the transport actor's acknowledgement with a monotonic deadline instead of assuming that 1,000 scheduler yields are enough. The existing primary-response latency assertion stays before that wait; production behavior and timeouts are unchanged.

Verification on PR head `16164a7729e65f672bb591a531ae0897b3646f8e`:

- `swift test --filter SakanaUsageFetcherTests`: all 23 tests pass, followed by ten additional passing repetitions (230 test executions).
- `make test`: all 82 groups / 984 selections pass, with zero failures, retries, or timeouts. The same compiler used SwiftPM's native backend after the default backend's SDK stat-cache tool stalled; suite isolation and Keychain suppression remained enabled.
- `make check`: zero violations across 2,079 Swift files.
- Independent Codex autoreview at its configured P0 threshold and maintainer review of the cancellation boundary found no accepted findings.
- [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/33469912103): both macOS shards, Linux x64/arm64, lint, and aggregate checks pass; musl is correctly skipped for this test-only change. GitGuardian also passes.

The merged tree was verified identical to the tested PR tree. This is test-only, so there is no changelog entry, UI proof, or live-account test. It does not claim to fix the separate StatusMenu timeout or Kiro cleanup failure seen in an earlier full run.

PR #3350, opened 2026-09-01
feat: icloud device removal

### LINKED ISSUE #3234: No way to remove a stale duplicate device record in iCloud Sync
some part of body was generated by AI.

## Summary
Settings → iCloud Sync → "Macs" remains after reinstall.
Causes showing old usage card below normal usage card

Image
Image

**suggest adding feature: delete synced device in iCloud - Macs.**
## Environment
- CodexBar 0.55.1 (macOS)
- Single physical Mac, one Claude account
- Claude CLI fully functional; direct CLI usage checks return current data

## Steps to reproduce
1. use codexbar with icloud sync
2. re-install codexbar

## Observed behavior
- The stale device entry persists indefinitely with no way to delete it
  from either the CodexBar Settings UI or the app's Debug tab (checked
  Advanced → Show debug settings → Debug tab; no iCloud/sync reset option
  present there).
- Disabling iCloud Sync entirely (all checkboxes off) stops the stale
  entry from being *shown* in the menu, but does not appear to delete the
  underlying record — re-enabling sync would presumably surface it again.
- `codexbar diagnose --provider claude --format json --pretty` confirms
  the live local/web fetch path is healthy and returns current data, so
  the stale display is purely a leftover-snapshot issue, not a fetch
  failure.

## Expected behavior
- The "Macs" list in Settings → iCloud Sync should allow removing
  individual device records (especially duplicates for the same physical
  machine), so stale snapshots stop being retained/shown.
- Ideally, CodexBar would also detect and auto-merge duplicate records for
  the same hostname/device rather than requiring manual cleanup.

**consider adding feature: reset icloud data, so user could temporarily solve icloud related problem.**

Happy to provide `codexbar diagnose` output or a debug log bundle if useful.

kwnms04:
Closes #3234.
**This is not completely verified. see below: Verification**
**Generated by Claude, Reviewed by kwnms04 and Claude.**

## What this does

Settings → iCloud Sync → **Macs** now lists every synced Mac and lets you remove one, with confirmation. Removing a Mac deletes its Device Record and every Usage Snapshot it published, on every Mac on the account.

The list renders from local persistence, so it is visible even with syncing off — otherwise there is nothing to remove a stale device *from*, which is the state the issue reports.

## Why the row does not disappear immediately

An earlier revision removed the row optimistically and put it back on failure. That machinery was the source of most review findings, so it is gone. The row now goes on one signal: CloudKit confirmed the delete, or reported the record was already gone. That is the same path a delete made on another Mac already took, so both cases share one code path.

A Device Record delete is attempted **once**. Retrying would race the Mac that owns it re-registering itself and delete the record it just published. Failures that get no retry are surfaced in the pane instead of being dropped, and clear when a push completes with nothing failing in it.

## Scope

Two deliberate departures, both recorded as ADRs:

- **ADR 0002** — removal is offered for the current Mac too, though triage scoped this to "a selected non-current device". The current Mac re-registers on the next sync start, and `queueDeviceRecord` cancels its own pending delete, so the removal is self-healing rather than destructive.
- **ADR 0001** — the Device Identifier is derived from the Mac's hardware UUID rather than minted per install, so a reinstall stops stranding a duplicate. This is beyond what the issue asked for. **It has a cost the ADR states plainly: every Mac syncing today creates one more Stale Device at its next reinstall.** Existing duplicates are not merged — the issue lists auto-merge as out of scope, and triage confirmed "without automatic hostname merging".

Also out of scope, as the issue notes: "reset icloud data".

## Duplicates are the point

The confirmation dialog shows the device's absolute last-seen time, because the case this issue is about is two rows with the *same host name*. On the machine this was developed on, `kwnmbp.local` currently has four Device Records (0.49.4 → 0.55.1 → 0.55.1 → 0.56.1). Host name alone cannot tell them apart.

## Verification

- `make check` — swiftformat clean, swiftlint `--strict` 0 violations across 2075 files, locales 22 catalogs / 1494 keys, packaging and signing checks pass.
- `make test` — all 82 groups run green except `ClaudeOAuthCredentialsStoreNeverPromptCacheTests`, which shells out to `/usr/bin/defaults`; that tool cannot write any domain on this machine, so those two tests fail for reasons unrelated to this change. The other 11 suites in that group were run separately and pass.
- New tests cover the confirmed-delete rule, the no-retry policy for Device Records, the reporting rule, and that queueing a delete leaves the fleet row in place.

**Not verified locally:** the removal itself, the confirmation dialog, and the failure banner. iCloud entitlements are only embedded in identity-signed release builds of `com.steipete.codexbar`, and this machine has no signing identity. What was verified against a running build is that the Macs list renders with syncing off and the remove controls are correctly disabled without the entitlement.

## Follow-ups, not addressed here

1. `CloudSyncState.removeDevice` is a closure on an `@Observable` state object — behaviour where state belongs.
2. `CloudSyncSnapshotMigration` now owns Device Record delete policy as well as snapshot migration; it has two reasons to change.
3. Record kind is carried as a `"device-"` name prefix rather than a type.
4. Snapshots the local cache never saw are not deleted, since removal reads the local envelope rather than querying CloudKit.
5. Removal stays unavailable while syncing is off; the footer says so, but the issue's original complaint is only partly answered.
6. A delete queued but not sent before the app quits is only requeued through the snapshot push, which is gated on snapshot syncing being on.
7. `macsFootnote` concatenates two localized sentences, which is not translatable as a unit.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -25,6 +25,11 @@ _Avoid_: Device ID, install ID, client ID
 The set of Devices syncing under a single iCloud account.
 _Avoid_: Cluster, group, network
 
+**Fleet Cache**:
+This Mac's copy of the Fleet, as of its last fetch. It is authoritative up to that moment and
+no further: anything published afterwards is missing from it until the next fetch.
+_Avoid_: Local state, mirror, replica
+
 **Usage Snapshot**:
 What one provider account's usage was at one moment, as seen on one Device. Snapshots are
 per-Device: two Devices reporting the same account hold separate Snapshots.
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -1,0 +1,32 @@
+# CodexBar
+
+Menu bar app that reports AI provider usage and credits. This glossary covers terms whose
+meaning is specific to CodexBar and easy to confuse with a neighbouring concept.
+
+## Language
+
+### iCloud Sync
+
+**Device**:
+One physical Mac in the user's fleet. A Device outlives reinstalls of CodexBar: reinstalling
+does not produce a second Device.
+_Avoid_: Machine, installation, node, client
+
+**Device Record**:
+The CloudKit record that represents a Device. One per Device.
+_Avoid_: Mac record, host record
+
+**Fleet**:
+The set of Devices syncing under a single iCloud account.
+_Avoid_: Cluster, group, network
+
+**Usage Snapshot**:
+A point-in-time usage reading for one provider account, taken on one Device. Snapshots are
+per-Device: two Devices reporting the same account hold separate Snapshots.
+_Avoid_: Usage record, sample, reading
+
+**Stale Device**:
+A Device Record that no live Mac will ever refresh, because the installation that created it
+is gone. Its Usage Snapshots keep being projected into the menu even though nothing updates
+them.
+_Avoid_: Orphan, ghost, dead device, duplicate
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -8,7 +8,7 @@ meaning is specific to CodexBar and easy to confuse with a neighbouring concept.
 ### iCloud Sync
 
 **Device**:
-One physical Mac in the user's fleet. A Device outlives reinstalls of CodexBar: reinstalling
+One physical Mac in the user's Fleet. A Device outlives reinstalls of CodexBar: reinstalling
 does not produce a second Device.
 _Avoid_: Machine, installation, node, client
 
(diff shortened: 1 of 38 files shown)


kwnms04:
Pushed 39bc8228a addressing two of the review findings.

**Clear resolved unknown-item deletion errors** (`CloudSyncEngine.swift:849-852`) — the banner now clears when the remaining delete failures are non-reportable, and confirmed deletes count as push progress. An all-`unknownItem` result no longer leaves a red banner on a removal that landed.

**Retain a recovery handle for partial snapshot deletion** (`CloudSyncEngine.swift:1353-1358`) — resolved by removing the problem rather than keeping the row: a confirmed Device Record delete now drops the Usage Snapshots that device left behind (`CloudSyncDeviceRemoval.orphanedSnapshotNames`), so no snapshot is projected into the fleet menu without a Macs row to remove it from. Holding the row open would have offered a retry for a delete that is deliberately not retried. Covered by `snapshots of a removed device go with it`.

Still open:
- **[P1] Reconcile all remote snapshots before completing removal** — removal still works off the local `fleetSnapshots` cache.
- **Real behavior proof** on an entitled signed build (confirmed / partial-failure / already-absent).
- **Owner decision** on the device-identity migration and current-Mac removal scope.

I'm trying to close those 3 findings.

kwnms04:
Pushed 39bc8228a and bb885b559, addressing all three findings.

**[P1] Reconcile all remote snapshots** (`CloudSyncEngine.swift:1307`) — removal now fetches zone changes first and builds the delete set from what comes back, rather than from the local cache alone. A failed fetch stops the removal instead of deleting against an unverified set. A CKQuery would guarantee no more: both are authoritative only up to the moment they run.

**[P2] Retain a recovery handle for partial snapshot deletion** (`:1353-1358`) — resolved by removing the problem rather than keeping the row: a confirmed Device Record delete now drops the Usage Snapshots that device left behind, so none is projected with no Macs row to remove it from. Holding the row open would offer a retry for a delete that is deliberately not retried.

**[P2] Clear resolved unknown-item deletion errors** (`:849-852`) — the banner clears when the remaining delete failures are non-reportable, and confirmed deletes count as push progress.

Still open, both outside the diff:
- Real behavior proof on an entitled signed build (confirmed / partial-failure / already-absent).
- Owner decision on the device-identity migration and current-Mac removal scope.

steipete:
Thanks @kwnms04 for the implementation and its documented verification limits. I recommend closing this version while keeping #3234 open. The patch permits current-Mac removal, selects deletions from a cache whose queued fetch events may still be pending, and hides local snapshots even when CloudKit rejected their deletion. The helper tests do not establish the complete fetch/send/retry/relaunch behavior. A bounded implementation should preserve existing installation IDs, reject current-device deletion, and first introduce a fake transport/event-completion seam for synthetic deletion tests.

PR #3361, opened 2026-09-01
fix(menu): reject non-finite saved status item positions

steipete:
## Summary

Reject non-finite saved menu-bar positions before status-item creation. NaN bypassed both numeric comparisons, and positive infinity was retained when no display bound was available. A single finite-value check closes those holes for both the current autosave key and its matching legacy key.

Finite positions, screen-bound calculation, padding, missing keys, unrelated items, and visibility policy are unchanged. The tests use isolated defaults suites and explicitly verify that the non-finite values survive the defaults round-trip before cleanup.

This was found while investigating #3355, but does **not** establish or fix that report's recurrence involving a finite position. That issue must remain open for runtime evidence. No user preferences were reset, no live status items were created, and no provider or Keychain access was used.

## Verification

- Baseline regression: three invalid-value cases produce nine expected assertion failures; finite controls pass.
- `swift test --no-parallel --filter 'status item placement preflight|ProviderArchitectureGatekeeperTests'`: 53 tests across two suites pass (43.399 seconds), including six non-finite cases and four finite controls.
- Codex autoreview is scoped-clean at its configured P0 threshold; maintainer review and an independent feasibility check agree with the narrow owner-level change.
- `make check`: passed with zero violations across 2,083 Swift files; all 95 process-cleanup tests passed with one expected Linux-only skip.
- Exact-head [CI run 33537432141](https://github.com/steipete/CodexBar/actions/runs/33537432141) passed on `fa8625d6bf415887b6d801075f0933555bd13382`: both complete macOS shards, Linux x64/arm64 builds and tests, Linux musl build, lint, and the aggregate gate. Hosted macOS discovered 988 selections; the full-suite merge gate is this hosted run, not a claim that the local full run passed. Both plugin engines also passed their 79-test golden suites.
- The first full run timed out in group 1 after 600 seconds; a process sample showed an unrelated Antigravity fixture blocked in a filesystem rename during an atomic temporary-file write. The automatic retry sequence was interrupted through the runner's cleanup path and all observed owned processes exited. An unchanged Antigravity diagnostic later passed all seven tests in 1.978 seconds, but it did **not** prove an internal-disk recovery: the task-local Swift launcher was still overriding the requested `TMPDIR`. That launcher override has now been corrected. No product source or test deadline was changed for this environment repair.
[...]

PR #3363, opened 2026-09-01
fix(poe): use the refresh clock for history windows

steipete:
## Summary

Use Poe's supplied host refresh clock consistently for the rolling 30-day history cutoff and UTC Today row. The plugin previously called wall-clock `Date.now()` twice, ignoring the runtime's explicit `now` argument and potentially crossing a time boundary within one refresh.

The existing parity golden exposed this on September 1: its August 2 record aged out under wall-clock time even though the fixture supplied an August 4 refresh date. Keep that golden unchanged, capture `ctx.date.now()` once, and add regressions for Today, exact-cutoff inclusion, exclusion after the cutoff, and expired-history balance retention.

Request endpoints, authentication, pagination bounds, optional-history failure handling, and rolling-duration semantics are unchanged. This is separate from the non-finite placement fix in #3361, whose full-suite gate exposed the problem. No live provider, browser, or credential access was used.

## Verification

- Baseline: existing golden plus new clock cases reproduce five assertion failures across four test functions.
- `swift test --no-parallel --filter 'Poe|ProviderPluginDetailsParityTests|date now'`: all 27 tests across six suites pass under QuickJS (2.721 seconds) and JavaScriptCore (4.384 seconds), including optional-history failure preserving the required balance.
- `make check`: passed with zero lint violations across 2,083 Swift files; all 95 process-cleanup tests passed with one expected Linux-only skip.
- Codex autoreview is scoped-clean at its configured P0 threshold; maintainer review and the independent clock-contract audit found no actionable issue. An initial bundle preparation detected concurrent index staging and refused before review; the completed review used the final stable candidate.
- Full `make test` and exact-head hosted CI: pending completion. Do not merge until green.

## Release-note context

Poe: use one refresh timestamp for recent-history filtering and Today totals. No changelog edit; release generation owns that work.

### DIFF
--- a/Sources/CodexBarCore/Resources/Plugins/poe.js
+++ b/Sources/CodexBarCore/Resources/Plugins/poe.js
@@ -13,6 +13,7 @@ defineProvider({
   ],
 
   async fetchUsage(ctx) {
+    const now = ctx.date.now();
     const balanceResponse = await ctx.http.getJSON("https://api.poe.com/usage/current_balance");
     if (balanceResponse.status === 401 || balanceResponse.status === 403) {
       throw new Error("Invalid or expired Poe API token");
@@ -68,7 +69,7 @@ defineProvider({
     const entries = [];
     try {
       let cursor = null;
-      const cutoff = Date.now() - 30 * 86400000;
+      const cutoff = now.getTime() - 30 * 86400000;
       for (let page = 0; page < 5; page += 1) {
         const query = cursor ? `?limit=100&starting_after=${encodeURIComponent(cursor)}` : "?limit=100";
         const response = await ctx.http.getJSON(`https://api.poe.com/usage/points_history${query}`);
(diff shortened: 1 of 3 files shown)


steipete:
Superseded by the independently landed main commit a0d5d2f66794e0ec4210d020d5ba12e52a7dbeb5, which now uses the frozen host refresh clock for both the history cutoff and Today totals.

I verified that the integrated candidate carries main's Poe implementation and tests unchanged, then ran `swift test --no-parallel --filter 'Poe|date now|status item placement preflight|ProviderArchitectureGatekeeperTests'`: all 65 tests across eight suites pass, including the original Poe golden and the new retention/Today test under QuickJS and JavaScriptCore. All requests were synthetic; no live account or credential access was used.

Closing this duplicate without merging it. The placement fix remains separate in #3361 and still requires its complete full-suite and refreshed CI gates. The extra local cutoff cases from this branch remain available in its commit history; they are not claimed as landed.

Issue #3362, opened 2026-09-01
Antigravity CLI probe falls back to offline under systemd on Linux


srijits:
## Summary

On Linux, CodexBarCLI v0.56.2 retrieves Antigravity quota data correctly from an interactive shell, but the same binary and command run by systemd reports a `parse` failure for the Antigravity CLI strategy and falls back to `source=offline`.

The underlying `agy models` command succeeds in the same systemd environment, so Antigravity authentication and network access are available there.

## Environment

- CodexBarCLI 0.56.2, Linux musl x86_64 release
- Linux Mint 22.3 (Ubuntu 24.04 base), x86_64
- Antigravity CLI installed and already authenticated
- Reproduces in both a persistent user service and a transient `systemd-run --user` unit

## Reproduction

1. From an interactive shell, run the Antigravity provider explicitly:

```sh
CodexBarCLI usage --provider antigravity --source cli --json
```

This returns the live Antigravity quota windows.

2. Run the same binary and arguments in a user systemd unit (persistent or transient), with the same `HOME`, `PATH`, locale, working directory, and an explicit `ANTIGRAVITY_CLI_PATH`.

3. Observe that CodexBar returns offline data instead of live quotas. A redacted `CodexBarCLI diagnose` shows:

- Antigravity CLI fetch attempted
- `errorCategory: parse`
- fallback to `source=offline`

4. In that same systemd environment, run:

```sh
agy models
```

It succeeds and lists the models/quota information.

## Troubleshooting already tried

The result is unchanged after:

- setting `ANTIGRAVITY_CLI_PATH` explicitly
- forcing `--source cli`
- matching interactive `PATH`, `TERM`, locale, working directory, and SSH-related environment
- using both persistent and transient user units
- removing systemd hardening directives
- granting the service write access to the Antigravity CLI state directory

## Expected behavior

If `agy models` succeeds and produces valid output in the service environment, CodexBar's Antigravity CLI strategy should parse it and return the same live quota windows that it returns interactively. If parsing fails, diagnostics should ideally include enough sanitized context to identify which expected shape differed.

## Impact

This prevents using CodexBar's Antigravity provider as the backend for a continuously running Linux service. A separate native Antigravity adapter is currently required as a workaround.

I can test a diagnostic build or provide additional redacted output if useful.

srijits:
[...]

PR #3364, opened 2026-09-01
fix(antigravity): recover Linux port discovery after lsof failures

steipete:
## Summary

Recover Antigravity Linux port discovery when an installed `lsof` exits with namespace warnings or otherwise cannot identify listeners. The existing fallback previously ran only when `lsof` was absent, so the reported systemd refresh failed before any quota endpoint was called.

The Linux resolver now tries the existing process-scoped `/proc` lookup after launch/nonzero/empty-result failures. Successful `lsof` results retain precedence. Cancellation, timeout, and output-limit errors still propagate. If the process has not opened its listener yet, CLI readiness polling retries within its existing deadline and retains the last warning if startup never succeeds; an authentication or endpoint-readiness error takes precedence. Ordinary empty discovery keeps its existing no-listeners classification. Socket inodes must belong to the target PID and match that same PID's TCP tables; discovery does not broaden to unrelated processes or namespaces. macOS kernel enumeration, quota parsing, authentication, and `agy` process ownership are unchanged.

Refs #3362. Thanks @srijits for the precise sanitized trace identifying the failure before endpoint discovery.

## Verification

- Baseline: six expected failures across the warning/empty/missing-executable cases and pre-cancelled fallback; existing success/error/timeout controls passed.
- Cold-start regression was independently red: the production readiness loop stopped after the first empty `/proc` lookup instead of waiting for the second poll's listener. The repair adds a typed pending-discovery state plus deadline-diagnostic, ordinary-empty classification, and endpoint-error precedence controls.
- Candidate `e012297`: focused port/proc/CLI/deadline/warm-reuse coverage passed 181 tests across seven suites; the test-layout follow-up passed all 41 CLI fetch tests. `make check` passed with zero violations across 2,083 Swift files and all 95 process-cleanup tests (one expected Linux-only skip).
- Codex autoreview is scoped-clean at its configured P0 threshold for the final branch; independent feasibility review and maintainer code inspection agree with the bounded repair. The accepted earlier cold-start review finding is fixed.
- [CI run 33548055692](https://github.com/steipete/CodexBar/actions/runs/33548055692) tests the combined candidate and current-main tree. Both macOS shards passed all 989 selections across 248 groups with no failures, timeouts, or retries; both plugin engines passed their 79-test goldens. Linux ARM64 passed all 469 tests, including the actual process-owned loopback listener and delayed-listener regressions. Musl build and lint passed.
- Linux x64's first parallel run crashed inside `libdispatch` without an assertion failure. Its unchanged rerun passed all 469 tests, including real `/proc` listener recovery and cold-start polling; the initial crash has not been attributed to a source defect. A separate AWS diagnostic expired during setup and ran no tests; its lease was released.
- Final fixture hardening `05e39f8` uses the existing fake-executable trampoline to avoid a known concurrent-fork writable-executable race. Assertions and production code are unchanged; this is not claimed as the cause or cure of the `libdispatch` crash. The scoped fixture review is clean, its pinned formatter check passes, and the final focused run passes 50 tests. Final `make check` passes all 95 process-cleanup tests and reports zero Swift violations after serialized validation.
- Linux x64 crashed again in the same `libdispatch` event-loop region in [CI run 33560449029](https://github.com/steipete/CodexBar/actions/runs/33560449029); ARM64 and musl passed. Exact runtime disassembly maps both crashes to adjacent instructions traversing a pipe HUP registration. A standalone correct-API stress case reproduces the crash in the official Swift 6.3.3 Ubuntu 24.04 x86_64 image without any CodexBar code.
- Follow-up `6c93ae6` removes CodexBar's Linux pipe capture from that runtime path: one owned nonblocking polling thread drains and closes each read descriptor, while callbacks stay coalesced on their separate queue. New controls cover buffered data at HUP, silent open-pipe shutdown, and stop-before-start cleanup. Darwin is unchanged. The scoped review is clean; 54 subprocess/readiness tests and final `make check` pass, including all 95 process-cleanup tests and zero violations across 2,083 Swift files.
- Independent clean x64 proof on the exact head passes the affected 20-test Linux slice nine consecutive times and the complete 472-test Linux binary. It used the same pinned official Swift 6.3.3 Ubuntu 24.04 image as the standalone red reproduction. The remote lease was released after success.
- [Exact-head CI run 33570192400](https://github.com/steipete/CodexBar/actions/runs/33570192400) is green on the current-main merge tree: both macOS shards passed all 989 selections across 248 groups without a failure, timeout, or retry; both plugin-engine passes ran 79 goldens. Linux x64 and ARM64 each passed all 472 tests, including the new pipe-lifecycle controls and real `/proc` listener recovery. Musl, lint, security, and the aggregate gate passed.
- Local full-suite attempts are not reported as passing. On the final head, groups 1–13 passed before an existing Claude account-invalidation group timed out at the unchanged 600-second bound; seven isolated controls passed and the Claude suite itself then timed out without an assertion. Earlier attempts encountered existing Claude fake-CLI/PTY failures, a recovered Antigravity timeout, and heavy filesystem contention. All owned processes exited. The task-local launcher disables debug information only; repository settings, assertions, and deadlines are unchanged. Hosted CI uses the normal toolchain and clean runners for the complete gate.

Tests use synthetic scripts, temporary `/proc` fixtures, and local sockets only. No real `agy` account, systemd provider installation, credentials, or quota endpoint was probed. The Linux listener regression establishes port-discovery behavior, not an authenticated provider refresh.

## Release-note context

Antigravity on Linux: retain CLI quota discovery when `lsof` encounters inaccessible mount-namespace warnings, using the existing process-owned `/proc` fallback. Credit @srijits for the report. No changelog edit before release.

### DIFF
--- a/Sources/CodexBarCore/Host/Process/ProcessPipeCapture.swift
+++ b/Sources/CodexBarCore/Host/Process/ProcessPipeCapture.swift
@@ -19,11 +19,9 @@ package final class ProcessPipeCapture: @unchecked Sendable {
     private var isStopping = false
     private var continuation: CheckedContinuation<Void, Never>?
     #if os(Linux)
-    private let readerQueue = DispatchQueue(label: "com.steipete.CodexBar.process-pipe-capture.reader")
     private let callbackQueue = DispatchQueue(label: "com.steipete.CodexBar.process-pipe-capture.callback")
-    private var readSource: DispatchSourceRead?
-    private var sourceStarted = false
-    private var sourceCancelled = false
+    private var readerStarted = false
+    private var readerExited = false
     private var callbackScheduled = false
     private var callbackRequested = false
     #endif
@@ -48,6 +46,14 @@ package final class ProcessPipeCapture: @unchecked Sendable {
 
     #if os(Linux)
     package func start(linuxDescriptorSetup: @Sendable (Int32) -> Bool) {
+        self.condition.lock()
+        guard !self.readerStarted, !self.isFinished, !self.isStopping else {
+            self.condition.unlock()
+            return
+        }
+        self.readerStarted = true
+        self.condition.unlock()
+
         let fileDescriptor = self.handle.fileDescriptor
         guard linuxDescriptorSetup(fileDescriptor) else {
             self.finishFailedLinuxStart()
@@ -55,20 +61,14 @@ package final class ProcessPipeCapture: @unchecked Sendable {
         }
 
         // FileHandle.readabilityHandler duplicates the descriptor on Linux and traps if dup(2) returns EMFILE.
-        // A dispatch read source monitors the pipe's existing descriptor, so descriptor exhaustion cannot trigger
-        // Foundation's precondition failure.
-        let source = DispatchSource.makeReadSource(fileDescriptor: fileDescriptor, queue: self.readerQueue)
-        source.setEventHandler { [weak self] in
-            self?.handleLinuxReadableData(fileDescriptor: fileDescriptor)
+        // DispatchSourceRead avoids that duplication, but Swift 6.3's Linux epoll backend can unregister and free
+        // a pipe's shared muxnote on its target queue while the manager queue is still traversing the HUP event.
+        // Poll on an owned thread instead: the reader keeps sole descriptor-close ownership, while the bounded poll
+        // interval keeps stop() responsive without allocating a wakeup descriptor that would break the EMFILE path.
+        Thread.detachNewThread { [self] in
+            Thread.current.name = "CodexBar pipe capture"
+            self.runLinuxReader(fileDescriptor: fileDescriptor)
         }
-        source.setCancelHandler { [weak self] in
-            self?.finishLinuxSourceCancellation()
-        }
-        self.condition.lock()
-        self.readSource = source
-        self.sourceStarted = true
-        self.condition.unlock()
-        source.resume()
     }
     #endif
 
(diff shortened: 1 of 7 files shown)


steipete:
Landed as `9e7af518796d16939f4939ef6e2ff47d1252406a` (tree `5fd27477f79f303fcd94c54a25042577e3427af5`). The visible `main` checkout was fast-forwarded and matches the CI-tested merge tree.

The reported systemd failure happened before any quota endpoint was called: executable `lsof` returned mount-namespace warnings and exit 1, while the old `/proc` fallback only ran when `lsof` was absent. Linux discovery now falls back only through the target PID's socket inodes and TCP tables, retains successful `lsof` precedence, retries a not-yet-bound listener within the existing readiness deadline, and preserves authentication/endpoint error precedence.

Validation uncovered a second real Linux failure: the new concurrent subprocess cases made Swift 6.3.3's `DispatchSourceRead` HUP race reproducible. Both hosted crashes mapped to adjacent instructions in the same epoll muxnote traversal. A standalone correct-API HUP/cancel stress case also crashed with SIGILL in the pinned official Swift 6.3.3 Ubuntu 24.04 x86_64 image without CodexBar code. Linux pipe capture now uses an owned nonblocking polling thread; the reader alone closes its descriptor, while callbacks remain coalesced separately.

Proof:

- `swift test --no-parallel --filter 'ProcessPipeCapture|SubprocessRunner|AntigravityCLIHTTPSFetchStrategyTests'`: 54 tests across two suites passed.
- `make check`: all 95 process-cleanup tests passed (one expected Linux-only skip), with zero violations across 2,083 Swift files.
- Independent clean x64 proof on the exact head: the affected 20-test Linux slice passed nine consecutive runs; the complete `swift test --skip-build --parallel` binary passed 472 tests across 65 suites.
- [Exact-head CI run 33570192400](https://github.com/steipete/CodexBar/actions/runs/33570192400): Linux x64 and ARM64 each passed all 472 tests; musl, lint, security, and the aggregate gate passed. Both macOS shards passed all 989 selections across 248 groups without a failure, timeout, or retry; both plugin-engine passes ran 79 goldens.
- `make test` was run locally but is not reported as passing: groups 1–13 passed, then an existing Claude account-invalidation group and its isolated retry exceeded the unchanged 600-second bound without an assertion. Clean hosted macOS shards provide the complete full-suite gate.

Tests used synthetic scripts, temporary `/proc` fixtures, and local sockets only. No real Antigravity account, credentials, provider endpoint, or Keychain data was accessed. No changelog was edited; release-note context and @srijits credit remain in the PR body for the next release.

PR #3365, opened 2026-09-01
fix(test): isolate app-group migration from user state

steipete:
## Summary

Prevent app-group and settings tests from discovering or copying real user state. This is a test-safety fix found during the cached-menu-render timing investigation in #3360, not a rendering change.

- Skip app-group discovery, shared-defaults bridging, and migration during ordinary `SettingsStore` test initialization. Production migration scheduling and logic remain unchanged.
- Share one complete test-runner detector between settings isolation and Keychain safety, including Swift Testing and XCTest process/environment variants.
- Keep automatic test suppression separate from explicit Keychain policy, preserve deliberate live-test opt-in, and inject a paired read/write policy so fixtures never race through process-global Keychain state.
- Use dictionary-backed defaults, explicit synthetic home and snapshot URLs, and a recording `FileManager` that rejects operations outside the migration fixture.
- Document the durable isolation contract in `AGENTS.md`.

## Regression coverage

Covers every supported runner signal, automatic suppression versus deliberate live opt-in, stored Keychain policy in both states, snapshot copying once, preservation of existing target defaults and snapshot bytes, missing legacy data, unavailable targets, settings initialization leaving app-group migration untouched, and fresh versus existing-install refresh defaults.

## Validation

- `make check` passed, including SwiftFormat and strict SwiftLint.
- Focused settings, app-group, Keychain, detector, and architecture-gate suites passed.
- Full local `make test` passed all 994 selections in 83 groups on the first attempt, with zero retries or timeouts. The run used a task-owned SwiftPM scratch path and a 300-second local group ceiling to accommodate this machine's slower AppKit and cost-history groups.
- Exact-head CI passed all jobs on `7e86e9aa48a6e1cdac21c2d900dfaa1c551ec665`: lint, both macOS shards, Linux x64/arm64, Linux musl, path detection, and aggregate verification ([run 33598496755](https://github.com/steipete/CodexBar/actions/runs/33598496755)).
- Final branch autoreview found no actionable P0-P2 issues.

No live provider, browser-cookie, or real Keychain validation was run; all proof used contained fixtures and explicit test isolation.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -24,6 +24,7 @@
 - After any code change, run `make check` and fix all reported format/lint issues before handoff.
 - Prefer CLI/focused tests over app-bundle live tests when behavior can be verified without relaunching CodexBar.
 - Never run tests/checks or ad-hoc validation that can display macOS Keychain prompts. Live provider probes, browser-cookie imports, `codexbar usage` against real accounts, and real SecItem reads must be explicitly requested; otherwise use parser tests, stubs, test stores, or `KeychainNoUIQuery`.
+- App-group migration tests must inject dictionary-backed defaults, both snapshot URLs, a synthetic home, and a contained recording FileManager. UUID defaults suites and Keychain isolation flags do not isolate defaults search domains or filesystem access. Ordinary SettingsStore tests must not discover shared defaults or run app-group migration.
 - macOS CI is brittle around headless AppKit status/menu tests. Prefer covering menu behavior through stable state/model seams (`MenuDescriptor`, `ProvidersPane`, `CodexAccountsSectionState`, etc.) instead of constructing live `NSStatusBar`/`NSMenu` flows unless the AppKit wiring itself is the thing under test.
 
 ## Commit & PR Guidelines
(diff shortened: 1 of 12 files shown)


steipete:
Landed as `98b86f39db63d230e2a7f78cdb013a845af33213`.

The final patch prevents ordinary settings tests from discovering real app-group defaults or running migration, shares a complete test-runner detector with Keychain safety, and uses a paired injectable Keychain policy so contained fixtures neither read nor mutate process-global gate state. Production migration and Keychain behavior remain unchanged.

Proof on reviewed head `7e86e9aa48a6e1cdac21c2d900dfaa1c551ec665`:

- `make check` passed.
- Full local `make test` passed 994 selections across 83 groups with zero retries or timeouts.
- Exact-head CI passed every job: [run 33598496755](https://github.com/steipete/CodexBar/actions/runs/33598496755).
- Final branch autoreview found no actionable P0-P2 issues.
- The squash merge tree exactly matches the reviewed head.

No live provider, browser-cookie, or real Keychain validation was used.

PR #3366, opened 2026-09-01
test: fix MiMo cache publication hooks across Python versions

steipete:
## Summary

Fix the interpreter-dependent MiMo cache tests without changing production cache behavior. Python 3.9's `Path.replace` calls a prebound filesystem accessor, so patching `os.replace` neither injected the publication failure nor reliably synchronized concurrent writers.

The tests now patch `Path.replace`, the boundary actually invoked by the writer. Concurrent writers rendezvous with distinct, complete temporary payloads before delegating to the real replacement method. Explicit assertions reject a bypassed hook or barrier. Both partial-write and publication-failure cases verify that the previous cache and another writer's temporary file remain byte-for-byte intact, while the failed writer's temporary file is removed.

Only `Scripts/test_mimo_usage.py` changes; `Scripts/mimo-usage.py` is unchanged.

## Validation

- Reproduced the original `stage='replace': AssertionError: OSError not raised` under Xcode Python 3.9.6.
- Both focused tests pass with Python 3.9.6, 3.12.14, 3.13.15, and 3.14.7 using `PYTHONDONTWRITEBYTECODE=1 <interpreter> Scripts/test_mimo_usage.py -v`.
- Twenty concurrency repetitions pass per interpreter (80 total).
- Twelve synthetic negative controls are rejected as expected: bypassed publication hook, skipped barrier, and no-op publication on each interpreter.
- `make check` passes with Homebrew Python 3.14.7. On Xcode Python 3.9.6 it now passes the MiMo tests, then encounters the separate existing Swift-runner requirement for `os.waitid` with `WNOWAIT`.
- Local full `make test` was deliberately not run: this checkout's test initialization can run app-group migration against real saved data. All task-specific runtime proof uses temporary directories and synthetic data; no live provider probes, real-cache reads, or Keychain tests were used.

### DIFF
--- a/Scripts/test_mimo_usage.py
+++ b/Scripts/test_mimo_usage.py
@@ -2,7 +2,6 @@
 
 import importlib.util
 import json
-import os
 from datetime import datetime, timezone
 from pathlib import Path
 import tempfile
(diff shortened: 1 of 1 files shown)


steipete:
Landed as ec8534de0ce067daaab316d733837a7f0d36ebc6. The landed tree matches the current-main merge tree computed before merging.

Proof:

- Current `main` reproduced the Python 3.9.6 failure exactly: the replace-stage case completed without raising because patching `os.replace` did not intercept the writer's `Path.replace` call.
- The PR passed both focused tests on Python 3.9.6, 3.12.14, 3.13.15, and 3.14.7.
- An additional 50 repetitions each on Python 3.9.6 and 3.14.7 passed (100 repeated runs total).
- Exact-head CI run 33549256368 was fully green across lint, Linux x64/ARM64, both macOS shards, security, and the aggregate gate.

Only `Scripts/test_mimo_usage.py` changed; production cache behavior is unchanged.

PR #3371, opened 2026-09-01
feat(provider): add Muse local token usage

krazybean:
## Summary

Adds Muse as a first-class CodexBar provider with local token and request usage history derived from Muse session telemetry.

Muse does not currently expose subscription quota, reset-window, or billing/cost data suitable for CodexBar's normal quota model, so this integration intentionally reports token history only rather than fabricating quota or spend information.

## How it works

Muse stores local session telemetry under its session directory as JSONL records. The provider scans `runtime.session` / `model_completed` events and aggregates:

- Input tokens
- Output tokens
- Cache-read tokens
- Reasoning tokens
- Request counts
- Model attribution
- Daily usage history

`model_completed` is treated as the authoritative usage event. `goal_usage_attribution` records are deliberately ignored to avoid double-counting the same completion.

Both the direct JSONL format and Muse's wrapped `retained_frame` / `record_json` format are supported.

The scanner only extracts usage telemetry required for aggregation; prompt, source, and tool-output contents are not surfaced through the provider.

## Incremental session cache

Muse session histories can become quite large, so rescanning every session file on each CodexBar refresh is impractical.

This adds a persistent incremental cache for Muse session usage. It tracks per-file state and daily/model contributions so subsequent refreshes can:

- Reuse unchanged files without reopening them
- Parse only newly appended JSONL content
- Detect truncation or replacement and reparse safely
- Remove contributions from deleted or expired files
- Rebuild safely after timezone changes or corrupt/incompatible cache state
- Avoid persisting partially updated state if a scan is cancelled

File replacement detection includes a lightweight prefix fingerprint so a larger file replacing an existing session at the same path is not mistaken for an append.

JSONL boundary handling also preserves incomplete trailing writes for the next scan while consuming complete final records correctly.

On a local Muse history containing ~9,500 session files / ~5.8 GB of session data, an initial scan is necessarily expensive, but unchanged warm refreshes drop to roughly 3–4 seconds with zero session files reopened.

## Provider behavior

Muse intentionally exposes:

- Token usage history
- Request history
- Model attribution

Muse intentionally does **not** expose:

- Fabricated subscription quotas
- Fabricated reset windows
- Estimated dollar spend without authoritative pricing data

This follows CodexBar's existing token-history path for providers where local token telemetry is available without corresponding quota/cost information.

## Tests

Adds coverage for:

- `model_completed` usage extraction
- `goal_usage_attribution` double-count prevention
- Cache/reasoning token accounting
- Epoch-microsecond timestamp bucketing
- Malformed/unrelated records
- Sensitive payload fields not affecting telemetry parsing
- Subagent session discovery
- Lookback/date-directory pruning
- Wrapped `record_json` records
- Incremental append parsing
- Unchanged-file cache reuse
- New/deleted/truncated session files
- Larger same-path file replacement
- Incomplete trailing JSONL writes
- Cache expiration/pruning
- Timezone invalidation
- Corrupt cache recovery
- Atomic cache writes
- Cancellation safety

`MuseLocalSessionScannerTests`: 13/13 passing.

The incremental cache regression tests also pass individually; one cancellation test can encounter a `swiftpm-testing-helper` signal 5 issue in the local Swift Testing environment, despite the underlying synchronous cancellation path completing correctly.

## Configuration

Session root resolution supports:

1. `MUSE_HOME`
2. `CODEXBAR_MUSE_HOME`
3. Muse's default local data location

Muse CLI version detection is also included for provider metadata.

## Scope

This PR is limited to the Muse provider and its local usage/cache implementation.

Changes needed to support packaging CodexBar forks with non-upstream Developer ID identities are intentionally kept in a separate PR.


steipete:
Thank you for the Muse work. Closing this overlapping implementation in favor of #3340’s better-proven local-token approach. This patch retains cross-provider fallback, deleted-cache invalidation, history-coverage and append-I/O defects. One scanner/cache owner is preferable; useful independent improvements can be incorporated with your credit.

PR #3372, opened 2026-09-01
fix(packaging): support non-upstream Developer ID signing

krazybean:
## Summary

Allows CodexBar forks to package/sign the app with a non-upstream Developer ID without embedding upstream provisioning or entitlement metadata.

The existing packaging flow assumes CodexBar's upstream signing team and provisioning profile. That works for upstream releases, but a fork signed with a different Developer ID can end up with a locally signed app that still contains upstream team-bound provisioning/entitlement values, which macOS can reject at launch.

## What changes

- Derives the signing Team ID from the configured Developer ID identity instead of hard-coding a fork-specific value.
- Embeds `Scripts/profiles/CodexBar-DeveloperID.provisionprofile` only when packaging with CodexBar's upstream team (`Y5PE65HELJ`).
- Generates local app/widget entitlements using the active signing team.
- Avoids claiming upstream iCloud, keychain, or provisioning resources for non-upstream signers.
- Preserves the existing strict upstream release-signing behavior.

## Timestamp fallback

For non-upstream signing only, packaging now falls back to signing without a timestamp if Apple's timestamp service is unavailable.

Upstream (`Y5PE65HELJ`) release signing remains strict and does not receive this fallback, so the upstream release path is not weakened.

## Launch smoke test

The packaged-app launch verification now detects environments where `sandbox-exec` itself is unavailable/forbidden and skips that sandboxed smoke step in that specific case.

The existing explicit:

`CODEXBAR_SKIP_LAUNCH_SMOKE=1`

override remains unchanged.

Normal environments where `sandbox-exec` works continue to run the smoke test.

## Why

This surfaced while testing CodexBar from a fork with a different Developer ID.

The resulting app was correctly signed by the fork's identity, but still contained an upstream provisioning profile/team metadata. Signature verification could succeed while launchd rejected the app at spawn time.

With this change, upstream and fork packaging paths remain internally consistent:

- upstream signer → upstream provisioning/profile behavior
- non-upstream signer → derived local team entitlements, no upstream profile

## Scope

This PR only changes fork/release packaging behavior.

The Muse provider work that exposed the issue is submitted separately in #3371.

### DIFF
--- a/Scripts/package_app.sh
+++ b/Scripts/package_app.sh
@@ -230,6 +230,25 @@ if [[ "$SIGNING_MODE" == "adhoc" ]]; then
 fi
 WIDGET_BUNDLE_ID="${BUNDLE_ID}.widget"
 APP_TEAM_ID="${APP_TEAM_ID:-Y5PE65HELJ}"
+# If a signing identity is explicitly configured, prefer its team over the
+# upstream default. This keeps a fork's Developer ID from
+# claiming Y5PE65HELJ's provisioning profile / CloudKit container.
+if [[ -n "${APP_IDENTITY:-}" && "${APP_TEAM_ID}" == "Y5PE65HELJ" ]]; then
+  if [[ "${APP_IDENTITY}" =~ \(([A-Z0-9]{10})\) ]]; then
+    APP_TEAM_ID="${BASH_REMATCH[1]}"
+  elif [[ -n "${APP_IDENTITY}" ]]; then
+    _team_subject="$(security find-certificate -c "${APP_IDENTITY}" -p 2>/dev/null | openssl x509 -noout -subject -nameopt RFC2253 2>/dev/null || true)"
+    if [[ "${_team_subject}" =~ (^|,)OU=([A-Z0-9]{10})(,|$) ]]; then
+      APP_TEAM_ID="${BASH_REMATCH[2]}"
+    else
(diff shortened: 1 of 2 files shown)


steipete:
The resolved-team provisioning mismatch is now fixed on main in #3461 (`02c073a22ab`), with your co-author and changelog credit preserved. The actual entitlement-generation/profile-copy code passes all 16 upstream/alternate-team, debug/release, identity/ad-hoc, profile-present/absent cases; full tests and exact-head CI passed.

The narrower repair also derives app/widget groups from the resolved bundle ID. This PR remains open for the separate direct-identity discovery path. Timestamp fallback, widget omission and sandbox-smoke bypass were not needed for the verified profile-selection defect.

steipete:
Adopted the safe identity-discovery remainder in https://github.com/steipete/CodexBar/pull/3803. Direct packaging now resolves the configured Developer ID identity, derives its Team ID, rejects missing/ambiguous/mismatched selections, and passes the same identity through release packaging. This complements the upstream-only entitlement/profile selection already landed in #3461.

The 56 synthetic signing/selection cases pass, including regression coverage against the original scripts, and [CI passed](https://github.com/steipete/CodexBar/actions/runs/35566032201). Widget builds, timestamping, hardened runtime, signature verification, and sandbox smoke validation remain enforced. The automatic timestamp-less fallback, widget omission, and sandbox-denial bypass are intentionally rejected. The old teamless development setup is retired with migration documentation. Thanks @krazybean for the original work.

PR #3373, opened 2026-09-02
Improve reliability of Codex subscription metadata capture

emanuelst:
## Summary

- Improve the reliability of Codex plan renewal and expiration dates.
- Return the authenticated dashboard usage snapshot without waiting for optional subscription metadata.
- Enrich the attached Codex snapshot asynchronously after the dashboard result is available.
- Preserve subscription metadata for direct web dashboard callers, including CLI `--source web`.
- Re-check account authority, refresh guards, and the active dashboard before applying late metadata.
- Preserve explicit empty-versus-unavailable parsing, including clearing stale dates only after an explicit empty response.
- Keep ambiguous same-email dashboards display-only; they never persist dashboard-derived state.

## Why

Usage data can refresh while subscription metadata is temporarily unavailable or slow. Separating the optional metadata enrichment keeps refresh completion responsive while retaining reliable plan dates when the authenticated billing route succeeds. Direct web callers receive the same metadata after usage data is secured, bounded by the remaining dashboard deadline.

## Design

The required dashboard path owns usage, credits, and account attachment. Subscription metadata is an optional follow-up task with its own cancellation token. It first tries the authenticated subscription API and then the existing WebView billing capture when needed. A result is applied only when the original account guard still matches and the dashboard remains authorized; otherwise it is discarded before persistence. A late result changes only the two subscription fields and does not overwrite newer usage data.

The direct Codex web strategy performs the same optional metadata fetch after the dashboard has been secured, using only the remaining overall deadline before returning its result. This keeps CLI JSON and other direct strategy consumers consistent without moving the optional wait back into the core dashboard fetch or delaying the required scrape.

## Maintainer improvements

Updated against current main, preserving the 0.56.5 extra-credit accounting fixes. Late billing results now require the original dashboard, account/workspace guard, and enabled web access. Concurrent capture publishes one response's status and fields together; numeric renewal flags are rejected as malformed. Optional work preserves cancellation and its deadline, and its dedicated WebView lease owns teardown.

## Validation

- `make check`: passed, including formatting and strict lint.
- Focused regression and architecture run: 48 Swift Testing tests and seven XCTest cases passed.
- Signed native proof: all seven lifecycle/proof tests passed; before/after captures inspected.
- Independent Codex review: no actionable P0 findings in the complete candidate diff.
- Full `make test`: all 85 groups passed on the first attempt; zero retries and timeouts.
- [GitHub CI](https://github.com/steipete/CodexBar/actions/runs/33917690957): all nine checks passed, including both macOS test shards and all Linux builds.

The regression fixtures reproduced newer-dashboard rollback, publication after disabled access, and same-email workspace crossover before the guard fixes. Tests also cover unavailable versus explicitly empty metadata, preserving newer usage fields, replacement task ownership, required usage completing before metadata, API schema validation, and the capture script installed by the production WebView factory. Settings, credentials, caches, and widget saves use isolated fixtures.

## Native before/after proof

Signed Developer ID test bundle, production attachment and menu-card rendering, synthetic data only. The installed app and real provider accounts were not used for this proof.

| Usage available while billing is pending | Billing date attached |
| --- | --- |
| Before | After |

## Original contributor runtime proof

The contributor supplied redacted production-path logs showing subscription API failure followed by successful authenticated billing capture, authorized persistence, and ambiguous-account denial. The maintainer's repeatable proof above is synthetic; it does not claim a fresh authenticated ChatGPT request.

Thanks @emanuelst for the original feature and this follow-up. Implementation and review were AI-assisted.


emanuelst:
Author decision: accept the bounded post-data subscription fallback. Once returnable dashboard data is secure, it may delay the final snapshot for at most 8 seconds, or the earlier overall deadline, so the authorized snapshot can include reliable subscription metadata. The cap is now explicit and regression-tested.

PR #3376, opened 2026-09-02
Add Nous Portal through a bundled provider plugin

asispan:
## What changes

Add disabled-by-default Nous Portal support for monthly subscription usage and purchased credit balances. HTTP, billing parsing, quota mapping, and credit rows live in the bundled TypeScript plugin; QuickJS runs it on macOS and Linux, with JavaScriptCore covered by the same fixtures. Swift only discovers the existing Hermes credential, reports credential errors, and registers the provider. There is no second native billing fetcher or parser.

Hermes remains the credential owner: CodexBar reads its access token and never refreshes or writes it. Custom Hermes homes stay isolated; stored destinations are restricted to Nous hosts; explicit portal overrides must be HTTPS origins. Credential-pool selection follows Hermes's agent-key expiry, access expiry, and priority order.

The original proposal defaulted missing amounts to zero. A response containing only a monthly grant could therefore show 100% used and a zero top-up balance. The plugin leaves missing amounts unavailable, rejects malformed/nonfinite money, and only shows a quota when both grant and remaining balance are known. Purchased credits remain separate, in the generic Credits detail section.

Thanks @asispan for the original provider, credential reader, fixtures, and authenticated endpoint proof. Contributor credit is preserved in the commit.

## Validation

- Focused Nous/plugin and provider-architecture tests passed; plugin fixtures exercise both QuickJS and JavaScriptCore.
- Final credential-pool regression and Nous suites passed: 26 tests.
- `make check` passed, including generated manifests/JavaScript, TypeScript, documentation, formatting, and strict SwiftLint.
- Independent code review passed with no actionable findings after fixing credential-pool selection.
- `make test` passed on the final head: 1,181 selections / 107 groups; zero failures, retries, or timeouts. The first run caught a missing monthly-only menu-metric expectation; that test registration was corrected and the full suite rerun.
- Fresh synthetic fixtures render through the plugin and production provider-card view. No real account, browser cookies, Keychain reads, or app restart were used for these captures.

The images below compare the original proposal's golden output for a partial response with the plugin output for that same response, followed by a complete account fixture. They are synthetic card renders, not a claim of newly performed authenticated provider verification. The contributor's earlier live proof established the account endpoint and token contract; this rewrite has been verified with captured-shape fixtures and contained runtime tests.

Original Swift proposal: missing balances became exact zero

Plugin: partial response shows only the reported monthly grant

Plugin: complete synthetic account with separate monthly and purchased credits


asispan:
Fixed the P1 changelog finding in 6b655b4.

The earlier `main` merge resolved `CHANGELOG.md` in favour of the branch, which silently dropped the current Unreleased `Fixed` bullets **and** the whole `0.56.4` release section. I took `main`'s changelog verbatim and re-added the Nous entry under its own `Added` heading alongside the existing sections.

The diff against `main` is now purely additive — zero deleted lines:

```
$ git diff origin/main --stat -- CHANGELOG.md
 CHANGELOG.md | 3 +++
 1 file changed, 3 insertions(+)

$ git diff origin/main -- CHANGELOG.md | grep -c "^-[^-]"
0
```

No Swift files changed in this commit. `swift test --filter 'Nous|ProviderArchitectureGatekeeper'` still passes (69 tests).

That leaves only the maintainer product decision: whether CodexBar wants to adopt a disabled-by-default Nous Portal provider. Happy to keep it out-of-tree as a plugin instead if that fits the project better.

steipete:
The plugin rewrite is ready at `64d41c6c62058`: independent code review is clean, all 1,181 local selections / 107 groups passed without retries, and [all eight CI jobs passed](https://github.com/steipete/CodexBar/actions/runs/35318132924).

The remaining gate is a real-account check through this rewritten production path. Please verify the current head with an existing Hermes login: the CLI/provider should return the same monthly quota and purchased-credit meaning as the Portal, and expired credentials should direct renewal back to Hermes. No credential refresh or credential-file write is performed by CodexBar.

A concise result tied to this SHA is sufficient; keep tokens, account identity, full payloads and private account amounts out of the public report. The existing synthetic card images prove rendering and the unknown-balance fix, but are not being substituted for this live check.

asispan:
Live verification of `64d41c6c62058b063e8baa909a1747114540fcc8` with an existing Hermes login (real Nous account, macOS, debug `CodexBarCLI` built from this exact head).

**Real-account run** — `codexbar usage --provider nous --verbose`, one request, `portal=portal.nousresearch.com credential=hermes`. I compared every rendered field against the same `/api/oauth/account` response fetched moments earlier by Hermes's own client (`hermes_cli.nous_account`), which is what the Portal `/products` page reflects:

| CLI row | Portal field | Result |
| --- | --- | --- |
| Subscription credits *remaining* of *grant* | `subscription.credits_remaining` / `subscription.monthly_credits` | match to the cent |
| Monthly credits % left | `round(remaining / grant)` | match |
| Resets in / Renews | `subscription.current_period_end` | match (local date) |
| Top-up credits | `purchased_credits_remaining` | match to the cent, rendered separately from the monthly meter |
| Total usable | `paid_service_access.total_usable_credits` | match; also equals remaining + top-up |
| Plan | `subscription.plan` | match (name only) |
| Rollover row | `subscription.rollover_credits` | correctly hidden when the reported value is zero |

No email, organisation, or other identity appears in the CLI output.

**Expired credentials → back to Hermes** (isolated temp `HERMES_HOME` copies with synthetic expired JWTs; the real auth file was never pointed at):

- Expired token in `auth.json` → `Nous Portal access token in <path>/auth.json has expired. Run \`hermes\` so Hermes Agent refreshes it.` exit 1
- Expired `NOUS_PORTAL_ACCESS_TOKEN` → `NOUS_PORTAL_ACCESS_TOKEN has expired. Export a fresh token or unset it to use the Hermes Agent login.` exit 1
- Empty custom `HERMES_HOME` while `~/.hermes` holds a valid login → `nous.api (api) unavailable`; the default profile's token was not used.

**No request before the expiry check** — differential test with `NOUS_PORTAL_BASE_URL` set to an unresolvable HTTPS host: the valid token yields a DNS error (`A server with the specified hostname could not be found`, i.e. a request was attempted), while both expired cases return the expiry messages above, so no request is built.

**No refresh, no credential-file write** — `~/.hermes/auth.json` and `~/.hermes/shared/nous_auth.json` have identical SHA-256 and mtime before and after all runs. The tampered stored-host case still reports `rejectedStoredHost=stored.example` and sends only to `portal.nousresearch.com`.

`swift test --filter 'Nous|ProviderArchitectureGatekeeper'`: 67 tests, 3 suites, all passed on this head.

Thanks for the rewrite — the unknown-balance fix is a real improvement over my zero-default.

steipete:
Merged as 88aab2ba6a25959fc876fbe6bf8b637779eae788. The bundled plugin now owns monthly quota/balance requests and parsing, while the native adapter reuses Hermes credentials read-only. Contributor credit is preserved.

Validation: `make check`; full `make test` (1,181 selections / 107 groups, no failures/retries/timeouts); both-engine fixtures and synthetic card renders; final independent review clean; [all exact-head CI jobs passed](https://github.com/steipete/CodexBar/actions/runs/35318132924). Current-account verification of the rewritten path was unavailable locally; this landing retains that limitation and the prior contributor service-contract evidence. No release was published.

PR #3330, opened 2026-08-31
perf(test): prefilter provider architecture scan

brzvsk:
Closes #3327

## Summary

- derive the small set of provider IDs that can occur on each source line before running the expensive dotted-reference classifier
- normalize quoted provider literals once instead of testing every provider ID
- preserve the legacy scanner for symbolic Swift identifiers such as emoji or punctuation-bearing IDs through a precomputed exact-match fallback; all 69 current provider IDs stay on the fast path
- cover dotted cases, qualified cases, raw literals, interpolations, URLs, identifier suffixes, quoted text, empty IDs, and unrelated lines

This changes only the architecture gatekeeper test scanner. Production code, test-runner behavior, sharding, retries, and time allowances are unchanged.

## Equivalence proof

I temporarily kept the pre-change scanner beside the candidate and compared their exact `[ProviderReference]` output for every shipped Swift source, using all 69 current provider IDs and the existing provider-owned-folder exclusion. I repeated the comparison after maintainer hardening with a symbolic-ID corpus covering emoji, punctuation, empty IDs, suffix boundaries, log literals, and URLs. Both comparisons produced no differences in matched file/line positions, occurrence multiplicity and order, provider IDs, or newly-recognized occurrence tracking.

The temporary legacy scanner was then removed. Permanent regressions retain the relevant lexical, symbolic-ID, and false-positive boundaries without carrying the old lines × providers implementation.

## Performance

The contributor's paired baseline measured the gatekeeper suite at a 26.605s median and the candidate at 4.379s. After rebasing and hardening, three local no-build runs measured the final suite at 4.566s, 4.244s, and 4.366s (4.366s median); warm command wall times were 5.27s and 5.36s. The symbolic fallback is empty for today's provider set, so it does not change the fast path.

## Verification

- focused `ProviderArchitectureGatekeeperTests`: 41/41 passed
- temporary legacy-vs-candidate shipped-source and symbolic-corpus differential: exact match
- `make check`
- final Codex autoreview: no actionable P0-P2 findings
- `make test` under isolated file, session, defaults, and Keychain settings with non-timeout retries disabled: 994 selections, 83/83 groups passed on the first attempt, 0 retries, 0 timeouts
- hosted CI: https://github.com/steipete/CodexBar/actions/runs/33611499077

Receipts:

- base: `98b86f39db63d230e2a7f78cdb013a845af33213`
- rebased contributor commit: `0cce565d290854f8d4c754a8495282acb989f0a0`
[...]

PR #3378, opened 2026-09-02
fix(test): prevent widget timeline reloads during tests

steipete:
## Summary

Tests that inject a temporary widget snapshot URL currently save to that file and then request a real WidgetKit timeline reload. The in-memory save override already returned before reload, so file injection was the missing side-effect boundary.

This change separates snapshot persistence from reload permission:

- test processes may opt into an in-memory save or a test-owned file, but never reload real WidgetKit timelines
- production still awaits the file save before requesting exactly one reload
- the reload callback is immutable, per-store, and `@MainActor`, avoiding process-global test overrides
- the helper takes an explicit test-mode decision so save/reload ordering can be proven with a temporary file and fake callback
- widget documentation records the isolation contract

The architecture audit catalog also refreshes source-line anchors shifted by the added code. Anchor text, provider fingerprints, reasons, and audit rules are unchanged.

There is no intended user-visible production behavior change.

## Verification

- rebased onto `main` after #3365 and #3330, retaining the broader shared test-process detector and app-group defaults isolation
- focused nonparallel run: 58 tests across `WidgetSnapshotTestIsolationTests`, `WidgetSnapshotBoundedIOTests`, `TestProcessSafetyTests`, `SettingsStoreKeychainPreferenceTests`, and `ProviderArchitectureGatekeeperTests` passed
- `make check`
- final Codex autoreview: no actionable P0-P2 findings
- `make test` with isolated file, session, defaults, and Keychain settings and non-timeout retries disabled: 994 selections, 83/83 groups passed on the first attempt, 0 retries, 0 timeouts
- hosted CI: https://github.com/steipete/CodexBar/actions/runs/33616860899

The first local full-suite attempt exposed only a task-owned native-build resource lookup problem: the executable-adjacent `CodexBarCore` bundle link was absent. After linking `.build/debug` to the actual scratch products, the clean no-retry run above passed in full.

Receipts:

- base: `33fd05ba73edf08ff92a7dd9d1205594d5ee62ab`
- final candidate: `7b1ffa9b6ce7941982a0efb4d8684fdf67d726a1`
- final tree: `d6b51d595bb822b45219217f0ecd7f67221dde54`

### DIFF
--- a/Sources/CodexBar/UsageStore+WidgetSnapshot.swift
+++ b/Sources/CodexBar/UsageStore+WidgetSnapshot.swift
@@ -50,18 +50,36 @@ extension UsageStore {
                 return
             }
 
-            let widgetSnapshotURL = self.widgetSnapshotURL
-            await Task.detached(priority: .utility) {
-                if let widgetSnapshotURL {
-                    WidgetSnapshotStore.save(snapshot, to: widgetSnapshotURL)
-                } else {
-                    WidgetSnapshotStore.save(snapshot)
-                }
-            }.value
-            #if canImport(WidgetKit)
-            WidgetCenter.shared.reloadAllTimelines()
-            #endif
+            await Self.saveWidgetSnapshot(
+                snapshot,
+                to: self.widgetSnapshotURL,
+                isRunningTests: SettingsStore.isRunningTests,
+                reloadTimelines: self.widgetTimelineReloader)
         }
+    }
+
+    static func saveWidgetSnapshot(
+        _ snapshot: WidgetSnapshot,
+        to url: URL?,
+        isRunningTests: Bool,
+        reloadTimelines: @MainActor () -> Void) async
+    {
+        await Task.detached(priority: .utility) {
+            if let url {
+                WidgetSnapshotStore.save(snapshot, to: url)
+            } else {
+                WidgetSnapshotStore.save(snapshot)
+            }
+        }.value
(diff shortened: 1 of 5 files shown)


steipete:
Landed as `4cc3373f3804f19615677cddc1c4eca67b26ee97` after rebasing onto the #3365/#3330 test-safety foundation.

The fix closes the missing side-effect boundary: tests may persist widget snapshots to memory or a test-owned file, but neither opt-in can call real WidgetKit timeline reloads. Production still awaits the snapshot save and then requests exactly one reload through the per-store `@MainActor` callback.

Proof on exact head `7b1ffa9b6ce7941982a0efb4d8684fdf67d726a1`:

- focused nonparallel run: 58 tests across widget isolation/bounded-I/O, shared test-process detection, SettingsStore Keychain isolation, and architecture-gate suites
- `make check`
- full local suite with retries disabled: 994 selections, 83/83 groups first-pass green, 0 retries, 0 timeouts
- final autoreview: no actionable P0-P2 findings
- hosted CI fully green: https://github.com/steipete/CodexBar/actions/runs/33616860899

The first full-suite attempt only exposed a task-owned native-build resource-link setup problem; after restoring the executable-adjacent `CodexBarCore` bundle path, the clean no-retry run passed in full. The merge tree `d6b51d595bb822b45219217f0ecd7f67221dde54` exactly matches the reviewed head tree.

PR #3386, opened 2026-09-02
fix(codex): preserve selected workspace ownership

steipete:
## Summary

- make CodexBar's selected managed workspace the canonical remote owner across reconciliation, refreshes, credits, history, dashboards, and promotion
- fail closed when a managed account's selected workspace differs from the saved auth default instead of silently using or overwriting the wrong workspace
- revalidate a concurrently inserted promotion-collision destination before replacing its auth, while preserving missing/unreadable-home repair
- preserve legacy managed-account compatibility while normalizing workspace IDs and migrating pre-normalization history keys

## Root cause

CodexBar intentionally stores the user-selected workspace as app-owned metadata while leaving the managed `auth.json` default unchanged. Several downstream paths still treated the auth-file default as authoritative. A selected workspace could therefore be fetched under one owner but published, cached, promoted, or reconciled as another owner with the same email.

This centralizes the effective remote owner and carries it through the full account lifecycle. Credits and history reuse stable provider ownership across ordinary token rotation, but workspace or same-workspace member changes still fail closed. Promotion now refuses ambiguous selected/default combinations rather than rewriting account metadata under the wrong identity.

## Testing

- `make check`
- 182 focused tests across reconciliation, refresh guards, credits, history ownership, and plan migration
- 43 promotion-path tests, including missing, matching, and conflicting post-planning collision races
- focused provider architecture gate
- full no-retry suite on final ancestry: 997 selections / 84 groups, all first-pass green, zero retries and timeouts
- independent review and final autoreview: clean

No screenshot: this changes account ownership and error-path behavior without changing visual layout.

### DIFF
--- a/Sources/CodexBar/CodexAccountPromotionCoordinator.swift
+++ b/Sources/CodexBar/CodexAccountPromotionCoordinator.swift
@@ -88,6 +88,10 @@ final class CodexAccountPromotionCoordinator {
                 L("CodexBar could not find saved auth for that account. Re-authenticate it and try again.")
             case .targetManagedAccountAuthUnreadable:
                 L("CodexBar could not read saved auth for that account. Re-authenticate it and try again.")
+            case .targetManagedAccountWorkspaceDiffersFromAuthDefault:
+                L(
+                    "That managed account uses a workspace that differs from its saved Codex auth default. " +
+                        "Keep it as a managed account; CodexBar will not rewrite Codex-owned auth to promote it.")
             case .liveAccountUnreadable:
                 L("CodexBar could not read the current system account on this Mac.")
             case .liveAccountMissingIdentityForPreservation:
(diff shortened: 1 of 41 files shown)


steipete:
Landed as `6ae54e88ab9ff3302b1e4d9169fb002703ede68f`.

Verified exact reviewed head `7b229737db9586a5c4f2dec292007b501733df04` and tree `fa800933280d4914930f5cb148b4cc7782457bbf`. CI run 33671516143 completed green across lint, both macOS test shards, aggregate, Linux x64/arm64, musl, changes, and security checks.

Local proof on the same head: `make check`; 43 promotion-path tests including missing, matching, and conflicting post-planning collision races; and the full isolated no-retry suite with 997 selections across 84 groups, all first-pass green with zero retries and zero timeouts.

PR #3391, opened 2026-09-02
fix: route About to Settings

elijahfriedman:
The application menu's About command used SwiftUI's standard macOS About panel while the status menu already opened Settings → About. Route both entry points through the existing Settings presentation controller, which selects About and reuses an open Settings window. Remove the uncalled standalone panel implementation.

This preserves the existing menu-tracking deferral and Settings-window ownership. The UI guide and Unreleased changelog now describe the behavior and credit @elijahfriedman.

## Verification

- Signed synthetic before/after bundles on macOS 26.6.2: application-menu About opens the standard panel before the change and selects About in the existing Settings window afterward. The General and About panes retained the same WindowServer window ID.
- All providers disabled, synthetic configuration, isolated credential/session files, and test Keychain access suppression during UI proof.
- Three existing Settings-window controller tests passed, covering About selection, window reuse, and presentation failure.
- Full isolated `make test`: 997 selections across 84 groups, all first-pass green, zero retries and timeouts.
- `make check` passed with zero Swift format/lint violations.
- Scoped Codex autoreview passed without actionable findings.

## Before

Application-menu About opens the separate standard panel

## After

Application-menu About selects the About pane in the existing Settings window

Original contributor recording:
https://github.com/user-attachments/assets/f7195d93-c547-46ec-bed6-253e99503b50


elijahfriedman:
@clawsweeper re-review

PR #3396, opened 2026-09-03
fix(antigravity): validate aligned step timestamps

steipete:
## Summary

Follow-up to #3266: require step-table timestamp evidence to agree with any embedded generation timestamp at the same stored occurrence before recovering later turns.

The UUID fallback already used every generation occurrence to align reused step IDs, but an occurrence with an inline timestamp contributed only its position. If its aligned step timestamp disagreed, a later timestamp-less turn could still be recovered and the database reported as complete. That converts contradictory date evidence into trusted history.

## Fix

- retain each generation occurrence's stored index and optional embedded timestamp
- resolve step timestamps against the complete ordered occurrence list
- compare every allocated step timestamp with existing inline evidence
- withhold UUID recovery when duplicate rows, missing candidates, or any aligned timestamp conflict makes the mapping ambiguous
- preserve the existing single-timestamp sharing and lowest-index behavior when all available evidence agrees

## Verification

- The parameterized regression was red on current main only for the conflicting case: main reported complete and published both rows. The matching control already passed.
- After the fix, matching evidence remains complete while conflicting evidence retains only the embedded diagnostic row and marks coverage partial.
- Focused Antigravity matrix: 76 tests across six suites passed.
- Full isolated local suite: 999 selections across 84 groups, all first-pass green with retries disabled, zero failed groups, and zero timeouts.
- `make check` passed with clean formatting and zero lint violations.

All verification used synthetic fixtures and repository isolation flags; no real local account data, credentials, browser imports, or installed-app launches were accessed.

### DIFF
--- a/Sources/CodexBarCore/Providers/Antigravity/AntigravityLocalSQLite.swift
+++ b/Sources/CodexBarCore/Providers/Antigravity/AntigravityLocalSQLite.swift
@@ -145,13 +145,13 @@ extension AntigravityLocalReader {
                 throw failure
             }
             if hasSteps {
-                var neededStepUUIDCounts: [String: Int] = [:]
+                var neededStepOccurrences: [String: [StepOccurrence]] = [:]
                 for stepUUID in Set(rows.pendingTimestampRows.map(\.stepUUID)) {
-                    neededStepUUIDCounts[stepUUID] = rows.stepOccurrenceRows[stepUUID]?.count ?? 0
+                    neededStepOccurrences[stepUUID] = rows.stepOccurrences[stepUUID] ?? []
                 }
                 let stepScan = try self.readStepTimestamps(
                     database,
-                    neededStepUUIDCounts: neededStepUUIDCounts,
+                    neededStepOccurrences: neededStepOccurrences,
                     progress: progress)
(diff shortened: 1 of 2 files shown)


steipete:
Maintainer closeout: reused step-UUID recovery now validates the step timestamp allocated to every generation occurrence, including turns that already carry an embedded timestamp. Contradictory inline and step evidence therefore withholds later recovery and marks coverage partial instead of publishing trusted history.

Proof on exact head `d3a38a03574159bc5fb5f43d89dd759ac43111ec`:

- The conflicting regression reproduced red on main: the source was incorrectly complete and published rows 0 and 1; the matching control passed.
- Focused Antigravity matrix: 76 tests across six suites passed after the fix.
- Full isolated local suite: 999 selections / 84 groups, all first-pass green, zero retries and zero timeouts.
- `make check` passed with clean formatting and zero lint violations.
- [CI run 33731379813](https://github.com/steipete/CodexBar/actions/runs/33731379813) passed lint, both macOS test shards, both Linux builds/tests, musl, the aggregate gate, and security checks.

Merged as `8d874e4f7ca97f3b18440d4b7def8894b898c9f8`; `main` was pulled and the merge tree verified clean locally.

Issue #3316, opened 2026-08-31
Spend Dashboard no-progress pause auto-restarts, loops at 100% CPU


heyajulia:
## Summary

The Usage & Spend Codex cost-history worker can enter a restart loop after a bounded pass reports no semantic progress. The UI repeatedly returns to `Refreshing N / N`, CodexBar consumes about one full CPU core, and the same `all pending account caches stalled` warning is logged every several seconds.

Further inspection found a material data-publication gap behind that stall: one large Codex rollout is fully parsed and contains token snapshots through August 30/31, but its persisted `usage_rows` and `file_day_aggregates` stop on August 29. Roughly 601M locally observed input-plus-output token deltas are therefore missing from the main Codex menu's daily graph, 30-day token total, and estimated spend. ChatGPT Profile independently attributes about 594M tokens to the corresponding day; the small difference is plausibly due to accounting and day-boundary semantics.

Canceling catch-up stops the CPU loop immediately, but does not materialize the missing usage. No cache deletion or app restart was required.

This is adjacent to #2815, but the observed store state is different: the large rollout file itself is marked complete and is not in the pending migration queue, while its cost rows end thousands of token-snapshot events before the file ends.

## Environment

- CodexBar 0.56.1 (build 132), Stable channel
- macOS 26.6.2 (25G83), Apple Silicon
- Codex local cost tracking enabled
- 537 indexed session files; affected rollout about 410.7 MB
- Current `main` inspected at `5fbe3b3ec`

### Main-menu undercount

The affected rollout dominates the already-materialized usage on three days:

| Local day | Materialized tokens | Estimated cost |
| --- | ---: | ---: |
| 2026-08-27 | 250,546,039 | $145.11 |
| 2026-08-28 | 361,651,554 | $220.04 |
| 2026-08-29 | 272,280,667 | $162.65 |

Those rows account for 92–99% of all locally indexed tokens on their respective days. However, the same rollout's raw token snapshots continue beyond the final aggregate row:

| Local day | Raw `last_*` token deltas not represented by usage rows | Approximate cost at CodexBar's cached pricing |
| --- | ---: | ---: |
| 2026-08-30 | 569,074,962 | $336.72 |
| 2026-08-31 | 32,188,457 | $21.43 |
| **Total omitted** | **601,263,419** | **$358.15** |

The ChatGPT Profile UI reports approximately 594M tokens for the corresponding usage day. That independently corroborates the size of the missing range while allowing for different token accounting and day boundaries.

[...]

PR #3400, opened 2026-09-03
Fix duplicate session progress accounting

mauricio-lxlaunch:
## Summary

- retain a compact completed cache record when a session fragment has no unique usage because another file already contributes the same session
- count that record in the exact scan inventory so bounded catch-up can finish instead of remaining one file short
- reparse the suppressed file if the original contributing file later disappears
- validate fork parent dependencies before reusing a suppressed record
- use one validated live-contributor index per scan, including contributors outside the current bounded batch
- persist the suppression marker in the existing file details payload

Partial fix for #3316.

## Root cause

The scanner intentionally removed a fully parsed duplicate fragment from the file cache. The current pass also removed that path from its inventory, but the next bounded pass discovered the physical file again. The saved progress total could therefore remain one above the number of completed cache files.

A bounded batch can contain the empty fragment without the file that already contributes the same session. The scanner therefore also needs a validated index of live cached contributors. The index prevents a repeated parse without adding a full-cache search for each file.

The new marker preserves the existing usage de-duplication rule. It stores no extra usage. It only keeps the file completion evidence that progress accounting needs.

## Safety

- a suppressed file is reparsed if its live contributor is missing or changed
- cached identity validation still runs before the suppression shortcut
- fork parent dependency validation still runs before the suppression shortcut
- contributor lookup accepts only complete, unchanged, live cache entries with usage
- physical aliases are excluded by file identity

## Tests

- added a 600-file bounded regression with an empty aborted fragment that shares a session ID
- added a regression where the empty fragment is scanned before its contributor in a later refresh
- added a recovery regression that removes the original contributor and verifies the suppressed duplicate becomes the contributor
- added a fork regression that changes the parent and verifies a suppressed child refreshes its dependency key
- `swift build --product CodexBarCLI` passed
- `swiftc -parse Tests/CodexBarTests/CostUsageBoundedProgressTests.swift Tests/CodexBarTests/CostUsageScannerBreakdownTests.swift` passed
- `./Scripts/lint.sh lint-macos` passed
- SwiftFormat passed for all Sources and Tests
- the parser hash check passed

## Real-history proof

I replayed a copy of the affected local 30-day history with an isolated cache. No app cache or source history was changed.

- exact inventory: 185 of 185 complete
- catch-up pending: false
- affected empty fragment: complete, zero usage rows, suppression marker stored
- parser hash: `d69345c635f28aa8`

No session content, private path, or cache file is included in this pull request.

## Local test limits

Full Swift tests are blocked locally because only Command Line Tools are installed. The test target cannot load the Swift Testing module or the SwiftUI preview macro plug-in.

Full `make check` reached its Swift test sharding self-test, then the host Python runtime reported that `waitid` with `WNOWAIT` is unavailable.

SwiftLint is also blocked locally because the Command Line Tools installation does not provide `sourcekitdInProc.framework`.

The generated parser hash is updated so existing affected caches rebuild with the corrected file state.


mauricio-lxlaunch:
@clawsweeper re-review

I moved fork dependency validation before the suppression shortcut, added the requested fork regression, and added a redacted real-history replay result to the PR description. I also fixed the bounded-batch case where the contributor is outside the current scan batch.

mauricio-lxlaunch:
@codex review

mauricio-lxlaunch:
Closing this pull request because I opened it from the wrong GitHub account. I will recreate it from mauriciopolvora.

steipete:
Correction: this PR was already closed and recreated as #3402 under @mauriciopolvora. I initially followed its old metadata and opened #3404, but am consolidating the focused repair into #3402 now that the replacement PR and its maintainer-edit permissions are verified.

It retains ordinary completed, rowless file records without a suppression flag or contributor index. Retention checks the raw parser rows as well as the final cache record, excludes resumable accounting and deferred buffers, and preserves all existing fork/identity freshness checks. If a retained fragment grows, it reparses from the start.

The added bare-usage regression caught a real token-loss case in an intermediate implementation: filtering a genuine duplicate to an empty record could lose its usage when the original contributor disappeared. The final guard fixes that; all 114 focused scanner tests and `make check` pass. Full local tests and CI are running before landing.

This PR remains closed. The active repair is tracked in #3402; #3316's missing token-history tail and aggregate recovery remain separate. Thanks for identifying the empty-fragment completion problem and providing the reproduction context.

PR #3401, opened 2026-09-03
Fix repeated Tailscale discovery crashes

### LINKED ISSUE #3397: Agent Sessions: Tailscale 1.102.3 app-binary probe can SIGTRAP every 60s
## Summary

With Agent Sessions enabled, CodexBar 0.56.2 probes Tailscale every 60 seconds. On this installation, the first candidate (`/usr/local/bin/tailscale`) resolves directly to the Tailscale macOS app binary and exits via `SIGTRAP`. CodexBar then falls through to `/opt/homebrew/bin/tailscale`, which succeeds, so discovery completes but the crash repeats on every refresh.

This appears to be a successor to #2098 / #2100: the `SHLVL` safeguard added there is not sufficient with Tailscale 1.102.3 on this machine. Tailscale now documents `TAILSCALE_BE_CLI=1` as the explicit override for scripted app-binary invocations, and that override prevents the crash here.

## Environment

- macOS 26.6.2 (25G83), Apple silicon
- CodexBar 0.56.2 (build 133)
- Tailscale 1.102.3 (build 101.102.3), standalone app
- Agent Sessions enabled
- `/usr/local/bin/tailscale` is a direct symlink to `/Applications/Tailscale.app/Contents/MacOS/Tailscale`
- `/opt/homebrew/bin/tailscale` is the standalone Homebrew CLI

## Actual behavior

Each refresh produces this sanitized sequence in `CodexBar.log`:

```text
Subprocess start binary=tailscale label=Tailscale session host discovery timeout=5.0
Subprocess failed binary=tailscale duration_ms=29 label=Tailscale session host discovery status=5
Subprocess error binary=tailscale duration_ms=30 label=Tailscale session host discovery
Subprocess start binary=tailscale label=Tailscale session host discovery timeout=5.0
Subprocess exit binary=tailscale duration_ms=61 label=Tailscale session host discovery status=0
```

The retained DiagnosticReports contain 70 matching crashes. Every inspected report has this shape:

```text
Process:             Tailscale
Path:                /Applications/Tailscale.app/Contents/MacOS/Tailscale
Parent Process:      CodexBar
Responsible Process: CodexBar
Exception Type:      EXC_BREAKPOINT (SIGTRAP)
Termination Reason:  Namespace SIGNAL, Code 5, Trace/BPT trap: 5
Thread 0:            libswiftCore _assertionFailure
```

Disabling Settings -> Menu -> Agent sessions stopped the discovery probes; no further probe appeared over more than two expected refresh cycles.

## Controlled reproduction

Tailnet output was discarded; only exit behavior was compared.

| Invocation | Result |
|---|---|
| `/usr/local/bin/tailscale status --json` with inherited `TERM=xterm-256color`, `SHLVL=1` | `SIGTRAP` (shell exit 133) |
| `/opt/homebrew/bin/tailscale status --json` | exit 0 |
| `TAILSCALE_BE_CLI=1 /Applications/Tailscale.app/Contents/MacOS/Tailscale status --json` | exit 0 |

This also explains the apparent "retry": current `RemoteSessionFetcher` tries executable candidates in order and falls through after the app binary crashes; the successful second call is a different executable.

## Current implementation gap

`tailscaleCLIEnvironment(from:)` currently relies on `TERM`/`SHLVL`, and returns the environment unchanged if either already exists:

https://github.com/steipete/CodexBar/blob/main/Sources/CodexBarCore/RemoteSessionFetcher.swift

Tailscale's macOS CLI documentation says scripted invocation can explicitly force CLI mode with `TAILSCALE_BE_CLI=1`:

https://tailscale.com/docs/reference/tailscale-cli?tab=macos

## Expected behavior

Tailscale host discovery should remain headless and should not generate a crash report for a failed candidate before falling through to a working one.

## Suggested fix

Set `TAILSCALE_BE_CLI=1` unconditionally in the environment used for every Tailscale discovery probe. Retaining the existing shell marker may be useful for compatibility, but the explicit override should not be skipped when `TERM` or `SHLVL` already exists.

Suggested regression coverage:

- a minimal environment receives `TAILSCALE_BE_CLI=1`;
- environments already containing `TERM` still receive it;
- environments already containing `SHLVL` still receive it;
- caller-provided unrelated environment values remain intact.

Canonicalizing executable candidates could additionally avoid probing the same app binary through both `/usr/local/bin/tailscale` and the app-bundle fallback, but it is not required to address the immediate crash.

steipete:
## Summary

Fixes #3397. Thanks @tzioup for the controlled reproduction.

Force `TAILSCALE_BE_CLI=1` at the shared Tailscale discovery environment boundary, including PATH candidates and symlinks to the macOS app executable. Preserve inherited terminal settings, unrelated environment values, and the existing shell marker for older installations. No candidate-order, discovery-frequency, SSH, or persistence changes.

The previous shell-marker safeguard is no longer sufficient for the reported Tailscale 1.102.3 installation: the app executable can trap before discovery falls through to a working standalone CLI. The explicit override follows [Tailscale's documented scripted invocation contract](https://tailscale.com/docs/reference/tailscale-cli?tab=macos). A larger discovery refactor is unnecessary for this bug.

## Verification

- Regression first: six of seven synthetic environment cases fail against the old helper, including existing TERM/SHLVL and inherited disabled/empty override values.
- Fixed helper: all 12 TailscaleSessionTests pass, including all seven environment cases and unchanged candidate fallback/parser/SSH-sanitization coverage.
- `make check`: passes; SwiftLint reports zero violations in 2,095 files.
- Full `make test`: all 999 selections across 84 groups pass on the first attempt; zero retries or timeouts.
- [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/33775418000): all checks pass on `4f331bd539b87b588eb2806663e489ec34b03f7a`, including both macOS shards, Linux x64/ARM64/musl builds, lint, and the aggregate CI gate.

Tests use synthetic data with Keychain and session-file isolation. No live Tailscale or account probe was run, and no GUI behavior changed. Updated session documentation and the Unreleased changelog entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+- Agent sessions: explicitly force Tailscale CLI mode during remote-host discovery, preventing repeated app-binary crashes on newer Tailscale installations while preserving existing terminal settings (#3397). Thanks @tzioup!
+
 ## 0.56.4 — 2026-09-03
 
 ### Fixed
--- a/Sources/CodexBarCore/RemoteSessionFetcher.swift
+++ b/Sources/CodexBarCore/RemoteSessionFetcher.swift
@@ -244,20 +244,15 @@ public struct RemoteSessionFetcher: Sendable {
 
     /// Environment that keeps the dual-mode Tailscale app binary in CLI mode.
     ///
-    /// With no shell/terminal marker present the binary boots the full menu-bar GUI
-    /// (SkyLight/WindowServer, status icon) instead of running the CLI: it never emits
-    /// JSON, the probe times out, and the Tailscale icon flickers on every refresh. A
-    /// set `TERM` or `SHLVL` forces CLI mode (argv[0] casing and `XPC_SERVICE_NAME` do
-    /// not). `SHLVL` is what the app's own `/bin/sh` CLI wrapper injects, so we mirror it here.
-    ///
-    /// Applied to every probe, not just the app-binary fallback: it is redundant but harmless for the
-    /// CLI wrapper (itself a `/bin/sh` script that already exports `SHLVL`), and injecting it
-    /// unconditionally keeps CLI mode guaranteed regardless of which binary `tailscaleBinary` resolves.
-    /// An existing `TERM`/`SHLVL` (real terminal context) is left untouched.
+    /// Shell markers alone can still select the GUI path and crash newer app binaries. Force the
+    /// documented CLI override for every candidate, including symlinks to the app. Retain the shell
+    /// marker for older installations without changing an existing terminal context.
     package static func tailscaleCLIEnvironment(from environment: [String: String]) -> [String: String] {
-        guard environment["TERM"] == nil, environment["SHLVL"] == nil else { return environment }
         var environment = environment
-        environment["SHLVL"] = "1"
+        environment["TAILSCALE_BE_CLI"] = "1"
+        if environment["TERM"] == nil, environment["SHLVL"] == nil {
+            environment["SHLVL"] = "1"
+        }
         return environment
     }
 
--- a/Tests/CodexBarTests/TailscaleSessionTests.swift
+++ b/Tests/CodexBarTests/TailscaleSessionTests.swift
@@ -44,13 +44,35 @@ struct TailscaleSessionTests {
 
     @Test
     func `cli environment preserves an existing terminal context`() {
-        // Already CLI-safe: leave TERM alone and don't fabricate a SHLVL…
+        // Leave TERM alone and don't fabricate a SHLVL…
         let withTerm = RemoteSessionFetcher.tailscaleCLIEnvironment(from: ["TERM": "xterm-256color"])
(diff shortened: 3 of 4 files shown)


steipete:
Fixed on main in #3401, commit 98bea29e9c1c05dd7221298f4a989f1c9d3fb865. Every Tailscale discovery candidate now receives the explicit CLI-mode override, including a PATH symlink to the app binary. Existing TERM/SHLVL values, candidate fallback, and SSH behavior are preserved.

The regression failed in six inherited-environment cases before the change and passes in all seven cases afterward. All 12 focused Tailscale tests, `make check`, and the full local suite (999 selections, 84 first-pass groups, zero retries/timeouts) passed. [All exact-head CI checks passed](https://github.com/steipete/CodexBar/actions/runs/33775418000), including both macOS shards and Linux x64/ARM64/musl builds. Independent Codex autoreview found no blocking issue.

Verification used synthetic inputs and the documented Tailscale CLI contract; I did not repeat a live probe against your affected installation. The fix is recorded under Unreleased and is not part of the already-published 0.56.4. Thanks @tzioup for the controlled reproduction.

PR #3403, opened 2026-09-03
fix(antigravity): recover step timestamps via exact bot ID join

WeGoToMars:
## Summary

Fix Antigravity local token history being assigned to the wrong timestamp when auxiliary or reordered step rows share a generation's step UUID. The previous reader accepts an ordered step prefix; a count mismatch alone does not necessarily make it withhold history.

Use the generation usage `bot_id` and the matching step metadata `bot_id` to recover the standard seconds/nanos timestamp within the same UUID. This does not interpret opaque context-meter fields, use session creation time, or change token counts, pricing, credentials, discovery roots, or stored data.

The maintainer revision retains the contributor's exact-ID approach and strengthens its safety boundaries:

- Count identities across every parsed generation, including embedded timestamps and rows without UUIDs. Repeated generation IDs retain the guarded legacy fallback.
- Withhold contradictory, cross-UUID, or timestamp-less duplicate step evidence. Keep ambiguous positional slots in place instead of shifting later rows forward.
- Check generation-unique exact matches against embedded timestamps in UUIDs needing recovery, without invalidating unrelated embedded-only UUIDs or requiring positional alignment for a clean exact join.
- Treat malformed auxiliary IDs as unavailable, persist that invalidity across repeated envelopes, and preserve otherwise valid usage. Keep counter/framing validation strict and parser-state validity out of copied-row equality.
- Update the local-history documentation and Unreleased changelog. Thanks @WeGoToMars!

## Proof

Ran the shipped 0.56.4 CLI and the rebuilt CLI against three synthetic SQLite histories, with a temporary home, account-file/Keychain test isolation, network denied, and real home-directory reads denied. No real accounts or history were used.

| Synthetic history | 0.56.4 | Repaired CLI |
| --- | --- | --- |
| Auxiliary step before the exact turn | 198 tokens assigned to August 26 | Same 198 tokens correctly assigned to August 28 |
| Duplicate ID with conflicting timestamps | Incorrectly establishes complete history | Withholds rows; coverage unavailable, not confirmed zero |
| Legacy history without per-turn IDs | 198 tokens on August 28 | Unchanged |

The final CLI run left all three input database hashes unchanged. Assertions cover input, cache-read, output, and reasoning buckets as well as dates and coverage.

Focused regression tests first failed on the contributor proposal, including embedded-ID uniqueness, malformed repeated IDs, positional compression, cross-UUID fallback, and missing-time duplicates. Additional tests caught and repaired over-withholding of repeated generation IDs and false copied-row conflicts.

- `swift test --filter Antigravity`: 401 tests across 34 suites passed (383 macOS tests and 18 portable tests).
- `make check`: passed, zero violations in 2,098 files. A local Homebrew Bash here-document stall was isolated to a temporary system-Bash command path; no repository toolchain change.
- Required Codex autoreview: no accepted/actionable findings at the configured priority.
- Full `make test`: all 1,002 selections across 84 groups passed first try in 861.9 seconds, with zero failures, timeouts, or retries.
- Exact-head CI [33790377020](https://github.com/steipete/CodexBar/actions/runs/33790377020): all checks passed, including both macOS shards, Linux x64/ARM64/musl, lint, the aggregate check, and GitGuardian.

The original author's live-data observations are useful schema evidence but were not independently replayed against private history. This revision's executable proof is synthetic and uses the actual CLI entry point.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 ### Fixed
+- Antigravity: match local token-history timestamps by per-turn IDs when auxiliary or reordered steps would otherwise assign usage to the wrong day, while withholding conflicting evidence and preserving legacy timestamp recovery (#3403). Thanks @WeGoToMars!
 - Codex: retain completed empty session fragments during cost-history scans instead of repeatedly dropping and rediscovering them, without suppressing usage-bearing duplicates or later appended usage (partial fix for #3316; #3402). Thanks @mauriciopolvora!
 - Agent sessions: explicitly force Tailscale CLI mode during remote-host discovery, preventing repeated app-binary crashes on newer Tailscale installations while preserving existing terminal settings (#3397). Thanks @tzioup!
 
--- a/Sources/CodexBarCore/Providers/Antigravity/AntigravityLocalSQLite.swift
+++ b/Sources/CodexBarCore/Providers/Antigravity/AntigravityLocalSQLite.swift
@@ -153,16 +153,16 @@ extension AntigravityLocalReader {
                     database,
                     neededStepOccurrences: neededStepOccurrences,
                     progress: progress)
-                guard stepScan.isComplete else {
+                guard stepScan.isComplete,
+                      self.embeddedTimestampsAgree(neededStepOccurrences, with: stepScan, botIDUses: rows.botIDUses)
+                else {
                     rows.source.isComplete = false
                     return rows.source
                 }
                 let recoveredCount = self.appendRecoveredEvents(
-                    to: &rows.source,
+                    to: &rows,
                     session: session,
-                    pendingRows: rows.pendingTimestampRows,
-                    stepTimestamps: stepScan.timestamps,
-                    stepOccurrences: rows.stepOccurrences)
+                    stepScan: stepScan)
                 if recoveredCount < rows.pendingTimestampRows.count {
                     rows.source.isComplete = false
                 }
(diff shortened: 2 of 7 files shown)


steipete:
Repaired the existing branch at bda9352066562266c20ed883cdbf8a3d672ebc8e and rebased it onto current main, retaining @WeGoToMars's contribution.

Both outstanding inline findings are now fixed: generation-side identity counts include embedded and UUID-less rows, and malformed auxiliary IDs stay invalid across repeated fields/envelopes. Regression tests reproduced both failures before the fix. The revision also preserves ambiguous positional slots, rejects missing-time and cross-UUID conflicts, checks embedded exact-ID agreement without blocking unrelated UUIDs, and keeps optional parser state out of copied-row equality.

The shipped 0.56.4 CLI reproduced the wrong-day assignment on a synthetic SQLite fixture. The rebuilt CLI assigns the same 198 tokens to the correct day, withholds conflicting evidence, and preserves legacy no-ID history. No private history or accounts were used, and the final CLI run left the fixture databases unchanged. The PR body has the proof matrix and scope limitations.

All 401 Antigravity-focused tests, lint, and the required autoreview have passed. The final full suite is running; the reviewed-head CI run has been approved. I will wait for both before merging.

steipete:
Merged as 0b181ccbaef2be9b65e18a0e0257024a6d4935f5. The clean main checkout has the same file tree as reviewed/tested head bda9352066562266c20ed883cdbf8a3d672ebc8e. Thanks @WeGoToMars!

Final verification:

- `swift test --filter Antigravity`: 401 tests across 34 suites passed.
- `make check`: passed with zero violations in 2,098 files.
- `make test`: all 1,002 selections across 84 groups passed first try in 861.9 seconds, with no failures, timeouts, or retries.
- Required Codex autoreview: no accepted/actionable findings at the configured priority. Both previous inline review findings are resolved.
- [Exact-head CI 33790377020](https://github.com/steipete/CodexBar/actions/runs/33790377020): all checks passed, including both macOS shards, Linux x64/ARM64/musl, lint, aggregate, and GitGuardian.

The actual shipped and rebuilt CLIs were exercised with `cost --provider antigravity --format json --refresh` against isolated synthetic SQLite histories. The fix moves the same 198 tokens from the incorrect day to the matching turn's day, withholds conflicting IDs, and leaves legacy no-ID recovery unchanged. Input database hashes remained unchanged. No real account, private history, network request, or Keychain probe was used.

Unknown or contradictory metadata intentionally remains partial coverage instead of acquiring a guessed date. This does not validate opaque relative-clock fields or fix unrelated Antigravity quota/auth reports. The docs and Unreleased changelog record that boundary.

PR #3404, opened 2026-09-03
Preserve completed empty Codex session fragments

steipete:
## Summary

Focused maintainer extraction from #3400, with credit to @mauricio-lxlaunch. The contributor fork currently disables maintainer edits. This addresses the empty-fragment portion only; #3316 and the broader duplicate-suppression proposal remain open.

Keep fully consumed empty Codex session fragments in the existing file cache even when another file contributes the same session. They retain scan-completion evidence, contribute no usage, and reparse from byte zero if they grow. Updated the parser hash, local-scan documentation, and Unreleased changelog.

## Correctness boundary

An empty filtered result is not proof of an empty source. Retention requires no raw parsed rows, no cached rows or token snapshots, no accumulator baselines, and no partial JSONL or fork/subagent buffers. Existing metadata, priority, identity, and fork-dependency checks still run before reuse.

There is no new suppression flag, contributor index, or cache schema. Usage-bearing duplicates keep their existing handling. This avoids two problems in broader suppression designs: losing append ordinals after filtering away rows, and treating an unrelated same-session contributor as proof that all omitted rows remain covered.

## Verification

- Baseline red: fully consumed empty records disappear after initial parsing and after an empty append; the out-of-window bare-usage fragment also loses its completion record.
- An intermediate candidate failed the added bare-usage duplicate/removal regression, proving that checking only filtered rows could lose genuine usage. The final guard checks raw parsed rows too.
- Regression coverage includes 600-file bounded batches with empty fragments before/after contributors, subsequent real usage, token and bare-usage duplicate appends/removal, out-of-window bare-usage ordinals, and rejection of partial scans, token state, and deferred buffers. The 600-file case is coverage, not a claimed baseline permanent-stall reproduction.
- `make check`: passes; zero SwiftLint violations across 2,096 files.
- Focused scanner tests: all 114 tests across four suites pass, including the expanded bare-usage regression.
- Full `make test` and exact-head CI results are recorded before merge.

All fixtures use isolated temporary session/cache roots, with Keychain and session-file isolation enabled. No live accounts, credentials, original session history, or app caches were probed or changed.


steipete:
Consolidated the reviewed narrow repair into the contributor's active PR #3402, now at `07d269eb00ef52cf31e66093dea00299f12776ea`. Closing this redundant maintainer PR so the implementation, contributor credit, review, and final CI stay together there.

#3400 was already closed by its author when they recreated it from their intended account. This repair addresses completed empty-fragment caching only; it does not claim to resolve every accounting or recovery concern in #3316.

Issue #3411, opened 2026-09-04
Spend Dashboard Codex publish delayed for hours — stale scanSinceKey reseeds lookback queue on every scan


kesslerio:
## TL;DR

The Usage & Spend dashboard took ~5 hours to publish Codex spend for one of my two Codex accounts because the lookback queue reseeded on every scan pass. Root cause: a stale `scanSinceKey` in the persisted lookback state triggered `activeLookbackStateWasReset` on every scan, which reseeded the entire queue (3,367 files) even though all files were already parsed and priced. The queue eventually drained fully between two scan boundaries and the dashboard published, but the recovery should take one scan pass, not five hours.

### Bug Type

behavior bug

### Release Blocker

- [x] No

### Steps to Reproduce

1. Have a large Codex session corpus (3,879 files, ~3,500 in ambient `~/.codex/sessions`, rest in CodexBar-managed homes). Two Codex accounts configured.
2. Let CodexBar build its cost cache. `scan_metadata` reaches `completedFiles: 3879 / totalFiles: 3879`.
3. Observe `lookback_state.pendingFilePaths` drain to zero during background passes (queue shrinks ~250-500 files per ~15 min pass).
4. Trigger a new scan (open Settings → Usage & Spend, or wait for the next background cycle). The scan re-enumerates and re-enqueues the entire lookback set. `pendingFilePaths` jumps back to ~3,367.
5. `catchUpPending` stays `true` for hours. The dashboard shows "Spend unavailable" for the affected account row until a drain pass happens to finish the whole queue between scan boundaries.

### Expected Behavior

Once all files are complete and the lookback queue is empty, `catchUpPending` should flip to `false` and the dashboard should publish Codex spend on the next refresh. The queue should not re-enqueue files that were already processed and marked complete. A scan-window change should not reseed already-completed files.

### Actual Behavior

The queue reseeded on every scan pass for ~5 hours before one drain pass completed between scan boundaries.

Timeline from a single session (all times UTC):

| Time | pending | completed | trigger |
| --- | ---: | ---: | --- |
| 04:42:48 | 0 | 3879/3879 | background pass, queue drained |
| 04:46:03 | 3,367 | 3878/3879 | pane opened, scan re-enqueued entire lookback set |
| 04:46:36 | 3,262 | 3878/3879 | draining again |
| 04:48:14 | 307 | 3878/3879 | continued draining |

### Root cause: `scanSinceKey` mismatch triggers lookback state reset on every scan

[...]

PR #3417, opened 2026-09-04
fix(codex): retain pending scan range across history requests

steipete:
Alternating narrow and wide history requests against the same Codex cache can repeatedly restart an unfinished discovery cycle. A synthetic two-file reproduction shows an already completed file being requeued on every 90↔365-day switch, even when the time budget permits no parsing.

Retain the active scan's earlier start while the roots, timezone, and ending day remain compatible. Planning, pricing, parsing, discovery, and inventory validation use that same work range; the returned report still uses the requested dates. Upper-bound changes, root/timezone changes, and forced rescans retain normal reset behavior. Regenerate the required parser fingerprint and adopt the compatible previous-main cache without rebuilding stored usage, retained reports, or partial-scan checkpoints.

Partial fix for #3411. The report's claim that bounded scans do not persist progress is not supported: they already save each successful pass. This fixes independently reproduced range churn. The original managed-account trigger remains open; the separate settings-invalidation path that left a stale Refreshing banner landed in #3418. Thanks @kesslerio for the detailed report.

Validation:
- The alternating-window regression fails on main in all three passes: two queued candidates instead of the one unfinished file. It passes with the fix, preserving completed discovery and reaching exact completion.
- Additional production-scanner fixtures verify older pending history is discovered, narrow reports exclude those older totals, and expansion into a new day discovers new sessions.
- `swift test --no-parallel --filter 'CostUsageBoundedProgressTests|ProviderArchitectureGatekeeperTests'`: 53 tests passed.
- `make check`: passed after regenerating the parser fingerprint and correcting test formatting.
- `make test`: the final candidate passed all 1,008 selections across 84 groups, with zero failures, retries, or timeouts. The scanner fixtures also verify the older discovered file retains its exact input/cache/output token counts.
- Cache-upgrade regression: the previous-main fingerprint `f043ae98075c8e4d` fails adoption before the repair. With the repair, populated snapshots and resumable JSONL checkpoints survive with zero rebuilds and zero session-head parses. The released `v0.56.4` fingerprint was already compatible.
- Final upgrade/scanner/architecture regression set: 55 tests passed; `make check` and required P0 autoreview passed. The final full suite passed. All nine exact-head CI checks passed: [run 33864019935](https://github.com/steipete/CodexBar/actions/runs/33864019935), head `682b1b2d1079237ec1727abeb74d83f7c7173c4f`.
- The Developer ID–signed test bundle passed the scanner range and cache-adoption regressions with `swift test --skip-build --no-parallel`; all fixtures use contained synthetic files.
- Independent source review checked discovery, pricing, report scope, and day rollover. Required Codex autoreview is scoped-clean at P0.

All scan proof uses contained synthetic session files and SQLite stores; no live account data or saved user caches are touched. Changelog and Codex documentation are updated.


steipete:
I checked the report-window finding against the final candidate. It is a false positive: this range already carries separate scan and report bounds.

- [`retainingScanStart`](https://github.com/steipete/CodexBar/blob/d6515ebfd3034fb4672b3d425ed3488482c1e854/Sources/CodexBarCore/Vendored/CostUsage/CostUsageScanner.swift#L2042) copies the range and changes only `scanSinceKey`. `sinceKey` and `untilKey` remain the caller's original report dates.
- [`saveCodexCache`](https://github.com/steipete/CodexBar/blob/d6515ebfd3034fb4672b3d425ed3488482c1e854/Sources/CodexBarCore/Vendored/CostUsage/CostUsageScanner.swift#L5560) passes the retained scan bounds as `requestedScanWindow` and the unchanged report bounds as `reportWindow`.
- The [budget fallback](https://github.com/steipete/CodexBar/blob/d6515ebfd3034fb4672b3d425ed3488482c1e854/Sources/CodexBarCore/Vendored/CostUsage/CostUsageStore%2BCodexCache.swift#L188) forwards that explicit report window. [`previousReport`](https://github.com/steipete/CodexBar/blob/d6515ebfd3034fb4672b3d425ed3488482c1e854/Sources/CodexBarCore/Vendored/CostUsage/CostUsageStore%2BCodexCache.swift#L990) uses it for both report construction and persisted identity, falling back to cache scan bounds only when no report window is supplied.

The existing separation therefore remains intact through budget-driven persistence. The regressions also verify older discovered token counts stay in the cache while the narrow returned report excludes that older day. No production change is needed for this finding.

@clawsweeper re-review

steipete:
Addressed the cache-adoption finding. The scanner change does not alter stored row, checkpoint, or retained-report semantics, so `f043ae98075c8e4d` now joins the compatible predecessor set.

The extended production-store regression failed before this repair: opening the populated predecessor database rebuilt it and lost its saved file/checkpoint. The scanner adoption regression also reparsed session heads. Both now pass: the full saved snapshot is preserved, rebuild count is zero, parser/schema stamps advance, the partial JSONL resumes correctly, and the warm scanner performs zero session-head parses. All 55 focused scanner/adoption/architecture tests passed, and the required Codex autoreview is scoped-clean at P0. The full suite and final-head CI are being rerun before landing.

One attribution correction: the `v0.56.4` tag contains `e3fca1e6d81137d6`, which was already compatible. `f043ae98075c8e4d` is the later main producer from #3402. This repair protects upgrades from that main build as well.

The earlier report-window finding remains resolved: only `scanSinceKey` widens, while the original `sinceKey`/`untilKey` continue to scope reports and persisted fallback identity.

steipete:
Maintainer decision: accept adoption of the previous-main cache fingerprint. This PR changes compatible pending-scan scheduling; it does not change row decoding, token/cost semantics, checkpoint encoding, or retained-report encoding. The requested report dates remain unchanged, and the existing schema/integrity checks still gate adoption.

The compatibility invariant is now covered by the populated-store and production-scanner regressions: identical saved snapshot, zero rebuilds, successful continuation of partial JSONL input, and zero session-head reparses. Final head `682b1b2d1079237ec1727abeb74d83f7c7173c4f` passed the full 1,008-selection suite without retries, plus the Developer ID–signed scanner/adoption run. The final-head review reports no remaining code findings. Landing remains gated on required CI.

steipete:
Landed as 30f881aee0e8dce08690d8eb971a2a44ef1285be after all nine final-head checks passed.

Compatible history requests now share the unfinished scan's discovery range while each report retains its original requested dates. The updated parser fingerprint also adopts the previous-main cache without rebuilding stored usage or losing partial-scan checkpoints.

The persisted scanner fixture fails before the change: alternating 90/365-day requests requeue an already completed file in all three bounded passes. The candidate preserves only pending work and reaches exact completion. The added predecessor fixture separately reproduced a database rebuild before the compatibility repair; afterward it preserves the full snapshot, resumes partial JSONL, and reparses zero session heads.

Validation: 55 focused scanner/adoption/architecture tests; `make check`; all 1,008 full-suite selections in 84 groups with zero retries/timeouts; 14 regressions from a Developer ID–signed test bundle; independent source review; final Codex autoreview scoped-clean at P0.

CI: https://github.com/steipete/CodexBar/actions/runs/33864019935

All storage and scanner proof uses contained synthetic session files and SQLite fixtures. #3411 remains open for the original managed-account trigger; the separate settings-invalidated dashboard activity path landed in #3418.

PR #3418, opened 2026-09-04
fix(cost): clear abandoned dashboard catch-up activity

steipete:
Usage & Spend can retain a Refreshing banner after its catch-up worker exits. If account or cost settings invalidate the context during a scan, the completion guard returns without publishing a terminal activity. Cleanup previously removed the task and token while leaving its indexing activity visible, including at 100% progress.

Clear abandoned indexing activity in the owning worker's cleanup. Preserve completed and paused states, and keep the token guard so an obsolete task cannot erase a replacement worker's activity. The dashboard view, accounting, and authentication behavior are unchanged.

Follow-up for #3411, separate from the scan-range churn fixed in #3417. The settings-invalidation path is reproduced independently; this does not establish which trigger caused the reporter's original banner. Keep #3411 open. Thanks @kesslerio for the report.

Validation:
- A gated production-worker regression fails on main: after settings change during a pending pass, the task is nil but indexing activity remains. The candidate clears it.
- Replacement-worker coverage verifies the old task preserves the replacement's indexing state after normal and cancellation exits. Existing user/error/no-progress/thermal/low-power pause tests remain intact.
- `swift test --no-parallel --filter 'UsageStoreSpendDashboardCodexCostCatchUpTests|ProviderArchitectureGatekeeperTests'`: 67 tests passed.
- `make check`: passed. `make test`: all 1,008 selections across 84 groups passed, with zero retries or timeouts.
- Developer ID–signed test bundle: all dashboard Codex catch-up worker tests passed with `swift test --skip-build --no-parallel --filter UsageStoreSpendDashboardCodexCostCatchUpTests`.
- Final-head CI: https://github.com/steipete/CodexBar/actions/runs/33861790233 passed; all nine checks are green.
- Independent ownership review found no actionable issues; required Codex autoreview is scoped-clean at P0.

The regression uses dictionary-backed settings, synthetic account paths, injected scan results, and a controlled async gate. It does not access live credentials or saved account history. Changelog and Codex documentation are updated.


steipete:
Landed as 88f0e58bf4bb970729c662e787a9a3f3745588ba after all nine final-head checks passed.

The owning dashboard catch-up task clears indexing activity when it exits after scope invalidation. Paused and completed states remain available, and the ownership token prevents an obsolete task from clearing a replacement's activity.

Proof: the gated settings-invalidation regression fails before the fix with a nil task and a stale 100% indexing activity, then passes with cleanup. All 67 focused worker/architecture tests passed; the full suite passed all 1,008 selections in 84 groups without retries or timeouts. A Developer ID–signed test bundle then passed all 26 dashboard-worker tests, including obsolete/replacement ownership cases. `make check`, independent review, and required scoped-clean P0 autoreview passed.

CI: https://github.com/steipete/CodexBar/actions/runs/33861790233

#3411 stays open because this proves one settings-invalidation path, not the original managed-account trigger. Worker proof uses synthetic settings and scan results without saved credentials or account history.

PR #3439, opened 2026-09-05
Fix stacked bar corner radius

elijahfriedman:
The daily and hourly spend charts round each provider segment independently, leaving notches at internal stack boundaries and rounded baselines. This fix makes the internal seams and baseline square while preserving rounded outer tops.

The shared selector identifies the highest visible segment per day/hour. Native verification found that clipping alone did not fix the seams: the original rounded mark had already lost those corner pixels. Both charts now disable native mark rounding before applying the per-segment top-only clip. A realistic trailing-zero regression ensures an invisible segment cannot replace the visible top.

Thanks @elijahfriedman for the original contribution. Contributor commits and human credit are preserved, and the Unreleased changelog is updated.

### Native behavior proof

`SpendStackedChartsNativeProofTests` hosts the existing production `SpendDashboardCurrencySection` in a signed Developer ID XCTest window. It builds the same data through `SpendDashboardModel.build` for both captures; it does not recreate the chart drawing code. Before uses chart/model source identical to main b921ed2ec733, and after uses the final candidate.

The fixture covers three providers, normal stacks, single visible providers, all-zero buckets, zero-height trailing/middle segments, and very thin segments. Assertions verify all 18 daily and 18 hourly points, contiguous offsets, $137.15 daily total, and $74.40 hourly total. Captures contain synthetic data only.

Before: internal rounded seams

After: flush internal seams

### Validation

- `swift test --no-parallel --filter 'SpendStackedBarChartTests|SpendDashboardModelTests'`: 50 tests in two suites passed.
- Signed opt-in native proof: one XCTest passed for the baseline and final candidate; both full captures visually inspected, with interior-corner pixel samples confirming the correction.
- `make check`: passed, zero violations across 2,128 files.
- Final managed Codex autoreview: scoped clean at the default P0 threshold.
- `make test`: all 1,024 selections across 86 groups passed on the first attempt, with no retries or timeouts.
- [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/33991911801): all checks passed, including both macOS shards, Linux x64/ARM64/musl, lint, and aggregate gate.
- Fresh pre-landing P0 autoreview is scoped clean; the candidate combines cleanly with the independent ElevenLabs fix now on main.


steipete:
Signed native before/after proof using the production Usage & Spend currency section and synthetic data. The fixture includes three providers, trailing zero-cost segments, and very thin segments. Both daily and hourly charts retain their totals, series, and geometry.

Before (chart/model source identical to main b921ed2ec733):

Before: rounded internal provider boundaries

After (6ecb6f8aaa390b0871c00bdbae714d9fce35814d):

After: square internal boundaries and rounded outer tops

Native proof exposed why clipping alone was insufficient: Swift Charts had already rounded the underlying marks. The final change sets the native corner radius to zero before applying the top-only clipping shape. Both captures were inspected in full and contain synthetic fixture data only.

steipete:
Merged as 211781be79d37fd942bc4ee623abd6d5c8baeffd. Daily and hourly stacked spend bars now have square internal seams and baselines, with rounded outer tops. Native proof exposed the missing step in the original proposal: disabling built-in mark rounding before applying the top-only clip.

Validation: `swift test --no-parallel --filter 'SpendStackedBarChartTests|SpendDashboardModelTests'` passed 50 tests. `make check` passed with zero violations across 2,128 files. `make test` passed all 1,024 selections across 86 groups on the first attempt, with no retries or timeouts. All nine exact-head checks passed in [CI run 33991911801](https://github.com/steipete/CodexBar/actions/runs/33991911801). Pre-landing managed Codex review was scoped clean at P0.

The signed native fixture hosts the actual production currency section and passed before and after. [Inspected synthetic before/after captures](https://github.com/steipete/CodexBar/pull/3439#issuecomment-5554781251) cover three providers, zero-height and thin segments; both attachments are anonymously downloadable with hashes matching the inspected files. Fixture totals and offsets stayed unchanged. The final tree combines cleanly with the independent ElevenLabs fix.

The Unreleased changelog is updated, and @elijahfriedman's contribution and commit credit are preserved.

PR #3443, opened 2026-09-05
Simplify chart menu layout sizing

elijahfriedman:
## Summary
- The cost/history submenu (`CostHistoryChartMenuView`) had two independent `ScrollView`s stacked vertically: one clipping the per-day model-breakdown rows, one clipping the "Conversations" session list. Combined with the native `NSMenu`'s own scroll gutter (`CostHistoryMenuScrollView`), a user could end up with three nested scrollable regions in one panel.
- Removed the session list's own `ScrollView` and fixed-height clipping (`sessionsBlock` now just flows as plain content). The existing outer `CostHistoryMenuScrollView` — the native `NSScrollView` that already measures and wraps the whole hosted view — is now the single scroll surface for the conversations list.
- Left the model-breakdown rows' `ScrollView` in place: its bounded height is load-bearing, not incidental — `detailViewportRowCount` caps display at a fixed row count regardless of how many models appear on the currently-hovered day, so the menu's height doesn't jump as the user hovers across chart bars. Removing it broke `cost history fitting height stays stable across compact overflow and mode selections`.
- Net effect: down from up to three nested scroll regions to two (native menu scroll + one bounded breakdown-rows scroll), with the session list no longer independently scrolling.
- Guard the hover-detail block (breakdown rows + "Hover a bar for details" fallback) behind `!model.points.isEmpty` so it can't render alongside "No data available" when `daily` has entries but none produce a chartable token/cost value.

## Verification
- `swift build` succeeds
- `swift test --filter CostHistoryChartMenuViewTests` (44/44 passed)

### Screenshots
https://github.com/user-attachments/assets/3739b1a0-a3e1-4c8a-bc79-24caf593b672

### DIFF
--- a/Sources/CodexBar/CostHistoryChartMenuView.swift
+++ b/Sources/CodexBar/CostHistoryChartMenuView.swift
@@ -118,6 +118,7 @@ struct CostHistoryChartMenuView: View {
             historyCoverageIsEstablished: self.historyCoverageIsEstablished)
         let selectedDateKey = self.selectedDateKey.flatMap { model.pointsByDateKey[$0] == nil ? nil : $0 }
             ?? Self.defaultSelectedDateKey(model: model)
+        let detail = self.detailContent(selectedDateKey: selectedDateKey, model: model)
         VStack(alignment: .leading, spacing: Self.outerSpacing) {
             if model.points.isEmpty {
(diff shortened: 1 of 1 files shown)


elijahfriedman:
@clawsweeper re-review

PR #3412, opened 2026-09-04
feat(antigravity): estimate local history costs from models.dev prices

WeGoToMars:
## Summary

This PR adds estimated dollar costs to Antigravity's existing local token history.

The estimates use public API prices from models.dev. They are shown as **API-equivalent estimates**, not as billed spend or subscription value.

## How it works

CodexBar continues to read Antigravity token history from the same local SQLite and JSONL files. This PR does not change history discovery, token counting, timestamps, credentials, or account attribution.

For each recorded model:

- Supported Gemini models use Google pricing.
- Supported Claude models use Anthropic pricing.
- `gpt-oss-120b` uses OpenAI pricing.
- Unknown or unsupported models keep their token counts but do not receive a dollar estimate.

When a period contains both priced and unpriced requests, CodexBar shows the coverage explicitly. For example:

```text
Coverage: 1 estimated · 1 unpriced
```

Dollar-related labels and explanations appear only when at least one cost can be estimated. Completely unpriced history continues to be presented as token history.

## Pricing catalog and network behavior

Prices come from CodexBar's on-disk models.dev catalog.

The current implementation follows the existing Codex and Claude pricing-cache policy: it uses a fresh cached catalog when available and may refresh models.dev when the catalog is stale or missing. A failed refresh keeps the existing catalog.

This changes Antigravity's previous fully local behavior and needs a maintainer decision:

- Keep the shared automatic refresh behavior; or
- Make ordinary Antigravity reads cache-only and provide a separate opt-in refresh path.

The local history files themselves are never uploaded or modified.

## Reliability

On Linux, catalog updates now use a same-directory atomic rename(2). Readers therefore see either the previous complete catalog or the new complete catalog, never a missing or partially written file.

The implementation follows the existing CostUsageClaudeCache pattern. Tests cover concurrent readers and writers and verify that no temporary or backup files remain.

## User-facing changes

- codexbar cost --provider antigravity can show API-equivalent cost estimates.
- JSON output includes the existing cost provenance and coverage fields.
- Menu and provider-detail views show the estimate explanation only when costs are available.
- Unknown models remain visible with their original token counts.

## Testing

Run on macOS at final head aab2de50:

- [x] make check — 0 violations across 2,100 files
- [x] make test — 1005 selections / 84 groups, first-pass all green, no retries
- [x] Focused Antigravity, pricing, menu, architecture, and settings tests — 128 passed
- [x] Concurrent catalog replacement test passed
[...]

PR #3451, opened 2026-09-06
fix(antigravity): skip steps without stepUUID during token scan

horosin:
### Summary
I've been wanting to see Antigravity usage and cost tracking in CodexBar work just like Codex does. I saw the recent improvements in #3403 to fix token counts and the ongoing work in #3412 to pull dollar amounts. However, on my machine, local Antigravity token counts were still not showing up at all, withholding with:
```text
Antigravity Token History
Local token history is unavailable or incomplete.
Local token history · dollar costs unavailable
```
In anticipation of full Antigravity cost support landing, I dug into why token history was still withheld and found a false-positive check during SQLite step scanning that this PR fixes.

### Problem
In `readStepTimestamps` (`Sources/CodexBarCore/Providers/Antigravity/AntigravityLocalSQLite.swift`), encountering a row in the `steps` table without a `stepUUID` set `rowsAreValid = false`.

Antigravity legitimately logs non-conversational steps in its SQLite `steps` table (such as internal lifecycle markers, model config transitions, or auxiliary events). These records have timestamp fields (field 1) and step types (field 3), but omit field 12 (`stepUUID`) because they are not LLM turns.

When `readStepTimestamps` encounters any such step row without a `stepUUID`:
1. It flags `rowsAreValid = false`.
2. `stepScan.isComplete` evaluates to `false`.
3. The database is marked `isComplete = false`.
4. Because `result.isComplete` across all historical databases is an aggregate boolean `AND`, a single auxiliary step in any conversation database cascades to invalidate all databases across the entire machine, completely withholding token history.

### Solution
- Skip `steps` table rows without a `stepUUID` instead of invalidating the entire database scan.
- Generation rows in `gen_metadata` that actually depend on external step timestamps remain fully protected by the existing `recoveredCount < rows.pendingTimestampRows.count` check.
- Track `botID` ambiguity before skipping UUID-less rows: if an auxiliary step row carries a `botID` but lacks a `stepUUID` or timestamp, it is recorded as ambiguous so it cannot be mistakenly used as exact timestamp evidence.
- Added regression test with synthetic auxiliary step bytes matching production traces (`[0x0A, 0x0C, 0x08, 0xA7, 0x8C, 0xE6, 0xD4, 0x06, 0x10, 0xC0, 0xD5, 0xA0, 0xCB, 0x03, 0x18, 0x05, 0xD2, 0x01, 0x00]`), and tests for duplicate bot IDs on UUID-less rows in either row order.
- Updated documentation in `docs/antigravity.md` and `CHANGELOG.md`.

#### Before Fix
Token history was withheld across the entire machine with coverage marked incomplete:
```text
Antigravity Token History
Local token history is unavailable or incomplete.
Local token history · dollar costs unavailable
```

#### After Fix
Local SQLite scan recovers full token history across all conversations:
```text
➜  CodexBar git:(fix/antigravity-step-uuid-withholding) swift run CodexBarCLI cost --provider antigravity
[1/1] Planning build
Building for debugging...
[5/5] Emitting module CodexBarCLI
Build of product 'CodexBarCLI' complete! (8.86s)
Antigravity Token History
Today: 73M tokens
Last 30 days: 1.1B tokens
Local token history · dollar costs unavailable
```

### Validation
```bash
swift test --filter AntigravityLocalReaderTests
swift test --filter AntigravityBotIDValidationTests
swift test --filter Antigravity
make check
```
All 403 Antigravity-focused tests and `make check` (SwiftFormat, SwiftLint, Oxlint) pass cleanly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 0.56.8 — Unreleased
 
+### Fixed
+- Antigravity: skip auxiliary and lifecycle steps lacking a step UUID during local token scans instead of invalidating the entire database history.
+
 ## 0.56.7 — 2026-09-06
 
 ### Highlights
--- a/Sources/CodexBarCore/Providers/Antigravity/AntigravityLocalSQLite.swift
+++ b/Sources/CodexBarCore/Providers/Antigravity/AntigravityLocalSQLite.swift
@@ -331,18 +331,18 @@ extension AntigravityLocalReader {
                 rowsAreValid = false
                 continue
             }
+            if let botID = parsed.botID {
+                self.recordExactBotID(
+                    botID,
+                    stepUUID: parsed.stepUUID,
+                    timestampMs: parsed.timestampMs,
+                    exact: &exactByBotID,
+                    ambiguous: &ambiguousBotIDs)
+            }
             // Auxiliary/lifecycle steps in the steps table may legitimately lack a stepUUID.
             // Skip them rather than failing the entire database scan.
             guard let stepUUID = parsed.stepUUID, !stepUUID.isEmpty else {
                 continue
-            }
-            if let botID = parsed.botID {
-                self.recordExactBotID(
-                    botID,
-                    stepUUID: stepUUID,
-                    timestampMs: parsed.timestampMs,
-                    exact: &exactByBotID,
-                    ambiguous: &ambiguousBotIDs)
             }
             if neededStepUUIDCounts[stepUUID] != nil {
                 stepTimestamps[stepUUID, default: []].append(StepTimestamp(
@@ -372,13 +372,13 @@ extension AntigravityLocalReader {
 
     private static func recordExactBotID(
         _ botID: String,
-        stepUUID: String,
+        stepUUID: String?,
         timestampMs: Int64?,
         exact: inout [String: ExactStepTimestamp],
         ambiguous: inout Set<String>)
     {
         guard !ambiguous.contains(botID) else { return }
-        guard let timestampMs else {
+        guard let timestampMs, let stepUUID, !stepUUID.isEmpty else {
             exact.removeValue(forKey: botID)
             ambiguous.insert(botID)
             return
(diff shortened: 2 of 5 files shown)


horosin:
@clawsweeper re-review

steipete (changes requested):
The bot-ID ordering fix addresses the earlier finding, and the rebuilt CLI output is useful availability proof. One date-attribution boundary still needs to be established before landing this skip.

`parseStepMetadata` does not identify a lifecycle step type. The new guard skips every syntactically valid record without a UUID, including the unknown-field-only record in the existing regression. With two pending generations sharing a UUID, one identified step timestamp, and one unidentified occurrence, skipping the latter leaves the single-timestamp fallback free to assign the identified timestamp to both generations. `recoveredCount` still reaches the expected count, so it does not prove that attribution was valid. The changed regression now explicitly accepts that case.

Please restrict the skip to positively identified auxiliary records while preserving unidentified occurrence evidence, or provide a producer contract establishing that UUID-less records cannot represent relevant generation occurrences. Keep coverage for unknown-field-only and timestamp-bearing unidentified records alongside the harmless lifecycle fixture. Restored total token counts alone do not verify the dates assigned to those generations.

Reviewed head: `e305c32673e9ef74d42c72e4dcc631778070676f`. This is a source/test-contract concern; I am not claiming to have inspected the private history used in the CLI demonstration.

steipete:
Thanks @horosin for identifying this and providing the live CLI trace. The reported UUID-less bookkeeping-row failure and bot-ID ambiguity handling are now addressed on main by #3462 (a1391fc3016d2c1a278eb117a1d77716e512cf3b).

The landed fix also retains ambiguous positional slots and withholds shared timestamps for reused UUIDs when unidentified rows leave the evidence incomplete. It passed 406 focused Antigravity tests, the full suite, and CI. I compared this PR's final diff with that implementation; closing this overlapping proposal in favor of the landed fix.

horosin:
Thank you @steipete! Really appreciate active maintaining of the project

PR #3452, opened 2026-09-06
fix(claude): preserve historical account usage and explicit repair

QuantIntellect:
Claude-swap can report historical usage after a fetch or credential failure, but CodexBar dropped those measurements, spend and disabled-slot status. This repair renders the source's last-good numbers with their original capture age, beside the current diagnostic. It also exposes the existing explicit re-authentication transaction when the active slot reports foreign credentials.

Historical provenance now travels through parsing, quota retention, cache reload, account context, full and compact menu cards, and terminal cards. Historical quotas do not feed the menu icon, ready-account recommendations, or brief terminal warnings/reset summaries. Explicit source last-good measurements take precedence over locally inferred at-limit retention; the latter keeps its existing identity and reset guards. Malformed optional fields cannot suppress valid live windows. Existing cache payloads remain readable.

The header gives historical age its own line so a long error cannot truncate it, and reserves space for actions. **Re-authenticate** uses the same validated `cswap --switch-to <slot> --json` command as existing explicit activation. Active-segment selection remains inspection-only; source-chosen strategies remain held separately. `unavailable` now states the known outcome without inventing a polling-failure cause. Account aliases remain hidden by Hide Personal Info. A typed plan override also prevents an absent action from falling back to the adapter login-method name as a subscription badge.

Thanks @QuantIntellect for the original implementation. This maintainer repair preserves the contributor's human credit and integrates the current account-model architecture.

## Validation

Candidate: `cad7a26889b12b24fb5137e517db33f6f5b0b8dd`. Isolated Codex review is clean through P2; `make check` reports zero violations across 2,268 files. The final build passes, 606 focused tests / 78 suites, including the architecture gate, pass in 14.563s, and two Developer ID signed native capture tests pass in 4.435s. [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34751675779) passed all nine checks. The full `make test` run passed 78 groups, then failed the architecture allowlist check because this change moved existing source anchors. Its 18 line coordinates are now corrected and independently reviewed; anchor text, provider sets, fingerprints and all detector assertions stay intact. There were no timeouts or retries. Groups 80–104 were not reached. The new exact-head full run passed all **1,128 selections / 104 groups first-pass in 936.4s**, with zero failures, retries or timeouts (62.2s discovery, 874.1s execution). All nine exact-head CI checks passed. All ten repeated native captures are byte-identical to the published candidate images; neither production code nor the gate assertions changed in this follow-up. The complete merge gates are now satisfied on this head. Baseline regressions reproduced missing bars, spend and capture-age advancement. Signed native captures use an owned `cswap` fixture, empty credential scope, dictionary defaults and isolated storage; the production reader, projection, account model and SwiftUI views run against synthetic data. A separate production-store test verifies active-slot repair arguments and reconciliation.

The first candidate's captures exposed the long-error/action clipping described above; final native assertions recognize the rendered pixels on-device, in addition to model and command assertions. No real provider, Keychain or browser-cookie probe was run.

An earlier check run timed out while tearing down an unrelated process-fixture sentinel. The same focused fixture passed unchanged (2.368s), and the final complete check passed. No assertion, fixture corpus, timeout, retry policy or check was weakened.

```sh
swift build --build-tests
make check
CODEXBAR_ALLOW_TEST_KEYCHAIN_ACCESS=0 make test-skip-build \
  FILTER='ProviderArchitectureGatekeeper|ClaudeSwap|AccountMenuLayoutPlanner|MenuCard|CLICards|UsageStoreMenuCardModel|ClaudeOAuthRotationError'
CODEXBAR_ALLOW_TEST_KEYCHAIN_ACCESS=0 \
  CODEXBAR_CLAUDE_SWAP_RICH_SCREENSHOT_DIR=/tmp/codexbar-proof \
  make test-skip-build FILTER=ClaudeSwapRichUsageNativeProofTests
```

Native proof uses macOS 27 / Apple Swift 6.4 with a four-job native build; CI covers the pinned supported toolchains. All captures contain synthetic data and were inspected before upload. The complete diagnostics remain available through the existing copy control when the bounded error text truncates; the historical timestamp and action remain visible.

Four baseline and ten final candidate captures are attached below.

## Native before / after

Same synthetic adapter payload, fixed clock, 340-point card width. Source parsing previously dropped the historical numbers and spend.

| Appearance | Before | After |
|---|---|---|
| Light | Before: Claude account cards, Light | After: dated Claude account usage, Light |
| Dark | Before: Claude account cards, Dark | After: dated Claude account usage, Dark |
| Hide Personal Info | Before: Claude account cards, Hide Personal Info | After: dated Claude account usage, Hide Personal Info |
| Highlighted | Before: Claude account cards, Highlighted | After: dated Claude account usage, Highlighted |

<details>
<summary>Compact account rows and explicit active-account repair</summary>

Compact historical accounts, Light

Compact historical accounts, Dark

Compact historical accounts, Hide Personal Info

Compact historical accounts, Highlighted

Explicit active-account re-authentication, light

Explicit active-account re-authentication, dark

</details>


QuantIntellect:
## Real behavior proof

All output below is from a build of this branch, against a real **claude-swap 0.26.0** install with three accounts. Emails are redacted; nothing else is edited.

### 1. Cards are dated by claude-swap's own measurement, not the refresh tick

`CodexBarCLI dashboard --identity redacted`, reformatted for readability:

```
generatedAt (refresh tick): 2026-09-07T08:56:31Z

claude-swap:1    active=True   updatedAt=2026-09-07T08:55:03Z   session=5%  weekly=1%  fable=0%
claude-swap:2    active=False  updatedAt=2026-09-07T08:54:23Z   session=25% weekly=30% fable=22%
claude-swap:3    active=False  updatedAt=2026-09-07T08:54:23Z   session=36% weekly=34% fable=59%
```

Three distinct `updatedAt` values, all older than `generatedAt`, because claude-swap serves a per-account cache and had last fetched those rows 88s and 128s earlier. Before this change every row was stamped with the refresh tick, so all three read "updated just now" regardless of how old the measurement was. This is the `usageFetchedAt` / `usageAgeSeconds` parsing in commit 1.

### 2. `switched: false` really does come back with a non-null destination

A live `--strategy best` run, which is the exact argument array the new submenu executes:

```console
$ cswap --switch --strategy best --json
{
  "schemaVersion": 1,
  "switched": false,
  "from": { "number": 1, "email": "<redacted>" },
  "to":   { "number": 1, "email": "<redacted>" },
  "strategy": "best",
  "reason": "already-best",
  "message": "Already on the account with the most remaining quota (Account-1).",
  "warnings": []
}
```

This is the shape the review flagged: `switched: false` **with a non-null `to`**. `already-best` is benign, so it confirms on the card. The blocked reasons (`candidates-exhausted`, `usage-unavailable`, `no-valid-target`, `only-one-account`, `unmanaged-account`) return the same non-null shape — see the `_switch_noop` call sites in [cswap 0.26.0 `switcher.py`](https://github.com/realiti4/claude-swap/blob/v0.26.0/src/claude_swap/switcher.py) — which is why classification is by `reason` and not by a non-null target. `ClaudeSwapSwitchResultClassifier` covers every one of those reasons, and treats an unrecognized future reason as blocked rather than confirming it as success.

### 3. What changed since the last review

- **Blocked outcomes no longer read as success.** Previously every `switched: false` became "Already Active".
- **Strategy errors have somewhere to render.** They carry no requested slot, so `lastErrorAccountID` was nil and the card renderer never matched; they now render on the section beside the switch actions.
- **Retention no longer promotes stale data to live.** An `unavailable` row retaining a last-known fallback kept the `.live` classification and slipped past the menu-bar exclusion. Provenance now survives both retention and the `RetainedUsageStore` disk round-trip. The added test fails without the fix.
- **A completed switch now shows up immediately.** The completion handler awaited the full ambient Claude refresh (OAuth + CLI + web) before redrawing, so the cards kept rendering the pre-switch state for seconds and the click looked inert. claude-swap is re-listed first; the ambient reconcile still runs, afterwards.

### Commands run

```
swift build      # clean
make check       # 0 violations in 2137 files, locale parity across 23 locales
make test        # full sharded suite, 1032 selections, 0 failed groups
swift test --filter ClaudeSwap   # 160 tests, 14 suites
```

QuantIntellect:
@clawsweeper re-review

All three findings from the previous review are addressed, and real behavior proof is posted above.

- **Surface strategy-switch errors without requiring an account ID** — errors that belong to no single account card now render on the claude-swap section beside the switch actions, instead of being stored with a nil `lastErrorAccountID` that no card renderer could ever match.
- **Preserve last-known provenance through retention and reload** — retention carries `usesLastKnownUsage` forward instead of reclassifying a retained fallback as live, and `ClaudeSwapRetainedUsageStore.Record` persists it so a relaunch cannot launder it either. The added test fails without the fix.
- **Classify blocked strategy outcomes before confirming success** — classification is now by `reason`, not by a non-null target. Benign `already-active`/`already-best` confirm on the card; every other non-switch, including a reason a future claude-swap release adds, is reported as a failure with its explanation.

Also fixed since the last revision: a completed switch was invisible for seconds because the completion handler awaited the full ambient Claude refresh before redrawing the cards.

On the merge-risk decision: the split you recommended is already in place. Commit 1 is the display work and stands alone inside the accepted Phase 1 contract; the source-chosen switching is commits 2, 5 and 6 plus the "Proposed: source-chosen switch targets" section, and can be dropped together without touching commit 1.

steipete:
While reviewing this proposal, we isolated and landed the failed-switch feedback and canceled-read fixes in #3580 (816d495e8ba4e3ac2ab0ace7ef8c6e83bf7d6108), preserving @QuantIntellect’s contributor credit. Known activation errors now appear before ambient refresh finishes, with the existing transaction guard held through reconciliation. The broader usage-display and source-chosen target proposal remains here.

The focused repair passed controlled command/refresh regressions, a full 1,091-selection serial suite, lint, independent review, and all hosted CI checks. Details and proof are recorded in #3580.

Issue #3209, opened 2026-08-26
I am using v0.55.0 - now after using it for a whole day, the token/spend chart does not refresh. no error message. it shows "updated just now" after refreshing, but only the session/weekly use updates. (I am not sure if I needed to create new issue, as the previous one was already closed)


vinschger:
I am using v0.55.0 - now after using it for a whole day, the token/spend chart does not refresh. no error message. it shows "updated just now" after refreshing, but only the session/weekly use updates.

_Ursprünglich gepostet von @vinschger in https://github.com/steipete/CodexBar/issues/3176#issuecomment-5413588985_

vinschger:
I have noticed that not only Claude but also ChatGPT usage did not refresh the token spend chart since 22nd of August, although the session/weekly usage refreshes perfectly. I do not understand why. 

<img width="1228" height="836" alt="Image" src="https://github.com/user-attachments/assets/65dd5e9e-d190-4d4e-af57-a2293c30bb50" />

steipete:
Thanks, the screenshot helps: this is the regular provider-menu cost chart and its submenu, not the separate Usage & Spend dashboard. “Updated just now” refers to the quota snapshot, so it does not prove that local token/cost history refreshed.

Could you confirm the exact installed version, whether global “Track costs” is enabled, and whether Codex's “Local session cost estimates” option is enabled? Also, after a provider-specific Refresh and closing/reopening this chart submenu, does its latest day move forward for either Codex or Claude?

We have identified a separate Codex publication bug in the supported local-estimates-only configuration, and are repairing scan fairness for busy Codex histories. Neither establishes the cause of your Claude report, so this issue remains open. Please don't post credentials or session contents; the settings above are enough for the next diagnostic step.

steipete:
The separate Codex local-estimates-only publication bug is now fixed on main in [#3230](https://github.com/steipete/CodexBar/pull/3230), commit [3992cbc8d834](https://github.com/steipete/CodexBar/commit/3992cbc8d834a567092d853ab520825cc8460f15). It applies when Codex's local estimates are enabled while global cost tracking is off: completed scans were repeatedly rejected as stale. The focused regressions, full local suite and exact-head CI passed. The busy-history scan fairness repair in #3207 has also landed.

Neither change establishes the cause of the Claude symptom, so this issue stays open. The version/toggle/refresh-and-reopen details requested above are still the next useful diagnostic step. These fixes are on main; this is not a new release announcement.

[...]

PR #3464, opened 2026-09-07
fix(claude): discard cached rows from replaced transcripts

steipete:
Claude transcript inventory detected atomic file replacement, but the scanner discarded file identity and treated a larger replacement as an append. A normal refresh could merge stale rows with the new suffix: the baseline fixture reported 1,017 input tokens where a full rescan correctly reported 24, across warm and cold loads.

Match persisted device/inode identity before reusing rows or a parse offset. Store identities atomically with rows in the Claude/Vertex JSON artifact, preserving its existing flat format. Missing legacy identities trigger one rebuild before reuse, including during refresh debounce. Consolidate full and incremental parsing into one path and prune identities alongside removed files.

Review caught a cross-provider compatibility problem in the initial shared-model approach. The final implementation leaves both the shared cache model and generated Codex parser fingerprint byte-identical to main and v0.56.7. The generator runs unchanged; no allowlist, hand-edited fingerprint, or unrelated Codex/Pi cache reset is introduced.

CI initially caught one timestamp-test adapter line omitted from the commit despite being present in the passing local tree. That line is now included, its four focused timestamp tests pass, and the committed tree matches the complete local candidate.

All 118 focused tests across four suites pass: warm/cold larger replacements, same-size/same-mtime replacement, actual incremental append, flat legacy-cache upgrade within debounce, empty parsed rows and pruning, released Codex row/checkpoint retention, and released Pi pricing-key reuse. `make check` passes and independent review is clean. The full `make test` suite also passes against the final isolated-cache revision: 1,028 selections across 86 groups, no retries.

Production code grows by 14 lines after the parsing consolidation. This is funded by nine remaining lines from separate cleanup #3465 and five of the eleven remaining lines from #3456; #3465 landed as debe5cdfbc7d6e2686c60f1b281575f14eca26a2. Changelog and Claude documentation are updated, including the literal-directory meaning of `CLAUDE_CONFIG_DIR`.

Related investigation: #3209 remains open. These filesystem fixtures prove transcript replacement handling, not that this is the reporter's specific cause. Arbitrary same-inode prefix rewrites are outside this fix.


steipete:
The reviewed compatibility blocker is resolved by keeping identity metadata inside the Claude/Vertex artifact. The generated parser fingerprint remains identical to the release, and the final full suite passes with 1,028 selections across 86 groups and no retries, alongside the 118 focused cache/upgrade tests.

The remaining prerequisite is also complete: #3465 landed as debe5cdfbc7d6e2686c60f1b281575f14eca26a2. Nine of its remaining cleanup lines plus five previously unspent lines from #3456 fund this PR's 14-line production increase.

vinschger:
in codexbar 0.58.0 the problem can still be observed

Issue #2871, opened 2026-08-11
bug(z.ai) Reset time shows 10 hrs window instead of 5 hours (again in 0.49.2)


carolitascl:
Reset time shows 10 hrs window instead of 5 hours

This issue was fixed but came back in version 0.49.2

carolitascl:
Codebar 

<img width="817" height="190" alt="Image" src="https://github.com/user-attachments/assets/2eacb402-5aa5-473a-8ed3-f02a259d435a" />

ZCode - Offcial app from z.ai 

<img width="918" height="323" alt="Image" src="https://github.com/user-attachments/assets/00450c85-4fe7-4eb3-85f6-248a7e4270b8" />

steipete:
Thanks—the side-by-side screenshots now prove the visible defect: the five-hour percentage agrees, but CodexBar's reset is about four hours later than ZCode's and even falls outside a possible five-hour window.

Before changing timestamp semantics, please capture these at the same refresh time:

1. The redacted `data.limits` array from the z.ai monitor response
2. Your macOS timezone identifier (for example `America/New_York`)
3. z.ai region/scope and whether you configured an endpoint override
4. The simultaneous ZCode and CodexBar reset clocks

Please remove the bearer token, Authorization header, organization/project IDs, request URL secrets, and unrelated response fields. The essential fields are the limit type, `unit`, `number`, usage/remaining fields, and exact `nextResetTime`.

If the raw epoch maps to CodexBar's later time, we need the official ZCode conversion rule before correcting it. If that rule remains unavailable, the safe mitigation is to suppress a five-hour reset that is more than five hours plus small clock skew in the future—not guess a timezone offset.

carolitascl:
<img width="807" height="330" alt="Image" src="https://github.com/user-attachments/assets/05ac4852-20a2-4933-a581-df6ed7b21162" />

<img width="353" height="707" alt="Image" src="https://github.com/user-attachments/assets/8177b48d-08ec-4c5e-b795-74241b6558a8" />

<img width="509" height="387" alt="Image" src="https://github.com/user-attachments/assets/bad1f656-5f43-4dcd-831c-9f84a26b5703" />

>  **curl -H "Authorization: Bearer {API_KEY}" -H "Accept: application/json" "https://api.z.ai/api/monitor/usage/quota/limit"**

steipete:
[...]

Issue #2522, opened 2026-07-31
z.ai / GLM Coding Plan V3 (points-based, launched 2026-07-30): confirm the zai provider handles the new plan's usage shape


LeoLin990405:
## Summary

Zhipu launched **GLM Coding Plan V3** on 2026-07-30 — a token-metered ("points" / 积分, token-consumption as the basis) revamp with new tiers (Lite ¥49 / Pro ¥149 / Max ¥469). Existing subscribers keep their current (V2) plan until expiry; **new subscribers get V3**. Filing a heads-up because the `zai` provider parses a specific usage shape, and V3 may return a different one.

## What still works today (existing V2 account, BigModel CN)

On an existing BigModel CN Coding Plan, `api/monitor/usage/quota/limit` returns three **windowed** limits, and CodexBar renders all three with no error:

```jsonc
{"data":{"limits":[
  {"type":"TOKENS_LIMIT","unit":3,"number":5,"percentage":..,...},   // 5-hour window
  {"type":"TOKENS_LIMIT","unit":6,"number":1,"percentage":..,...},   // weekly window
  {"type":"TIME_LIMIT","unit":5,"number":1,"percentage":..,          // NEW monthly "MCP" quota
   "usageDetails":[{"modelCode":"search-prime"},{"modelCode":"web-reader"},{"modelCode":"zread"}]}
],"level":"max"}}
```

So the current parser (`TOKENS_LIMIT` / `TIME_LIMIT`) handles this shape fine — 5-hour / weekly / monthly-MCP all render. (The monthly/MCP window's missing `windowMinutes` is already addressed by #2514.)

## What needs a V3 account to confirm

V3 is described as **points/积分-based** (token consumption as the basis). Two things I can't verify — I'm on an existing V2 plan and won't cancel + repurchase just to get a V3 account:

1. Does V3's `usage/quota/limit` return the **same windowed `TOKENS_LIMIT`/`TIME_LIMIT` shape**, or a different **points-pool** shape the current parser doesn't handle?
2. If the shape differs, does the `zai` card still render, or go blank / wrong?

This is the same failure mode as **StepFun's #2491 / #2492**: a plan revamp that moved to a credit/token pool silently broke the windowed renderer until the classifier was made shape-driven. Flagging proactively so it doesn't surprise a V3 subscriber.

## Ask

[...]

PR #3467, opened 2026-09-07
fix(zai): contain invalid optional model analytics

steipete:
Optional z.ai model analytics could invalidate an otherwise valid quota snapshot when the response exceeded native chart limits, contained an invalid label, or overflowed an aggregate. The existing JavaScript catch ran before native snapshot validation, so it did not contain these failures.

Validate optional analytics before attaching the section and reuse normalized positive token counts. A pure native `ctx.isDetailLabel` predicate shares the exact existing Unicode grapheme and whitespace rules across QuickJS and JavaScriptCore. Invalid optional sections are omitted while primary and weekly quotas remain available; chart periods and model names are not truncated.

Both native engines failed the 121-point baseline while the 120-point control passed. All 54 focused z.ai/runtime tests pass, covering chart limits, invalid whitespace and labels, finite aggregates, non-string inputs, and 120-character decomposed/emoji labels. `make check` and the full `make test` suite pass (1,028 selections across 86 groups, no retries). Independent review found no actionable P0–P2 findings. No live account probes were used.

The config-watcher dependency #3469 has landed. This branch is integrated with main, and its diff is now standalone. The earlier combined full suite passed all 1,028 selections across 86 groups without retries; both macOS shards and the glibc CLI matrix passed on the final commit. The static Linux SDK installer crashed in upstream Swift/OpenSSL code after downloading the SDK, before project compilation; the failed jobs passed on retry on the same commit. Final CI run 34112526311 is green, including both macOS shards and all Linux builds.

The standalone z.ai change grows production code by 14 lines, funded by the separate 23-line localization cleanup in #3465, landed as debe5cdfbc7d6e2686c60f1b281575f14eca26a2. Documentation and changelog are updated. Investigation was adjacent to #2522 and #2871; their original reset/API-version symptoms are not claimed fixed.


steipete:
The production-growth prerequisite is satisfied: #3465 landed as debe5cdfbc7d6e2686c60f1b281575f14eca26a2, removing 23 production lines. Fourteen fund this fix.

The label predicate is an additive host API, shipped together with the bundled z.ai plugin that uses it. Existing third-party plugins retain their current API behavior. A third-party plugin choosing to adopt a newly added helper must target a supporting host or feature-detect it, as with any additive JavaScript API; this change does not require existing plugins to adopt it. Both engines exercise the same native validator, and review found no blocking correctness defect.

PR #3474, opened 2026-09-07
Show Venice subscription credits from the Web source

audreyt:
Add an explicit Web source for Venice subscription credits, spending this cycle, monthly refill, bank cap, and next refill. Auto/API retain API-key behavior. Saved API accounts remain stored but are passive for implicit Web requests; explicit account overrides still use the API. Cookie source Off prevents all credential loading and requests, and canceled refreshes never begin a browser scan.

Web credit details keep the available balance distinct from the monthly refill. Spending above one refill does not imply an exhausted quota when banked credits remain, and refill dates are not labeled as subscription renewals. Shared cookie settings and existing detail rendering replace duplicated settings plumbing and provider-specific menu/widget branches. Chrome is the default import source; expired, missing-token, and unsubscribed profiles can fall through to a valid profile.

A signed candidate web fetch was attempted, but no usable Venice browser session was available. Authenticated session compatibility remains unverified locally.

Synthetic fixtures cover parser/HTTP behavior, account authority in app and CLI, complete settings assembly, cancellation, multi-profile fallback, and rollover presentation. The before image shows the empty card before web data is available; the after image uses synthetic subscription claims in the real shared card.

Validation: `make test` passed 1,188 selections in 108 groups without failures, retries, or timeouts. Final main integration passed focused app/CLI, settings, presentation, history, and architecture tests, `make check`, and isolated review. Signed native proof was inspected in light and dark appearances. Final-head [CI 35335124847](https://github.com/steipete/CodexBar/actions/runs/35335124847) passed lint, Linux builds, and both macOS shards.

| Before | After |
| --- | --- |
| Before | After |


audreyt:
@clawsweeper re-review

Addressed all four findings: token-account rejection before import, multi-session fallback, Linux auto exemption with CLI preflight coverage, and full app exposure (usage-source setting, picker, sourceMode routing) plus presentation label wiring and timeout threading. Details in the updated PR body.

audreyt:
@clawsweeper re-review

Addressed the two new findings: selected token accounts now resolve to the API script at the credential-adapter seam (no ambient fetch, no mislabeling; covered), and the Preferences detail shows the live source label. Details in the updated PR body.

audreyt:
Live proof (head f3de3df13): ran the real `VeniceWebUsageFetcher.fetchUsage` against the live `outerface.venice.ai/api/user/session` endpoint using the signed-in venice.ai session cookie. HTTP 200, valid JWT, parsed snapshot (numbers only, no secrets):

- Used this cycle: 15,349 / Monthly allowance: 22,500 (used 68.2%)
- Subscription remaining: 7,151 | Bank cap: 67,500 | Total credits: 7,651 | Plan: MAX

Scratch live test passed (`liveSessionReturnsRealQuota`, worktree-only, removed afterwards); the JWT `bundledCreditsUsage` claim carries all fields the parser requires. No fixture involved — live shape matches the parser contract exactly.

Caveat for QA: the session came from the user's relayed browser, not the local Chrome profile, so `--source web` cookie auto-import on this machine would still report missing-credentials until venice.ai is signed in via local Chrome. The fetcher + parser themselves are proven live.

audreyt:
Follow-up (8b46ff6ee): Venice web import now tries Chrome, then Brave. Rationale: the live session used for the proof above lives in the user's Brave, and local-Chrome-only import would report missing-credentials for Brave-only users. Chrome stays first per repo cookie policy; Brave-second matches the Alibaba/Augment precedent. Adds `VeniceBrowserOrderTests`; focused Venice batch (33 tests) green, `make check` clean.

audreyt:
Follow-up (eed155514): the web strategy now honors a stored manual cookie header (Abacus pattern) instead of forcing browser import. Motivation: a valid session cookie can exist in config while the browser profile is unreadable (sandboxed relay browsers, TCC-blocked profiles). Adds a `VeniceProviderSettings` snapshot section wired through the descriptor; browser import (Chrome, then Brave) remains the path when the cookie source is auto. Verified live: stored header returns HTTP 200 with real quota through the new path. Venice batch 36/36 green, `make check` clean. Note: session cookies rotate, so a stale manual header surfaces the sign-in error until refreshed.

audreyt:
Follow-up (4f8100d52): fix launch trap. Registering the Venice settings section without an app-side snapshot contribution tripped the startup assembly precondition (`Missing settings snapshot section`) and SIGTRAPed on launch; the CLI was unaffected. Mirrors the Abacus wiring (SettingsStore snapshot helper + implementation override) and adds a regression test pinning that the registration accepts the contribution. Venice batch 37/37 green, `make check` clean, packaged app launch-verified.

PR #3485, opened 2026-09-08
fix(claude): preserve claude-swap measurement timestamps

steipete:
Repeated claude-swap refreshes displayed cached measurements as freshly updated because the parser discarded `usageFetchedAt` and the projection always used the current clock. Preserve the optional source timestamp using the existing parser; older adapters and malformed optional timestamps keep their existing fallback without dropping valid usage windows.

Extracted from #3452, with credit to @QuantIntellect. Its remaining last-known usage display and source-chosen account switching stay separate. This changes no adapter arguments, credential access, cache schema, or account activation behavior.

Validation: the regression fails on the baseline and passes after the fix, including fractional timestamps, repeated refreshes, exhausted-window retention, and malformed optional fields. Focused ClaudeSwap suites pass. `make check` passes using macOS system Bash (Homebrew Bash stalled in the existing signing fixture's here-document). Independent P0–P2 review is clean. The full suite passed all 1,032 selections across 86 groups on the first attempt, with no retries or timeouts. Exact-head CI passes all jobs, including both macOS shards, all three Linux builds, lint and the aggregate gate.

Live proof: the actual built CLI `dashboard --identity redacted` with an isolated synthetic adapter reproduced both baseline account timestamps being replaced by the poll time. The candidate preserves each distinct source measurement timestamp and the same quota values. The Developer ID–signed app also launched in an isolated synthetic profile and opened its real Settings menu through native automation. No real provider credentials, browser sessions, or personal history were used.

The changelog entry will be carried in #3488 to keep sibling branches independent.

### DIFF
--- a/Sources/CodexBarCore/Providers/Claude/ClaudeSwap/ClaudeSwapAccountList.swift
+++ b/Sources/CodexBarCore/Providers/Claude/ClaudeSwap/ClaudeSwapAccountList.swift
@@ -7,7 +7,7 @@ import Foundation
 /// are decoded: slot number, display email, display-only `organizationName`,
 /// optional display-only `alias`, active state, usage status, and the
 /// 5-hour/7-day windows and optional model-scoped weekly windows (percent +
-/// reset timestamp). Everything else in the payload is ignored; unknown schema
+/// reset timestamp), and optional usage measurement time. Everything else in the payload is ignored; unknown schema
 /// versions and partial top-level shapes are rejected.
 public struct ClaudeSwapAccountList: Equatable, Sendable {
     public let activeAccountNumber: Int?
@@ -34,6 +34,8 @@ public struct ClaudeSwapAccountRow: Equatable, Sendable {
     public let fiveHour: ClaudeSwapUsageWindow?
     public let sevenDay: ClaudeSwapUsageWindow?
     public let scoped: [ClaudeSwapScopedUsageWindow]
+    /// The adapter's measurement time, which can precede this list refresh.
+    public let usageFetchedAt: Date?
 
     public init(
         number: Int,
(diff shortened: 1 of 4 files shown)


steipete:
Validation on `d519d1a225bbd55b346040a3475889cdf4fb337c`:

- `make test`: all 1,032 selections / 86 groups passed on the first attempt, with zero failures, retries, or timeouts (1,130.8 seconds including discovery).
- `make check`: passed, zero SwiftLint violations. The existing signing fixture hung inside Homebrew Bash `heredoc_write`; a fresh private TMPDIR reproduced it, and the identical checks passed with macOS system Bash selected through a task-local PATH.
- Focused ClaudeSwap tests pass. The new raw-JSON regression failed before the fix, then passed with fractional timestamp, repeated-refresh, retained exhausted-window, and malformed optional-field coverage.
- Independent local and committed-branch autoreview through P2: no actionable findings.

Live built-binary proof used only an owned synthetic `cswap --list --json` fixture and isolated HOME/config/cache roots. Baseline `dashboard --identity redacted` replaced both accounts’ measurement dates with its poll time. The candidate kept `2026-09-07T08:54:23Z` and `2026-09-07T08:55:03Z`, retaining the same 25% and 50% usage. The Developer ID-signed app also launched in an isolated synthetic profile; native automation opened its real Settings menu and verified the General window before quitting. That is an app startup/navigation smoke check; the changed timestamp behavior was proven through the shared production CLI pipeline.

[Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34175168692) passed every job, including both macOS shards, all three Linux builds, lint and the aggregate gate. The current PR head still matches the tested commit and is mergeable. This PR is land-ready; merge authorization remains with the orchestrator. The changelog is collected in #3488, which should land after this PR and the dependency update #3487. No real credentials, browser cookies or personal history were accessed for these proofs.

Exact-head receipt tightened: after the unrelated main fast-forward, the committed d519d1a225bbd55b346040a3475889cdf4fb337c was rebuilt and the same built CLI and signed-app Settings smoke were repeated successfully.

PR #3486, opened 2026-09-08
fix: reject unrepresentable numeric usage safely

steipete:
Usage parsers could admit an unrepresentable rounded integer because Double(Int.max) rounds upward. Replace the trapping conversions in MiMo, Pi/OMP, OpenCodex, and Bedrock with failable conversion after each source's existing rounding step. Oversized finite subprocess timeouts now saturate safely.

Preserve valid neighboring fields, MiMo/OpenCodex exact integer payloads, Pi/Bedrock nearest rounding, and MiMo/OpenCodex truncation. OpenCodex's established negative-fraction fallback remains zero. Version the OpenCodex parse cursor so values clamped by an older parser are reparsed once from unchanged logs; subsequent reads reuse the corrected cache. The generation lives in the cursor transaction so an older writer cannot leave a misleading current-generation marker behind.

The refactor removes 16 production lines. Includes boundary, valid-field, rounding, and legacy-cache regressions, plus changelog and parsing documentation.

Validation: the added negative-fraction regression failed before correction, and the legacy-cache regression failed for both absent and old generations before invalidation was added. The repaired focused suites passed. Full make test passed all 1,035 selections across 87 groups on the first pass, with no retries or timeouts. make check passed. After integrating latest main, 145 tests in 9 affected suites passed; the numeric production code was unchanged. Independent P0–P2 review passed on the final integration. All checks passed for commit 9264b28cd9ec2f1cdff10a5888f9197467c3e9d6: https://github.com/steipete/CodexBar/actions/runs/34197381254.

This patch fixes numeric conversion and reuse of older parsed rows. The separate, pre-existing unchecked additions in OpenCodex report aggregation are addressed separately in #3501; this PR does not claim comprehensive aggregate-overflow protection.


steipete:
Checking the landing path for this completed fix. The PR still targets the old CLI-login branch even though #3484 is on main. I am retargeting it to main, preserving the numeric changes and their validation, then checking the synchronized candidate before landing.

PR #3490, opened 2026-09-08
fix(codex): resume interrupted warm cost scans

steipete:
A timed warm Codex cost refresh could validate one file, discard the remaining current-window cursor, and mark the cache pending. The next wider request then rediscovered and requeued already-validated files. On large histories this repeats catch-up work and delays spend publication.

Save the unvalidated candidates before the existing lookback completion logic removes successful work. Preserve completed discovery and use the established same-ending-day scan range when compatible timed requests alternate between shorter and wider reports. Each report still renders its requested interval. The separate exact-inventory pass remains mandatory before publication.

The generated parser fingerprint changes because the shared scanner is a hash input. Native parsing and stored-row semantics are unchanged, so the immediate predecessor fingerprint is adopted through the existing compatibility mechanism, with adoption, receipt-invalidation, and zero-head-parse regression coverage.

Related to #3411. This fixes the independently reproduced timed-warm-refresh path; it deliberately does not close the broader managed-account report Peter kept open. Thanks @kesslerio for the report.

Validation:

- `make check` passed, including the regenerated parser hash.
- All 14 focused `CostUsageBoundedProgressTests` passed. They cover the lost cursor, alternating windows, preservation of older coverage, requested-range filtering, and the exact-validation gate.
- Local and committed-branch autoreview are clean through P2. The full suite exposed a byte-backlog admission regression in the first repair; limiting the new marker to non-forced timed warm scans fixed it. All 22 bounded-progress/fairness tests pass serially. Final full `make test` passed 1,033 selections across 87 groups, with one unrelated display-only settings group recovering on its single retry. [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34185523120) succeeded on `1fb4f893155f4192ed630908b80d2df5823bf15a`.
- A separately compiled executable linked to the actual production core ran against 513 synthetic session files and a real SQLite cache. Before the fix, interrupting after one validation lost the cursor; the next pass rediscovered 512 files, including the validated first file. After the fix, the interrupted pass persists the other 512 files, the next pass records zero discovery visits, and draining the queue still leaves publication pending until exact validation. All phases retain 56,430 tokens.
- Real built CLI upgrade proof: the base binary created a cache with fingerprint `9ca89383b9957b07`; the new binary adopted it as `7f00691fa96c78d1` while every owned session file was chmod 000. Both returned established coverage, 56,430 tokens, and the same USD 0.177498 total. Modes were restored and fixtures removed afterward. This proves compatible rows survive without reparsing source files.

The changelog line is collected in the batch's separate final notes PR.


steipete:
Maintainer verification complete on `1fb4f893155f4192ed630908b80d2df5823bf15a`.

- `make check` passed; local and committed-branch autoreview are clean through P2. The final full `make test` passed 1,033 selections across 87 groups; one unrelated display-only settings group recovered on its single retry. The 22 bounded-progress/fairness tests also passed.
- A production-core executable and real SQLite cache reproduced the lost cursor across 513 synthetic files. The repaired timed warm pass preserves the 512 unfinished files, resumes with zero discovery visits, and still requires exact inventory validation before publishing established coverage. Every phase retains 56,430 tokens.
- The real built CLI adopted a predecessor cache from `9ca89383b9957b07` to `7f00691fa96c78d1` while its synthetic transcript files were unreadable. Established coverage, 56,430 tokens, and USD 0.177498 remained unchanged; file modes were restored afterward.
- [CI run 34185523120](https://github.com/steipete/CodexBar/actions/runs/34185523120) succeeded on this exact head. The earlier byte-backlog fairness regression was repaired before this head; no test or validation requirement was weakened.

Land-ready. Keep #3411 open for its broader managed-account report. Changelog is consolidated in #3491; no merge was performed here.

PR #3473, opened 2026-09-07
Add Grok token refresh

audreyt:
Rotates expired Grok OIDC access tokens via the xAI refresh endpoint (pure refresh call plus a needsRefresh predicate with 5-min leeway).

Previously the parsed refresh token was never used, so expired logins just failed.

Ownership note: CodexBar deliberately does not write auth.json — the CLI remains the sole writer, so there is no cross-process write race by construction. Persistence and live fetch wiring are follow-ups once a CLI coordination contract exists.

Verified: GrokTokenRefresherTests 6/6, ProviderArchitectureGatekeeperTests 41/41, SwiftFormat/SwiftLint clean.

PR #3497, opened 2026-09-08
Recover expired Grok sessions through the CLI owner

audreyt:
## Status: draft — credential-owner coordination gate remains

Integrated successor to #3473. This is not the standalone refresher again: the production OAuth fetch path now performs recovery, and a real CLI-owned renewal plus billing recovery has been exercised.

**Do not merge yet.** Happy-path renewal and subsequent CLI reuse are proven. The credential owner's concurrent account-change/logout contract is not cleared for release. Read-side checks cannot establish persistence safety inside another process. Owner-side findings are being kept out of public reproducer material in accordance with [xAI's security reporting policy](https://github.com/xai-org/grok-build/blob/main/SECURITY.md).

## Implementation / ownership boundary

- The CLI remains the only refresh-token consumer and `auth.json` writer. CodexBar does not POST to an OIDC token endpoint or save credentials.
- An expired xAI OIDC record in the production OAuth strategy invokes `grok agent --leader stdio`, then ACP `_x.ai/auth/getBearerToken`. The shared owner is separate from the short-lived probe client.
- Check the captured account/generation before requesting recovery. Afterward, reload the CLI-written record and require an unexpired token matching the export, scope, user, email, team, principal, issuer, and client.
- Billing, bearer fallback, identity, and subscription enrichment all share the validated post-recovery capture. Team identity-only fallback retains principal metadata.
- Pasted tokens, browser cookies, legacy sessions, and explicit auth-source overrides are not silently migrated to this flow.
- Stop logging raw inbound Grok RPC responses, since the new response carries a bearer token.

The read-side account/generation checks intentionally are **not** represented as a cross-process compare-and-swap or a guarantee about the CLI's own final write.

## Real behavior proof (2026-09-08, official Grok 1.0.13 / 5e9a58528b76)

The initial live protocol probe caught a bug the original stub missed: `x.ai/auth/getBearerToken` returns `-32601`; `_x.ai/auth/getBearerToken` succeeds. The response also contains both the JSON-RPC result and the extension's nested result envelope. The updated regression failed with the old wire method, then passed after the correction.

For the live expiry exercise, only the existing auth record's expiry timestamp was temporarily marked expired under the CLI's file lock. No refresh token was copied into a second credential store. The real CLI contacted the real token service and persisted the renewed generation. The harness was guarded to restore only the expiry of an unchanged generation; no rollback was needed.

Redacted result from the newly built `CodexBarCLI usage --provider grok --source oauth --json`:

```json
{
  "exit_code": 0,
  "elapsed_seconds": 2.776,
  "access_token_rotated": true,
  "refresh_token_rotated": true,
  "same_identity": true,
  "unexpired": true,
  "source": "grok-cli-proxy",
  "usage_present": true,
  "subscription_tier_present": true
}
```

A separate subsequent connection to the installed CLI exported the same persisted bearer and successfully called `_x.ai/billing`:

```json
{
  "bearer_matches_auth_file": true,
  "cli_reused_existing_token": true,
  "cli_billing_supported": true,
  "cli_billing_result_present": true
}
```

This proves real token exchange, persistence by the owner, resumed billing, and CLI reuse after a **locally forced expiry timestamp**. It is not a claim that the server-side JWT expired naturally during the test, or that the owner-side concurrency gate passed. No tokens, account identifiers, or private transcripts are included here.

## Validation

- `swift test --filter 'GrokAccountContextTests|GrokSessionRecoveryTests'`: 11 tests passed, including the real subprocess protocol fixture, Team recovery, account-switch/removal/generation rejection, and existing account-capture cases.
- After merging current `main` (only `CHANGELOG.md` conflicted), `swift test --filter Grok`: 156 tests in 15 suites passed. The newly built production CLI also returned usage and subscription tier with no error.
- Final integrated `make check`: passed, including SwiftFormat and strict SwiftLint (0 violations across 2,142 Swift files).
- Full `make test` on the implementation commit: did not pass. The sharded run selected 1,035 selections in 87 groups and exited on a 180-second timeout in `UsageStoreCachedTokenHydrationTests`. The same selection also timed out at 180 seconds on unchanged base `0cb8c425e`, after a clean rebuild in an isolated worktree with the same test-isolation environment. This timeout therefore reproduces without the Grok changes; the full suite is not reported as green.

## Before marking ready

- Resolve the credential owner's concurrent account-change/logout coordination guarantee. Do not infer that guarantee from the read-side tests or the happy-path runtime transcript.
- Keep this PR in draft until that release gate is cleared.


steipete:
Thanks @audreyt for the CLI-owned recovery work and the protocol investigation. Recovery is still absent from the reviewed main snapshot, and this patch keeps credential writes with the CLI. The remaining blocker is the explicitly unresolved CLI contract for renewal concurrent with logout or account replacement. CodexBar’s before/after reads cannot establish the owner’s final-write behavior, and the injected tests do not exercise it. We are closing this revision without adopting it; maintainer follow-up should resume once that upstream contract is established. This review did not run live account probes.

Issue #3524, opened 2026-09-09
Codex cost: subagent/forked rollouts are counted at their final cumulative total_token_usage, inflating fan-out days 30-150x


vnnkl:
## Summary

Rollout files created by Codex Desktop subagent fan-outs (`thread_source: "subagent"`, `forked_from_id` set) inherit the **parent thread's cumulative `total_token_usage`** in their very first `token_count` event. CodexBar appears to count each file at its final cumulative total instead of the delta from that inherited baseline, so every fork re-counts its parent's entire token history. On days with many parallel forks of a long-lived parent this inflates tokens (and therefore estimated cost) by 30–150×.

My 30-day Codex estimate currently shows **$138k–170k / 175–204B tokens**; the true numbers from the same rollout files are **~$6k / 9.6B tokens**, and OpenAI's own `account/usage/read` confirms the account's *lifetime* total is 29.6B tokens with a peak day of 1.64B.

## Reproduction / evidence

Environment: macOS, CodexBar 0.54.0 and 0.57.0 (both affected; 0.57.0 slightly worse), rollouts written by Codex Desktop cli_versions 0.148.0-alpha.9 … 0.153.1.

A fork's first `token_count` event (ordinal 2, before any turn) already carries the parent's cumulative counter — note `last_token_usage` all zeros:

```json
{"type":"session_meta","payload":{"id":"01a00602-…","forked_from_id":"019faece-…","thread_source":"subagent","source":{"subagent":{"thread_spawn":{"parent_thread_id":"019faece-…","depth":1}}}}}
{"type":"event_msg","payload":{"type":"token_count","info":{"total_token_usage":{"input_tokens":3153916657,"cached_input_tokens":3113606656,"output_tokens":3991590,"total_tokens":3157908247},"last_token_usage":{"input_tokens":0,"cached_input_tokens":0,"output_tokens":0,"total_tokens":9035}}}}
```

Per-day comparison — "naive sum" = Σ over the day's files of the *final* cumulative `total_tokens`; "true delta" = Σ of (last − first) per file:

| day | files | CodexBar 0.57 claims | naive sum | true delta | OpenAI `account/usage/read` |
|---|---|---|---|---|---|
| 2026-09-04 | 62 | 30,541,006,796 | 30,380,536,401 | 821,257,479 | 1,637,928,247 |
| 2026-09-05 | 14 | 7,763,819,245 | 7,844,073,190 | 264,351,645 | 359,188,681 |
[...]

PR #3527, opened 2026-09-09
fix: respect explicit subagent history boundaries

steipete:
Codex subagent rollouts can contain an explicit child-history boundary without any child-owned records yet. The parser treated that case like a missing boundary and could fall back to an earlier delivery marker inside inherited context, attributing copied usage to the child.

Honor the explicit ordinal even for a prefix-only file. Earlier records contribute no child usage; a later append crossing the boundary contributes its genuine delta. Legacy files without an explicit boundary retain their existing independent-counter and inferred-boundary handling.

Older cached results also need correction when the source file has not changed. Replace the whitespace-only cache marker with a native parser revision, using the existing bounded reparse and row-replacement paths. The 0.57.0 SQLite row/checkpoint format is adopted without rebuilding the store.

Related to #3524. Thanks @vnnkl for the numeric fixture. The reporter subsequently confirmed that the original file is prefix-only and that the larger inflation came from stale cached rows, supporting both corrections here. The issue remains open for a separate owned-usage undercount already observed on 0.57.0; that case needs an additional fixture and scan-completion evidence.

The producer defines the boundary as the [first child-owned rollout ordinal](https://github.com/openai/codex/blob/73a1148c9c775c2a4616ce5096291740a00ed68a/codex-rs/protocol/src/protocol.rs#L3108) and [excludes earlier records from child history](https://github.com/openai/codex/blob/73a1148c9c775c2a4616ce5096291740a00ed68a/codex-rs/thread-store/src/local/thread_history_materialization.rs#L221).

Validation:

- Red/green reproduction: the minimal prefix counted 80 inherited tokens before and zero after. Direct parsing of the supplied fixture changed from 748,286 tokens/two rows to zero tokens/zero rows, without parent lookup. Raw fixture text remains outside the repository; committed fixtures are synthetic.
- 96 focused tests passed, including prefix-only records, first owned usage after ordinary/bounded append, cold/warm/forced scans, stale-revision cache replacement, 0.57.0 store compatibility, existing independent/replay accounting, and whitespace regressions.
- Cache tests verify obsolete SQL rows are removed while the stored file record and source bytes remain intact, with no store rebuild.
- `make check` passed; zero lint violations across 2,151 files.
- Independent P0–P2 review found no actionable issues.
- `make test` passed all 1,044 selections in 87 groups on the first attempt, with zero retries or timeouts.
- [Final-head hosted CI passed](https://github.com/steipete/CodexBar/actions/runs/34370412274): both macOS shards, both glibc Linux builds, musl, lint, and security checks.


steipete:
Landed as c856e4b127f2a356cfc3df7a2d5d02f79bdb11fe. Thanks @vnnkl for the fixture and follow-up discrimination of stale cached rows.

Final-head proof:

- `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 CODEXBAR_TEST_CODEX_FILE_ISOLATION=1 CODEXBAR_TEST_SESSION_FILE_ISOLATION=1 swift test --filter 'CodexSubagentOrdinalBoundaryTests|CodexCompactSubagentAccountingTests|CodexSubagentAccountingIntegrationTests|CostUsageScannerWhitespaceTests|CostUsageWhitespaceRowReplacementTests|CostUsageStoreTests'`: 96 tests passed across six suites.
- `make test`: all 1,044 selections passed in 87 groups on the first attempt; zero retries or timeouts.
- `make check`: passed; zero lint violations across 2,151 files.
- [Final-head CI](https://github.com/steipete/CodexBar/actions/runs/34370412274): both macOS shards, both glibc Linux builds, musl, lint, and security checks passed.
- Independent P0–P2 review: no actionable findings.
- Before/after direct parsing of the supplied prefix changed 748,286 tokens/two rows to zero tokens/zero rows. Synthetic tests retain the first genuine appended child delta and verify cold/warm/forced/bounded behavior, removal of obsolete SQL rows, unchanged source bytes, retained file records, and compatible-store adoption without rebuilding.

Main is synchronized locally. #3524 remains open for the separate owned-usage undercount already reported on 0.57.0; this merge does not claim that remaining case resolved.

PR #3528, opened 2026-09-09
perf: avoid decoding usage events for Codex activity

brzvsk:
Cached Codex token activity now reads scoped daily aggregates instead of loading and decoding every usage-event payload. Account/root and time-zone validation, coverage, pending scans, retained retries, identity reconciliation, and detailed monetary/project/session reports keep their existing behavior. The nearby cached-read methods also return typed optional executor results directly, removing redundant double-optional wrapping.

This addresses one verified refresh-time CPU and memory hotspot related to #3247. It does not claim to resolve that report’s separate WebKit or whole-app energy symptoms. No parser hash, schema, pricing, refresh cadence, or credential-ownership change is introduced.

Maintainer proof:

- On unchanged main, the new real-fetcher regression recorded 1,024 event reads, 190,528 payload bytes, and 1,024 decodes. The retained-retry case also read eight unused events. All four assertions failed before the fix.
- After the fix, 70 focused tests passed: zero event payload reads/decodes, identical totals and coverage, foreign-account exclusion, time-zone and root rejection, and incomplete/retry handling.
- `make check` passed with zero violations; the Codex parser hash remains `4969a789db679c93`.
- `make test`: all 1,057 selections across 89 groups passed on the first attempt, with zero retries or timeouts (1,008.5 seconds). Exact-head CI remains the final pre-merge gate.
- Independent P0–P2 review found no actionable issues, including the optional-return cleanup and ownership boundaries.

The contributor’s release-core benchmark on independent private-cache copies measured median activity-call CPU of 1.385110 seconds before and 0.234250 seconds after, with complete output parity and unchanged stored-record fingerprints. Those are contributor measurements of this read path, not a prediction of whole-app battery savings.

Includes documentation and an Unreleased performance entry. Thanks @brzvsk!


steipete:
Landed as 5c0d7b4f87d64dfde01043955cce73f78f6b5ecb. Thanks @brzvsk for the focused projection fix and release-core measurements.

The activity reader now loads scoped daily aggregates without materializing usage-event payloads. Account/root, time-zone, coverage, incomplete-scan, and retry checks remain in place, and detailed reports retain their row-level history. The nearby optional-return cleanup preserves nil/error behavior.

Verification:

- Before the fix, the real-fetcher regression failed on 1,024 event reads, 190,528 payload bytes, 1,024 decodes, and eight unnecessary reads in the retained-retry case. Afterward those work counts are zero while totals and coverage match.
- `swift test --filter 'CostUsageStoreReadWorkTests|CostUsageFetcherCacheSnapshotTests|CostUsageStoreReadViewTests'`: 70 tests passed across two suites.
- `make test`: all 1,057 selections across 89 groups passed on the first attempt; zero retries/timeouts, 1,008.5 seconds total.
- `make check`: passed, zero violations across 2,169 files.
- `Scripts/regenerate-codex-parser-hash.sh --check`: unchanged `4969a789db679c93`.
- Independent P0–P2 review: no actionable findings.
- [Final-head CI](https://github.com/steipete/CodexBar/actions/runs/34437664280): macOS shards, Linux x64/arm64/musl, lint, aggregate gate, and security checks passed on `4b9ee5e081a98e8345e3c41e7adc20869db27149`.

Swift validation used Keychain suppression and Codex/session-file isolation through the test wrapper. Native UI/provider probing was unnecessary for this cache-read-only change. The contributor’s benchmark remains their measurement of this one read path; #3247 stays open for its broader CPU and WebKit symptoms. Documentation and the 0.58.1 Unreleased notes include the change and credit.

Issue #3523, opened 2026-09-09
Managed Codex workspace has no automatic renewal path when native OAuth expires (0.56.8)


zhulijin1991:
## Summary

On official CodexBar **0.56.8**, a managed Codex Team workspace becomes **Unavailable** when its native access token needs renewal. The error promises recovery through the Codex CLI, but both automatic-mode CLI fallback and explicit-OAuth native CLI recovery are disabled when a managed workspace is selected.

Preserving workspace scope is important. This report asks for a scope-safe renewal path, not removal of the existing guard.

## Observed on the installed release

- macOS, official `/Applications/CodexBar.app`, bundled CLI 0.56.8; three configured Codex accounts.
- Only the Team workspace failed; the two personal accounts continued returning OAuth usage.
- The CLI returned `Codex auth.json needs refresh` and said CodexBar would retry through the Codex CLI.
- The managed auth file existed and contained a refresh token. Its presence does **not** establish that the token was still redeemable; no direct refresh-token redemption was attempted during this incident.
- A prior repair record reported the Team access token expiring at **2026-09-09 06:13:35 UTC**. Today's failure was observed after that time. The prior expiry comes from that repair record, not a fresh inspection of the now-replaced token.
- Running official `codex login` with the target managed home, completing ordinary browser authorization with a localhost callback, and selecting the same workspace restored it.
- Afterwards, all three accounts returned `source=oauth`, `error=null`, and `dataConfidence=exact`; the Team workspace identity was checked, and the app's persisted account snapshots were also error-free after restarting CodexBar.

Device-code login was unavailable under this workspace's policy. Ordinary browser login worked without changing that policy. This is a recovery constraint, not the cause of the automatic-renewal gap.

## Tagged source explaining the gap

1. [`needsRefresh`](https://github.com/steipete/CodexBar/blob/v0.56.8/Sources/CodexBarCore/Providers/Codex/CodexOAuth/CodexOAuthCredentials.swift#L51-L61) recognizes native JWT expiry with a five-minute renewal window.
2. [`prepareCredentialsForUsage`](https://github.com/steipete/CodexBar/blob/v0.56.8/Sources/CodexBarCore/Providers/Codex/CodexProviderDescriptor.swift#L442-L457) throws `nativeRefreshRequired` for native credentials. Its comment correctly keeps the CLI as refresh-token owner to avoid rotation conflicts.
[...]

PR #3379, opened 2026-09-02
Refresh managed Codex credentials without losing workspace scope

Yuxin-Qiao:
## Summary

Renew expired or HTTP-401-rejected native Codex credentials through the selected `CODEX_HOME` app-server, then reload that store and retry the workspace-scoped HTTP usage request. Renewal has a dedicated 30-second timeout. HTTP 403 remains terminal and does not trigger renewal or CLI usage recovery; shared transport errors use endpoint-agnostic wording.

Concurrent callers for the same home share renewal. Each canceled caller now returns promptly; another active caller keeps its renewal, and canceling the final waiter drains the operation until the app-server exit is confirmed. An unconfirmed exit quarantines the home and retains the cross-process lease until Foundation observes termination. A per-home advisory `flock` serializes the desktop app and separate `codexbar usage` processes that share a `CODEX_HOME`; lock acquisition is bounded at 15 seconds and reports a retryable busy result instead of waiting forever. Read-only CLI commands cannot mutate native credentials.

## Validation

Current PR head: `01fd1883b`; base: `639b155226`.

- The focused Codex renewal, fallback, OAuth-expiry and transport suites passed: 40 tests in four suites.
- The architecture gate passed: 41 tests in one suite. Its `UsageFetcher.swift` suppression anchor was updated to the current `providerID: .codex,` line.
- `make check` passed: SwiftFormat 0/2318 files require formatting, SwiftLint 0 violations in 2317 files, and all repository/package checks passed.
- `make test` ran 1,159 selections in 107 groups. The only failure was the unchanged main-branch `CLIServeWebUITests` case `shared costs and diagnostics survive account grouping without sharing credits` (four assertions). On this host `zh-Hans_US` makes JavaScript currency formatting produce `US$2.00`/`US$5.00` while the unchanged test expects `$2.00`/`$5.00`; no PR file is involved. The failing group was retried with the same result.
- Tests use isolated fixtures and injected transports/processes. No real credentials, Keychain reads, token rotation or live provider probe was performed in this validation run.

## Remaining merge gates

- A genuinely denied or revoked workspace still needs authorized real-behavior evidence showing terminal HTTP 403 with no renewal and no successful usage publication. The earlier cross-account negative attempt returned 2xx and is not claimed as 403 proof.
- Fresh-home/existing-home isolation and independent-process runtime evidence are still maintainer/provider-auth review gates; the implementation has per-`CODEX_HOME` locking and deterministic fixture coverage.
[...]

PR #3534, opened 2026-09-10
Fix misleading Codex credential recovery guidance

steipete:
The shared native-credential refresh error no longer promises that CodexBar will retry through the CLI. Managed workspace selection deliberately disables that unscoped fallback, so the previous promise was false in both Auto and OAuth modes. The error now points to reauthenticating the affected account or running `codex login` in the same Codex home.

This corrects the diagnostic portion of #3523. It preserves `nativeRefreshRequired`, all strategy selection, credential ownership, and the existing reauthentication flow. The broader automatic-renewal work remains separate in #3379; #3523 should stay open.

Proof:
- The existing production-pipeline fixture failed eight added wording assertions before the change, across expired/near-expiry managed credentials in Auto and OAuth modes.
- 80 focused tests passed afterward, including default-account CLI delegation, managed-scope suppression, zero transport requests for stale credentials, and unchanged auth-file bytes, modification time and inode.
- Formatting/lint and independent P0–P2 review are clean. The full local suite passed all 1,058 selections in 89 groups on the first pass, with no retries or timeouts. All exact-head CI checks passed, including both macOS shards and the Linux builds: https://github.com/steipete/CodexBar/actions/runs/34451275400.

Documentation identifies the existing account-specific Reauthenticate action and same-home CLI recovery. The 0.58.1 Unreleased entry credits @zhulijin1991 for the report. No live credentials were used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 ### Performance
 - Codex activity: read scoped daily totals without decoding unused event history, preserving account boundaries, coverage, and incomplete-scan checks (#3528, related to #3247). Thanks @brzvsk!
 
+### Fixed
+- Codex accounts: replace a misleading automatic CLI-retry promise with account-specific reauthentication guidance when native credentials need renewal (related to #3523). Thanks @zhulijin1991!
+
 ## 0.58.0 — 2026-09-09
 
 ### Highlights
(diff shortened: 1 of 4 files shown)


steipete:
Landed in f5c94a11834b11c386177b554d26b1fa8c667f63. The shared diagnostic now gives account-specific recovery guidance instead of promising a CLI retry that managed workspace scopes deliberately suppress. No credential, fallback, or auth-file behavior changed.

The regression failed eight added wording assertions before the fix. Afterward, 80 focused tests and the full 1,058-selection suite passed, with all 89 groups successful on the first pass and no retries or timeouts. Formatting/lint and independent P0–P2 review were clean. All checks passed in [CI run 34451275400](https://github.com/steipete/CodexBar/actions/runs/34451275400).

#3523 remains open: this fixes its misleading guidance, while automatic renewal remains separate in #3379. Provider documentation and the Unreleased changelog are updated, with credit to @zhulijin1991 for the report.

PR #3537, opened 2026-09-10
fix(kimi): recover monthly Total usage from Desktop Local Storage

### LINKED ISSUE #3536: Kimi Code shows unused weekly/5h windows after monthly membership is exhausted
Current Kimi Desktop no longer writes a `kimi-auth` cookie. The live web access JWT (`typ=access`, `aud` includes `kimi.com`) lives in Desktop Local Storage instead.

CodexBar 0.56.8 only reads `~/Library/Application Support/kimi-desktop/Cookies` for `kimi-auth`. On a machine with a valid Desktop session and a working Kimi Code API key, Auto enrichment therefore never calls `GetSubscriptionStats`.

That leaves the menu on the Code API 7-day / 5-hour rate-limit windows. Those windows can reset to `0/100` while the shared monthly membership pool is already empty:

- Kimi Code CLI returns `403 You've reached your monthly usage limit`
- `GetSubscriptionStats` reports `subscriptionBalance.amountUsedRatio = 1`, notice “本月额度已用完”, and `overdrawn: true`
- CodexBar still shows 7-day / 5-hour as 100% remaining

Even after monthly `Total usage` is present as an extra window, Kimi's automatic menu-bar resolver only considers primary/secondary, so an exhausted month stays hidden behind a freshly reset 5-hour window.

## Expected
- Auto Desktop import should recover the Local Storage access JWT (read-only; skip refresh tokens and expired JWTs).
- API-key / CLI usage should still enrich monthly `Total usage` from `GetSubscriptionStats` when that session exists.
- Automatic menu-bar mode should prefer an exhausted monthly window over unused Code rate-limit windows.

## Repro
1. Stay signed into current Kimi Desktop (no `kimi-auth` cookie; session only in Local Storage).
2. Enable Kimi with a Code API key, Cookie source Automatic.
3. Exhaust the monthly membership pool.
4. Refresh CodexBar: 7-day / 5-hour look unused; monthly Total usage is missing.

I have a fix ready and will open a PR against this issue.

OttoPrua:
## Summary
- Current Kimi Desktop keeps the web access JWT in Local Storage and often no longer writes a `kimi-auth` cookie, so Auto never calls `GetSubscriptionStats`.
- Code API 7-day / 5-hour windows can reset to unused while the shared monthly membership pool is already empty.
- Read the Desktop Local Storage access JWT automatically (skip refresh / expired tokens), and prefer an exhausted extra window (`Total usage`) in automatic menu-bar mode.

Fixes #3536

## Test plan
- [ ] Stay signed into current Kimi Desktop with no `kimi-auth` cookie; Cookie source Automatic; API key configured
- [ ] Confirm CodexBar recovers a live Desktop access JWT and enriches monthly `Total usage` from `GetSubscriptionStats`
- [ ] Exhaust the monthly pool and confirm automatic menu bar shows 100% `Total usage` instead of unused 5-hour / 7-day windows
- [ ] Confirm an existing Cookies `kimi-auth` still wins over Local Storage, and refresh / expired JWTs are ignored
- [ ] `swift test --filter KimiDesktopAuthTokenTests` and `swift test --filter KimiProviderTests`

### DIFF
--- a/Sources/CodexBarCore/Providers/Kimi/KimiDesktopAuthToken.swift
+++ b/Sources/CodexBarCore/Providers/Kimi/KimiDesktopAuthToken.swift
@@ -5,11 +5,18 @@ import SQLite3
 #elseif canImport(CSQLite3)
 import CSQLite3
 #endif
+#if os(macOS)
+import SweetCookieKit
+#endif
 
 #if canImport(SQLite3) || canImport(CSQLite3)
-/// Read-only access to the official Kimi Desktop Chromium cookie store.
+/// Read-only access to the official Kimi Desktop Chromium session store.
 public enum KimiDesktopAuthToken: Sendable {
     private static let log = CodexBarLog.logger(LogCategories.provider(.kimi, scope: "cookie"))
+    private static let localStorageOrigins = [
+        "https://www.kimi.com",
+        "https://kimi.com",
+    ]
 
     public static func cookiesDatabaseURL(
         homeDirectory: URL = FileManager.default.homeDirectoryForCurrentUser) -> URL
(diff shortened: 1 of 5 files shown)


steipete:
I extracted and landed the independently verifiable display correction in #3543, with credit to @OttoPrua. A known exhausted monthly pool now takes precedence in Auto, with regression coverage and signed native before/after proof.

This PR remains open for the Desktop authentication work. That portion still needs evidence that the selected credential belongs to the current active session and that sign-out/deletion state is respected before it can be adopted.

steipete:
Thanks @OttoPrua for the report and proposal. The display correction is already in #3543. I reviewed the remaining authentication patch and cannot adopt it as written: substring audience matching admits lookalike hosts, and raw LevelDB candidate scanning can recover deleted historical credentials. Removing that raw scan is insufficient to establish current-session ownership: a synthetic fixture against the pinned reader selects an older sequence when its file modification time is newer. Closing this PR while keeping the acquisition problem tracked in #3536. The follow-up belongs in the shared Local Storage reader, followed by a bounded Kimi integration with current-record, sign-out, origin, token-type, expiry, and source-control coverage.

PR #3538, opened 2026-09-10
Add one-shot SSH Codex cost summaries to the CLI

ECuteri:
Compare native Codex cost history on this machine with one SSH host using `codexbar cost --provider codex --remote <host>`. Reports remain separate, so mirrored session logs cannot silently inflate a combined total. Each host keeps its own day boundaries, known token/cost values, pricing provenance, and coverage gaps.

This narrows the original proposal to an explicit CLI operation. It adds a versioned `--summary-only --format json` producer and a strict receiver without background polling, remote settings, or a second app-level cost store. The summary contains no account labels, project paths, model rows, or session text. Unknown prices remain unknown, and a remote failure retains the successful local report in one JSON document with a nonzero exit.

System SSH uses existing trusted hosts and batch authentication, no PTY or forwarding, closed stdin, a minimal environment, and a constant remote command. The remote executable is selected before the scan, avoiding a second scan on failure. Capture retains at most 16 KiB plus an overflow sentinel per stream; excess output is rejected after process completion, and the local SSH operation has a 60-second timeout. SIGINT/SIGTERM cancel and reap local helpers before re-raising the original signal; remote process lifetime remains the SSH server's responsibility.

Thanks @ECuteri for the proposal and original implementation.

Validation:
- `make check` passed. Independent review found no actionable P0–P2 findings.
- `make test` passed all 113 groups / 1,236 selections after integrating current main, with no retries or timeouts.
- Twelve focused tests cover schema validation, unknown/zero values, unsafe host arguments, coverage overflow, one remote invocation, cancellation, independent host results, and one-day text output.
- Real local CLI plus real SSH transport to a controlled Linux receiver passed success, remote failure, oversized valid JSON stdout, oversized stderr, SIGINT, and SIGTERM cases. Every case invoked the receiver once. Failure cases preserved local results and did not disclose the receiver's private stderr sentinel.
- SIGINT and SIGTERM stopped local SSH in 0.056s and 0.170s respectively. The controlled peer survived disconnection in both cases and was explicitly cleaned up. This proves local cleanup, not guaranteed peer-side cancellation.
- The Linux transport receiver was a fixture, not a compiled Linux CodexBar CLI. CI separately validates Linux builds and portable tests.
- The original local app and QA state were restored. Temporary receiver files and processes were removed, and the reused host was returned to its original hibernated state.

The screenshots below show actual captured CLI stdout displayed read-only in a local document window. All usage data is synthetic, and the host label is redacted.

**Before: local report**

Before: captured local CLI report

**After: separate local and SSH-host reports**

After: captured separate host CLI reports


steipete:
The username-case repair is now on main through #3700, merged as b88c78718da1372c341dae2cde6d53b830a1c813. Destinations such as `user@host` and `USER@host` stay distinct, while hostname-only case differences still deduplicate.

The regression reproduced the missing destination before the fix. `make test-fast FILTER='TailscaleSessionTests|RemoteSession'` passed 12 tests; `make check` and independent P2 review passed. All macOS and Linux CI checks succeeded at https://github.com/steipete/CodexBar/actions/runs/35170771260. These were sanitizer fixtures; no SSH connections were made.

Thanks @ECuteri. This fixes only destination identity handling; the broader remote cost-report proposals in #3538 and #3545 remain open.

PR #3540, opened 2026-09-10
fix: restore shared Codex spend in account menus

kays0x:
Restores the shared ambient Codex ledger beneath multi-account cards. Both stacked and compact account layouts now reach one Codex-owned trailing section, labeled **This Mac**, without assigning the same spend to individual accounts. Only the existing `codex:ambient` scope qualifies; managed-account and profile-home history remains isolated.

The maintainer pass also fixes the proposal’s display-mode mismatch: inline-only uses the existing inline dashboard, submenu-only uses the existing compact Cost row, and Both shows each once. The generic compact renderer no longer needs a Codex-specific trailing callback. Account override-card isolation is unchanged.

Validation:
- The new layout/preference regressions produced 12 failing assertions against the proposal. 53 focused tests then passed, including override isolation and architecture checks.
- `make check`: zero violations across 2,173 Swift files. Independent P0–P2 review is clean.
- Full `make test`: 1,061 selections across 89 groups passed on the first attempt, with no retries or timeouts (787.6 seconds).
- Developer-ID-signed native proof exercised two-account stacked and five-account compact menus, all three display modes, and opening the Cost history submenu. The native test passed with no failures. Only synthetic accounts and spend data were used.
- Native proof source matches commit `9b44b2fe9a6ae85712fbbc7242b700997c11cbdc`. [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34524643609) passed all Linux, musl and macOS checks. Merged as `cebf9995f672d2c0c9ff3501b24721855b86b16c`.

Documentation and the Unreleased changelog explain the ambient scope and credit @kays0x.

<details>
<summary>Inspected synthetic native before/after proof</summary>

Legacy stacked rendering omits spend:

Before: stacked accounts without shared spend

Stacked accounts with shared inline spend and Cost history:

After: shared spend below stacked accounts

Compact accounts with the Cost history submenu open:

After: compact accounts and history

Submenu-only mode keeps the compact Cost row:

After: submenu-only preference

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Codex activity: read scoped daily totals without decoding unused event history, preserving account boundaries, coverage, and incomplete-scan checks (#3528, related to #3247). Thanks @brzvsk!
 
 ### Fixed
+- Codex accounts: restore shared local spend below multi-account cards in stacked and compact layouts, honoring the selected cost display mode and preserving account-scoped history isolation (#3540). Thanks @kays0x!
 - Website: keep Arabic and Persian hero copy clear of the illustration, preserve natural text direction, and avoid an oversized tablet popover during its reveal (#3514, fixes #3511). Thanks @devYRPauli!
 - Codex accounts: replace a misleading automatic CLI-retry promise with account-specific reauthentication guidance when native credentials need renewal (related to #3523). Thanks @zhulijin1991!
 - Menus: refresh cached status menus when macOS appearance changes, including previously opened submenus, so the first opening matches Light/Dark and accessibility appearances (#3526). Thanks @emanuelst!
--- a/Sources/CodexBar/StatusItemController+CodexStackedMenu.swift
+++ b/Sources/CodexBar/StatusItemController+CodexStackedMenu.swift
@@ -1,4 +1,5 @@
 import AppKit
+import SwiftUI
 
 extension StatusItemController {
     func addStackedCodexMenuCards(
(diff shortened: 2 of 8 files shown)


kays0x:
@codex review

The implementation changed substantially since the last review, so the previous findings no longer describe the code.

- The override-card fallback is gone. `StatusItemController+MenuCardModel.swift` is now unchanged from main, so the isolation rule is untouched. Testing against two live accounts showed that giving each override card the ambient snapshot rendered the same total four times over in a two-account menu, because `tokenUsage` drives both the inline usage dashboard and the cost block.
- The machine-wide ledger is now rendered once below the account cards, gated on `UsageStore.tokenCostIsAccountAgnostic(for:)`.
- Both earlier findings are addressed: the tests build their stores via `testSettingsStore`/`withStatusItemControllerForTesting` so no Keychain access is reachable, and the compact layout shares the section through a `trailingSection` hook on `addCompactAccountMenuRows`.

Worth scrutinising: whether the shared section belongs in `addCompactAccountMenuRows` versus the common caller, and whether gating on `tokenCostScope` returning `codex:ambient` is the right signal for "this spend belongs to no single account".

steipete:
Merged as cebf9995f672d2c0c9ff3501b24721855b86b16c. The shared ambient ledger now appears once beneath stacked or compact Codex accounts, with inline-only, submenu-only and Both preferences respected. Account override cards and managed-profile history remain isolated.

The proposal failed twelve layout/preference assertions before the maintainer fix. All 53 focused tests, the full 1,061-selection local suite, lint, independent P0–P2 review and all Linux/musl/macOS CI checks passed. Signed synthetic native proof covers both account layouts, all display modes and opening Cost history; inspected images are in the PR body.

Provider docs and the Unreleased changelog are updated. Thanks @kays0x!

PR #3541, opened 2026-09-10
fix: wire Devin manual auth through the CLI on Linux

Waseemilyas:
Fixes the Devin CLI config path as well as the Linux platform gate. A configured manual bearer token previously never reached the provider's typed settings, so exempting prebuilt settings in the gate alone did not make the CLI work.

Devin now reads manual auth and organization through the existing descriptor credential adapter. Availability, the Linux gate, and the request share token precedence. Only a usable token in manual mode permits Linux quota requests; automatic browser import remains macOS-only. An empty primary environment override continues to shadow fallback values.

The regression decodes real provider config, checks the resulting settings, and exercises the existing fetcher with a synthetic HTTP transport. It verifies the final organization URL, Authorization header, and daily/weekly result. Linux CI also checks both Auto/Web CLI source modes, explicit Auto/Off auth, and empty overrides using those config-derived settings.

Validation so far: the original proposal failed the config-to-settings regression; 26 existing Devin tests and four new tests (eight cases) pass locally. `make check` is clean and independent P0–P2 review found no actionable defects. Full `make test` passed all 1,060 selections across 89 groups on the first attempt, with no retries or timeouts (954.4 seconds). [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34527317199) passed all Linux x64, Linux ARM64, musl and macOS checks. The Linux x64 log confirms all eight new config/request cases executed successfully. Merged as `c9f2eca8da03b525f097e89f6b98a55fed96b9fd`.

Includes Linux setup documentation and an Unreleased changelog entry. Thanks @Waseemilyas for the original Linux-gate fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
 - Codex accounts: replace a misleading automatic CLI-retry promise with account-specific reauthentication guidance when native credentials need renewal (related to #3523). Thanks @zhulijin1991!
 - Menus: refresh cached status menus when macOS appearance changes, including previously opened submenus, so the first opening matches Light/Dark and accessibility appearances (#3526). Thanks @emanuelst!
 
+- Devin CLI: load manual bearer and organization settings from config and allow manual quota requests on Linux while keeping browser import macOS-only (#3541). Thanks @Waseemilyas!
+
 - Documentation: point Spark and Daily Routines visibility instructions to the current per-item controls and include Overview in their scope.
 
 ## 0.58.0 — 2026-09-09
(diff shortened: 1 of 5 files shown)


steipete:
Merged as c9f2eca8da03b525f097e89f6b98a55fed96b9fd. The fix covers the missing descriptor credential reader as well as the Linux gate, so manual bearer and organization config reaches the actual fetcher. Availability, the platform gate and requests now share token precedence; automatic browser import remains macOS-only.

The new regression failed on the proposal before the config reader was added. Final validation passed: 26 existing Devin tests, four new config/request tests with eight cases, all 1,060 full-suite selections with no retries/timeouts, lint, independent P0–P2 review and every Linux/musl/macOS CI check. The Linux x64 log confirms the config-derived gate and synthetic final-request cases ran successfully.

Linux setup docs and the Unreleased changelog are updated. Thanks @Waseemilyas!

PR #3547, opened 2026-09-10
feat(sessions): make failed-host hiding opt-in

warthurton:
Add an opt-in Settings → Menu → Agent sessions → “Hide unreachable hosts” control. It is off by default for both new and existing installations, so connection, authentication, incompatible-CLI and decoding failures remain visible unless the user explicitly chooses to hide them. Fetching and retries continue unchanged; turning the option off restores the diagnostic rows.

Reconciled with current main, preserved explicit saved choices, completed the two strings in all 22 required locales, and added isolated default/persistence/menu-observation coverage. The native proof uses the production settings section and live controller descriptor path, fed only by an injected synthetic session store. Thanks @warthurton for the original contribution. The release note is centralized in #3567.

The focused descriptor/settings suites pass 19 tests, locale validation passes, and independent P0–P2 review of the current implementation is clean. Signed interactive proof passed visible → hidden → restored through the actual control and production controller path. The full local suite passed 1,074 selections in 90 groups, with zero failures or retries. Make check and 41 architecture tests pass. Exact-head CI is green: https://github.com/steipete/CodexBar/actions/runs/34658787278. Native screenshots and the detailed proof are in https://github.com/steipete/CodexBar/pull/3547#issuecomment-5641870346. This is a menu visibility preference; it does not change SSH discovery or authentication policy.


warthurton:
I have compiled and tested this and shown working.  There might need to be some more fidly filtering, but it does reduce a list of 20 tail nodes that don't have codexbar or claude on them from being displayed.

steipete:
Maintainer native proof passed with the repaired opt-in behavior.

Built the current CodexBar sources and launched a Developer ID–signed, isolated XCTest host using the production settings section and the same controller descriptor path used by the live menu. An injected session store supplied only `ready.example.invalid` and `offline.example.invalid`; no real SSH discovery or credentials were involved. Drove the actual controls with Peekaboo:

- Initial/default off: the failed-host row and its diagnostic tooltip are retained.
- Opt in: the failed-host row disappears; the reachable empty host remains.
- Turn off again: the failed-host row returns. The explicit choice is written to the injected defaults store.

The interactive XCTest passed with zero failures. The 19 descriptor/settings tests and 41 architecture-gate tests pass; default and explicit true/false reloads have isolated regression coverage. Locale coverage, make check, and independent P0–P2 review pass. The attached captures contain synthetic data only. The first proof attempt left a native popup tracking; the verified run closes it before changing the setting and yields during the interactive wait.

The final branch also reanchors existing architecture checks without changing their provider memberships or fingerprints. The changelog line and contributor credit are in #3567. Exact-head CI remains the final readiness gate.

Default off: failed host remains visible

Opt in: failed host hidden, reachable host retained

steipete:
Final CI passed on `44dbc7e9c22c840ec5b11a6e152955856dbd24b0`: https://github.com/steipete/CodexBar/actions/runs/34658787278.

The full local suite also passed 1,074 selections in 90 groups with zero failures or retries. Signed native visible → hidden → restored proof and inspected synthetic screenshots are recorded above. Make check, all 41 architecture-gate tests, and final independent P0–P2 review pass. Ready for squash with @warthurton's credit; keep #3567 last for the centralized notes.

Issue #3549, opened 2026-09-11
Merged menu keeps a stale height after switching provider tabs (macOS 26 table-backed menus never shrink)


zenibako:
### Summary

With Merge Icons enabled and two or more providers whose menus differ in height, switching provider tabs leaves the menu the wrong size. Switching to a shorter tab leaves a block of empty space below the last item; switching to a taller tab clips the bottom so **Quit and About become unreachable**. The menu is correct on first open, and switching *accounts* within a tab is unaffected.

### Environment

- macOS 26.6.2, MacBook (1512x949 logical, `visibleFrame` 919pt)
- Merge Icons enabled, providers: Claude (claude-swap, 3 accounts) + Codex, plus the Overview tab

### Steps to reproduce

1. Enable Merge Icons with two providers whose menus differ substantially in height
2. Open the menu — it is correctly sized
3. Switch provider tabs (Overview / Claude / Codex)
4. Observe trailing empty space on the shorter tab, or clipped items on the taller one

### Root cause

macOS 26 renders menus as a table inside a scroll view:

```
NSMenuItemViewer < NSCustomMenuItemView < NSTableRowView
  < NSTableViewBackedMenuRepresentation < NSClipView < NSMenuScrollView
  < NSView < NSNextStepFrame
```

`NSTableViewBackedMenuRepresentation` **grows to fit new content but never shrinks**. It floors near the tallest height it has held during the session. The window, content view and clip view all resize correctly — only the table's document height goes stale.

Settled state, sampled 1.2s after populate so these are not mid-animation values:

| Tab | `menu.size` | window | contentView | NSClipView | NSTableViewBackedMenuRepresentation |
|---|---|---|---|---|---|
| Overview | 776 | 776 | 776 | 776 | 776 |
| Claude | 606 | 606 | 606 | 606 | **779** (+173) |
| Codex | 564 | 564 | 564 | 564 | **774** (+210) |

The grow/shrink asymmetry, same session:

```
menu.size=536   table 441 -> settled 536    grow:   correct
menu.size=776   table 779 -> settled 776    shrink: correct (3pt)
menu.size=606   table 802 -> settled 779    shrink: stuck
menu.size=564   table 779 -> settled 774    shrink: stuck
```

### Why CodexBar triggers it

[...]

PR #3498, opened 2026-09-08
fix(claude): honor segmented account menu layout

steipete:
Claude-swap ignored Menu → Multi-account layout and always rendered stacked/compact cards. Honor Segmented with stable account buttons and one source-owned detail card, reusing the existing serialized activation path.

Keep the active marker tied to adapter output. Unavailable accounts remain inspectable without activation; pending and failed switches retain the requested account’s usage and diagnostics. Use stable slot labels with Hide Personal Info. Closing inspection invalidates the cached menu so a fresh persistent menu returns to the active card on reopen without another fetch. With no active or explicitly selected account, show the no-active notice without falling back to ambient usage. Stacked/compact menus and CLI cards keep their existing behavior.

Fixes #3382. Thanks @thatlev for the report and acceptance criteria. Release notes are in #3499, to land after the code PRs.

Validation: make check passed; make test passed all 1,036 selections in 87 groups on the final head with no retries, failures, or timeouts. The persistent-menu regression failed before repair for both provider and merged menus, then passed. The architecture guard remains unchanged apart from four exact-line offsets. Independent branch review is clean through P2.

Developer ID-signed native proof exercised successful/failed switching, retained usage, privacy labels, sentinel inspection, merged menus, four-account wrapping, and no-active behavior. Reusing the same menu across inspection and reopen restored the active card with no change to provider revision or adapter invocation count. The built CLI continued to render every synthetic account, including the expired slot. All proof uses synthetic data and a fake cswap executable.

| Existing stacked rendering (preserved) | Segmented rendering |
| --- | --- |
| Existing stacked Claude cards with synthetic accounts | Segmented Claude cards with the same synthetic accounts |

| Inspect an unavailable account | Reopen the same menu, without a fetch |
| --- | --- |
| Inspecting a synthetic expired account | Persistent menu restored to the active account |

[No active account: no ambient quota card](https://github.com/user-attachments/assets/bf0c06bb-8af6-472d-b850-f9624588d56d).

Issue #3382, opened 2026-09-02
Apply segmented multi-account layout to claude-swap and provider account sources


thatlev:
## Summary

Make the existing **Settings → Menu → Multi-account layout → Segmented** preference apply to every multi-account source shown in the provider menu, including `claude-swap` accounts. At minimum, Claude should get the same compact side-by-side account switcher UX that managed Codex accounts already have.

Today the preference says “Choose segmented account switching or stacked account cards,” and `segmented` is the default, but `claude-swap` always renders one full card per account vertically. With two Claude subscriptions this makes the menu unnecessarily long and gives Claude different behavior from Codex under the same global preference.

## Current behavior and source path

- `MultiAccountMenuLayout` already defines `.segmented` and `.stacked` in `Sources/CodexBar/SettingsStore.swift`.
- Token accounts and managed Codex accounts honor that setting through `TokenAccountMenuDisplay` / `CodexAccountMenuDisplay`, `addTokenAccountSwitcherIfNeeded`, and `addCodexAccountSwitcherIfNeeded`.
- `ClaudeSwapMenuPrecedence.prefersClaudeSwap(...)` intentionally suppresses Claude token-account presentation when the adapter has eligible accounts.
- `StatusItemController.addMenuCards(...)` then calls `addClaudeSwapMenuCards(...)` before the normal token-account branches.
- `Sources/CodexBar/Providers/Claude/StatusItemController+ClaudeSwapMenu.swift` chooses only full stacked cards or the four-plus-account compact list. It never consults `settings.multiAccountMenuLayout` and has no segmented switcher path.
- Therefore changing the existing preference to **Segmented** has no effect on `claude-swap`, even though the setting is presented as provider-neutral.

## Expected UX

When a provider has multiple activatable account snapshots and the global layout is **Segmented**:

1. Show account buttons side by side at the top of that provider's menu, matching the existing Codex switcher visual language.
2. Highlight the active account and show only that account's full usage card below the switcher.
3. Clicking an inactive Claude button must call the existing serialized `UsageStore.switchClaudeSwapAccount(_:)` path (`cswap --switch-to <slot> --json`), not introduce a second switching mechanism.
4. Preserve immediate loading/selection feedback, refresh the ambient Claude snapshot and all `claude-swap` account snapshots after success, and surface a switch failure on the requested account without losing retained usage.
[...]

PR #3550, opened 2026-09-11
feat(menu): unify Claude and Codex account switching (view-only segments, shared System Account submenu)

zenibako:
claude-swap-view-only-light
claude-swap-view-only-dark

### Why

Claude and Codex should behave the same way, so expectations don't differ per provider. Before this PR they didn't:

- Clicking a **Codex** segment changes which account CodexBar follows. It never touches the Codex CLI login; that happens in the **System Account** submenu.
- Clicking a **Claude** (claude-swap) segment ran `cswap --switch-to`, changing Claude Code's login from a click that looks like "show me this account".

This PR gives both providers one model: **segments select, the System Account submenu switches**.

### What

**1. Claude segment clicks are view-only.**

- Clicking a claude-swap segment only changes which account's details are shown. It never runs `cswap`.
- The viewed account is keyed by `ProviderAccountIdentity`. It survives refreshes, reordering and menu closes for the session, and is not persisted across restarts. It never changes which account drives the menu bar quota.
- It is dropped when the adapter is disabled or its executable changes. If its slot disappears, it falls back to the source-reported active account, and it never invents one.
- Unavailable accounts stay inspectable. A view selection outranks a pending or failed switch, and a switch error stays attached to the account that produced it.

**2. One shared account switcher for Claude and Codex.**

- `AccountSegmentedSwitcherView` replaces `CodexAccountSwitcherView` and `ClaudeSwapAccountSwitcherView`. It owns the layout, selection styling, click tracking, hit-testing, tooltips and accessibility.
- Providers only supply labels. Codex's width-aware email/workspace/discriminator titles and private ordinals moved into `CodexAccountSwitcherLabeling` as pure functions.
- Shared vocabulary: the filled segment is **Selected**, and a leading `●` marks the **System** account, the one the CLI uses (Codex `isLive`, claude-swap active slot).
- `CodexAccountSwitcherView` remains as an override-free subclass, so menu reuse still recognizes the Codex row.

**3. One shared System Account submenu for switching.**

- Codex's existing submenu becomes a shared builder behind two defaulted `ProviderImplementation` hooks.
- A provider-neutral `requestSystemAccountSwitch` action replaces `requestCodexSystemPromotion`.
- **Codex:** promotion eligibility and `CodexAccountPromotionCoordinator` are unchanged. Submenu titles now honor Hide Personal Info; previously they showed raw emails. The submenu title is now localized.
- **Claude:** claude-swap accounts appear while claude-swap owns Claude account presentation. The active slot is checked, and inactive slots with usable credentials are enabled. Choosing one awaits the existing serialized `cswap --switch-to <slot> --json` transaction.
- **Claude cards:** they no longer offer "Switch Account…" for inactive slots. They keep #3452's accent **Active** badge and its explicit **Re-authenticate** repair action for an active slot claude-swap marks re-activatable (`foreign_credential`).

**4. Switch feedback.**

- The target card's subtitle shows "Switching Claude Code to Account 7…" and then "Account 7 is now the System account", for every layout:
  - segmented and single live cards;
  - stacked and compact Codex account cards;
  - claude-swap account cards.
- An error already on the card outranks progress and success. That includes a claude-swap switch error published before the follow-up Claude refresh finishes.
- A card showing switch feedback stops following the live refresh monitor's provider status, so the feedback is what renders. The header resolves its subtitle through `MenuCardLiveSubtitle.resolve`.
- When no menu is open at completion, a notification reports success or failure. If a failure notice can't be delivered (notifications not authorized), the alert Codex showed before is presented instead.

**5. The claude-swap executable has a real default, and its settings are grouped.**

- `~/.local/bin/cswap` was only the settings field's placeholder, so enabling the adapter without typing a path silently did nothing.
- The default now applies, but only when something executable is at that path.
- An explicit path always wins, including when it is missing.
- The toggle, executable field, adapter status and discovered accounts render as one settings section through `supplementarySettingsContent`, the slot Codex already uses.

### Behavior changes for existing users

- **Claude:** clicking a segment no longer switches accounts, and cards no longer offer "Switch Account…"; use **System Account** in the menu. Re-authenticate stays on the active card. This is covered in the changelog and `docs/claude.md`.
- **Codex:**
  - the live account gains a `●` in the switcher;
  - a failed promotion shows on the card while the menu is open;
  - when the menu is closed, a failed promotion is reported by notification, with the previous alert as fallback when notifications aren't allowed.
- **Upgrades with the claude-swap adapter enabled and an empty saved path** start using `~/.local/bin/cswap` if it is installed there. Otherwise nothing runs, as before.

### Proof

The signed synthetic native proofs are opt-in via `CODEXBAR_CLAUDE_SWAP_VIEW_PROOF_DIR`. They require credential isolation and use synthetic accounts and stub executables only.

- **Screenshots:** the production switcher in Light and Dark, viewing slots 2, 7 and 9 while slot 2 is System.
- **Segments vs submenu (`menu-owner.json`):**
  - Clicking every segment through the real menu's switcher view starts **0 activations**: no switch task, and no stub `--switch-to` invocation.
  - The System Account submenu checks slot 2, enables only slot 7, and disables unavailable slot 9.
  - Choosing slot 7 runs exactly `--switch-to 7 --json` and posts one notice: "Claude Code now uses Account 7".
- **Executable discovery (`discovery.json`),** run through the production config loader, resolver and reader:

  | Config | Result |
  |---|---|
  | Fresh config, cswap at the default | listed 3 accounts |
  | Fresh config, nothing at the default | idle, nothing run |
  | Upgraded config with a saved empty path, cswap at the default | listed 3 accounts |
  | Explicit path, default also installed | explicit binary used |
  | Explicit path that does not exist | "Missing CLI" error; the default is never run |

### Testing

- Rebased on `main` `62a7fa1c`.
- `make check`: 0 violations, including the app locale checker.
- Full sharded suite (`Scripts/test.sh`): 106/106 groups pass on the first pass, with nothing skipped.
- **Review fixes, each with a test that fails without it:**
  - a published claude-swap switch error outranks progress;
  - stacked and compact Codex account cards show progress and failure for the promoted account only;
  - an applied subtitle wins over a live `MenuCardRefreshMonitor`.
- **Alert fallback:** an undelivered failure notice raises the fallback alert, and an undelivered success notice does not. Both tests use injected delivery; no UI is shown.
- **Other new coverage:**
  - the shared switcher view, the feedback model, the submenu builder, and the applied subtitle;
  - the Codex adapter (promotion plus an unknown account) and redacted Codex submenu titles;
  - the Claude adapter (entries, success, failure, unavailable slot, controller notice) and the card repair action being offered only for active repair.
- **Migrated:** existing Codex and Claude switcher, menu and proof tests.
- No real `cswap`, provider probe or Keychain read in any test.

### Notes

- **Docs:** `docs/claude.md`, `docs/claude-multi-account-and-status-items.md` and `docs/codex.md` are updated.
- **Localization:** 7 new strings are translated for all 23 app locales.
- **Changelog:** entries are under 0.60.2 Unreleased.
- Follows #3382 / #3498 (segmented multi-account layout for claude-swap) and #3452 (claude-swap repair and last-known usage).

### Not included

The merged-menu row-height bug from #3549 is fixed separately in #3616.

https://claude.ai/code/session_012T91QRonsX591fKr6UuN5R


steipete:
The existing System Account submenu now honors Hide Personal Info on main through #3702, merged as 141ecf642a97d9ce4dadfd70102a790eb24578a5. It uses stable private account labels while keeping promotion targets and menu state intact; privacy off retains the original display names.

Validation: `make test-fast FILTER='CodexSystemAccountPrivacy|CodexAccountSwitcherRedaction|ProviderArchitectureGatekeeper'` passed 56 tests. Signed synthetic native before/after proof is at https://github.com/steipete/CodexBar/pull/3702#issuecomment-5707505346. `make check`, final independent P2 review, and every macOS/Linux check passed: https://github.com/steipete/CodexBar/actions/runs/35174630916. The only integration conflict was in the changelog; source and test blobs remained unchanged.

Thanks @zenibako. This repairs the existing submenu's privacy omission; the broader account-switching redesign in #3550 remains open.

steipete:
Thanks @zenibako for the implementation and proof. Closing this broad unification proposal. Main already fixes Codex System Account label privacy in 141ecf642a97 and clarifies Claude’s explicit chip switching and reconciliation feedback in 9a2f5e420ab3 (#3740). The remaining changes are a larger UX redesign: view-only chips, relocated activation, executable discovery, grouped settings, and shared notification state. Those are not all superseded, but we are retaining the current provider-specific behavior rather than adopting this combined redesign.

PR #3552, opened 2026-09-11
Improve Codex multi-account dashboard visibility

xxchan:
## Summary

Expose Codex multi-account details consistently in the dashboard and menu bar Overview while preserving privacy and existing provider-level metrics.

## Changes

- Include all visible Codex accounts in the dashboard snapshot and render them as separate account cards.
- Carry the actual active-account selection through the usage payload instead of assuming the first sorted account is active.
- Redact account labels and expose opaque account IDs when dashboard identity redaction is enabled.
- Add per-account credit balances, reset-credit counts, and expiry data; the WebUI shows the earliest expiry while retaining the provider-level financial summary.
- Add provider-scoped WebUI display settings for hiding Spark windows and reset-credit counts.
- Expand Codex stacked accounts into one row per account in CommandBar Overview, with safe activation IDs and no misleading provider-level history submenu.
- Add multi-account snapshot coverage for active selection, redaction, opaque IDs, credits, and reset details.

## Validation

- `swift build --product CodexBar`
- `swift build --product CodexBarCLI`
- `make check` (passed on the initial revision; source changes are format-safe under `git diff --check`)
- Focused dashboard snapshot assertions added for the multi-account projection.

The full test target was started but its first complete rebuild exceeded the local run window; no test failure was reported before it was stopped. The two production products compile successfully after the review fixes.

### DIFF
--- a/Sources/CodexBar/StatusItemController+Menu.swift
+++ b/Sources/CodexBar/StatusItemController+Menu.swift
@@ -578,12 +578,35 @@ extension StatusItemController {
         // interaction closures must always reference the live menu they end up serving.
         let interactionMenu = captureMenu ?? menu
         let providerScopes = self.overviewProviderScopes(enabledProviders: enabledProviders)
-        let rows: [(provider: UsageProvider, model: UsageMenuCardView.Model)] = providerScopes.visible
-            .compactMap { provider in
-                guard let model = self.menuCardModel(for: provider) else { return nil }
-                guard !model.isOverviewErrorOnly else { return nil }
-                return (provider: provider, model: model)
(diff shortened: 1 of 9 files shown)


xxchan:
@clawsweeper re-review

steipete:
Thanks @xxchan for working on multi-account visibility. Closing this patch without merging it. The WebUI account feature is not already implemented by the native Usage & Spend work, but the proposed selection projection loses ownership on fetch failures: error payloads omit accountActive, then the snapshot builder treats the first row as active. A failed inactive first row can therefore produce two active accounts, and a failed active second row can produce none. A focused follow-up should carry selection metadata through success and failure and cover both cases before adding the WebUI presentation.

PR #3555, opened 2026-09-11
fix(deepseek): restore daily Platform usage and preserve fallback

DrSmoothl:
Use the DeepSeek Platform per-key daily endpoints for rolling 30-day usage, request/key counts, and token/spend charts. Preserve the existing Gregorian UTC monthly fallback and its “This month” label, while the daily path uses the same fixed offset for request bounds and bucket labels across DST. Preserve cancellation, currency separation, string/object key identifiers, fractional-cent cost precision, and API-key-only balance behavior.

Local validation passed `make check`, 1,076 selections across 90 groups, 130 focused tests across 12 suites, and independent review. A signed native host drove the production fetcher/parser with synthetic endpoint responses and rendered the actual usage/spend cards alongside balance-only behavior.

DeepSeek daily usage and balance-only cards

The direct Platform-token importer found no available session for an authenticated per-key request in this environment. The new endpoint proof is fixture-based; it does not claim a live Platform schema check.

Changelog and contributor credit are centralized in #3572. Exact-head CI remains a merge gate.


steipete:
Maintainer validation for `911356b14477d2bc2e9f3a91324c32a26dedddd8`.

Full local run: 1,076 selections / 90 groups, all first-pass green. After the calendar/parser/precision refinements, the final focused run passed 130 tests / 12 suites. `make check` and independent P0–P2 branch review passed.

The signed native host called the production fetcher/parser with synthetic per-key responses, then displayed the production daily Usage/Spend charts and balance-only card. Actual fallback requests are covered across positive/negative UTC offsets and month/day/DST boundaries, including cancellation before fallback. Fractional-cent usage remains visible.

No Platform session was available to the direct importer for an authenticated per-key endpoint check. The new endpoint proof is explicitly fixture-based; no live daily-API claim is made.

Exact-head [CI run](https://github.com/steipete/CodexBar/actions/runs/34668050977) completed successfully, including both macOS shards and all Linux/lint gates. Changelog/credit are in #3572.

deepseek-native-final

PR #3557, opened 2026-09-11
fix(devin): honor hidden daily quotas

### LINKED ISSUE #3542: CodexBar should not show Devin Daily limit on Max plan
What I see is that the daily limit is always 100%, and the weekly is behaving like Daily. 

Image

Update:

Max plans only have a weekly limit yes.
On pro plans you can also see a daily limit.

steipete:
Fixes #3542. CodexBar showed a Daily row even when Devin explicitly hid that quota. The parser now honors a root Boolean `hide_daily_quota: true` before both current-field and fallback parsing, while preserving weekly usage, reset timestamps, extra balance and percentage scaling. False, absent and malformed flag values retain their previous behavior. No plan-name heuristic is needed.

The existing hidden-daily test omitted daily data, so it could not catch the defect. The strengthened regression and new current/fallback cases fail six assertions on the old parser. Removing the single-use tuple wrapper keeps this fix at **six fewer production lines**. Provider docs and the 0.59.1 Unreleased changelog are updated. Thanks @dzienisz for the dashboard comparison.

Validation:
- 69 focused parser and architecture tests passed, including actual JSON Boolean/numeric/string/null flags and missing-visible-window behavior.
- `make check` passed with zero violations across 2,182 files. Two earlier test argument-formatting violations were corrected; those edits were verified whitespace-only.
- Independent P0–P2 review of the final staged patch is clean.
- Developer-ID-signed synthetic native before/after tests passed in Light and Dark appearances. The actual production parser feeds production menu cards: the hidden-daily card loses only Daily, Weekly remains 10% left, and the visible-daily card is unchanged. All four full captures were inspected before upload.
- Full `make test` passed all 1,069 selections across 90 groups on the first attempt, with no retries or timeouts (842.7 seconds).
- [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34616966632) passed all checks.

The reporter supplied a same-time dashboard comparison and confirmed weekly-only Max versus daily-and-weekly Pro; the exact live response was not shared. The proof here is synthetic and does not use real provider credentials. The separate account-switching question is outside this quota-visibility repair.

<details>
<summary>Signed synthetic native before/after proof</summary>

Before, Light: synthetic Devin quota cards

Before, Dark: synthetic Devin quota cards

After, Light: synthetic Devin quota cards

After, Dark: synthetic Devin quota cards

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.59.1 — Unreleased
 
 ### Fixed
+- Devin: honor hidden daily quotas even when the response includes daily usage, preserving weekly limits and extra balance (#3542). Thanks @dzienisz!
 - Codex accounts: honor Hide Personal Info in switcher labels and tooltips, redact embedded workspace emails, and preserve distinct account numbers in narrow menus (#3551). Thanks @zenibako!
 - Codex spend: keep waiting history files ahead of repeated migration revisits so large histories can finish bounded catch-up, preserving stored rows and checkpoints (#3548, related to #3411). Thanks @SergeiNikolenko!
 
--- a/Sources/CodexBarCore/Providers/Devin/DevinUsageSnapshot.swift
+++ b/Sources/CodexBarCore/Providers/Devin/DevinUsageSnapshot.swift
@@ -103,9 +103,15 @@ public enum DevinUsageParser {
     }
 
     public static func parse(_ object: Any, organization: String?, now: Date = Date()) throws -> DevinUsageSnapshot {
-        let current = (object as? [String: Any]).map(self.currentQuotaWindows)
-        let daily = current?.daily ?? self.findWindow(in: object, matching: self.isDailyKey)
-        let weekly = current?.weekly ?? self.findWindow(in: object, matching: self.isWeeklyKey)
+        let dictionary = object as? [String: Any]
+        let hideDaily = (dictionary?["hide_daily_quota"] as? NSNumber)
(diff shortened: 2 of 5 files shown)


steipete:
Landed as [1c4650d1b42](https://github.com/steipete/CodexBar/commit/1c4650d1b421bda3f0e495e8c14b69dc1b3de81d). The parser now honors Devin’s explicit daily visibility flag before either daily parsing path. Weekly usage, resets and extra balance are preserved, and the single-use tuple helper is removed for six fewer production lines.

Verification used the credential/session-isolating test wrapper:

```sh
swift test --filter 'DevinUsageFetcherTests|ProviderArchitectureGatekeeperTests|DevinQuotaVisibilityNativeProofTests'
make check
make test
```

The new regressions failed six assertions on current main. The fixed focused run passed 69 tests; the full suite passed all 1,069 selections across 90 groups first time, with no retries or timeouts (842.7 seconds). Final lint had zero violations after correcting two whitespace-only test formatting issues. Independent P0–P2 review was clean, and [exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34616966632) passed.

The separately launched Developer-ID-signed native fixture passed before and after in Light and Dark mode, using the actual parser and production menu cards with synthetic data. The hidden-daily card now shows Weekly alone; the visible-daily card is unchanged. All four inspected captures are in the PR body. No real provider credentials were used, and the exact live response was not supplied by the reporter. Docs and 0.59.1 Unreleased notes are updated. Thanks @dzienisz.

PR #3562, opened 2026-09-11
fix(grok): skip discarded work after terminal billing failures

### LINKED ISSUE #3345: Grok local scan: signals.json undercounts by 54.5x, and costUsdTicks is a usable real-spend source (divisor verified as 1e10)
# Grok local scan: `signals.json` undercounts by 54.5x, and `costUsdTicks` is a usable real-spend source (divisor verified as 1e10)

## Context

Not a PR proposal. #3135 and #3236 are already rewriting the Grok local scanner and are in review; I'm not opening a competing branch. I measured an independent corpus and got two results worth putting on the record — one corroborating #3135, one that I think argues against a design choice all three open PRs currently share.

Corpus: 51 session directories under `~/.grok/sessions`, 934 completed turns, 2026-07-22 → 2026-08-25, models `grok-4.6-build` and `grok-4.5-build`.

## 1. `main` still reads `signals.json` — 54.5x undercount

`docs/grok.md:115` and `GrokLocalSessionScanner.swift` on current `main` aggregate `totalTokensBeforeCompaction`, `contextTokensUsed`, and `modelsUsed`. Those are context-window occupancy snapshots, not per-turn consumption.

| Source | Tokens |
|---|---:|
| `signals.json` (`totalTokensBeforeCompaction` + `contextTokensUsed`) | **44.9M** |
| `turn_completed` in `updates.jsonl` (934 turns) | **2.45B** (2.44B in / 7.9M out) |

**54.5x.** #3135 reports 653K vs 54.1M (83x) on its own corpus — different ratio, same failure mode, so this is not specific to one usage pattern.

Mechanism, stated explicitly because it explains why the ratio varies: `totalTokensBeforeCompaction` counts each compaction event's context snapshot once, while billed input is the full context **re-sent every turn**. My corpus is 97.6% `cachedReadTokens` (2.38B of 2.44B) — cache reads are real billed input that `signals.json` has no field to represent. The ratio therefore scales with turns-per-compaction, which is why my 54.5x and #3135's 83x differ while both being correct.

## 2. `costUsdTicks / 1e10 == USD` — verified, and it is the only real-spend source

All three open PRs avoid `costUsdTicks` (#2407: "missing ticks are **not** estimated"; #3135: models.dev list price labeled an estimate; #3236: token-only, no dollars), and [ccusage documents ignoring it because the scale is undocumented](https://ccusage.com/guide/grok/). I want to submit evidence that the scale *is* determinable, because the consequence is material.

**Method.** Price each turn from the [published xAI rates](https://docs.x.ai/docs/models) and compare against `costUsdTicks / 1e10`. Two structural facts must be modeled or the check fails:

1. **Long-context tiering.** At 200k tokens the *entire* request bills at 2x ($2/$0.50/$6 → $4/$1.00/$12 for 4.6).
2. **A promotional rate from 2026-08-15**, at 17% of list. (xAI shipped Grok 4.6 on 2026-08-12 with a [launch promotion for Grok Build users](https://www.layer3labs.io/guides/grok-4-6-pricing).)

Taking the best of {tier1, tier2} × {list, 17%} per turn:

| Model | turns | median error | within 1% |
|---|---:|---:|---:|
| grok-4.6-build | 645 | **0.0000%** | 82.2% (530/645) |
| grok-4.5-build | 286 | **0.0000%** | 79.4% (227/286) |

The promotional rate is not a fitted parameter — it falls out of the data as a clean date boundary:

| Date | turns priced at list | turns priced at 17% |
|---|---:|---:|
| 07-22 → 08-14 (8 active days) | **445** | **0** |
| 08-15 → 08-25 (6 active days) | 9 | **477** |

Zero discount-priced turns before 08-15; 9 list-priced turns after. A fitted constant does not produce that.

**A caution on my own result:** the 200k tier does *not* key off `inputTokens` as recorded. 212 turns with `inputTokens >= 200k` price at tier 1. So the tier predicate is something else (uncached portion, or per-`modelCalls` context). This does not affect the conclusion — it means the tier must be *read* from `costUsdTicks` rather than *recomputed*, which is precisely the argument for using the field.

### Why this matters for #3135 and #3236

List-price reconstruction cannot see a promotional rate. On my corpus:

| Period | Real spend (`costUsdTicks`) | List-price reconstruction | Overstatement |
|---|---:|---:|---:|
| through 08-14 (list-rate period) | $697.56 | $822.92 | 1.18x |
| 08-15 onward (promo period) | **$264.82** | **$1,762.74** | **6.66x** |
| total | **$962.38** | $2,585.66 | 2.69x |

A user inside the promo window would see about 6.7x their actual spend. #3135's "list-price estimate, explicitly labeled" framing is honest about being an estimate, but a 6.7x gap is not a rounding artifact — it is the difference between a bill and a sticker price during any promotional period, and xAI runs these at model launches. The residual 1.18x in the list-rate period comes from the tier predicate noted above: reconstructing the 200k tier from `inputTokens` over-applies the 2x rate.

`costUsdTicks` already encodes tier and promo. Suggested handling: use it when present, fall back to list price only when absent, and label which one produced each figure. Absence is rare — **2 of 934 turns (0.21%)** had `costUsdTicks == 0`, both early in the corpus.

I'd also gently flag that this cuts against #3236's "do not invent dollar/API-rate estimates for subscription credits" rationale: reading `costUsdTicks` is not inventing an estimate, it is reading the number the CLI recorded. Whether to *display* dollars for subscription users stays a product call — but the data is there and it is accurate.

## 3. Two bounds notes for the scanners under review

- **`costUsdTicks` coverage is high**: 2/934 turns (0.21%) missing. #2407's "missing ticks are not estimated" stays correct but will rarely trigger on a mid-2026 corpus.
- **7 of 51 session directories have no `turn_completed` rows at all** (aborted/short sessions). Keying off `turn_completed` presence rather than directory existence avoids carrying empty entries.

## Environment

- CodexBar `main` as of 2026-09-01 (`docs/grok.md:115`, `GrokLocalSessionScanner.swift`)
- Grok Build: 51 session dirs, 934 turns, 2026-07-22 → 2026-08-25
- macOS 15.6 (Darwin 25.6.0)

## Ask

Nothing beyond recording this, and specifically that the `costUsdTicks` scale question be treated as answered rather than open. Happy to re-run any query against my corpus if it would settle a review question — field distributions, per-version behavior, specific edge-case records. Not planning a competing PR.

steipete:
Avoids a local-history scan whose result is discarded when CLI billing has already failed. The provider can reach its fallback sooner, without scanning the same history once in the failed CLI attempt and again in the later web strategy.

The probe now evaluates its existing acceptance policy first: successful billing or the narrowly defined unexpired Team identity-only fallback. Accepted results share one scan/version/settings path and snapshot constructor. Team eligibility is rechecked after local work at the original decision point; the identity admitted before optional settings enrichment remains captured afterward. Original errors, credential filtering, diagnostics, plan enrichment, public APIs and legacy history semantics are preserved. This is a **four-line production reduction**.

This extracts the lazy-scan observation from #3236; it does not implement that PR’s new log format, expand automatic reads, or claim to fix #3345’s token totals. Thanks @Yuxin-Qiao. Docs and 0.59.1 Unreleased notes are updated.

Validation:
- The actual Grok RPC client talks to a synthetic local CLI child. Four terminal cases reproduce an unwanted scan on the old ordering: personal method-unavailable, Team authorization failure, initialization failure, and expired Team credentials.
- The fixed tests require zero scan/settings calls for those failures and preserve exact errors. Successful billing and eligible Team fallback each scan once and retain summary, version and identity/diagnostic behavior.
- 48 focused tests across seven suites passed, including auth, account context, billing, settings, plan and child-process coverage. Five additional regression assertions caught changed expiry checkpoints in the first refactor; a nine-case matrix now preserves both late eligibility rejection and already-admitted identity using deterministic markers, without clock sleeps.
- `make check` passed with zero violations; final independent P0–P2 review is clean.
- The final expiry-checkpoint revision passed all 1,070 selections / 90 groups on the first pass in 926.5 seconds, with zero retries or timeouts, using an isolated temporary home.
- Superseded candidate CI was canceled; [final-head CI](https://github.com/steipete/CodexBar/actions/runs/34642009256) passed before merge.

All credentials and RPC responses in the proof are synthetic. Settings transport is injected; no live provider request or personal history is used. Successful presentation is unchanged, so no UI screenshot is claimed for this failed-path work reduction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Devin: honor hidden daily quotas even when the response includes daily usage, preserving weekly limits and extra balance (#3542). Thanks @dzienisz!
 - Codex accounts: honor Hide Personal Info in switcher labels and tooltips, redact embedded workspace emails, and preserve distinct account numbers in narrow menus (#3551). Thanks @zenibako!
 - Codex spend: keep waiting history files ahead of repeated migration revisits so large histories can finish bounded catch-up, preserving stored rows and checkpoints (#3548, related to #3411). Thanks @SergeiNikolenko!
+- Grok: skip discarded local-history scans and version probes after terminal CLI billing failures, allowing fallback to start sooner (extracted from #3236). Thanks @Yuxin-Qiao!
 
 ## 0.59.0 — 2026-09-10
 
--- a/Sources/CodexBarCore/Providers/Grok/GrokStatusProbe.swift
+++ b/Sources/CodexBarCore/Providers/Grok/GrokStatusProbe.swift
@@ -84,6 +84,8 @@ public struct GrokStatusProbe: Sendable {
     }
 
     var settingsTransport: any ProviderHTTPTransport = ProviderHTTPClient.shared
+    var identityOnlyFallback: @Sendable (GrokCredentials?, Bool, Error?) -> Bool =
+        GrokStatusProbe.shouldUseIdentityOnlyFallback
 
     public init() {}
 
@@ -128,10 +130,7 @@ public struct GrokStatusProbe: Sendable {
             rpcError = error
         }
 
-        let isIdentityOnly = billing == nil && Self.shouldUseIdentityOnlyFallback(
-            credentials: credentials,
-            billingAttempted: billingAttempted,
-            error: rpcError)
+        let isIdentityOnly = billing == nil && self.identityOnlyFallback(credentials, billingAttempted, rpcError)
         // Terminal CLI failures must reach the provider's web fallback without scanning discarded history.
         guard billing != nil || isIdentityOnly else {
             throw rpcError ?? GrokRPCError.notAuthenticated
@@ -139,13 +138,17 @@ public struct GrokStatusProbe: Sendable {
 
         let localSummary = try await self.localSummary(env)
         let cliVersion = Self.detectVersion(env: env)
+        // Preserve the original eligibility checkpoint after potentially slow local work.
+        if isIdentityOnly, !self.identityOnlyFallback(credentials, billingAttempted, rpcError) {
+            throw rpcError ?? GrokRPCError.notAuthenticated
+        }
         let subscriptionTier = try await Self.loadSettingsTier(
             credentials: credentials,
             session: self.settingsTransport)
         return GrokUsageSnapshot(
             billing: billing,
             webBilling: nil,
-            credentials: Self.credentialsForSnapshot(
+            credentials: isIdentityOnly ? credentials : Self.credentialsForSnapshot(
                 credentials: credentials,
                 billing: billing,
                 webBilling: nil),
--- a/Sources/CodexBarCore/Providers/Grok/GrokStatusProbe.swift
+++ b/Sources/CodexBarCore/Providers/Grok/GrokStatusProbe.swift
@@ -79,6 +79,12 @@ public struct GrokStatusProbe: Sendable {
     public static let usageUnavailableMessage =
         "Grok usage is unavailable because its billing sources did not report a usage percentage."
 
+    var localSummary: @Sendable ([String: String]) async throws -> GrokLocalSessionSummary? = {
+        try await GrokLocalSessionScanner.summarizeOffMainThread(env: $0)
+    }
+
+    var settingsTransport: any ProviderHTTPTransport = ProviderHTTPClient.shared
+
     public init() {}
 
     public static func detectVersion(env: [String: String] = ProcessInfo.processInfo.environment)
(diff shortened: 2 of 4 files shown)


steipete:
Landed as ad8b1745257ebb6c970e7be7dc4443e2ec6d815a.

The failed CLI billing path now rejects terminal errors before local history scanning, version lookup or settings enrichment. Successful billing and the existing Team identity-only fallback share one result path. Team eligibility is still checked at the original post-scan/version point, and identity admitted before optional settings work stays captured afterward. This removes four production lines without changing the public API or legacy history semantics.

Synthetic RPC child-process tests first reproduced four unwanted scans. Further deterministic regressions caught expiry-checkpoint changes before landing. The final 48 focused tests, full suite (1,070 selections / 90 groups, zero retries or timeouts), format/lint and independent review passed. No live provider credentials or personal history were used.

Thanks @Yuxin-Qiao for the lazy-scan observation in #3236. This is a narrow extraction: that PR remains open for its broader history format, and #3345 remains open for token-counting correctness. The Unreleased changelog and Grok documentation are updated.

[Final-head CI](https://github.com/steipete/CodexBar/actions/runs/34642009256) passed across both macOS shards, plugin-engine checks and Linux builds. Superseded CI was intentionally canceled for the expiry-checkpoint correction. The canonical main checkout is updated; its changed production/test/docs files match the reviewed head, and contributor credit is present in the squash commit and changelog.

Issue #3508, opened 2026-09-08
Codex token snapshot can stay stale after #3314 while directory discovery is pending


kernnel:
## Summary

On the official signed CodexBar 0.56.8 (139), today's Codex token total can remain stale for hours after #3314 when bounded file work is complete but current-window directory discovery is still pending.

Quota data and the outer widget timestamp continue refreshing. A forced full refresh with the bundled CLI completes in about 10 seconds and immediately publishes the current totals, so the source data is readable.

## Sanitized runtime evidence

- Global refresh frequency: five minutes.
- The widget snapshot's outer `generatedAt` advances normally while the Codex `tokenUsage.updatedAt` remains more than three hours old.
- The persisted advancing cache contains a newer current-day total while the visible snapshot remains the preserved report.
- Scan metadata has `catchUpPending = true` with `completedFiles == totalFiles` and `processedBytes == totalBytes == 0`.
- `lookback_state.pendingFilePaths` is empty, but `currentWindowNextDayKeyByRoot` is still roughly four weeks behind the current day.
- A forced full CLI refresh clears the pending state and updates the visible total immediately.

No account identifiers, usernames, email addresses, absolute paths, exact usage totals, credentials, device identifiers, session IDs, or screenshots are included.

## Root cause on current main

Verified against `2033edb37`.

#3314 correctly freezes per-file byte targets so a growing rollout cannot make one generation infinite. This remaining case is directory discovery rather than file-tail growth:

- automatic catch-up runs for two seconds and then waits about 1,998 seconds on AC power;
- while discovery is incomplete, the scanner preserves and returns the previous established report;
- `UsageStore.retainsEstablishedTokenHistory` rejects an incomplete same-scope snapshot entirely;
- therefore an advancing current-day cache cannot reach the menu or widget until historical directory discovery completes.

The result is another all-or-nothing publication starvation path even when the reported file and byte work is already complete.

## Expected behavior

Historical discovery may continue at low duty cycle, but same-scope current-day token growth should become visible within the normal token refresh cadence. Established historical days must not disappear, lower partial values must not overwrite published values, and the final stable snapshot must remain authoritative for full reconciliation (including legitimate downward corrections).

## Proposed fix shape

- Keep the preserved-report behavior as the default for normal cached readers.
- When a foreground Codex refresh is incomplete, read the advancing cache explicitly.
- Merge only per-day monotonic token increases into the same-scope established snapshot.
[...]

PR #3566, opened 2026-09-11
fix(codex): exclude queue waiting from catch-up delays

steipete:
Automatic Codex history catch-up currently counts time waiting behind another account or provider on the shared serial scan queue as active work. Its duty-cycle policy then multiplies that wait into additional sleep. Measure the scan inside the executor and use its own active duration in both usage and Spend Dashboard workers.

Preserve the existing power/thermal policy, initial delay, scan budgets, cancellation, and complete-history publication boundary. Refs #3508 and #3411; these broader reports remain open. This does not change held PR #3509 or enable its dormant accelerated path. The changelog entry will be centralized in the final batch notes PR.

Validation so far: 48 focused regression tests passed, independent P0–P2 review clean, and a real integration executable using the production executor and policy parsed 10,000 synthetic JSON records. A 460 ms total await contained only 32 ms of active work: the old delay was 459 seconds, while the corrected delay was 32 seconds. The final full local suite passed 1,072 selections in 90 groups with zero failures or retries. Signed native menu smoke, make check, 41 architecture-gate tests, and final branch review pass. Exact-head CI is green: https://github.com/steipete/CodexBar/actions/runs/34658785497.

### DIFF
--- a/Sources/CodexBar/UsageStore+CodexCostCatchUp.swift
+++ b/Sources/CodexBar/UsageStore+CodexCostCatchUp.swift
@@ -172,11 +172,10 @@ extension UsageStore {
                         return
                     }
 
-                    let passStartedAt = ContinuousClock.now
                     self.codexCostCatchUpPassIsRunning = true
-                    let nextStatus: CostUsageFetcher.CodexScanCatchUpStatus
+                    let result: CostUsageScanExecutor.TimedResult<CostUsageFetcher.CodexScanCatchUpStatus>
                     do {
-                        nextStatus = try await self.advanceCodexCostCatchUp(
+                        result = try await self.advanceCodexCostCatchUp(
                             now: Date(),
                             codexHomePath: context.codexHomePath,
                             historyDays: context.historyDays)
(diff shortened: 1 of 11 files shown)


steipete:
Verified head `cdfefc954488afff6ed9b16eddf839dee652935f`: exact-head CI passed at https://github.com/steipete/CodexBar/actions/runs/34658785497.

The final full local run passed all 1,072 selections in 90 groups, with zero failures, retries, or timeouts. The targeted scheduler/executor tests and 41 architecture-gate tests pass; existing provider memberships and fingerprints are unchanged. Make check and final independent P0–P2 branch review pass.

A separately built integration executable used the production serial executor and unchanged scheduling policy while parsing 10,000 synthetic JSON records behind another queued task. Its output was:

```text
parsed_records=10000
awaited_seconds=0.459539541 active_seconds=0.031833417
old_delay=runAfter(459.080001459) fixed_delay=runAfter(31.801583583000003)
```

The old-delay value feeds the measured total await into the existing policy, matching the old workers' input; the fixed value uses the executor's own measured work. A freshly built Developer ID–signed app also passed native menu interaction in the isolated synthetic smoke mode.

Ready for squash. This corrects queue-wait debt only: retain the initial automatic delay, resource safeguards, and complete-history publication. Leave #3508 and #3411 open, and keep #3509 held. The release note is in #3567.

PR #3576, opened 2026-09-12
fix(groq): reject unavailable and overflowing metric rates

steipete:
Groq metrics can contain unavailable Prometheus values or malformed samples, and finite query rates can overflow when combined or converted to per-minute units. These previously produced nonnumeric usage or invented zero. Eager payload decoding could also hide the server’s actual query error.

Check query status before decoding metric data, validate successful tuples and finite nonnegative aggregates, and validate all derived per-minute rates before publishing the snapshot. Genuine empty vectors remain zero. Numeric decoding and decimal-label formatting are simplified while preserving their output; production code is **net -4 lines**. Documentation and the Unreleased changelog are updated.

The [Prometheus API format](https://prometheus.io/docs/prometheus/latest/querying/api/#expression-query-result-formats) documents nonfinite strings. Parser and injected-transport regressions reproduce 14 original failures; four additional cases prove API-error precedence over malformed scalar, tuple, result, and data shapes. Magnitude-based label precision and finite-rate mapping remain covered. Focused tests, `make check`, and independent review through P2 pass on the synchronized candidate. The updated-main full suite passed 1,090 selections in 91 groups (1,435.0 seconds). One existing RPC teardown fixture hit its initialize deadline and passed the runner’s single group retry; no groups timed out. [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34680063354) passed every check. No live provider credentials are used.

Validation cleanup removes the file-watcher fixture’s competing initial-write retry, records callback write errors, and asserts final replacement bytes as well as notification. Production file-watching behavior is unchanged.


steipete:
Landed as 6b83e08637de85f11887934fb37e1bf7db559e01. Groq now checks API status before metric decoding, rejects malformed/unavailable/nonfinite samples and overflowing aggregates, and validates the final per-minute rates. Genuine empty vectors remain zero, API messages are preserved, and rate-label precision is unchanged. Production code is net -4 lines.

Validation: `swift test --no-parallel --filter 'Groq|CloudSyncSettingsTests'`, `make check`, and `make test` passed in isolated test homes using synthetic parser/HTTP fixtures. Eighteen failing-before cases cover the original rate defects and the added error-precedence boundaries. The final full run passed 1,090 selections in 91 groups; one existing RPC initialization-deadline test passed on the runner’s single group retry, with no timed-out groups. Independent review through P2 was clean. [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34680063354) passed every macOS/Linux/lint/aggregate check.

The file-watcher cleanup is test-only: it removes a competing fixture write and verifies the callback’s replacement bytes. No real provider credentials or live queries were used. Main was pulled with `--ff-only` and is clean, with the merge verified reachable.

PR #3578, opened 2026-09-12
fix(spend): honor civil days and bucket time zones

steipete:
Daily calendar arithmetic and display formatting still had several independent failures after #3565. Around Santiago's skipped midnight, two covered dates counted as one, chart ranges extended an hour into the following day, and historical pace backfill could replace following-day credits with zero. A bucket time zone different from the host also put chart bars/labels and the selected-day subtitle on the wrong dates.

Use actual calendar-day interval ends, civil-day ordinals, and one group calendar throughout chart ranges, bar binning, axis/accessibility labels, and the selected-day subtitle. Pass one calendar through historical parsing and reuse each day's end as the next cursor. The cleanup leaves production code **net -1 line**. Documentation and the Unreleased changelog are updated; this follows Gabriel Rojas's #3565 heatmap repair.

Validation: seven new calendar/credit assertions fail on main and pass after the fix; 145 focused tests cover dashboard, historical pace, and architecture, with additional 23/25-hour and year-boundary controls. Signed native component rendering on a Los Angeles host with synthetic Santiago data verifies1/7→2/7 coverage, Sep5→Sep6 selected-day text, and correctly aligned Sep6–7 bars. The renderer supplies no calendar/timezone workaround. Independent review through P2 is clean. `make check` and the final full local suite passed: 1,076 selections in 90 groups, all first pass with no retries/timeouts. [CI for the reviewed head](https://github.com/steipete/CodexBar/actions/runs/34672989978) passed with all checks green.

Validation also exposed a test-only cleanup race: its two-second interrupt could fire before fixture ownership was observed. A delayed-start regression proves it; interruption now starts after the existing ownership handshake, keeping cleanup/sentinel/grace assertions.

Before:

Synthetic bucket-calendar baseline

After:

Synthetic bucket-calendar corrected output

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - **Safer account menus:** preserve privacy in Codex account labels and reauthenticate the credential source shown by the selected row.
 
 ### Changes
+- Usage & Spend: count covered calendar days, align chart labels and ranges with the selected bucket time zone, and preserve historical-pace credits after midnight daylight-saving transitions (follow-up to #3565).
 - Usage & Spend: preserve heatmap coverage, token totals, and daily ledger rows across midnight daylight-saving transitions, retaining real gaps and unscanned days (#3565). Thanks @gabrielrojasc!
 - Cursor on Linux: restore automatic authentication from the signed-in app, honor absolute XDG/HOME paths, and preserve manual-cookie precedence and explicit web-mode isolation (#3539). Thanks @DonnieFi!
 - Codex spend: exclude time waiting behind other scans from automatic catch-up sleep calculations while preserving scan budgets, power safeguards, and complete-history publication (#3566, related to #3508 and #3411).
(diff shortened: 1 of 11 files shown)


steipete:
Landed as f9bc62c344b3ba5fa758632c94a5c43af8fae90e. Calendar-day coverage, credit backfill, and chart binning/labels now share the correct civil-day boundaries and bucket time zone; production code is net -1 line.

Validation used isolated test homes and no provider credentials: `swift test --no-parallel --filter 'SpendDashboardMidnightDSTTests|SpendDashboardModelTests|HistoricalUsageMidnightDSTTests|HistoricalUsagePaceTests|ProviderArchitectureGatekeeperTests'`, `make check`, and `make test`. The final full run passed 1,076 selections in 90 groups on the first pass with no retries or timeouts. Independent review through P2 was clean. [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34672989978) passed every check.

Signed native before/after renders attached above use synthetic Santiago buckets on a Los Angeles host: coverage changes from 1/7 to 2/7, the selected date from Sep 5 to Sep 6, and bar dates align correctly. Earlier intermediate captures exposed the remaining axis/binning errors and were replaced by the final proof. An earlier local run was invalidated by overlapping a rebuild; the final frozen run is the passing result reported here. Main was pulled with `--ff-only`, and the merge is reachable in a clean checkout.

PR #3579, opened 2026-09-12
fix: recover Claude menu refresh interactions

HP-network:
Fix Claude menu refresh behavior after credential recovery failures.

Summary:
- Keep refresh-all retries for providers that are not shown in the open menu in background interaction.
- Allow prompt-capable Claude recovery only when Claude is one of the rendered providers and its provider policy recognizes the persisted credential error.
- Cover MCP-only Keychain state alongside the existing unreadable and missing-credential recovery cases.
- Keep the existing account-action decision as the single source for Claude sign-in presentation.
- Add regression coverage for visible-provider scoping and MCP-only recovery.

Validation:
- `git diff --check` passes.
- `swiftc -frontend -parse` passes for the changed implementation files.
- `swift test --filter StatusMenuInstantOpenTests` is blocked locally because the repository requires Swift 6.2 and the installed toolchain is Swift 6.1.
- `swiftformat` and `swiftlint` are not installed in this environment.


HP-network:
@clawsweeper re-review

PR #3580, opened 2026-09-12
fix(claude): publish switch errors before ambient refresh

steipete:
A failed explicit claude-swap activation can keep its known error hidden until a slow or queued ambient Claude refresh finishes. Independent adapter reads can also pass the publication guard after cancellation when they have no provider-refresh generation.

Publish the switch error and its stable requested account ID before awaiting reconciliation. Keep the transaction task and pending selection until reconciliation completes, and clear the error if configuration changes. Reject canceled list/version reads through one shared publication predicate. Reset transient state through its initializer while preserving the live credential transaction and version generation. Production code is **net -1 line**; documentation and the Unreleased changelog are updated.

This extracts the error-timing defect identified while reviewing #3452, with contributor credit to @QuantIntellect. The larger display/strategy-switch proposal stays in that PR; this patch uses the existing explicit numeric-slot activation contract.

Synthetic proof uses an isolated fixture, a controlled failing executable, and a suspended refresh. Four early-error assertions fail before the fix; the repaired state exposes the error while holding the transaction guard, blocks duplicate activation, and discards errors after configuration changes. A canceled-task regression also fails before the publication guard fix. Focused adapter/runtime tests, `make check`, and independent review through P2 pass. The final serial full suite passed 1,091 selections in 91 groups, all first pass with no retries/timeouts (1,085.2 seconds). [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34679336294) passed every check. No live credentials or provider accounts were used.


steipete:
Landed as 816d495e8ba4e3ac2ab0ace7ef8c6e83bf7d6108, preserving @QuantIntellect’s contributor credit. Known switch errors now publish before ambient reconciliation finishes, while the credential transaction remains serialized. Canceled adapter reads cannot publish stale state. Production code is net -1 line.

Validation: `swift test --no-parallel --filter 'ClaudeSwap|ClaudeProviderRuntimeTests'`, `make check`, and the final serial `make test` passed in an isolated home with synthetic command/refresh fixtures and Keychain access suppressed. The final full run passed 1,091 selections in 91 groups on the first pass, with no retries/timeouts. Independent review through P2 was clean. [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34679336294) passed all checks.

Two earlier simultaneous local runs interfered with shared preference fixtures; the unchanged candidate passed the complete serial rerun. The synthetic cases prove error publication before a suspended refresh, duplicate-transaction exclusion, configuration-change discard, and canceled-read rejection. No live credential transaction was performed. Main is synchronized and clean, and the merge and human co-author trailer are verified.

PR #3582, opened 2026-09-12
fix(stepfun): preserve credit plans without balance data

steipete:
A new StepFun credit plan without a populated balance is already classified as a credit plan, but display conversion required both that classification and a credit rate. A missing rate therefore fell through to the classic plan path and turned zero placeholder fields into two exhausted five-hour/weekly bars with epoch resets.

Select the credit branch by plan classification, then map the optional balance into an optional primary bar. Preserve genuine zero-credit exhaustion, live classic windows, plan identity, update time, reset fallback, and monthly pace eligibility. Nearby simplification removes redundant conversion temporaries: production code is **net -5 lines**. Documentation and the Unreleased changelog are updated.

The existing missing-pool fixture now checks its projected snapshot: both bar assertions fail on unchanged production and pass with the fix. Focused StepFun tests and `make check` pass, including existing exhausted-credit and classic-window controls. Independent review through P2 is clean. The full local suite passed 1,079 selections in 90 groups, all first pass with no failures, retries, or timeouts (1,181.3 seconds). [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34678411043) passed with every check green. Proof uses contained parser/projection fixtures with no live credentials or provider calls.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - Agent Sessions: add an opt-in setting to hide hosts whose session fetch failed, while keeping diagnostics visible by default (#3547). Thanks @warthurton!
 - Menu bar: align window and display coordinates so monitors above or below the primary display do not cause missed or false startup recovery.
 - Devin: honor hidden daily quotas even when the response includes daily usage, preserving weekly limits and extra balance (#3542). Thanks @dzienisz!
+- StepFun: keep credit plans without a populated balance from displaying exhausted five-hour and weekly quota bars, preserving genuine zero-credit balances and classic plan windows.
 - Grok: skip discarded local-history scans and version probes after terminal CLI billing failures, allowing fallback to start sooner (extracted from #3236). Thanks @Yuxin-Qiao!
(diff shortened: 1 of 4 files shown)


steipete:
Landed as 8667306a8569b112c0d608e778f14479ebc51f88. Credit-plan classification now remains authoritative when the credit balance is absent, so placeholder rolling-window fields no longer become two exhausted bars. A real zero-credit balance still shows 100% credit usage; classic/live-window behavior is preserved. Production code is net -5 lines.

Validation: the existing missing-credit fixture failed both new projection assertions before the fix. `swift test --no-parallel --filter StepFun`, `make check`, and `make test` passed in a contained test home with Keychain access suppressed. The full suite passed 1,079 selections in 90 groups on the first pass, with no retries or timeouts. Independent review through P2 was clean, and [exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34678411043) passed all macOS/Linux/lint/aggregate checks. Proof is parser/model based; no live provider credentials were used. Main is synchronized and clean, with the merge verified reachable.

PR #3587, opened 2026-09-12
feat(muse): add Meta Muse (Muse Code) usage and token cost provider

Srimi1:
## Summary
- Adds a **Meta Muse (Muse Code)** provider (`muse`) that reads the Muse Code CLI's local session logs — no network access, no cookies.
- Usage card: Today / Weekly token windows (estimates against per-plan token budgets; Meta publishes no hard quotas, so the snapshot is marked `estimated`), plan tier and email from `~/.config/muse/settings.json`.
- Token cost: Today and 30-day token totals + USD estimates from `~/.local/share/muse/sessions/**/session.jsonl` (also `~/.config/muse/sessions`, `~/.muse/sessions`), with a Muse pricing catalog (standard + Contributor rates), inline cost dashboard, and `codexbar cost --provider muse` / `codexbar usage --provider muse`.
- Follows the provider authoring guide: everything Muse-specific lives in `Sources/CodexBarCore/Providers/Muse/` + `Sources/CodexBar/Providers/Muse/`; descriptor-driven registration (credential adapter, `supportsTokenSnapshot`, `supportsInlineTokenCostDashboard`, `history: .alwaysTracked`, `supportsCostCommand`, `widgetSelectable: false`); manifests regenerated with `Scripts/regenerate-provider-manifests.sh`.

## Details
- Muse Code 1.1.x logs each model call twice — a `model_completed` runtime event (carries the model name) and a `goal_usage_attribution` record. The scanner and usage fetcher count `model_completed` and use attribution records only as a fallback, so calls are never double counted (covered by `MuseCostUsageScannerTests`).
- `MuseStatusProbe.probe()` is filesystem-only (safe from the main actor); `muse --version` runs only through the descriptor `versionDetector` via `ProviderVersionDetector.run`.
- Muse keeps a small JSON cache (`cost-usage/muse-v1.json`) beside the Claude/Vertex artifacts; Codex's `CostUsageStore` is untouched (the Codex parser hash is regenerated only because `CostUsageScanner.Options` gained `museSessionsRoots` and the dispatch switch gained a `.muse` case).
- Gatekeeper fingerprints/allowlists, provider count (69 → 70) across README/site/llms metadata and all site locales, `docs/muse.md`, `docs/providers.md`, and CHANGELOG (Unreleased → Added) updated.

## Commands run
- `make check` (`./Scripts/lint.sh lint`) — 0 SwiftFormat/SwiftLint violations, locales, manifests, parser hash, docs links, llms index OK.
- `swift build` (all targets).
- `swift test --filter "Muse|MusePricingTests|ProviderArchitectureGatekeeper|ProviderRegistryTests|CodexBarWidgetProviderTests|ProviderIconResourcesTests|InlineUsageDashboardBarColor"` — all passing.
- `make test` (full sharded suite) — run locally; see follow-up comment for the result.
- Manual: `codexbar usage --provider muse` / `codexbar cost --provider muse` against real Muse Code 1.1.1 logs; packaged app installed and launched.

## Screenshots
CLI output against real logs:
```
== Meta Muse 1.1.1 (local) ==
Today: 93% left [===========-]
Resets in 22h 36m
Weekly: 0% left [------------]
Resets in 6d 22h
Plan: Meta Account

Meta Muse Cost (API-rate estimate)
Today: $0.04 · 287K tokens
Last 30 days: $7.22 · 43M tokens
Estimated from local logs · may differ from your bill
```

## Notes for review
- Daily/weekly percentages depend on approximate per-plan token budgets in `MuseUsageFetcher.budgets(for:)`; happy to switch to token-count-only display or a user-configurable budget if preferred.
- Widget support intentionally deferred (`widgetSelectable: false`).


Srimi1:
Pushed `bf6565486` addressing the review findings:

- **Quota basis visible in both displays** — Muse now uses `usesDetailBackedWindow`; each window carries its token total and the unofficial budget it is scaled against (`287K tokens today · est. Standard budget 4M/day (unofficial)`, `43M tokens in the last 7 days · est. budget 28M (unofficial, rolling)`), rendered in the menu and the CLI. The menu additionally shows the existing localized `Quota estimated from local usage history` note via the descriptor's `usageNotesResolver`.
- **Weekly reset no longer advances per refresh** — the 7-day window is a rolling sum and carries no `resetsAt`; only "Today" has a real boundary (local midnight).
- **`subscriptionRenewsAt` is nil** — no billing metadata is invented.
- **Contributor pricing applied** — the scanner resolves the tier from Muse settings (or `Options.museIsContributor`) and passes it to `museCostUSD`; each tier keeps its own cache artifact (`cost-usage/muse-v1-<tier>.json`), so switching tiers never reuses stale costs.
- **Undated records rejected** — the usage fetcher now reuses the cost scanner's parser (`parseMuseFileCancellable`), so timestamp handling is shared and records without a parseable date are dropped instead of charged to today.
- **Unavailable telemetry ≠ zero usage** — missing session directories throw `MuseUsageError.sessionLogsUnavailable`, an existing directory with no logs throws `.noSessionLogs`; a log with no model calls is a measured zero. `MuseFetchStrategy.isAvailable` now checks the probe instead of returning `true`.
- **Checked arithmetic** — token counts must be non-negative integers ≤ 1e12 (else the record is rejected); nanodollar conversion uses `Int(exactly:)`; all accumulation uses `addingReportingOverflow` and drops overflowing contributions.

New/updated tests: contributor pricing + per-tier cache, corrupt count rejection (`1e16`, `Int.max`, negative, fractional), warm-cache reuse and rescan on change, estimated windows/reset semantics, missing-directory and empty-directory errors, zero-usage log.

Commands: `make check` (0 violations), `swift build`, `swift test --filter "Muse|ProviderArchitectureGatekeeper|ProviderRegistryTests|InlineUsageDashboardBarColor|CodexBarWidgetProviderTests|ProviderIconResourcesTests"` (123 tests pass), CLI against real Muse Code 1.1.1 logs.

Full `make test` locally: `AdaptiveRefreshTimerTests` (2 timing cases) fails and `AgentSessionsStoreSchedulerTests` times out identically on pristine `upstream/main` in this environment, so the grouped runner stops early here; I could not get a complete local run. CI should be authoritative once workflows are approved.

On the maintainer decision: I kept the estimated windows but labeled the basis explicitly; happy to drop the percentages entirely and defer quota to #3435 if you prefer that direction.

@clawsweeper re-review

Srimi1:
Pushed `448f3e645`, redesigned per the latest review round:

- **No invented quota windows** — Muse no longer reports `primary`/`secondary` rate windows at all. Meta publishes no local-derivable limits, so a remaining-capacity bar built from a guessed budget would still mislead. The account line instead shows the exact local token totals verbatim: `today <N> tokens · 7d <N> tokens`. This intentionally leaves subscription-window display to an authoritative source (e.g. #3435); local telemetry and subscription quota are different data sources and shouldn't be conflated in one strategy.
- **`.api` source mode removed** — the descriptor only advertises `.auto`/`.cli`; `--source api` now correctly errors instead of silently returning local estimates under an API label.
- **Weekly range is filtered, not just scan-padded** — both the cost scanner and the usage fetcher filter parsed rows against `range.sinceKey...range.untilKey` before accumulating, so the scanner's one-day padding can no longer leak an 8th day into a "7-day" total. Covered by a regression test with a day-7/day-8 boundary.
- **Readable telemetry tracked separately from filenames** — `MuseParseResult` now carries `decodedRecordCount`; a directory of only corrupt JSON/JSONL surfaces `MuseUsageError.noSessionLogs` instead of a silent zero-usage success. A log with zero model calls (but valid JSON) is still a measured zero, distinct from unreadable.
- **Contributor pricing never falls through to standard** — `musePricing(model:isContributor:)` resolves within one tier only; unknown models and `muse-code` both get a Contributor-tier fallback entry when `isContributor` is true (added `muse-code-contributor` to the catalog). Model names already carrying the `-contributor` suffix keep that tier regardless of the flag.
- Per-tier caching, checked arithmetic on token counts/costs, and shared parsing between the cost scanner and the usage fetcher are unchanged from the previous revision.

Commands: `make check` (0 violations), `swift build`, `swift test --filter "Muse|ProviderArchitectureGatekeeper|ProviderRegistryTests|InlineUsageDashboardBarColor|CodexBarWidgetProviderTests|ProviderIconResourcesTests|AntigravityPoolDeduplication"` → 133 tests pass. CLI against real Muse Code 1.1.1 logs:

```
== Meta Muse 1.1.1 (local) ==
Tokens: today 287K · 7d 43M

Meta Muse Cost (API-rate estimate)
Today: $0.04 · 287K tokens
Last 30 days: $7.22 · 43M tokens
Estimated from local logs · may differ from your bill
```
`--source api` now correctly errors: `Error: Source 'api' is not supported for muse.`

Rebased onto current `main` (`448f3e645` on top of `c9490ff8c`).

@clawsweeper re-review

steipete:
The local Muse history work is now consolidated and landed in #3340 (e92675860899938d623daf47e5c8ff2503332571), with credit to @Srimi1 and @sanjay3290. It preserves the shipped JavaScript subscription provider and adds bounded local token history, partial-data handling, and token-only presentation. The PR includes six synthetic native screenshots, passing full-suite results, and green macOS/Linux CI. Thank you for the contribution.

PR #3590, opened 2026-09-12
fix(zai): preserve unavailable quota limits

steipete:
Successful z.ai responses with empty or unrecognized quota limits were converted into a real zero-usage window, incorrectly showing 100% remaining. Preserve the absence of a quota measurement while retaining plan details and optional analytics. Explicitly reported zero usage remains visible.

Simplify primary quota selection to the first sorted Coding Plan limit, falling back to MCP. Existing secondary and extra-window behavior is retained. Production code is net negative by one line. This was found during the z.ai backlog review; it does not resolve the separate reset-timestamp report #2871.

Validation: an offline run of the actual plugin fails before the fix and passes afterward for empty and unknown-only responses. New native regressions exercise both plugin engines, analytics preservation, and genuine zero usage for token, credit, and MCP limits. All 33 focused native tests passed across JavaScriptCore and QuickJS. Formatting/lint and independent P0–P2 review passed. Full `make test` passed all 1,099 selections in 92 groups (three normal full-group retries recovered; no group timeouts). [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34705022221) passed, including both macOS shards and Linux x64/ARM64. Local verification used synthetic HOME and the repository credential-suppression flags with the native SwiftPM backend. No live credentials or provider requests are used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.60.1 — Unreleased
 
 ### Fixed
+- z.ai: keep missing quota limits unavailable instead of displaying 100% remaining, preserving real zero usage, plan details, and optional analytics.
 - Antigravity: suppress remote model variants that exactly mirror a known pool and reset, preserve distinct quota rows and their saved visibility, and prefer known usage over reset-only duplicates (#3583). Thanks @hhh2210!
 
 ### Development
--- a/Sources/CodexBarCore/Resources/Plugins/zai.js
+++ b/Sources/CodexBarCore/Resources/Plugins/zai.js
@@ -170,29 +170,28 @@ defineProvider({
       .sort((a, b) => (a.windowMinutes || Number.MAX_SAFE_INTEGER) - (b.windowMinutes || Number.MAX_SAFE_INTEGER));
     const timeLimit = limits.filter((item) => item.raw.type === "TIME_LIMIT").pop() || null;
     const tokenLimit = tokenLimits.length ? tokenLimits[tokenLimits.length - 1] : null;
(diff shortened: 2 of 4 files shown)


steipete:
Merged as [ed336215b4b5](https://github.com/steipete/CodexBar/commit/ed336215b4b5d0ce22b810cad67bcdd1ee286c10).

Empty or unknown-only successful z.ai quota responses now retain an unavailable window instead of reporting 100% remaining. Real zero usage, plan information, optional analytics, and existing Coding Plan/MCP ordering remain covered.

Verification on the reviewed head:

- Offline execution of the actual plugin reproduced the false-zero fallback before the fix and passed afterward.
- `swift test --build-system native --skip-build --no-parallel --filter 'ZaiProviderTests|ZaiReset|ProviderPluginDetailsParityTests'`: 33 tests passed, including both JavaScriptCore and QuickJS cases.
- `make check`: passed with zero lint violations.
- `make test`: all 1,099 selections across 92 groups passed; three ordinary group retries recovered, with no group timeouts. The separate menu-test race exposed during this run is corrected in #3591; the unchanged RPC fixture timing failures are not being called production fixes.
- Independent P0–P2 review was clean, all review threads were resolved, and [exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34705022221) passed both macOS shards and Linux x64/ARM64. The workflow skipped the unchanged musl path.

Local tests used synthetic HOME and repository credential-suppression flags. No live provider calls or real credentials were used. Production source is net negative by one line; changelog and provider documentation are updated. The separate #2871 reset-timestamp report remains open.

PR #3591, opened 2026-09-12
fix(dashboard): preserve disambiguated account labels

steipete:
The web dashboard replaced the server's disambiguated account label with the raw email. Claude accounts sharing an email therefore had identical card titles even when the snapshot correctly provided an alias or organization suffix. Prefer the existing projected label and retain the email and generic fallbacks. This completes the browser side of #3082 without adding new account projections or changing the identity/redaction contract.

The one-use title temporary is removed, making production code net negative by one line. Documentation now describes the existing projected-label contract. A native JavaScriptCore regression executes the production renderer for aliases, shared-email organization labels, redacted labels, and both fallbacks.

Live Chrome proof used the actual page template and identical synthetic snapshots served on loopback, without provider collection, credentials, or real account data. Before: both cards read `shared@example.com`. After: `Work` and `shared@example.com · Acme`. All six label/fallback cases also passed against the real DOM. The 32 focused native renderer/projection tests passed, and independent P0–P2 review is clean. The 10 renderer/menu checks passed, followed by full `make test`: all 1,099 selections in 92 groups passed on the first attempt, with no retries or timeouts (880.3 seconds). After integrating current main and resolving only a changelog conflict, 95 focused integration tests across six suites passed, along with a fresh build, formatting/lint, and P0–P2 review. [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34709550982) passed both macOS shards, Linux x64/ARM64, musl, and lint. Native tests used synthetic HOME, repository credential suppression, and the native SwiftPM backend.

Validation also exposed an existing menu-test race: after releasing a blocked refresh, the test asserted that a retry stayed pending while the real automatic scheduler was free to consume it. It now awaits the second automatic refresh directly, preserving the blocked-state, exactly-two-calls, background-interaction, and cleared-pending checks. This removes 12 test lines without changing production scheduling or increasing any deadline.

| Before | After |
| --- | --- |
| Synthetic shared-email accounts with duplicate titles | Synthetic accounts retaining alias and organization labels |


steipete:
Merged as [3a7ddffb9634](https://github.com/steipete/CodexBar/commit/3a7ddffb963424762c16da354bb9e9628e788e1e).

The browser now displays the existing projected account label, preserving aliases, organization suffixes, and redacted account labels instead of replacing them with the raw email. Email and generic fallbacks remain available. This completes the browser-side behavior established by #3082.

Verification:

- Actual dashboard template in Chrome, served with identical synthetic loopback snapshots: two duplicate `shared@example.com` titles before, `Work` and `shared@example.com · Acme` afterward. All six full/redacted/fallback cases passed against the real DOM; inspected synthetic-only before/after screenshots are in the PR body.
- Initial renderer/projection selection: 32 native tests passed. Final renderer/menu selection: 10 tests passed, including the corrected automatic-retry test.
- Full pre-integration `make test`: all 1,099 selections across 92 groups passed on the first attempt, with zero retries or timeouts (880.3 seconds).
- After integrating current main and resolving only the changelog conflict: fresh native build, 95 focused tests across six suites, `make check` with zero lint violations, and independent P0–P2 review passed. The selection covered the renderer, account projection, automatic menu retry, z.ai, test-process safety, and architecture gates.
- [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34709550982) passed both macOS shards, Linux x64/ARM64, musl, and lint. Final review threads were resolved before the SHA-pinned merge.

The menu fixture previously raced a legitimate automatic retry after releasing its blocked first refresh. It now waits for that actual retry, retaining the blocked-state, exactly-two-calls, background-interaction, and cleared-pending assertions. No deadline was raised. Production source is net negative by one line; the test cleanup removes 12 net lines. Documentation and changelog are updated. Native validation used synthetic HOME and repository credential suppression; no real provider credentials were used.

PR #3592, opened 2026-09-12
refactor: remove obsolete settings and menu code

steipete:
Settings fields still advertised lazy credential loading after credentials had moved into the config snapshot. Every activation callback was nil or called an empty loader, while retired menu and formatting helpers remained compiled without production callers. This removes those obsolete paths and reduces production code by 1,245 net lines.

- Remove 26 empty credential loaders, the field-activation callback API, and the unused token-account loading flag. Explicit reloads and legacy config migration remain authoritative.
- Route 40 credential properties through the existing typed config accessor, preserving provider identity, normalization, persistence, and logging. Specialized setters retain their provider-specific behavior.
- Delete unused menu views, provider helpers, and redundant layout/promotion state. Simplify menu-card projection while retaining account isolation and the existing Codex credit behavior.
- Retarget tests and screenshot fixtures to active rendering, aggregation, and publication paths. Preserve public report formats, framework callbacks, and deliberate test seams; relocate existing architecture allowances without widening them.

Validation:

- `make check`: passed, zero SwiftLint violations across 2,237 files.
- Focused verification: 616 tests across 43 suites passed.
- `PATH="/tmp/codexbar-dedup-swift:$PATH" make test`: all 1,111 selections in 93 groups passed on the first attempt, with no failures, retries, or timeouts (983 seconds).
- Independent complete-candidate Codex review: clean through P2.
- Mechanically verified the original provider/field mappings for all 40 delegated credential accessors and all 398 architecture anchors.

Local Swift tests used the native SwiftPM backend with the installed Xcode toolchain. Validation used fixtures and injected stores; no live account probes, real Keychain reads, or app-bundle launches were performed.

PR #3593, opened 2026-09-12
perf: reuse validated Codex activity store reads

brzvsk:
Codex status and activity refreshes reopened SQLite and decoded unchanged aggregates on every poll. Reuse a bounded read-side store and decoded activity snapshot while its database stamp remains current; this removes repeated integrity walks and decoding without transferring scanner or writer ownership.

Thanks @brzvsk (Nikolai Berezovskii) for the implementation and measured investigation. The [contributor’s pinned terminal proof](https://github.com/brzvsk/CodexBar/blob/e6b28e454204fdd7d611c7b657050799a4718467/.github/pr-proof/codex-activity-warm-read.log) records a private 126 MB, 2,270-file backup: warm median approximately 299 ms → 90 ms, with equal normalized totals. That is contributor evidence for steady-state reads, not a new cold-start claim.

Maintainer hardening covers a discovered availability edge: external compatible parser/schema metadata changes previously left the warm reader returning empty activity indefinitely. A regression fails on the original PR and verifies recovery without losing rows. The reader now requests reopen after its active database operation finishes, preserving handle lifetime and existing compatibility validation. Ordinary concurrent commits invalidate the raced view without rebuilding the database.

A four-entry MRU list replaces the clock-and-dictionary bookkeeping. Further regressions verify purpose widening, transient detailed reports, atomic source-file replacement, per-root eviction, and a real WAL commit during a read transaction. Filesystem reconciliation still runs on every returned view, and detailed report history is not retained.

Production changes total +94 lines. The separately landed cleanup #3592 removed 1,245 production lines, covering this necessary addition under the requested cleanup budget. Changelog and Codex storage documentation are updated; contributor credit is preserved.

Validation on the final source:

- Independent P0–P2 review: clean for the full implementation and the final fixture correction.
- `make check`: zero violations across 2,237 files; native test-bundle build passed.
- `make test` on a separate macOS 27 worker: all 1,111 selections across 93 groups passed on the first attempt, with no retries or timeouts. The storage/read-work suites passed all 139 tests.
- [Final macOS 26/Linux CI](https://github.com/steipete/CodexBar/actions/runs/34720400574): all checks passed.

An earlier local combined storage run timed out under host contention after its 62 read-work tests passed. The preceding run passed the 77 store tests on identical production code; the final full run independently passed both suites together. The source-replacement fixture was corrected to seed pending catch-up metadata, matching the existing read-only reconciliation contract; production behavior was not widened to discover arbitrary changed files.

Refs #3247; follow-up to #3528 and #3312. This addresses repeated activity-read work and does not close the broader CPU/memory investigation.


brzvsk:
@clawsweeper re-review

Pinned the sanitized exact-revision terminal proof, including raw steady-state timings, structural counters, red/green regression results, and recovery coverage. The new head is evidence-only; runtime remains `b69ade32`.

PR #3608, opened 2026-09-13
fix(vertex): preserve OAuth refresh values and validate responses

steipete:
Vertex refresh encoded form values using URL-query rules, so reserved characters could change the credentials sent to Google. It also treated a 200 response without a usable access token as success, retaining an expired token or accepting a blank one while assigning a fresh expiry.

Extract Copilot’s existing form encoder into one internal owner and use it for both providers. Require a nonblank access-token string before constructing refreshed Vertex credentials. The public refresh API, endpoint, timeout, existing error categories, default expiry, opaque token bytes, display-only identity handling, and gcloud-owned credential storage are preserved.

Validation:

- Before the fix, the new Swift regressions failed on form round-tripping and all six missing/unusable-token cases; input-guard and valid-response controls passed.
- An optimized standalone binary compiled the actual production refresher, credentials, identity decoder, HTTP client and form encoder. Its injected transport verified the original Google request and forwarded only to a loopback HTTP server, which independently decoded the form. Before: wire corruption plus missing/null/wrong-type/empty/blank success responses failed. After: all eight wire/response/default-expiry/identity cases passed. No real credentials or Google request were used.
- Focused command: `make test-fast FILTER='VertexAIRefreshTests|CopilotDeviceFlowTests|ProviderHTTPClientTests'` — final verification passed 31 tests in three suites in 9.101 seconds.
- Final isolated review through P2 passed with no accepted/actionable findings. `make check` passed with zero violations across 2,246 files.
- Complete local `CODEXBAR_TEST_RETRY_NON_TIMEOUT_FAILURES=0 make test` passed all 1,117 selections in 102 groups on the first attempt, with zero retries/timeouts (1,464.0 seconds total). The unchanged RPC teardown and subprocess-group suites passed in this run.
- [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34731321499) passed both macOS shards, Linux x64/arm64, musl, lint, and the aggregate check.

Runtime proof outputs:

```text
Before: failed cases wire, missing, null, number, empty, blank; defaults and identity passed.
After: PASS wire; PASS missing/null/number/empty/blank rejection; PASS defaults; PASS identity.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Codex activity: reuse validated SQLite and decoded aggregate state across unchanged refreshes, preserving replacement, writer, and compatibility invalidation (#3593, related to #3247). Thanks @brzvsk!
 
 ### Fixed
+- Vertex AI: preserve reserved characters in OAuth refresh credentials and reject successful responses that omit a usable access token.
 - Account cards: keep Grok's cached local costs out of an account card that has no usage snapshot.
 - Codex: prefer fresh authorized subscription dates over stale cache metadata, and keep empty account cards from inheriting another account's identity, usage, or credits.
 - Provider settings: preserve saved configuration for unavailable plugins when reordering providers, and localize Qoder and Qwen Cloud cookie-cache labels consistently.
--- a/Sources/CodexBarCore/FormURLEncoding.swift
+++ b/Sources/CodexBarCore/FormURLEncoding.swift
@@ -1,0 +1,18 @@
+import Foundation
+
+enum FormURLEncoding {
+    static func body(_ parameters: [String: String]) -> Data {
+        let pairs = parameters
+            .map { key, value in
+                "\(Self.encode(key))=\(Self.encode(value))"
+            }
+            .joined(separator: "&")
+        return Data(pairs.utf8)
+    }
+
+    private static func encode(_ value: String) -> String {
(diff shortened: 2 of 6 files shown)


steipete:
Landed in 237bf22a145641b2afacb088011b5cb54f89e123.

The form round-trip and malformed-success regressions failed before the fix. An optimized binary compiled from the production refresher/credentials/HTTP sources then passed all eight loopback HTTP cases after the fix: reserved-character form values, missing/null/numeric/empty/blank token rejection, default expiry, and display identity. No Google-account request or real credentials were used.

`make test-fast FILTER='VertexAIRefreshTests|CopilotDeviceFlowTests|ProviderHTTPClientTests'` passed 31 tests in three suites. `make check` passed with zero violations in 2,246 files. `CODEXBAR_TEST_RETRY_NON_TIMEOUT_FAILURES=0 make test` passed all 1,117 selections in 102 groups on the first attempt, with zero failures, retries or timeouts. Isolated Codex review was clean through P2. Exact-head CI passed both macOS shards and every Linux/lint/aggregate check: https://github.com/steipete/CodexBar/actions/runs/34731321499.

The public refresh API, default expiry, error categories, valid token bytes, display-only identity policy, and gcloud-owned persistence remain intact. Primary main was pulled and verified clean after merging.

PR #3613, opened 2026-09-13
test: fix shutdown heartbeat fixture ownership

steipete:
The shutdown fixture made the child create the heartbeat file, then expected that file after cleanup deliberately killed the child. A valid hard stop between PID publication and heartbeat opening therefore failed with a missing file.

The test now owns heartbeat initialization. The already-running signal-handler parent publishes the forked child's PID, and the existing scenario runs with immediate and delayed first writes. The observation extends beyond the deliberate delay so a surviving child is still detected. Any fallback cleanup captures and validates the child identity while the root still owns the process-group ID. Production code and its 0.2-second cleanup grace are unchanged; existing readiness, PID and stable-heartbeat assertions remain.

Validation:

- `make test-fast FILTER=SpawnedProcessGroupTests`: 18 tests passed, including both first-write cases.
- `make check`: zero violations in 2,258 files (macOS Bash selected through the task-local Bash-only PATH shim).
- Isolated Codex review through P2: clean.
- Standalone executable linked to 815 actual Core/dependency objects: all four controlled fixture cases passed. The after cases use the Python fixture extracted from this source file.

| Runtime case | Observed result |
|---|---|
| Original fixture, 1000ms delay inserted after PID publication | Missing heartbeat reproduced after valid cleanup |
| Fixed fixture, immediate first write | Heartbeat `4 → 4` after cleanup |
| Fixed fixture, 1000ms delayed first write | Heartbeat `0 → 0` after cleanup |
| Deliberate survivor control, soft signal only | Heartbeat `0 → 24`; the observation detects the survivor |

The survivor control changes only the readiness payload to expose its owned helper PID and deliberately omits hard cleanup before observation; all owned processes are cleaned afterward. Every process case used the same 15-second containment bound. The earlier unrelated ordinary-control launch stall remains disclosed and is not claimed fixed by this observation-file repair. No tests were skipped, isolated, weakened, or given a larger cleanup deadline.

[Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34745707453) passed on `72aac012a0fc9f861e68775caf32aab17294388f`: both macOS test shards, Linux x64/arm64, lint, security, and aggregate checks are green. The existing path filter skipped the unaffected musl build for this test-only diff.

The local full run did **not** pass: it retained all 1,122 selections / 103 groups, passed 46 groups initially, then the unchanged cost-store group exceeded its 180-second deadline. Nine individual retries passed; the tenth, `CostUsageStoreScaleProofTests`, also timed out under the same deadline. Groups 48–103 were not reached. The read-work retry passed 62 tests in 90.932s. A sample of the scale retry showed its 100 MiB SQLite write waiting in a WAL-checkpoint filesystem flush; host load was about 260 on 32 cores. This is supporting environment evidence, not a claim that the fixture repair fixes the cost-store timeout. Total local run: 1,367.8s.

Supplemental unchanged-head validation **passed** on a separate Mac through the SSH provider: all 1,122 selections / 103 groups passed first time, with zero failed groups, retries, or timeouts. Total 1,167.8s (373.1s discovery/build, 794.7s execution). Hydration was disabled; the run used the same Swift compiler, four build jobs, suppressed test Keychain access, and unchanged deadlines. The unchanged read-work and scale suites passed in 5.066s and 4.368s respectively, including the full 100 MiB write. This preserves the local failure record without attributing it to the heartbeat repair. No changelog entry for this test-only correction.


steipete:
Merged as 9ac949917d9bb04f92910e666a9a98ec018a9ee4 after clean isolated review through P2, [all applicable exact-head CI checks](https://github.com/steipete/CodexBar/actions/runs/34745707453), and compiled process proof on 72aac012a0fc9f861e68775caf32aab17294388f.

The test now creates the heartbeat before launching the process group. The signal-handler parent publishes the child PID, and immediate/delayed first-write cases retain the same 0.2-second production cleanup grace. The controlled executable reproduced the original missing-file race, showed stable heartbeats after both fixed cleanup cases, and detected a deliberately surviving child. All 18 focused tests and lint passed; no production code or test assertion was weakened.

The local full run did not pass: 46 groups passed, then a 180-second cost-store batch timeout triggered individual retries. Nine retries passed; the scale suite also timed out during its 100 MiB SQLite write. Its sampled stack was in a WAL-checkpoint filesystem flush, with host load around 260 on 32 cores. Groups 48–103 were not reached. This does not claim a root fix for that separate environment-sensitive timeout. Hosted macOS full validation passed on the exact head; an additional unchanged-head full run is underway on a separate Mac and will be recorded when complete.

steipete:
Supplemental full validation of the unchanged PR head 72aac012a0fc9f861e68775caf32aab17294388f has passed on a separate Mac through the SSH validation provider.

`make test` retained all 1,122 selections / 103 groups: 103 first-pass successes, zero failed groups, zero retries, and zero timeouts. Total 1,167.8s (373.1s discovery/build and 794.7s execution), using the same Swift compiler, four build jobs, hydration disabled, test Keychain suppression, and the original 180-second deadlines.

The cost-store suites that encountered the local host problem also passed unchanged: 62 read-work tests in 5.066s and the scale suite in 4.368s, including the complete 100 MiB write. The earlier local failure remains recorded; no test, corpus, assertion, timeout, or SQLite durability setting was changed to obtain this result.

PR #3616, opened 2026-09-13
fix(menu): re-measure merged-menu rows when a tab swap changes card height

### LINKED ISSUE #3549: Merged menu keeps a stale height after switching provider tabs (macOS 26 table-backed menus never shrink)

zenibako:
Fixes #3549.

Switching merged-menu tabs could leave a large gap around a short card or clip content and actions. The cached swap preserved an attached AppKit row while replacing its payload with a differently sized card; AppKit retained the row's old measurement.

The in-place swap now requires matching measured card heights. Different-height cards use the existing insert-before-remove path, so AppKit measures the incoming row. Equal-height swaps retain their attached views. Nearby tail, highlight, and metadata reconciliation is simplified; the complete PR is net **-2 production lines**.

Thanks @zenibako (Chandler Anderson) for isolating the cause, implementing the fix, and supplying the native diagnostic. The diagnostic now requires materialized card geometry, validates the intended tab selection, and compares row height against both the view frame and independently measured intrinsic height. Offscreen virtualized rows remain valid.

### Native before/after proof

Same six-provider synthetic fixture on macOS 26.6.2, opened through the production status-item menu with cost sections and a scrollable Overview. Both test bundles were signed with the matching Developer ID. The before control disables only the height check; the final source was restored and rebuilt afterward. No real accounts or provider credentials were used.

| Overview → Claude | Height check disabled | Fix enabled |
|---|---:|---:|
| Table / menu height | 735 / 516 pt | 516 / 516 pt |
| Card row / content height | 341 / 122 pt | 122 / 122 pt |
| Native assertions | Fail on stale geometry | All 11 transitions pass |

| Before | After |
|---|---|
| Before: retained tall row creates gaps and clips actions | After: content-sized Claude card and all actions |

### Validation

Reviewed head: `5af73dc90f6c655927f1a41b8781f6a061cf4f69`.

- Signed native diagnostic: 38.25 seconds, all 11 phases, zero failures. Negative control: 44.28 seconds, fails on the original 735/516 and 341/122 measurements.
- `make test-fast FILTER='StatusMenuTests|MenuRowContainerSizingTests'`: 182 tests in two suites passed.
- `make check`: zero violations in 2,276 files. Independent review and isolated Codex review completed; no accepted actionable findings remain.
- `make test` was attempted locally: 47 groups passed before an unrelated `CostUsageStoreScaleProofTests` 180-second timeout stopped the run. This is not a full local-suite pass.
- [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34764650337): passed: both macOS test shards, all three Linux builds, lint, and the aggregate gate.


steipete:
Landed as caad1ca38c237fc77426ad06cf56a2daa1dbbcb9. Thanks @zenibako for the root-cause isolation, fix, and native fixture.

The final head passed 182 focused tests (`make test-fast FILTER='StatusMenuTests|MenuRowContainerSizingTests'`), `make check`, and both complete macOS CI shards plus Linux builds in https://github.com/steipete/CodexBar/actions/runs/34764650337. The local full-suite attempt stopped at an unrelated scale-test timeout; that limitation is recorded in the PR body.

The matching Developer ID-signed macOS 26.6.2 diagnostic passed all 11 transitions in 38.25 seconds. Disabling only the height check reproduced the original table/menu mismatch (735/516 points) and card row/content mismatch (341/122 points), failing the diagnostic in 44.28 seconds. The inspected before/after screenshots in the PR contain only synthetic fixture data. Production changes are net -2 lines, and the Unreleased changelog credits the contributor.

Issue #3617, opened 2026-09-13
Codex cost: day blanked by unresolvedRowGroups on 0.60.1 with Codex CLI 0.154 (priority rows + repeated token_count snapshots)


BUKOWSKIREAL:
## Summary

On CodexBar 0.60.1 with Codex CLI 0.154.0, today's Codex cost is suppressed to `—` while the day's token count (198M) still displays. The day is also missing from the 30-day total and chart. The trigger appears to be the `unresolvedRowGroups` fail-closed path that #3150 explicitly left out of scope.

## Environment

- CodexBar: 0.60.1
- Codex CLI: 0.154.0 (Codex desktop app sessions, multi-agent v2)
- macOS, Apple Silicon

## Evidence

For the affected day the local scan produces ~951 usage rows for model `gpt-6-astra`:

- Canonical (deduplicated) day total: **198,456,263 tokens** — displayed correctly.
- Sum over retained per-file rows: **~448M tokens** (~2.26x canonical) — fails the row-ownership checks, so the group is distrusted.
- **628 rows carry `pricingMode: priority`** — Codex 0.154 stamps `service_tier: priority` on Fast turns directly in the rollout rows.
- With the rows distrusted, the group falls back to the aggregate path, which returns `nil` for this model (long-context tier — the same deliberate refusal as #2858), so the whole day's cost is suppressed while its tokens remain.

Two contributors to the row-sum inflation are visible in the raw rollouts:

1. **Repeated `token_count` snapshot rows** (the #1062 mechanism): in one rollout, 2,162 `token_count` events include **167 rows where `total_token_usage` is unchanged from the previous row but `last_token_usage` is non-zero**. Summing `last` across those rows alone adds ~53.4M phantom tokens.
2. **Fork/subagent rollouts** (the #2754 mechanism): 14+ rollout files whose `session_meta` carries `forked_from_id` / `source.subagent.thread_spawn` appear to contribute copied-parent rows to the per-file mode-split maps.

CLI output for reference:

```
Today: — · 198M tokens
Last 30 days: $2,823.16 · 1.9B tokens
```

## Expected behavior

The day should be priced from the trusted rows (or degrade to a priced partial estimate) rather than having the entire day's cost suppressed while its token count remains.

## Related

- #1062 — repeated `token_count` rows with unchanged `total_token_usage`
- #2754 — mode-split cost path re-inflates deduplicated fork usage
[...]

PR #3618, opened 2026-09-13
Fix Codex day cost suppressed by re-emitted usage rows

BUKOWSKIREAL:
**Held — the recovery still lacks sufficient event-identity evidence and must not be merged.**

This proposal repairs re-emitted Codex pricing rows when deduplicated token components exactly match a file's canonical aggregates. The refreshed candidate adds safeguards for zero-token charges, unpriced markers, observed priority metadata, and the current typed persistence planner. Thanks @BUKOWSKIREAL for the report and original repair.

The maintainer's earlier [evidence requirement](https://github.com/steipete/CodexBar/pull/3618#issuecomment-5654507027) remains unresolved. Aggregate equality cannot prove the request partition: stored `A,A,A,B,B` with 200K-input A rows, 400K-input B rows, and a 600K canonical total fits both three A requests and A+B. Those partitions can price differently. Conflicting monetary or pricing evidence also must remain unresolved instead of choosing a preferred copy. The candidate does not yet establish those guarantees.

A controlled native-log comparison through the signed CodexBarCLI confirmed the partition ambiguity for the released GPT-5.4 model: three 200K-input requests produce a $1.50 API-equivalent estimate, while a 200K/400K pair produces $2.50. Both contain exactly 600K input tokens, zero cached/output tokens, and use the same synthetic turn/timestamp. This is a synthetic counterexample, not a claim about the reporter's actual source file, and no inference requests were sent.

Required before landing: ordered source evidence or stronger canonical event identity, conflicting-evidence regressions, and verification that the affected store still needs this recovery after #3620. No change from this PR is being landed through the menu-position PR #3723.

Proof completed, with its limits:
- 114 focused tests passed. Three added regressions detected discarded priority metadata, a lost zero-token charge, and a lost unpriced marker before those local safeguards were added.
- `make check` passed. Independent review of the supplied patch reported no P0–P2 findings, but that result did not resolve the request-partition counterexample above.
- A freshly signed CLI recovered an isolated synthetic four-row cache to two rows, retained 3,300 tokens, restored a cost estimate, and retained that result on a second process run. That controlled fixture does not establish safe recovery for ambiguous real histories.
- The combined local test candidate passed all 1,180 selections in 107 groups without retries or timeouts. Passing tests do not substitute for the missing evidence boundary.

Related to #3617; the earlier producer/persistence repair landed separately in #3620.


steipete:
I found a separate, reproducible producer/persistence defect in #3620: normal rescans retain obsolete request rows, and SQLite can reuse changed history when counts and offsets match. Its regression now covers both the in-memory report and a later persisted read.

This recovery path still needs a stricter evidence boundary before I can land it. `codexUsageRowEventKey` excludes pricing fields, while `preferredCodexUsageRow` picks populated fields and then the larger event index. If two copies agree on token components but disagree on `knownCostNanos`, `pricingMode`, or `pricingModel`, matching canonical token totals does not establish which monetary evidence is correct. Interleaved A/A′/B/B′ can pass the exact-total check while changing the chosen cost when their indexes change. Please keep conflicting-evidence groups unresolved instead of selecting a winner. Missing timestamps/turn identities also should not be treated as positive evidence that two events are the same.

The affected-store measurements are useful. Now that #3620 has landed, the next check is whether that store still requires this recovery, ideally paired with sanitized ordered source rows. That will separate remaining old-cache repair from a still-active row producer defect. I am keeping this PR open; it is not being marked superseded by the narrower rescan fix.

Further review found an ambiguity even if copies have identical pricing fields and nonempty timestamps/turn IDs. Consider stored rows `A,A,A,B,B`, where A is a 200K-input request and B is a 400K-input request, with canonical input of 600K and zero cached/output tokens. Copies within each class can match on every proposed fingerprint field except event index.

Those aggregates fit both three legitimate A requests plus obsolete B rows, and one A plus one B with replayed copies. Deduplication chooses A+B and passes the exact component-total check. But three 200K requests and a 200K/400K pair can have different costs under request-size tiering. Matching aggregate tokens does not prove the request partition. This is a reasoned counterexample, not a claim about the reporter's actual source file.

The remaining repair needs ordered source evidence or stronger canonical event identity; stricter pricing-field equality alone does not establish a safe reconstruction.

steipete (changes requested):
The earlier evidence boundary remains unresolved: aggregate token equality cannot establish canonical event identity or request partition. A,A,A,B,B with A=200K input, B=400K input and canonical 600K fits both three A requests and A+B, which can price differently. Conflicting monetary/pricing fields must also remain unresolved rather than selecting a preferred copy. The new safeguards and passing tests do not resolve these cases. Please supply ordered source evidence or a stronger canonical event identity and cover the ambiguous/conflicting groups before landing. The updated PR body records completed checks and these limits.

A controlled native-log comparison through the signed CodexBarCLI confirmed the partition ambiguity for the released GPT-5.4 model: three 200K-input requests produce a $1.50 API-equivalent estimate, while a 200K/400K pair produces $2.50. Both contain exactly 600K input tokens, zero cached/output tokens, and use the same synthetic turn/timestamp. This is a synthetic counterexample, not a claim about the reporter's actual source file, and no inference requests were sent.

Issue #3586, opened 2026-09-12
Antigravity: agy CLI source broken by agy 1.2.2 — local server now enforces CSRF token


Haven2026:
## Summary

CodexBar's `agy` CLI usage source (`--source cli`, and the CLI leg of `auto`) broke today after `agy` auto-updated to **1.2.2**. Every local probe is rejected, and the provider card falls back to `Offline · N conversations`.

The same setup worked in the morning on the previous `agy` build; it broke in the afternoon, right after the on-disk binary was replaced (mtime 12:58 local).

## Root cause (verified)

`agy` 1.2.2's embedded local server now enforces a CSRF token on **all** of its ports. CodexBar's CLI-HTTPS probe sends none (`csrfToken: ""` in `AntigravityStatusProbe+LocalEndpoints.swift`), so every request fails auth.

Repro — spawn `agy` 1.2.2, take its two listening ports, probe both schemes:

```sh
$ curl -k -X POST https://127.0.0.1:49884/exa.language_server_pb.LanguageServerService/GetUserStatus \
    -H "Content-Type: application/json" -H "Connect-Protocol-Version: 1" -d '{}'
{"code":"unauthenticated","message":"missing CSRF token"}

$ curl -X POST http://127.0.0.1:49885/exa.language_server_pb.LanguageServerService/GetUserStatus \
    -H "Content-Type: application/json" -H "Connect-Protocol-Version: 1" -d '{}'
{"code":"unauthenticated","message":"missing CSRF token"}
```

`codexbar usage --provider antigravity --source cli --log-level debug` shows the full loop: CodexBar spawns `agy`, enumerates its ports, then retries every endpoint until the readiness deadline:

```
error=Antigravity session expired. Restart Antigravity and retry.
  path=/exa.language_server_pb.LanguageServerService/RetrieveUserQuotaSummary
  port=49884 scheme=https source=cli-https
error=A TLS error caused the secure connection to fail. port=49885 ...
```

then emits `{"source":"offline"}`.

## The token is not externally obtainable (as far as I can tell)

- It appears to be generated in-process and passed only to `agy`'s own sidecar children via the `ANTIGRAVITY_CSRF_TOKEN` env var at spawn time.
- Launching `agy` *with* `ANTIGRAVITY_CSRF_TOKEN` set does **not** make the server accept that value — it still answers `invalid CSRF token`, i.e. the server generates its own.
[...]

PR #3621, opened 2026-09-13
Fix Antigravity native usage probe

chenzhan4321:
## Summary

- Query `agy -p /usage --output-format json` before opening the legacy local HTTPS session, allowing its bounded 120-second eligibility check.
- Parse only successful, zero-turn, zero-token responses and preserve every quota group/window.
- Keep selected OAuth accounts on the existing identity-guarded path because native `/usage` does not provide an email.

## Validation

- `swift build --build-system native --sdk /Library/Developer/CommandLineTools/SDKs/MacOSX26.5.sdk --product CodexBarCLI`
- Real local read-only quota path: `CodexBarCLI usage --provider antigravity --source cli --format json --verbose` returned the Gemini and Claude/GPT five-hour and weekly windows.
- `swiftformat --lint` on the changed Swift files and `git diff --check`.

`make test` cannot start in this checkout because the active developer directory is Command Line Tools and lacks `actool`. `make check` completed its 101 Python tests and format checks, then SwiftLint could not load `sourcekitdInProc.framework` for the same missing-full-Xcode environment.


chenzhan4321:
Withdrawing this implementation PR. Keeping the report and validation evidence in the existing issue #3586: https://github.com/steipete/CodexBar/issues/3586#issuecomment-5654719652 . No merge is requested.

PR #3626, opened 2026-09-13
fix(chart): align the cost metric picker with content

elijahfriedman:
The Token/Cost segmented picker rendered at its natural width in the center of a wider reserved frame, leaving a gap before the chart content’s trailing edge. Align the control to that frame’s trailing edge while retaining the width reserved for localized labels.

The native regression now requires the actual NSSegmentedControl and checks both 296- and 360-point menu widths. Remove an unreachable empty-data accessibility branch inside the chart’s nonempty path; the real empty-state text is unchanged. Production code is net −2 lines. Thanks @elijahfriedman for the original fix and native regression; contributor attribution is preserved.

Validation:
- Before control: the native geometry assertion fails at both widths; restoring the alignment passes.
- Final focused chart and architecture checks: 86 tests across two suites pass.
- `make check`: zero violations. Independent review found no actionable P0–P2 issues.
- The images below are native NSHostingView renders using synthetic daily data and an explicit light background. They show the 296-point fixture; the regression also checks 360 points.
- Native full suite: all 1,133 selections across 104 groups passed after the runner isolated one storage-scale group. A preexisting write-timing assertion exceeded its 30-second budget; all 12 isolated retries passed. [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34790988562) passed, including both macOS shards, Linux x86_64/ARM64 and musl.

Before: native picker floats inside its reserved width (synthetic data)

After: native picker aligns with the content edge (synthetic data)


steipete:
Maintainer verification for the screenshot-access caveat: I inspected the locally generated native before/after captures attached to this PR. They contain only the production chart view with synthetic daily data and a light background. The actual NSSegmentedControl geometry fails before the fix and passes afterward at both 296- and 360-point menu widths.

The retained fixed width and trailing alignment are supported by that proof, the 86 focused chart/architecture tests, and independent code review. No new product decision is needed for this alignment correction. Final CI remains the landing gate.

steipete:
Landed as bb1c47195be5aaf1d8a3f50f698b39286d064549, preserving @elijahfriedman’s attribution. Main is synchronized and its full tree matches the reviewed merge preview, including the independently landed LiteLLM fix.

Verification used contained native test environments:
- Before/after actual NSSegmentedControl geometry at 296- and 360-point menu widths; the old frame failed both cases, and the fixed frame passed. The attached pictures are inspected native chart renders with synthetic data.
- `swift test --filter 'CostHistoryChartMenuViewTests|ProviderArchitectureGatekeeperTests'`: 86 tests across two suites passed.
- `make check`: zero violations. Independent code and final changelog reviews found no actionable P0–P2 issues.
- `make test`: all 1,133 selections across 104 groups completed successfully. One storage-scale write-timing failure recovered through isolated retries; all 12 isolated selections passed. Total 1,614.5 seconds.
- [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/34790988562) passed lint, both macOS shards, Linux x86_64/ARM64, musl and the aggregate gate.

The reserved localized-label width and actual empty-state accessibility remain intact. Production code is net two lines smaller, and the changelog is updated.

PR #3546, opened 2026-09-10
Recognize unexpired Grok Bot trial usage

vincent-peng:
Related to #3530.

Cursor's Grok Bot projection currently requires `hasNonZeroIncludedLimit == true`, which hides the reported trial percentage when that flag is false. This draft also accepts a top-level `sandTrialExpiresAt` strictly later than the snapshot time. Exhausted, unexpired trials remain visible; ended or malformed trial data falls back to the existing paid-allowance rule.

Trial windows omit duration and reset metadata. This avoids calling a seven-day expiry a weekly reset or creating a recurring weekly utilization series. Paid weekly windows, monthly Cursor bars, legacy request plans, and authentication/fetch behavior are preserved.

### Draft blocker: verify the web response

The official [Grok Bot 0.47.0 distribution](https://downloads.cursor.com/grokbot/stable/c1e7d7a46549956d25f53e9c0b9f59666e03aa3a/linux/x64/grok-bot_0.47.0_amd64.deb), linked from [Cursor's download page](https://cursor.com/download/bot), confirms a top-level trial-expiry field in the native `GetSandUsageStatus` RPC and uses a strict future-expiry check, independently of usage availability or cancellation eligibility. Its trial presentation takes precedence over included weekly timing.

That native RPC evidence does **not** establish the JSON shape returned by CodexBar's `/api/dashboard/get-sand-usage-status` endpoint. Issue #3530 abbreviates the trial fields as `sandTrial: <active trial data present>`. The new REST fixture is synthetic, and the decoder deliberately does not guess a nested trial schema. A sanitized response from that exact endpoint is needed before marking this ready or claiming #3530 fixed; if the shape differs, the decoder and fixture must be updated.

### Validation

- Focused parser, probe, card-model, and menu-selection tests: 104 tests across six suites passed.
- `make check`: passed, zero lint violations.
- `make test`: all 1,065 selections in 89 groups passed on the first attempt, with no retries or timeouts. Keychain access was suppressed and live-test opt-in was unset.

Focused command (with Keychain and local-session isolation enabled):

```sh
CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 \
CODEXBAR_ALLOW_TEST_KEYCHAIN_ACCESS=0 \
CODEXBAR_TEST_CODEX_FILE_ISOLATION=1 \
CODEXBAR_TEST_SESSION_FILE_ISOLATION=1 \
[...]

PR #3629, opened 2026-09-13
fix(cursor): show Grok Bot trial usage

### LINKED ISSUE #3530: Cursor Grok Bot trial usage is hidden when hasNonZeroIncludedLimit is false
## Description

CodexBar does not display Cursor Grok Bot usage for an active Grok Bot trial, even though Cursor's Sand usage endpoint returns a valid usage percentage.

The Grok Bot desktop application currently shows approximately 14% trial usage, while Cursor's API reports approximately the same value (~13.55%). However, the Grok Bot usage window is absent from CodexBar.

## Environment

- CodexBar: `0.57.0`
- Grok Bot: `0.47.0`
- Cursor plan: Free + active Grok Bot Trial
- macOS
- Cursor account authentication works correctly in CodexBar

## Observed behavior

Grok Bot desktop app:

```text
Current plan: Trial
Trial usage: ~14%
```

A sanitized response from:

```text
POST /api/dashboard/get-sand-usage-status
```

contains a valid usage percentage:

```text
usagePercent: ~13.55
hasAvailableUsage: true
hasNonZeroIncludedLimit: false
sandTrial: <active trial data present>
```

CodexBar successfully retrieves the normal Cursor usage for the same account:

```text
Total: 100% left
Cursor: 100% left
Third Party: 100% left
Plan: Cursor Free
```

but no `Grok Bot` rate window is displayed.

The same result was reproduced using both Cursor source paths:

```bash
codexbar usage --provider cursor --source auto --verbose
codexbar usage --provider cursor --source web --verbose
```

so this does not appear to be caused by browser/session source selection.

## Suspected cause

`Sources/CodexBarCore/Providers/Cursor/CursorSandUsage.swift` currently requires:

```swift
guard self.hasNonZeroIncludedLimit == true,
      let usagePercent = self.usagePercent else {
    return nil
}
```

For this trial account, Cursor provides a valid `usagePercent`, but:

```text
hasNonZeroIncludedLimit = false
```

so CodexBar drops the Grok Bot rate window entirely.

CodexBar already treats Grok Bot/Sand as an extra Cursor usage window and retrieves it from `/api/dashboard/get-sand-usage-status`. The current guard therefore appears to exclude trial quotas.

## Expected behavior

When Cursor reports an active Grok Bot trial together with a valid `usagePercent`, CodexBar should display the Grok Bot usage window, for example:

```text
Grok Bot: ~14% used
Trial ends / resets: <timestamp>
```

even if `hasNonZeroIncludedLimit` is `false`.

## Actual behavior

No Grok Bot usage is shown at all.

## Reproduction

1. Use a Cursor Free account with an active Grok Bot trial.
2. Confirm that Grok Bot → Usage & Billing shows non-zero Trial usage.
3. Run:

   ```bash
   codexbar usage --provider cursor --source auto --verbose
   ```

4. Observe that normal Cursor usage is returned but no Grok Bot window is present.
5. Inspect `/api/dashboard/get-sand-usage-status` using the same authenticated Cursor session.
6. Observe that `usagePercent` is populated while `hasNonZeroIncludedLimit` is `false` and trial data is present.

## Possible direction

It may be necessary to distinguish an active trial from an account with no Grok Bot allowance.

For example, the presence of active `sandTrial` data and a valid `usagePercent` could allow the rate window to be emitted even when `hasNonZeroIncludedLimit == false`.

I have intentionally not included cookies, tokens, account identifiers, or other authentication data.

JackHo12:
Cursor's current Grok Bot response uses `includedLimitZero` and `sandTrialExpiresAt`; requiring the removed allowance field hides valid usage, including active trials. Decode the current fields while retaining the released initializer and older payload fallback. The current allowance flag takes precedence.

A trial-only allowance is visible only while its parsed expiry is strictly in the future, including when it is exhausted. It has no recurring reset or duration because expiration does not replenish quota. An affirmative paid allowance retains its existing recurring timing even if a future trial date is also present. This preserves paid-account behavior without assuming mixed-state server semantics that the reported payload does not establish. Monthly Cursor bars, legacy request-plan suppression, and optional-fetch failure handling remain intact.

The shared duration calculation and snapshot clock simplify the implementation: production Swift is +19/−21, net −2 lines. Thanks @JackHo12 for the current response sample, implementation, and live trial screenshot. Contributor credit is preserved. Thanks also @vincent-peng for the expiry and timing investigation in #3546.

Validation on `95142fea827628f2fc578f34852cc09468595a5c`:

- `make test-fast FILTER='Cursor|ProviderArchitectureGatekeeperTests'`: 282 tests across 28 suites passed. Coverage includes current and older payloads, field precedence, strict expiry boundaries, malformed dates, missing percentages, exhausted trials, paid-plus-future-expiry timing, and fetch-to-snapshot projection for both paid and trial responses.
- `make check`: zero lint violations. Independent P2 review found no actionable issues after the paid-timing correction.
- The contributor's current REST payload and [live trial screenshot](https://github.com/steipete/CodexBar/pull/3629#issuecomment-5662521274) establish the reported failure and restored bar. Maintainer validation uses synthetic parser and transport fixtures; no live account credential probe was performed.
- Full local `make test` passed all 1,140 selections in 104 groups with zero retries or timeouts (1,002.1 seconds). [Final-head CI passed](https://github.com/steipete/CodexBar/actions/runs/34900601741), including both macOS shards and all Linux builds.

Fixes #3530.


JackHo12:
Hi @h3x89 and maintainers,

While debugging this locally, we discovered that Cursor has recently changed the response format of the `/api/dashboard/get-sand-usage-status` endpoint!

The original fields reported in this issue (`hasNonZeroIncludedLimit` and `sandTrial`) have been completely removed from the payload by the server. If we don't adapt to the new schema, the Grok Bot extra rate window will fail to display for **all** users, not just those on trials, because the old fields will always decode as `nil`.

The new JSON payload looks like this:
```json
{
  "currentPeriodStart": "2026-09-10T00:00:00Z",
  "nextResetTimestampUtc": "2026-09-21T09:12:32.776Z",
  "usagePercent": 12.3,
  "includedLimitZero": true,
  "hasAvailableUsage": true,
  "sandTrialExpiresAt": "2026-09-21T09:12:32.776Z"
}
```

In my latest commit, I've removed the obsolete fields and refactored `CursorSandUsageStatus` to rely on the new `includedLimitZero` and `sandTrialExpiresAt` fields. I have also updated all the associated test fixtures to match this new reality.

JackHo12:
Here is the live proof. As you can see in the screenshot below, the
  Grok Bot usage bar is now appearing correctly for my active trial account! It handles the new API schema fields (`includedLimitZero` and `sandTrialExpiresAt`) perfectly.
image_2026-09-14_122330476

JackHo12:
The inverted boolean test fixtures have been fixed. Thanks for catching that!

@clawsweeper re-review

h3x89:
@JackHo12 A few days ago, I decided to pay $20 for Cursor, and it looks like this now:
Screenshot 2026-09-14 at 16-36-01

steipete:
Landed as 6ffc81f6ea7d9af6abd556984be39a48a10817b9. The merged tree matches reviewed/tested head 95142fea827628f2fc578f34852cc09468595a5c, and main is synchronized and clean. Thanks @JackHo12, and thanks @vincent-peng for the expiry/timing investigation.

The fix decodes the current REST allowance and expiry fields, retains older payload support, requires a valid future expiry for trial-only usage, and keeps exhausted trials visible. Regular included allowances retain recurring timing even when a trial date is present. Production Swift is +19/−21, net −2 lines; changelog and provider documentation are updated.

Proof: `make test-fast FILTER='Cursor|ProviderArchitectureGatekeeperTests'` passed 282 tests/28 suites. `make check` passed with zero lint violations. Full local `make test` passed all 1,140 selections/104 groups in 1,002.1 seconds with no retries or timeouts. Independent P2 review is clean, and [final-head CI passed](https://github.com/steipete/CodexBar/actions/runs/34900601741), including both macOS shards and Linux builds.

The contributor's current REST sample and live screenshot establish the reported case. Maintainer tests use isolated synthetic parser/transport fixtures, including paid-plus-future-expiry and exhausted-trial controls; no live account credential probe was performed.

Issue #2, opened 2025-11-16
Homebrew


pygarap:
Can you make it available on Homebrew?

steipete:
Done. brew install --cask steipete/tap/codexbar

PR #3630, opened 2026-09-13
fix(keychain): retry transient no-UI preflight failures

ysyyork:
A transient no-UI Keychain preflight failure could block an already-authorized read for the rest of a refresh cycle. Retry only `temporarilyUnavailable`, with three total attempts and a 30 ms wait before each retry. Stable access grants, ACL rejection, missing items, and other failures return immediately. The operation memo stores the final outcome, so repeated checks do not add retries.

This applies to generic-password preflights used by browser storage, credential repair, and CodexBar caches. It adds at most two requested 30 ms waits per distinct inconclusive preflight, plus Security call time. Queries remain strictly non-interactive, background browser reads still require an explicit grant, and existing denial cooldowns and disabled-Keychain behavior remain intact.

The final scope is the transient preflight retry. Credential-cache retention, persisted rejection fingerprints, and authentication error classification keep their current behavior. Thanks @ysyyork for the original fix and regression coverage.

Validation: 51 focused tests passed across preflight retry, no-UI query construction, background browser gating, and cache handling; `make check` and independent review through P2 passed. Scripted outcomes and injected delays verify retry limits and memoization without real Keychain reads or sleeps. The full local suite passed all 1,247 selections across 114 groups with zero retries or timeouts. Exact-head macOS/Linux CI passed: https://github.com/steipete/CodexBar/actions/runs/35485146791.


ysyyork:
Pushed a fix for the finding from this review pass:

**[P2] Track skipped browsers separately from unrelated empty reads:** confirmed, and worse than it first looked. The aggregate `recoveryAttempted` boolean was set true the moment *any* browser candidate was genuinely reached — including Safari, which needs no Keychain decryption and sits in the default import order, so it's normally reached and tried even when Chrome is skipped by the gate. An unrelated empty Safari read was silently reverting to the original "sign in to claude.ai" message, exactly as flagged.

While fixing it I found the fix itself needed a second pass: my first attempt re-checked `BrowserCookieAccessGate.shouldAttempt` inside the loop, but `lazyCookieImportCandidates` (which the loop iterates) already filters out anything that check rejects — so the re-check was dead code, unreachable by construction. Fixed by inlining the browser-list filtering instead of going through that helper, so "not installed" and "installed but gate-rejected" can be told apart explicitly, tracked as a new `anySkippedByGate` signal set only in the latter case.

Also had to fix the two *existing* recovery tests along the way: they were passing for the wrong reason. A default-home `BrowserDetection` is deliberately safety-suppressed under `swift test` (to keep tests off real browser cookies), which makes every browser "unavailable" outright — the old aggregate flag couldn't distinguish that from a real gate skip, so the tests never actually exercised the gate at all. Moved both to an isolated synthetic home (matching the existing pattern in `ClaudeWebBrowserFallbackTests`), and added the specific mixed-outcome regression this review called for (Chrome skipped by an inconclusive preflight, real Safari read on the isolated profile comes up empty).

All `ClaudeWebBackgroundRecoveryTests` + `KeychainAccessPreflightRetryTests` pass, broader `Claude|BrowserCookie|Keychain` sweep green aside from confirmed pre-existing flakes unrelated to these files, `swiftformat`/`swiftlint --strict` clean.

On the "real behavior proof" ask: I don't have a way to attach runtime logs/recordings from an actual macOS run programmatically — that would need to come from a human with the packaged app running. Happy to have @ysyyork provide that if wanted.

ysyyork:
Fixed the Revision 2 finding (\`ClaudeWebAPIFetcher.swift:1463-1468\`) in `8869eb306`.

**Bug:** `fetchUsageSerialized` wrapped *both* session-key extraction and the subsequent `fetchUsage` API call in the same `do`/`catch`. If a Keychain preflight skipped one browser (e.g. Chrome, inconclusive) but a *different* browser (e.g. Safari) genuinely recovered a real session key, and the API then rejected that recovered session with a confirmed 401 (e.g. the cookie itself was actually revoked), the error was wrongly reclassified as `cachedSessionUnverifiedInBackground` — because `anySkippedByGate` was still `true` from Chrome's earlier skip, even though the failure had nothing to do with the gate at that point.

**Fix:** Split the `do`/`catch` so only `extractSessionKeyInfo`'s own failure is eligible for the `anySkippedByGate` reclassification. Once a session key is actually recovered, `fetchUsage`'s errors now propagate unmodified — a confirmed post-recovery auth failure is reported as what it is, not papered over as "unverified."

Also reordered the test-only browser-override check to run *after* the `BrowserCookieAccessGate.shouldAttempt` gate check (safe — it's `#if DEBUG` only, never active in production). This lets a test combine "browser genuinely skipped by the gate" with "a different browser's read satisfied by an override" in the same scenario, which was previously impossible since the override short-circuited the gate check entirely.

**New test:** `background refresh preserves a confirmed auth failure after a real session is recovered` — Chrome is skipped via an inconclusive Keychain preflight (`.temporarilyUnavailable`), Safari recovers a real session via override, and the HTTP stub rejects that recovered cookie with a real 401. Asserts the original `.unauthorized` propagates, not `.cachedSessionUnverifiedInBackground`.

**Verification:**
- `swift test --filter ClaudeWebBackgroundRecoveryTests` — 7/7 pass
- `swift test --filter "KeychainAccessPreflightRetryTests|KeychainCacheStoreTests"` — 32/32 pass
- `swift test --filter "Claude|BrowserCookie|Keychain"` — full sweep passes except the same two pre-existing, unrelated flaky tests noted in the previous round (`ClaudeOAuthPromptCoalescingTests`, `ClaudeCLISessionTests`'s "overlapping profile captures serialize PTY ownership"), which pass cleanly in isolation and don't reference any file touched by this PR.
- `swiftformat Sources Tests` / `swiftlint --strict` (via `Scripts/lint.sh`) — clean

ysyyork:
Fixed the new Revision 3 finding (\`ClaudeWebAPIFetcher.swift:1462\`) in \`9865c958d\`.

**Bug:** confirmed, and this was the real remaining piece of the original bug report. The unverified-session reclassification only fired inside \`if let invalidatedCacheError { if anySkippedByGate { ... } }\` — meaning it only applied when this specific refresh cycle had a cached cookie to invalidate in the first place. But that same invalidation clears the cache (\`clearIfCurrent\`), so a **second** consecutive background refresh under the exact same still-inconclusive Keychain preflight has nothing left to invalidate: \`invalidatedCacheError\` is nil, the gate skip falls through the outer \`if let\` entirely, and extraction's own \`noSessionKeyFound\` propagates as-is — which \`isClaudeWebSessionRefreshFailure\` doesn't recognize, so the misleading "Sign in to claude.ai" message reappears on the very next refresh. This is exactly the intermittent behavior originally reported: it looks "fixed" for one cycle and then comes back.

**Fix:** check \`anySkippedByGate\` first, unconditionally — before even looking at \`invalidatedCacheError\`. Whether recovery was left unverified is a fact about *this attempt* (was a browser skipped by the gate), not about whether there happened to be a cache to compare against. Now every refresh cycle under a persistently inconclusive preflight reports the same "unverified" state consistently, instead of only the first one.

Added a regression test for exactly this: no cached cookie at all (simulating the second cycle), Chrome skipped via the same \`.temporarilyUnavailable\` Keychain preflight override used in the existing "mixed" test, Safari genuinely tried and empty — asserting \`cachedSessionUnverifiedInBackground\` rather than \`noSessionKeyFound\`.

**Verification:**
- \`swift test --filter ClaudeWebBackgroundRecoveryTests\` — 8/8 pass
- Full \`Claude|BrowserCookie|Keychain\` sweep green aside from the same two pre-existing, unrelated flakes noted in prior rounds (confirmed passing in isolation again this round)
- \`swiftformat Sources Tests\` / \`swiftlint --strict\` (via \`Scripts/lint.sh\`) — clean

ysyyork:
Fixed all three new findings in \`bacc59f66\` — turns out they share one root cause, so one restructuring fixes all three.

**Root cause:** the stale cached cookie was cleared immediately when rejected, *before* recovery even ran. That discarded the only evidence ("was there ever a working session") the unverified-session classification needs, forcing every subsequent refresh cycle to reason from scratch with nothing to go on.

**[P2] Require evidence of a previously cached session:** confirmed — a first-ever refresh (or genuinely signed-out account) with no cache at all could still be reclassified as "unverified... showing last-known usage" purely because a browser was gate-skipped, even though no last-known usage ever existed. Worse, since \`cachedSessionUnverifiedInBackground\` isn't in \`shouldOpenBrowserForWebSessionError\`'s recognized session-error list, this silently removed the "Re-login at claude.ai" menu action a new or signed-out user actually needs.

**[P1] Carry the unverified state through Auto availability:** confirmed. \`ClaudeWebFetchStrategy.isAvailable\` probes \`ClaudeWebAPIFetcher.hasSessionKey\` before ever calling \`fetch\`. That probe has a cache fast path (a structurally-valid-looking cached cookie short-circuits straight to "available" without touching the network) — once the first rejection cleared the cache, that fast path found nothing, fell through to its own unaware extraction attempt, and reported the provider unavailable under the same gate-skip condition. The Auto pipeline then surfaced a different (OAuth) failure instead, so the unverified classification was never reached on any refresh after the first.

**[P1] Limit gate-skip masking to background refreshes:** confirmed. \`BrowserCookieAccessGate\`'s explicit-retry scope (used by a user-initiated Refresh click) can permit one browser through its cooldown while a *different* installed Chromium browser is still separately gate-skipped. The masking applied regardless of interaction context, so a user who explicitly clicked Refresh and got a real, completed answer could still see "...click Refresh to check now" — which is nonsensical (they just did) and hides both the real result and the sign-in affordance.

**Fix:** defer clearing the cache until the recovery outcome is known, instead of clearing it up front. When recovery is gate-skipped during a *background* refresh with a cookie that was actually rejected this attempt, leave the stale-but-structurally-valid cookie in place rather than clearing it:
- This is what lets a repeat cycle see the same evidence again, fixing #1 for genuine repeat cycles (a first-ever refresh with *no* prior evidence still correctly reports \`noSessionKeyFound\`).
- It also keeps \`hasSessionKey\`'s cache fast path reporting "available," which fixes #2 with no separate change needed in \`ClaudeProviderDescriptor\` — the Auto pipeline keeps calling \`fetch\` every cycle instead of only the first.
- The reclassification is now also gated on \`ProviderInteractionContext.current == .background\`, fixing #3 — a user-initiated refresh always gets the confirmed error.

A confirmed failure (not gate-skipped, no cache to begin with, or explicitly excluded by #3) still clears the cache immediately, same as before.

**Tests:** updated the existing skipped-recovery test to assert the cache is now retained (not cleared) and added a second consecutive attempt proving the repeat-cycle persistence; replaced the old (now-incorrect) "repeat cycle with no cache" test with one asserting a genuinely first-ever refresh reports \`noSessionKeyFound\`; added a direct assertion that \`hasSessionKey(browserDetection:)\`'s fast path stays true after an unverified classification (covering #2 without needing the full \`ClaudeWebFetchStrategy\`/\`ProviderFetchContext\` machinery); added a user-initiated mixed attempted-plus-skipped scenario asserting the confirmed error propagates (covering #3).

**Verification:**
- \`swift test --filter ClaudeWebBackgroundRecoveryTests\` — 9/9 pass
- Full \`Claude|BrowserCookie|Keychain\` sweep green aside from the same two pre-existing, unrelated flakes noted in every prior round (confirmed passing in isolation again)
- \`swiftformat Sources Tests\` / \`swiftlint --strict\` (via \`Scripts/lint.sh\`) — clean

ysyyork:
Fixed the new finding (\`ClaudeWebAPIFetcher.swift:1498\`) in \`ac0c16ce4\`.

**Bug:** confirmed. When the cached cookie was rejected and browser recovery found a *different*, real session key that the API also rejected, the \`.unauthorized\` propagated correctly (that part was already right) — but the original cached entry (deferred rather than cleared immediately, per the previous fix round) was never actually cleared. The only clear was inside the extraction-failure \`catch\`, which a *successful* extraction never reaches. So the now doubly-confirmed-dead original entry stayed cached, and a later cycle's gate skip could still reclassify it as merely "unverified" despite this attempt already confirming both keys were rejected.

**Fix:** wrapped the final \`fetchUsage\` call so a confirmed auth-type failure (\`.unauthorized\`, \`.noSessionKeyFound\`, \`.invalidSessionKey\`) also clears the deferred original entry before rethrowing. Verified this is safe regardless of outcome: \`fetchUsage\` only ever writes to the cache after a successful fetch (via \`persistSessionKeyIfNeeded\`, called after \`fetchOrganizationInfo\`/\`fetchUsageData\` succeed), so on a failure path the cache is never touched and the deferred entry's expected-observation stays valid for the clear.

Extended the existing "confirmed auth failure after a real session is recovered" test (which already covers this exact scenario — stale cache rejected, Safari recovers a real key, API rejects that too) to also assert the cache ends up empty afterward.

**Verification:**
- \`swift test --filter ClaudeWebBackgroundRecoveryTests\` — 9/9 pass
- Full \`Claude|BrowserCookie|Keychain\` sweep green aside from the same two pre-existing, unrelated flakes noted in every prior round
- \`swiftformat Sources Tests\` / \`swiftlint --strict\` (via \`Scripts/lint.sh\`) — clean

ysyyork:
Fixed the new finding (\`ClaudeWebAPIFetcher.swift:1480\`) in \`14ca6cfc1\`.

**Bug:** confirmed, and a good catch beyond what the earlier no-cache fix covers. \`cachedSessionUnverifiedInBackground\`'s fixed wording ("...showing last-known usage...") and the suppressed sign-in affordance it implies both assume there's real prior usage worth protecting. That's true once \`UsageStore\` has a snapshot for the provider — but \`UsageStore.snapshots\` starts empty on every launch (there's no persisted-snapshot restoration), so a background refresh that hits this classification before the *first successful fetch of the process* (e.g. a fresh launch with a leftover stale cookie from a previous session) has nothing to show and nothing to lose by asking the user to sign in. Previously it would still claim to be preserving last-known usage while showing nothing, and hide the "Re-login at claude.ai" menu action the user actually needed.

**Fix:** handled entirely at the app layer (\`UsageStore+Refresh.swift\`) rather than pushing \`UsageStore\`-level snapshot knowledge down into \`ClaudeWebAPIFetcher\` (a lower-level Core module that shouldn't need to know about app-level in-memory state). The existing suppression logic (\`preservesClaudeWebSessionFailure\`) already correctly required \`hadPriorData\` — only the *displayed text* stored into \`self.errors[...]\` didn't share that gate. Added \`claudeWebEffectiveErrorDescription\`, which remaps \`cachedSessionUnverifiedInBackground\` to the plain \`noSessionKeyFound\` message specifically when \`hadPriorData\` is false, so the existing session-error recognition in the login menu (\`shouldOpenBrowserForWebSessionError\`) picks it back up correctly with no further changes needed there.

**Test:** added a full end-to-end regression test exercising the real \`UsageStore\` refresh pipeline (not just \`ClaudeWebAPIFetcher\` directly) with a stubbed fetch strategy that throws \`cachedSessionUnverifiedInBackground\` on the very first refresh of a fresh store — asserting no snapshot exists and the displayed error is \`noSessionKeyFound\`'s message, not the unverified one.

**Verification:**
- \`swift test --filter ClaudeResilienceTests\` — 18/18 pass
- Broader \`Claude|BrowserCookie|Keychain|UsageStore\` sweep green aside from the two previously-noted pre-existing flakes plus one newly-observed one (\`ClaudeProviderRuntimeTests\`'s "replacement refresh cannot publish a cancelled older version probe" — a filesystem race in an unrelated claude-swap temp-dir fixture under heavy parallel load; confirmed passing in isolation, doesn't reference any file this PR touches)
- \`swiftformat Sources Tests\` / \`swiftlint --strict\` (via \`Scripts/lint.sh\`) — clean

ysyyork:
Fixed the new finding (\`ClaudeWebAPIFetcher.swift:1525\`) in \`99dc20ab4\`.

**Bug:** confirmed, and a genuinely deeper gap than a few more retries would close. \`clearInvalidatedCacheEntryIfNeeded\` discarded \`clearIfCurrent\`'s boolean result. The underlying Keychain delete can fail while temporarily unavailable — but unlike the read-side retry added earlier in this branch (3 attempts, 30ms apart, meant for a momentary ACL-inspection race), a locked screen or busy keychain daemon can stay unavailable far longer than that. When the clear failed, the confirmed-dead cookie stayed cached and could be loaded again on a later cycle; if that cycle's recovery was then gate-skipped, the already-confirmed sign-out would be reclassified back into merely "unverified" — indefinitely, for as long as the clear kept failing.

**Fix:** rather than retrying the clear further (which just moves the same race to a different, still-arbitrary timeout), track confirmed-dead cookie headers in an in-process set. This sidesteps the retry-timing question entirely: once a cookie is confirmed dead, no further Keychain operation is needed to *remember* that — the in-memory marker is authoritative regardless of whether the Keychain delete itself ever succeeds. Checked before ever reclassifying a gate skip as unverified; cleared once the entry is actually deleted (or was never added), so it doesn't linger forever once the Keychain recovers.

**Test:** simulated a failed Keychain clear via the existing \`KeychainCacheStore.withClearFailureStatusOverrideForTesting(errSecInteractionNotAllowed)\` override on a genuine (non-gate-skipped) rejection, confirmed the stale entry survives in the cache, then ran a second cycle under a real gate skip — asserting the confirmed \`.unauthorized\` still propagates rather than downgrading to unverified.

**Verification:**
- \`swift test --filter ClaudeWebBackgroundRecoveryTests\` — 10/10 pass
- Broader \`Claude|BrowserCookie|Keychain\` sweep green aside from the previously-noted pre-existing flakes plus one more newly-observed one this round (\`ClaudeDebugDiagnosticsTests\`'s "debug log uses user initiated interaction for OAuth prompt gate" — confirmed passing in isolation, doesn't reference any file this PR touches; same class of parallel-load flakiness as the others already noted)
- \`swiftformat Sources Tests\` / \`swiftlint --strict\` (via \`Scripts/lint.sh\`) — clean

ysyyork:
Fixed the follow-up finding (\`ClaudeWebAPIFetcher.swift:1539\`) in \`b6320a6d2\`.

**Bug:** confirmed — the previous fix's \`confirmedDeadCookieHeaders\` was in-process memory only, so it was lost on every app restart while the rejected cookie itself stayed in the cache (its clear having failed). If the Keychain was still unavailable right after relaunch and the very next refresh cycle's recovery happened to be gate-skipped rather than a clean retry, the new process had no memory of the earlier confirmation and could reclassify an already-confirmed sign-out back into merely "unverified" — exactly at the moment the marker was still needed.

**Fix:** backed the marker with \`UserDefaults\` instead of process memory alone, matching the pattern \`BrowserCookieAccessGate\` already uses for its own cross-restart cooldown state (\`browserCookieAccessDeniedUntil\`). Plain defaults writes have no ACL/lock-screen failure mode at all, so persisting the confirmation doesn't reintroduce the very Keychain-availability problem it exists to route around — it's a genuinely different, always-available storage layer, not a "try harder" retry on the same one. The in-memory \`Set\` is seeded from defaults once per process and kept in sync on every write, so ordinary checks stay a simple lock instead of round-tripping through defaults each time.

**Test:** extended the existing failed-clear regression test to assert the confirmation is actually written through to \`UserDefaults\` — the most direct way to prove restart-survival without literally restarting the test process. Added a small \`DEBUG\`-only reset helper (mirroring the codebase's existing \`...ForTesting()\` convention) so the test resets both the in-memory and persisted state on both sides, keeping runs isolated from each other.

**Verification:**
- \`swift test --filter ClaudeWebBackgroundRecoveryTests\` — 10/10 pass, including running the new test in isolation twice in a row to confirm clean isolation
- Full \`Claude|BrowserCookie|Keychain\` sweep green aside from the two originally-noted pre-existing flakes (no new ones this round)
- \`swiftformat Sources Tests\` / \`swiftlint --strict\` (via \`Scripts/lint.sh\`) — clean

ysyyork:
Fixed the security finding (\`ClaudeWebAPIFetcher.swift:1565\`) in \`a0b3df4d6\`. Good catch — this one mattered.

**Bug:** confirmed. The \`UserDefaults\`-backed marker from the previous commit stored the raw cookie header — including the actual Claude \`sessionKey\` credential — in the unencrypted preferences plist. That defeats the Keychain's protection precisely during the locked/unavailable scenario this marker exists to cover for, and leaves real credential material behind indefinitely, even after the cache entry is later cleared through a different path.

**Fix:** store \`CookieHeaderCache.credentialFingerprint(_:)\` instead of the raw header — a non-reversible SHA-256-based identifier already used elsewhere in this exact module for cache-scope comparisons without exposing the credential itself (its own doc comment: "Safe for cache-scope comparisons; never expose the cookie header itself to UI, logs, or persisted reports" — should have reached for it the first time). Also renamed the defaults key (\`claudeWebConfirmedDeadCookieFingerprints\` instead of \`...CookieHeaders\`) so a device that happened to run the previous, never-released version wouldn't keep reading a raw value back in under a same-named key, and added a one-time removal of the old key as defense in depth in case this ever ran against a real account before landing.

To be clear on exposure: this was caught in review before merge, so it never shipped in a release. I also checked my own local UserDefaults for the old key — nothing was ever written there in this session (my tests only ever used a fake test cookie value, never a real session).

**Test:** updated the regression test to assert the persisted value is the fingerprint of the test cookie, and explicitly assert the raw cookie string is *not* present in defaults.

**Verification:**
- \`swift test --filter ClaudeWebBackgroundRecoveryTests\` — 10/10 pass
- Full \`Claude|BrowserCookie|Keychain\` sweep green aside from the two originally-noted pre-existing flakes (no new ones this round)
- \`swiftformat Sources Tests\` / \`swiftlint --strict\` (via \`Scripts/lint.sh\`) — clean

ysyyork:
Fixed the new finding (\`ClaudeWebAPIFetcher.swift:165\`) in \`2d4e52000\`.

**Bug:** confirmed. \`cachedSessionUnverifiedInBackground\`'s description unconditionally claimed "showing last-known usage" and told the reader to "click Refresh." That's only true in the app's own \`UsageStore\`-backed refresh loop — the same fetcher also serves the CLI's one-shot \`codexbar usage\` (\`CLIUsageCommand.swift\` prints \`error.localizedDescription\` directly at every catch site), which has no snapshot to preserve, no automatic retry loop, and no Refresh control. It would exit with an error while claiming to be showing data it never had and pointing at a button that doesn't exist there.

**Fix:** made the fetcher-level description neutral — it now states only what's always true regardless of caller ("a local security check was inconclusive, not a sign-out"). Moved the app-specific framing into \`claudeWebEffectiveErrorDescription\` (already added in an earlier round for the \`hadPriorData\` remap): it now appends the "showing last-known usage... click Refresh" text on top of the neutral base only when \`hadPriorData\` actually holds — i.e. only in the one context (the app, with real prior data) where every word of it is true. The no-prior-data remap to \`noSessionKeyFound\` from the earlier fix is unchanged.

**Tests:** added a fetcher-level test asserting the raw description contains neither "last-known" nor "Refresh," and an app-level test that seeds a prior snapshot and forces two consecutive failures (the first is suppressed by the single-flake failure gate — confirmed via the existing \`ConsecutiveFailureGate\` tests, so this needed two refreshes, not one) to assert the full framing is still correctly shown when there really is prior data to describe.

**Verification:**
- \`swift test --filter "ClaudeResilienceTests|ClaudeWebBackgroundRecoveryTests"\` — 30/30 pass
- Broader \`Claude|BrowserCookie|Keychain|UsageStore\` sweep green aside from the previously-noted pre-existing flakes (same unrelated files, confirmed passing in isolation again)
- \`swiftformat Sources Tests\` / \`swiftlint --strict\` (via \`Scripts/lint.sh\`) — clean

steipete:
The recovered-session error correction is now on main in #3698, with your co-author and changelog credit preserved. A stale cached cookie followed by a recovered session's timeout, cancellation, HTTP 503, or Cloudflare challenge now surfaces that actual request failure. The four cases failed before the fix; 117 focused tests and full macOS/Linux CI passed.

This PR remains open for its separate Keychain preflight retry and persisted rejection-state work.

steipete:
The final patch is limited to bounded transient no-UI preflight retries. The twelve outdated review threads concern rejected-cookie persistence or error reclassification, which are absent from this revision; those source paths match current main. Resolving those threads on that basis.

The retry keeps stable authorization decisions, existing query flags, and background read policy intact. All 51 focused tests, strict checks, independent review through P2, all 1,247 local selections, and macOS/Linux CI passed on 5fb5215afa8c1a86000e56143dc92ea8e68bf4de. Thanks @ysyyork for the original retry fix.

steipete:
Landed the bounded preflight retry on main as 5c19739aa7de701db3f8c0ba16abbbb634be9aba, with @ysyyork's co-author and changelog credit preserved.

Only inconclusive no-UI checks retry, with three total attempts and two 30 ms waits. Stable outcomes stop immediately, the operation memo retains the final outcome, and background reads still require an explicit grant. Credential persistence and authentication-error classification keep their current behavior.

Validation: 51 focused tests, `make check`, independent review through P2, and all 1,247 local selections across 114 groups passed with no retries/timeouts. CI passed for the merged PR head: https://github.com/steipete/CodexBar/actions/runs/35485146791.

PR #3642, opened 2026-09-14
feat(widget): add large family to combined burn-down widget

karanravindra:
## Summary
- Add `.systemLarge` to the combined burn-down widget's supported families.
- Introduce family-dependent sizing (header, hero number, label column width, chart height) so the large tile grows its charts rather than stretching the medium row over twice the height.

## Commands run
- `git diff` review only; no test changes (UI-only widget layout change).


karanravindra:
Closing; opened on my fork instead.

PR #3645, opened 2026-09-14
fix(opencode): offer Monthly percentage in menu bar controls

### LINKED ISSUE #3644: Opencode Go: Add an option to show Monthly limit in Menu Bar
For me, both Weekly and Session limits are never the issue, but Monthly limit is the bottleneck, so it would be cool to see it in Menu Bar.

egigoka:
OpenCode Go users can now select Monthly in the menu-bar percentage picker and layout palette before the first Monthly snapshot arrives. The selection uses the existing direct tertiary percentage token, and missing quota data renders an unavailable dash. Tertiary choices use each provider's own label, including Perplexity's Purchased pool; Cursor keeps its existing snapshot requirement.

The picker preserves explicit pace/reset tokens, conditionals, and independently placed lane percentages. Layouts that already combine ordinary percentages with an independent tertiary percentage retain their normal controls and use the editor for that mixed selection. A retained popup binding rechecks the latest layout before applying an action, so newer custom edits cannot be overwritten. Existing layout encodings and older-release reconciliation continue to govern saved values.

Validation: three availability assertions reproduced the omission before the fix. `make test-fast FILTER='MenuBarPercentWindow|MenuBarLayout|MenuBarResetToken|OpenCodeGoEstimatedPace|OpenCodeMonthlyPicker|AbacusMonthlyPercent|ProviderArchitectureGatekeeper'` passed 238 tests. `make check` reports zero violations; independent P2 review is clean. Signed synthetic native proof opens the real popup, selects Monthly and shows 79%, then selects Weekly and shows 100%, with the production picker and renderer. Renderer tests also cover used percentages and unavailable data; layout tests cover round trips, stale bindings, custom tokens, persistence and older-release edits. Final macOS/Linux CI is the landing gate.

Fixes #3644. Thanks @egigoka; the original contribution and commits are preserved in this focused rewrite.


egigoka:
Before:
file-63559d4fa119d45958c198aab4f8ecf9
After:
file-ca0a524e718e38505e0f190cf14369d8
Menu Bar:
image

egigoka:
Follow-up 7e855865b addresses the review:

- applied() is now previous-selection-aware (percents always follow; pace/lane tokens only follow when they matched the previous selection), so layouts with an independent custom pace or direct tertiary lanes keep them when switching the picker, and re-picking Monthly after the earlier build converts the leftover session-pace token.
- Palette offers lanePace for the tertiary lane only; primary/secondary lanePace stay unavailable (renderer has no data for them).
- Fixed the global-layout restart bug: releasedCompatible() maps lanePace to ordinary pace before V1/V2 filtering, so both projections agree and preferredLayout survives reload — verified with a new test. V3 downgrade exposure matches the already-accepted windowReset precedent, so no new version gate.
- P2: current() ignores primary/secondary lane tokens instead of forcing Custom.

Tests: 4 new (mixed-layout preservation, monthly keeps custom pace, current ignores direct lanes, V2/V1 agreement + reload keeps V3), 2 expectations updated to the new preservation semantics. Targeted 160 + full suite green. Please re-review.

egigoka:
Follow-up 7090e5e42 fixes a real live bug found while verifying monthly pace: the menu bar showed bare `M 21%` while the dropdown Monthly row showed pace. Root cause: the V3 override contained `lanePace(.tertiary)`, but `preferredLayout`/`preferredOverrides` fell back to the stale V2 key forever, because `releasedCompatible()` now maps `lanePace`→`pace(.automatic)` (from the review fix) while the stored V2 predates that mapping — so the equality gate never agreed again. Fix: a projection stamp (`menuBarLayoutProjectionVersion`, currently 1). When the stamp is older than the running build and a V3 value exists, loaders trust V3 and repair-write all three keys; stamp matches on normal runs, so genuine downgrade edits are still honored. One accepted tradeoff: a downgrade edit made before the stamp existed gets repaired from V3 once. Bump the constant whenever the `releasedCompatible`/`legacyCompatible` mappings change. Tests: 4 new downgrade tests (prestamp lanePace triple repairs + reprojects V2/V1, older-only provider key preserved, prestamp global layout repairs, stamped downgrade edit still wins); full suite passes.

egigoka:
Follow-up 142e1c778: post-launch verification caught the shared stamp disarming overrides repair. The global layout triple (defaults, agreeing) stamped first on launch, so loadOverrides saw stamp==version and skipped repair, stranding lanePace layouts on stale V2 forever. Stamps are now per-triple (Layout/Overrides/Library keys; old single key kept as legacy and removed by stamp helpers, so this build re-repairs once per triple). New regression test 'layout load does not disarm overrides repair' covers the exact load-order failure. Re-review welcome.

egigoka:
Follow-up fix for a self-heal failure found on a live machine (commit 05cad5d5f):

**The bug behind 'still showing session pace after re-picking Monthly'.** The reviewer's preservation rule in applied() treats any pace that doesn't match the previous selection as a deliberate custom. But pace(.automatic) next to a Monthly lane is never deliberate — it's residue from older released projections ([lanePercent(.tertiary), pace(.automatic)] was exactly what the pre-mapping V2 fallback wrote). Re-picking Monthly hit the guard (auto != previous window) and 'preserved' the residue forever — Weekly→Monthly round-trips reproduced it every time.

**Fix:** in applied(.monthly), pace(.automatic) always heals to lanePace(.tertiary). Specific-window paces keep the match-guard, so the review contract (independent customs survive) is intact — auto pace renders the session pace anyway, which is incoherent next to a Monthly percent.

**Tests:** 2 new (direct Monthly re-tap on the degraded mix heals; Weekly→Monthly heals), full suite green (270+74+4). Re-requesting review.

steipete:
The real native popup now offers Monthly. Selecting its menu item changes the production preview to 79%; selecting Weekly changes it back to 100%. The explicit Weekly pace token remains configured, and its unavailable dash stays visible in this synthetic fixture.

The host contains the production picker leaf and renderer with local layout state. The baseline binary used current main plus the behavior-neutral leaf extraction. Captures are confined to that synthetic window; outside-window pixels were cleared.

Before — the popup has no Monthly option:
Before: the percentage popup offers Auto, 5-hour and Weekly

After — Monthly is available:
After: the percentage popup includes Monthly

After selecting Monthly through the native control:
Monthly selected with a 79 percent preview

After selecting Weekly through the same control:
Weekly selected with a 100 percent preview

The state receipt recorded `Monthly → 79% → Weekly → 100%`; no saved settings or account state were needed for this interaction proof.

steipete:
Landed as 1a57ef11cbe6d0765695bd37ef594c0f465e1478; #3644 is closed. Monthly now uses the existing tertiary percentage layout without overwriting independent pace/reset tokens or newer edits made while the popup was open.

Validation: `make test-fast FILTER='MenuBarPercentWindow|MenuBarLayout|MenuBarResetToken|OpenCodeGoEstimatedPace|OpenCodeMonthlyPicker|AbacusMonthlyPercent|ProviderArchitectureGatekeeper'` passed 238 tests; `make check` passed; independent P2 review was clean. Signed synthetic native popup proof recorded Monthly → 79% → Weekly → 100% in the [published before/after captures](https://github.com/steipete/CodexBar/pull/3645#issuecomment-5707993691). All macOS/Linux checks passed on head 35f98d0bf68c5bd9d23559b285fd3b0a8dd1df29 in [CI run 35178006970](https://github.com/steipete/CodexBar/actions/runs/35178006970). The proof uses synthetic data through the production picker and renderer.

Thanks @egigoka for the contribution.

PR #3648, opened 2026-09-14
fix(cursor): show Grok Bot trial usage without an included limit

### LINKED ISSUE #3530: Cursor Grok Bot trial usage is hidden when hasNonZeroIncludedLimit is false

Borisserz:
Trial Grok Bot usage is dropped because `extraRateWindow` requires `hasNonZeroIncludedLimit == true`. Cursor's sand endpoint still returns a live `usagePercent` and `hasAvailableUsage: true` for an active trial, so the menu never shows the Grok Bot window.

Keep hiding the window when both flags are off, so accounts with no Bot allowance stay unchanged.

Fixes #3530.

## Validation
- RED: `shows grok bot trial usage when the included limit flag is false` failed with `window?.id == nil` before the guard change.
- GREEN: `swift test --filter CursorSandUsageTests` — 5 tests passed.
- `make check` — SwiftLint 0 violations in 2287 files.


JackHo12:
Hi @Borisserz, just a heads up that I opened #3629 tackling the same issue.
The underlying cause is that the Cursor backend dropped `hasNonZeroIncludedLimit` in favor of `includedLimitZero` and `sandTrialExpiresAt`. My PR updates the schema to match this change and has been tested. Let me know if you’d like to consolidate our PRs!

Borisserz:
Hi @JackHo12 — thanks for the heads-up. Your schema (`includedLimitZero` / `sandTrialExpiresAt`) matches the backend change, and #3629 already has a live account test. I'll close this PR in favor of #3629 so we don't ship two patches for the same trial bar.

Borisserz:
Closing in favor of #3629, which matches the current Cursor schema and already has live proof.

JackHo12:
Thanks a lot for the quick check and understanding, @Borisserz! Really appreciate the smooth collaboration. 🤝

PR #3653, opened 2026-09-14
feat(codex): opt-in ping to auto-start the next 5h window

BartSu:
## Summary

Implements the "Optional future" item that `docs/refresh-loop.md` has carried for a while:

> Auto-seed a log if none exists via `codex exec --skip-git-repo-check --json "ping"` (currently not executed).

CodexBar already knows exactly when the Codex 5-hour window expires and already schedules a reset-boundary refresh at that moment. This PR adds an **opt-in** follow-up to that refresh: when the boundary that fired belongs to Codex's session window and the refreshed snapshot still shows the expired reset (i.e. nobody has started a new window), CodexBar runs one tiny non-interactive prompt so the next 5h window starts immediately instead of waiting for the user's next real prompt.

- New Codex provider setting **Auto-start next 5h window** (`codexWindowKeepAliveEnabled`), **off by default**, with a subtitle that spells out the cost: one small request per reset and a short Codex session on this Mac.
- `CodexWindowKeepAliveRunner` (CodexBarCore) runs `codex exec --skip-git-repo-check --sandbox read-only --json "ping"` in a scratch directory outside any repo, reusing the RPC executable resolution (`CODEX_CLI_PATH`, login-shell PATH, bundled desktop copies), the `CodexCLILaunchGate` cooldown, and the selected account's `CODEX_HOME` fetch environment. Two-minute timeout.
- `UsageStore+CodexWindowKeepAlive` holds the pure decision (`codexWindowKeepAliveSkipReason`) and the trigger. After a successful ping it refreshes Codex a few seconds later so the menu shows the new window.
- The reset-boundary candidate now carries the window that produced it (`ResetBoundaryWindow`), so the follow-up knows which provider/window expired without re-deriving it. Existing `runResetBoundaryRefresh(boundaryRefreshAt:)` callers are unchanged (new parameter defaults to `nil`).
- Docs: `docs/refresh-loop.md` (replaces the "Optional future" note), `docs/codex.md`, CHANGELOG.

- **Subscription only.** The ping is admitted only for a ChatGPT sign-in. `auth.json` holding an `OPENAI_API_KEY` is rejected (`apiKeyLoginUnsupported`): API keys have no 5-hour window and would be billed per request. A home with no readable login fails closed (`loginUnavailable`). So the ping can only ever spend the 5-hour window the user's subscription already includes; it never creates a bill.
- **Bound to the admitted login.** `CodexWindowKeepAliveAuthority` (fetch environment, `auth.json` SHA-256, account ID) is captured when the boundary admits the ping and re-verified on the main actor right before `codex exec` launches.

Note: this only moves the *start* of the next window earlier; it never adds quota.

## Admission checks (addresses the ClawSweeper / Codex review findings)

The ping is a paid, authenticated side effect, so it is admitted only when **all** of the following hold. Each is a separate `CodexWindowKeepAliveSkipReason` and is logged at info level when it blocks the ping.

| Finding | Fix |
|---|---|
| **P1 – managed workspace bypass** (`codex exec` only carries `CODEX_HOME`, so a selected workspace B could bill the `auth.json` workspace A) | `managedWorkspaceUnsupported`: any non-nil `managedWorkspaceAccountID` from `codexSettingsSnapshot` keeps the ping off, mirroring `CodexOAuthNativeRefreshCLIStrategy.isAvailable`. The toggle explains this under itself when such an account is selected. |
| **P2 – raw system Low Power flag** | `lowPowerMode` now reads `settings.backgroundWorkLowPowerModeEnabled`, the same resolved preference the reset scheduler uses. |
| **P2 – cached expired reset admitted a ping after a failed fetch** | New `UsageStore.lastSnapshotPublicationAt` is stamped only on the success path of `refreshProvider`. `runResetBoundaryRefresh` records when its pass started; `snapshotStale` rejects the ping unless Codex published a snapshot at or after that instant. |
| **P2 – toggle silently inert under Manual cadence** | `isEnabled` greys the toggle out when Refresh is Manual, the subtitle says an automatic cadence is required, and `manualRefreshCadence` blocks the decision as defense in depth. |
| **Consent / account changes racing a queued ping** | Right before the CLI launches, the detached task hops back to the main actor and re-checks `codexWindowKeepAliveEnabled` and the current login. Turning the toggle off also cancels the queued task (`cancelCodexWindowKeepAlive`). |
| **P1 (rev. 2) – same-home login replacement bypassed revalidation** (environment equality does not establish account identity) | `CodexWindowKeepAliveAuthority` captures the fetch environment, the SHA-256 of `auth.json` (`CodexAuthFingerprint`), and the login's account ID at admission. `codexWindowKeepAliveRemainsAdmitted` requires the freshly re-read authority to be identical; a replacement login in the same `CODEX_HOME`, a signed-out home, or a swap to an API key drops the queued ping before any paid I/O (`login-changed` / `login-unavailable` in the log). |
| **P2 (rev. 2) – stacked refreshes never stamped `lastSnapshotPublicationAt`** | `applySelectedCodexVisibleAccountOutcome` (the stacked-layout selected-account publisher) now stamps the publication on its success path only; preserved and failed publications stay unstamped. |
| **API key logins** (no 5h window; per-request billing) | `apiKeyLoginUnsupported` keeps the ping off whenever `auth.json` carries `OPENAI_API_KEY`; the toggle explains this under itself. |

## Test plan

- `CodexWindowKeepAliveTests` (Swift Testing):
  - exec argument shape; fail-closed when Codex is missing.
  - every skip reason of the pure decision, including the new `manualRefreshCadence`, `managedWorkspaceUnsupported`, and `snapshotStale` (nil publication, publication before the pass, publication at/after the pass).
  - `codexWindowKeepAliveRemainsAdmitted` for toggle-off and account-switch.
  - `loginUnavailable` and `apiKeyLoginUnsupported` in the pure decision; `codexWindowKeepAliveRemainsAdmitted` for toggle-off, account switch (different `CODEX_HOME`), same-home login replacement (different fingerprint), different account ID, swap to API key, and signed-out home.
  - `loadCodexWindowKeepAliveAuthority` against a fixture `CODEX_HOME` (`CodexCredentialFixtures`): missing file → nil, ChatGPT tokens → subscription authority with fingerprint and account ID, `OPENAI_API_KEY` → `isAPIKeyLogin`.
  - `windowKeepAliveStatusText` under Manual cadence, with an injected managed-workspace reconciliation snapshot (via `_test_codexAccountSnapshotLoader`, no filesystem or Keychain), for an API key login, and for a missing login.
  - store-level: pings once per boundary through the injected runner; no task for a stale publication; no task for a selected managed workspace; no task for an API key login; a queued ping is dropped when the toggle is turned off before launch; a queued ping is dropped when the same-home login changes before launch (authority loader swapped between scheduling and launch).
- `CodexStackedPublicationTimestampTests` (extension of `CodexAccountScopedRefreshTests`): a stacked-layout `refreshCodexVisibleAccountsForMenu` success stamps `lastSnapshotPublicationAt[.codex]` without seeding it; a failed stacked refresh leaves it unset.
- Existing `UsageStoreResetBoundaryRefreshTests` / `AdaptiveRefreshHeuristicsTests` cover the unchanged reset-boundary pipeline.
- `node Scripts/check-app-locales.mjs` passes; `swiftformat --lint` and `swiftlint --strict` pass on the changed files.
- I could not run `swift build` / `make test` locally: this machine only has Command Line Tools (Swift 5.8), no Xcode. Upstream CI for this fork PR is waiting for maintainer approval, so I ran this repository's CI workflow on the same commit in my fork: BartSu/CodexBar#4 → [run 35011321139](https://github.com/BartSu/CodexBar/actions/runs/35011321139) on `54a88be66` passed every job: lint, Linux CLI builds (x64, arm64, musl), and both macOS test shards (`swift-test-macos`, the same workflow file as upstream).
- The CI run also caught that `CodexWindowKeepAliveTests` needed `@testable import CodexBarCore` and `@MainActor` on the two settings-backed tests, and that the line-anchored `ProviderArchitectureGatekeeperTests` allowlist had to move with the inserted lines (31 anchors shifted; two cluster fingerprints changed for real: the `refreshProvider` publication cluster and the stacked selected-account publisher).

## Runtime proof

Still not supplied: a redacted transcript of a real reset needs the built app on a Mac with Xcode and a real ChatGPT-signed-in Codex, which I do not have on this machine. What *is* established without the CLI runner:

- The store-level tests drive the real callback path (`runResetBoundaryRefresh` → `scheduleCodexWindowKeepAliveIfNeeded` → launch-time re-check → runner) with only the CLI process and the `auth.json` read injected, and show rejection before I/O for: revoked consent, managed workspace, API key login, stale publication, and a same-home login replacement.
- `loadCodexWindowKeepAliveAuthority` is tested against real `auth.json` files in a fixture home (ChatGPT tokens, API key, missing).
- The fork CI run above compiles and runs the full suite on macOS.

If a maintainer wants this merged, I will gather the redacted real-reset transcript on a machine with Xcode, or a maintainer can produce it by enabling the toggle and grepping the log for `window-keepalive`: every admission or rejection is logged with its `reason`.

## Maintainer decision

The original "Optional future" note described a one-off log seed. This PR turns it into a recurring, opt-in, default-off request per reset that only ever spends the user's ChatGPT subscription window (never an API key, never a bill) and is bound to the exact login that admitted it. @steipete, please confirm you want that direction before this goes further, and please approve the CI workflow run for this PR.

https://claude.ai/code/session_015dPaX8gACdcUMvZt5hZ7Lh


ne0lines:
This would be a very neat feature!

BartSu:
> This would be a very neat feature!

Thanks. I think this feature could be practical for users with session limit.

BartSu:
@steipete two things when you have a moment:

1. **CI approval.** Upstream CI for this fork PR is still `action_required`. The identical workflow is green on my fork for the current head `54a88be66`: https://github.com/BartSu/CodexBar/actions/runs/35011321139 (lint, Linux x64/arm64/musl CLI builds, both macOS test shards).
2. **Direction.** Since the last ClawSweeper round the ping is now strictly subscription-only and login-bound: it is admitted only for a ChatGPT sign-in (an `OPENAI_API_KEY` in `auth.json` is rejected, so it can never produce a per-request bill), the admitting login's `auth.json` fingerprint and account ID are captured and re-verified right before `codex exec` launches, and stacked-layout refreshes now stamp the freshness timestamp the ping requires. Details and the updated admission table are in the PR body. If you do not want CodexBar to initiate this kind of recurring request at all, say so and I will close the PR.

Still outstanding on my side: a redacted real-reset transcript from a machine with Xcode, which I do not have locally.

steipete:
Thanks @BartSu for the careful follow-through on account binding, cancellation, and fixture coverage. I also verified that your fork CI passed on the current head.

I've decided not to add recurring automatic prompts to CodexBar. The benefit is moving the next window's start time; it does not add quota, while making the usage monitor responsible for ongoing inference and its configuration, billing, and lifecycle behavior. The old optional log-seeding note was a one-off idea, not a commitment to a recurring feature.

There is also a remaining implementation boundary: a ChatGPT login in auth.json does not constrain ordinary codex exec to that subscription. The selected home can configure a different model provider and its own credentials, so the subscription-only guarantee would require execution/configuration isolation as well.

Closing as not planned on product direction. Thank you for the work and for asking for that decision explicitly.

Issue #3450, opened 2026-09-06
Claude OAuth quota-warning notifications re-fire every refresh cycle instead of once per threshold crossing


JonLaliberte:
CodexBar 0.56.7, Claude provider, Usage source = OAuth API, all three notification toggles on (Quota depleted, Threshold warnings, Predictive pace warnings).

## What's happening

The session threshold warning ("Claude session quota low") keeps firing on every refresh (looks like every ~5 min, matching my refresh interval), not once when I cross the threshold like it's supposed to. Screenshot of the notification history:

<img width="718" height="1134" alt="Image" src="https://github.com/user-attachments/assets/7852b3bd-0cf9-4aec-adaa-64b49645d125" />

Remaining % only goes down over that window, it never crosses back above 50% in between. So by the app's own crossing logic this shouldn't be re-firing at all, it should've fired once at the 50% crossing and gone quiet until either I cross back above 50% or drop through the next threshold down.

## My best guess at what's going on (not confident, haven't confirmed with logs)

I went digging through the source a bit. The threshold-warning dedup state is keyed per account (`QuotaWarningStateKey`, in `UsageStore+QuotaWarnings.swift`), presumably so two Claude accounts don't step on each other's warning state. For OAuth-sourced usage, that account key comes from `warningAccountDiscriminator` in `UsageStore+Refresh.swift`, which for Claude calls `warningClaudeAccountDiscriminator` (`PredictivePaceWarnings.swift`), which in turn depends on `ClaudeOAuthActiveAccountObservation` being `.stable(identity:)` with a non-empty identity (`UsageStore+ClaudeOAuthHistoryTypes.swift`). If that observation comes back as `.stable(nil)` or `.changed` on some refreshes, and the `oauthHistoryOwnerIdentifier` fallback is also inconsistent, the resulting account key would change from refresh to refresh even though it's the same account. If that's what's happening, then every refresh looks like "never seen this account before," so of course it re-warns.

What makes me suspect this specifically: the same discriminator function hardcodes `nil` for Web/API-key/local-probe sources, only OAuth and CLI try to resolve a live "active account identity." So if this theory is right, switching Usage source away from OAuth (to Web, for example) should make it stop, since the key would be constant instead of churning.

I have NOT verified this yet, it's a read of the code, not a debug session. I can try switching to Web (cookies) and report back whether that actually makes it stop, if that'd help confirm or rule this out. Also happy to grab verbose logs around a refresh cycle if there's a specific thing to look for.

[...]

PR #3654, opened 2026-09-14
fix(claude): keep weekly fallback out of session warnings

steipete:
Claude OAuth promotes the weekly quota into `primary` when the five-hour payload is absent. Quota notifications treated that promoted value as a session sample: a healthy weekly quota could re-arm a session threshold, while a depleted weekly quota could emit a mislabeled session warning.

The session-notification path now uses the existing session-window eligibility rule and preserves its previous threshold history while the promoted weekly value is present. Weekly alerts, account scoping, authoritative missing-window cleanup, and quota-low hooks retain their existing behavior. Removing the redundant warning-key initializer keeps production Swift net negative: +3/−13, net −10 lines.

The synthetic parser-to-snapshot-to-notifier regression proves both cases:

- Same account, session usage `51 → missing → 52 → 81`, weekly usage 20: session warnings change from `50, 50, 20` to `50, 20`.
- Weekly-only usage 95: emit the weekly warning once, without an additional session warning.

Validation on `36720221294029e12d8218c034d674ba1337cbe5`:

- Baseline regression failed with four expected assertions; the fixed focused run passed 99 tests across seven suites, including account isolation, extra windows, hooks, disabled-provider cleanup, and architecture checks.
- `make check` passed with zero lint violations. Independent P2 review found no actionable issues.
- Tests use injected stores and a recording notifier; they perform no live credential probes or actual OS notification sends.
- Full local `make test` passed all 1,140 selections in 104 groups without retries or timeouts (888.7 seconds). [Final-head CI passed](https://github.com/steipete/CodexBar/actions/runs/34906502766), including both macOS test shards and all Linux builds.

Related to #3450. This repairs a demonstrated weekly-fallback path; it does not close the separate unresolved cases involving missing or changing account ownership.


steipete:
Landed as b202fc0c8ce7f6b39d6acc832f5cb15ede1e4dd2. The merged tree matches reviewed/tested head 36720221294029e12d8218c034d674ba1337cbe5; main is synchronized and clean.

The fix keeps promoted weekly OAuth values out of the session-notification lane. The same-account missing-session sequence now emits session thresholds `50, 20` instead of `50, 50, 20`; weekly-only data emits only the weekly warning. Production Swift is net -10 lines, and the changelog/provider docs are updated.

Proof: the baseline parser-to-notifier regression failed four assertions, while the fixed focused run passed 99 tests/7 suites covering account isolation, extra windows, hooks, disabled cleanup and architecture. `make check` passed with zero lint violations. Full local `make test` passed all 1,140 selections/104 groups in 888.7 seconds with no retries or timeouts. Independent P2 review is clean, and [final-head macOS/Linux CI passed](https://github.com/steipete/CodexBar/actions/runs/34906502766).

All testing used synthetic data, injected stores and a recording notifier. No live credential probe or actual OS notification was sent. #3450 stays open for its separate missing/changing-owner cases.

PR #3658, opened 2026-09-15
Support read-only Claude account adapters

benedictfrancis:
## Summary

- Add an optional supportsAccountSwitching capability to Claude account-list adapters.
- Preserve existing schema-v1 behaviour by defaulting the capability to true.
- Suppress account activation actions when an adapter declares itself read-only.

## Verification

- swift test --filter ClaudeSwapListParserTests
- swift test --filter ClaudeSwapAccountProjectionTests
- make check


steipete:
Thanks @benedictfrancis. Your implementation is adopted in #3846 (`86c1bd0b795e`), preserving the schema-v1 default and strictly rejecting nonboolean capability values. Read-only lists retain usage and account details while suppressing switching and credential repair. The regressions failed before the fix and passed afterward; synthetic coverage confirms no credential command starts. All 201 selected local tests, `make check`, independent review, and CI passed. Contributor credit is preserved. #3846 landed via #3867; closing as superseded.

PR #3659, opened 2026-09-15
Fix Claude proxy usage deduplication

MoridinBG:
Claude Code proxy transcripts can repeat cumulative usage for one response without a `requestId`. Counting each record inflated token and cost estimates. Use the exact, nonblank session and message IDs to identify those repeated snapshots, while retaining the existing behavior for explicit request IDs.

Typed request/session keys keep unrelated responses separate even when identifiers contain delimiters or match across the two identity forms. Initial parsing, appended records, and copied-file reconciliation use the same rule. Existing parent/non-sidechain/path preference is preserved. Older Claude and Vertex caches and persisted report memos rebuild once from the original transcripts so inflated totals cannot survive the upgrade.

Validation:

- `make test-fast FILTER='CostUsageScannerClaudeRegressionTests|CostUsageScannerTests|CostUsageScannerClaudeMemoTests|CostUsageScannerClaudeProxyIdentityTests|CostUsageScannerClaudeCacheUpgradeTests'`: 58 tests across five suites passed.
- New regression cases first reproduced 84 failed assertions across 21 collision/blank-identity cases on the submitted implementation. The repaired implementation passes them across parsing, appends, and cross-file reconciliation.
- Synthetic Claude and Vertex upgrade fixtures replace an inflated 495-token cache and matching old memo with 169 tokens. Transcripts remain byte-for-byte unchanged, and another cold restart reuses the corrected report without parsing or rewriting its cache.
- Coverage includes incomplete append/restart/completion, distinct IDs, copied-record preference, deterministic ordering, and existing explicit request behavior.
- `make check` and independent P2 autoreview passed.
- `make test`: all 105 test groups passed with system Bash.

All scanner and cache proof uses isolated synthetic transcripts. No UI behavior changes.


steipete:
Merged as [1c54cdded2ca](https://github.com/steipete/CodexBar/commit/1c54cdded2cafef985224f6f10b99488dd1b519d). Proxy response snapshots now deduplicate consistently during parsing, appends, and copied-file reconciliation. Typed request/session identities prevent unrelated responses from colliding, while explicit request behavior and existing source preference remain intact.

Verification on final PR head `7c63bfc939b1f0219c5ab5cb7abe903f91e5aa1b`:

- `make test-fast FILTER='CostUsageScannerClaudeProxyIdentityTests|CostUsageScannerClaudeCacheUpgradeTests|CostUsageScannerClaudeRegressionTests|CostUsageScannerClaudeMemoTests|CostUsageScannerTests'`: 58 tests in five suites passed.
- The new collision/blank-identity matrix first failed 21 parameter cases with 84 assertions against the submitted implementation, then passed after repair.
- Synthetic Claude and Vertex upgrade tests reconstructed an inflated 495-token cache and matching old memo to 169 tokens. Original transcripts remain unchanged; a subsequent cold restart performs no parsing or cache rewrites.
- `make test`: all 105 groups passed. `make check` and independent P2 autoreview passed.
- [CI run 35007202121](https://github.com/steipete/CodexBar/actions/runs/35007202121): all eight jobs passed, including both macOS shards, Linux x64/ARM64/musl builds, lint, and the final gate.

All scanner proof uses isolated synthetic data. Thanks @MoridinBG for the original fix and regression.

Issue #3323, opened 2026-08-31
Memory grows to ~2.6 GB RSS while idle (0.56.1)


CharlieLZ:
### Environment
- macOS 15.6.1 (24G90), Apple M4 Pro (Mac16,7), 48 GB RAM
- CodexBar 0.56.1 (build 132), menu bar only, default settings, no crash/kill events

### Symptom
RSS grows over time while the app is essentially idle (~1% CPU), for a process that only monitors quota:

- ~1.48 GB after several hours of uptime
- ~2.27 GB 20 minutes later
- ~2.65 GB 10 minutes after that (measured today, 2026-08-31, within one session)

App Support data is only ~214 MB (mostly `Backups/`), so the growth is live memory, not on-disk data.

For comparison, a fresh launch of 0.56.2 starts at ~500 MB RSS.

### Repro
1. Launch CodexBar with Codex + Claude providers enabled.
2. Leave it running in the menu bar (idle, no interaction).
3. Watch RSS over hours: `ps -o rss -p <pid>`.

It climbs steadily. This matters because the machine runs many agent sessions; a monitoring app eating 2.6 GB makes system memory pressure worse.

### Question
Is this a known issue / does the memory accumulation correlate with the local cost-history scan (parsing ~/.codex/sessions)? 0.56.2's release notes mention reducing CPU spent parsing session timestamps and decoding cached cost history — is there an accompanying memory fix, or should I keep reporting RSS over time?

apple-ouyang:
Root cause for the retained memory on this path, with measurements from a real local history (20 GB of Codex sessions, 16,144 JSONL files, 807 MB `cost-usage.sqlite`):

| table | rows |
| --- | --- |
| `usage_rows` | 870,436 |
| `token_snapshots` | 2,745,660 |
| `files` | 24,768 |

`heap` on the running app matched those tables one-for-one, and `leaks` reported only 614 leaks / 24 KB — so this is reachable retention, not a leak.

Two distinct retention sources:

[...]

PR #3661, opened 2026-09-15
perf(cost): skip token snapshots when reading the row-pricing cache

apple-ouyang:
The experimental Codex Workspaces indexer decoded the entire token-snapshot table even though it prices exact usage rows and daily aggregates. Skip those unused histories while retaining all pricing evidence and the existing full-read default for other callers.

The lean reader uses an ordinary SQLite read transaction. A concurrent scanner commit therefore preserves a consistent report instead of turning a rejected scanner baseline into empty project history. Scanner receipt validation remains unchanged. Regression tests cover both the cache read and the persisted Workspaces sidecar, plus incomplete caches, timezone checks, and unloaded-history fields.

Validation:

- The concurrent-read regressions fail on the original implementation: two files and 104 tokens become an empty cached report. Both pass after the repair.
- `make test-fast FILTER='CostUsageStoreReadWorkTests|CodexLocalProjectUsageTests|CostUsagePricingRaceTests'`: 105 tests passed.
- `make test`: all 104 test groups passed with system Bash.
- `make check`: passed with system Bash; the installed Homebrew Bash hangs in an existing packaging fixture.
- Independent P2 autoreview passed.
- Compiled and ran StoreStress against a synthetic 32-file history. Both readers retain 16,384 usage rows; full loads 16,384 token snapshots and lean loads zero. Separate processes using the debug core measured 30.1 MiB and 18.1 MiB respectively.

This affects the opt-in debug Workspaces inspector. The menu-bar scanner already loads token histories lazily; the broader idle-memory report in #3323 remains open. No UI behavior changed.


apple-ouyang:
Scope note after installing this build and profiling the running menu bar app (real history: 807 MB cache, 20 GB of sessions):

- The token-snapshot read this PR removes was reachable from the opt-in Codex Workspaces inspector (and from the StoreStress/tests path), **not** from the launch/menu-bar refresh. The menu bar path goes through `loadCodexScan` → `readCodexBaseline(loadTokenSnapshots: false)`, which was already lazy.
- So this PR fixes a real over-read with a large, measured effect on that path (-77% retained / -67% peak), but it should not be expected to change steady-state menu bar memory.

What still dominates the app's retained cost memory, for anyone following up: `usage_rows` are materialized in full by `readSnapshot` and retained as `CodexUsageFileUsage.codexRows` — 255 MB of `_ContiguousArrayStorage<CodexUsageRow>` plus ~117 MB of strings on this history, with a further ~308 MB of transient `usage_rows.payload` `Data` during `decodeCodexCache`. Unlike this PR, touching that changes pricing inputs (`turnID` / `pricingMode`) and the `persistFile` `.reuse` guard, so I left it out.

Details and numbers are also on #3323.

steipete:
Merged as [0fced33a1ae9](https://github.com/steipete/CodexBar/commit/0fced33a1ae99f8e6c7cf169b72f854552fcd52f). The lean read now preserves ordinary SQLite transaction semantics, so concurrent scanner commits cannot replace a valid Workspaces report with empty history. Scanner receipt safeguards remain unchanged.

Verification on final PR head `9432de16ab56d8ef787e0d8832fc762663f2e952`:

- `make test-fast FILTER='CostUsageStoreReadWorkTests|CodexLocalProjectUsageTests|CostUsagePricingRaceTests'`: 105 tests passed. The two concurrent-write regressions first failed against the submitted implementation, including persisted totals dropping from 104 tokens/two files to zero.
- `make test`: all 104 groups passed. `make check` passed. These ran with system Bash; the installed Homebrew Bash hung in an existing packaging fixture.
- Independent P2 autoreview passed.
- [CI run 34997715057](https://github.com/steipete/CodexBar/actions/runs/34997715057): both macOS test shards, Linux x64/ARM64/musl builds, lint, and the final gate passed.
- Compiled and ran StoreStress against 32 synthetic sessions: both readers retain all 16,384 usage rows. Full reads load 16,384 token snapshots at 30.1 MiB; lean reads load zero at 18.1 MiB, measured in separate processes using the debug core.

The improvement is scoped to experimental Workspaces indexing. The broader idle-memory report #3323 remains open. Thanks @apple-ouyang for the optimization and the clear scope correction.

PR #3666, opened 2026-09-15
fix(workspaces): preserve reports when raw cache scope is invalid

steipete:
Workspaces could overwrite saved usage with an empty report after a failed cost-cache read. When changing Codex homes during trace validation failure, it could also import the previous home's cached sessions into the new scope. The sidecar assigns the requested roots when rebuilding its cache, so the existing later scope check could not detect either invalid raw input.

Validate the raw cache roots immediately after reading, before cache reuse or any sidecar import. Invalid input fails the refresh through the existing scope error and preserves saved history. A valid empty scan remains publishable. Scanner validation, receipt safeguards, and transactional cache reads remain unchanged.

Validation:

- Regression cases reproduced both failures in normal and forced refreshes: a 104-token/two-session report became persisted empty history, or its sessions appeared under a different selected home.
- 139 focused tests passed across store read-work, local project usage, priority-cursor, and pricing-race suites. The six new cases also verify a genuinely empty source remains accepted.
- `make check` passed with no formatting or lint violations. Independent review through P2 found no actionable issues.
- Full `make test` passed its first 62 groups, then stopped on an unchanged Kiro PTY fixture waiting for a late-child PID. Its hard-stop latency assertion passed (1.16 seconds); the unchanged `KiroStatusProbeTests` retry passed all 57 tests.
- [CI run 35027539080](https://github.com/steipete/CodexBar/actions/runs/35027539080) passed on final head `58ef818a43c6f8d2600eaf61cf5d78edc0fc2fcd`: both full macOS test shards, lint, Linux x64/arm64, musl, and aggregate checks.
- Tests use isolated synthetic session files and SQLite faults; no live accounts or provider requests are involved.

This change is scoped to the experimental Workspaces indexer.


steipete:
The raw-cache guard now rejects failed reads and caches from another selected Codex home before Workspaces reuses or imports them. The regression tests reproduced persisted empty history and wrong-source session publication; both normal and forced refreshes now preserve the saved report. A valid empty scan still publishes normally.

Local validation:

- `CODEXBAR_ALLOW_TEST_KEYCHAIN_ACCESS=0 ./Scripts/test_fast.sh --filter 'CostUsageStoreReadWorkTests|CodexLocalProjectUsageTests|CostUsageScannerCodexPriorityCursorTests|CostUsagePricingRaceTests'`: 139 focused tests passed. The wrapper enabled file and Keychain isolation and used `swift test --no-parallel`.
- `CODEXBAR_ALLOW_TEST_KEYCHAIN_ACCESS=0 make check` passed with zero violations. The temporary PATH shim selected system Bash for the test scripts.
- Full `make test` passed its first 62 groups, then stopped on the unchanged Kiro hard-stop fixture waiting for a late-child PID. The hard-stop assertion itself passed in 1.16 seconds. The unchanged retry, `swift test --skip-build --no-parallel --filter KiroStatusProbeTests`, passed all 57 tests.
- Independent P2 review found no actionable issues. Tests used synthetic files and injected SQLite failures; no live provider or account access was used.

[CI run 35027539080](https://github.com/steipete/CodexBar/actions/runs/35027539080) passed on exact head `58ef818a43c6f8d2600eaf61cf5d78edc0fc2fcd`: both full macOS test shards, lint, Linux x64/arm64, musl, and aggregate checks. Merged as `a529d96408c1c6135b7c366311a59c608f191df6`.

PR #3677, opened 2026-09-16
Expose usage snapshot sharing from the menu overview

Chipagosfinest:
Overview now offers **Share Usage Snapshot…**, using the same filtered Usage & Spend sources as its summary. Hidden accounts and providers stay excluded, including during initial loading, and an account's saved history cannot borrow a different account's subscription label.

The existing local preview is shared with Settings. Image copying verifies PNG/TIFF publication, reports success or failure truthfully, and respects Reduce Motion. Export stays local until the user chooses where to share it. Thanks @Chipagosfinest for the original contribution.

Validation:

- Focused sharing, ownership, architecture, and clipboard tests passed. Both `make check` and the final full suite passed: 1,219 selections in 112 groups, with no retries or timeouts. All eight [CI jobs](https://github.com/steipete/CodexBar/actions/runs/35416511232) passed at `40acb895b1979bca5e6afbd5a114b43e863b3bee`.
- Independent Codex review found no actionable findings in the final integration.
- A Developer-ID-signed native test host opened the actual preview with synthetic data in English and German. The hidden $900 source was excluded; the visible $2 source remained. External Accessibility inspection verified the Copy Image and Copy Stats controls. No live account credentials were used.

Native preview, English:

English preview showing only the visible synthetic source

Native preview, German:

German preview showing only the visible synthetic source


Chipagosfinest:
Green and ready whenever you have a slot — ClawSweeper has this at "Ready for maintainer review" with no findings, and CI is 9/9 after merging current main.

One note on ordering, since the diff looks larger than it is: of the 44 files, **7 are source**. The rest are 23 localization catalogs (one key each), 11 proof artifacts under `docs/research`, and 3 test files.

#3678 is stacked on this one, so its diff currently carries this PR's changes too — that's most of why it reads as 74 files. Landing this first shrinks #3678 to just the widget entry point, which should make it a much smaller read.

PR #2647, opened 2026-08-04
feat(copilot): add optional seat credit allowances

KSEGIT:
Add optional per-account Copilot seat-credit allowances to the shared menu-card detail rows. With no allowance, consumption stays text; a configured allowance adds a progress bar using the existing `credits_used` value. The selected account can inherit the legacy default, and “Clear default allowance” removes that fallback while retaining explicit overrides.

Allowance edits atomically save settings and repair only a validated account cache, preserving usage, age, source, and error metadata. Clearing an account override resolves the effective default after saving. Credential or endpoint changes still invalidate cached data, and edits preserve independent reset baselines. This remains scoped to menu-card credit details using existing Copilot authentication. Refs #2593.

Local validation includes `make check`, 1,078 selections across 90 groups, and focused allowance/settings regressions. A signed native host used the production settings row and card while every refresh failed offline: inherited 3,000 → text → 5,000; account 2 retained 9,000; returning to account 1 retained 5,000. Usage and timestamps stayed unchanged across more than 300 failed refreshes.

Inherited default:

Inherited Copilot allowance while offline

After clearing the default:

Copilot credit text without an allowance

Explicit account allowance:

Copilot credit bar with a 5000 allowance

Second account:

Independent second Copilot account

Changelog and human contributor credit are centralized in #3572. Independent review and exact-head CI remain merge gates.

Successful responses already in flight resolve the current allowance at publication, so they cannot restore a cleared global default. A signed native delayed-response run verifies the live row, reset baseline, and account cache after the clear action and a subsequent failed refresh.

Observed metered zero credits remain hidden by default, survive persistence internally, and can become an allowance bar during offline edits or a late refresh. Clearing the allowance hides the zero row again. Final full validation and native results are in the proof comment.

PR #3681, opened 2026-09-16
fix(copilot): show configured credits in provider tabs

KSEGIT:
Copilot seats billed in AI credits can show `1351 / 3800` on the detail card while the provider tab has no usage bar. Use that existing configured allowance in Automatic mode when the tab has no real quota window.

The provider owns the seat-row lookup through the existing presentation capability. Real windows keep their current priority, explicit metric selections retain their behavior, and unknown allowances remain without a bar. Clearing or restoring an allowance updates the retained native switcher immediately. This does not invent quota windows, pacing, or an entitlement from the account plan.

Validation: the original eight credit assertions failed against main; 195 Copilot/switcher/architecture tests pass, with a final nine-test replay after explicit geometry typing. `make check` passes with zero violations, and managed independent review is clean through P2. [Native before/after proof](https://github.com/steipete/CodexBar/pull/3681#issuecomment-5707178553) uses identical synthetic data; published image bytes were verified. Full repository CI is the landing gate.

Follow-up to #2647. Thanks @KSEGIT for the original fix; contributor credit is preserved.


steipete:
Fresh production-view proof uses the same synthetic **1351 / 3800 configured seat credits** in both states. Before, the real main-branch projection returns no Copilot percentage. After, Automatic mode returns **35.55% used / 64.45% remaining**, and the native provider tab renders those proportions. The Claude control stays unchanged.

The original credit assertions reproduced the bug before the fix. The revised candidate passes 195 Copilot, switcher, and architecture tests, including explicit-metric controls, real-window priority, zero/overage usage, unknown allowances, and clearing/restoring the allowance on a retained native view. Both complete captures contain synthetic data only.

Before:
Before: Copilot has no provider-tab progress despite configured credits

After:
After: Copilot shows the configured credit ratio in used and remaining modes

steipete:
Landed as 5af5594e2ab6ded5769d9b67eacc978c35c41036 after every macOS/Linux check passed in https://github.com/steipete/CodexBar/actions/runs/35172248258.

The Copilot-owned fallback supplies seat-credit progress only for Automatic tabs without a real window. Explicit choices, real-window priority, unknown allowances, and account isolation retain their existing behavior. Thanks @KSEGIT.

Validation: `make test-fast FILTER='Copilot|SwitcherExhaustedQuotaTests|ProviderArchitectureGatekeeperTests'` passed 195 tests, followed by the final nine-test replay; `make check` and final independent P2 review passed. The attached native before/after images use the production switcher and configured synthetic credits. I accept that evidence with the reviewed unchanged app input wiring. The bot's broader installed-app settings-flow request is outside the proof claimed here; no live Copilot account was accessed.

PR #3684, opened 2026-09-16
fix(cost): use GPT long-context thresholds for Claude Code

MoridinBG:
GPT usage recorded through Claude Code could receive a long-context surcharge above 200,000 input tokens, even though the existing OpenAI model table applies it only above 272,000. Reuse that model-specific boundary while keeping catalog prices and Claude's separate uncached, cache-read, and cache-creation token counts. Other providers and unlisted models keep their catalog boundary.

Saved Claude reports are recalculated under the new pricing semantics. The shared parser fingerprint also changes, so the released 0.60.4 fingerprint is admitted through the existing predecessor-adoption path; retained native Codex history and checkpoints survive even after their transcripts have disappeared. This does not change native Codex row semantics.

Validation: the full suite passed all 1,159 selections across 107 groups without failures, retries, or timeouts. The new boundary tests fail against the old implementation with 12 expected assertion failures; the restored fix passes the final 122-test pricing, cache-upgrade, history-adoption, and architecture-gatekeeper selection. A synthetic version-2 report/cache is repriced without rereading or modifying its transcript, then reused after another cold restart. `make check` is clean, and independent P0–P2 review has no actionable findings. CI on this rewritten head must pass before merge.

The fixtures use synthetic catalog rates to prove that only the threshold comes from the bundled table. No real account or credential probes were run. Thanks @MoridinBG for the original fix; contributor credit is preserved in the commit and changelog.


steipete:
Landed in cd47c9efccdfd9a2469b1313da72c6113b074b3d. Thanks @MoridinBG!

Claude-recorded GPT usage now uses the same bundled OpenAI long-context boundary as native Codex while retaining catalog rates and separate input/cache accounting. Existing reports are repriced, and the released native Codex cache predecessor remains eligible for adoption so retained history survives missing source transcripts.

Verification: the full local suite passed 1,159 selections in 107 groups without failures, retries, or timeouts; 122 focused pricing/cache regressions passed; the old threshold failed the new boundary regression as expected; `make check` and managed P2 review passed. CI passed on final PR head 25dd00bf16ec9105a328e2a0b2cf50213f9a6c53: https://github.com/steipete/CodexBar/actions/runs/35135551074. Main was pulled and merge reachability verified. No live account or Keychain probes were used.

PR #3679, opened 2026-09-16
fix(grok): preserve measured billing periods for pace

sudoHG:
Grok's proxy reports a full billing period, but CodexBar discarded its start and duration. Near the end of a weekly period, the card could still say Weekly while losing its pace marker and forecast. Preserve matching start/end bounds through parsing, normalization, and billing enrichment so measured weekly periods keep their pace near reset.

This follows the measured-bounds scope accepted in #2455. Invalid or missing bounds remain unknown, and a measured monthly period with six days left stays Monthly without weekly pace. Contributor: @sudoHG.

The contributor's [sanitized live CLI comparison](https://github.com/steipete/CodexBar/pull/3679#issuecomment-5695642189) shows the same measured weekly bounds restoring pace against a real account. The Grok production files match that historically tested revision; the native renders below are separate synthetic display proof.

Validation on the maintainer candidate: 160 focused Swift Testing tests and one native screenshot test passed; `make check` passed with zero violations; independent code review found no actionable P0–P2 defects. Focused coverage includes parsing, enrichment/reset pairing, zero-usage handling, monthly/weekly pace policy, and CLI snapshots. Exact-head CI must finish successfully before merge.

Synthetic native card renders below use identical 90% usage and reset times. “Before” reproduces the previous durationless normalization; “after” uses the production proxy parser and mapper. No account credentials or live provider data are used.

| Period remaining | Before | After |
| --- | --- | --- |
| 30 hours | Before: weekly pace missing | After: deficit and forecast restored |
| 6 hours | Before: weekly reserve missing | After: reserve lasts until reset |

Reproduce the captures with `CODEXBAR_GROK_PACE_PROOF_DIR=/tmp/grok-proof make test-fast FILTER='GrokPaceScreenshotRenderTests|GrokMenuCardModelTests|GrokCreditsProxyFetcherTests'`.

PR #3674, opened 2026-09-16
fix(settings): stop repeated placeholder window closes

harjothkhara:
CodexBar can retain the empty SwiftUI Settings window after closing it. Every subsequent window-update notification then closes it again, producing repeated log messages and unnecessary main-thread work.

Track dismissed windows weakly and mark closes in flight before calling AppKit. Hidden retained placeholders settle after one close; the same instance remains eligible if macOS presents it again. The real Settings window and unrelated windows keep their existing exclusions. Related to #3671: this repairs the demonstrated close loop; the reported Claude refresh-starvation correlation remains unverified.

Validation includes the production guard with actual `NSWindow.close()`, real visibility, a synchronous update during close, 100 subsequent update notifications, and same-instance re-presentation. The native fixture uses synthetic windows and no provider credentials.

| Native observation | Original close behavior | Fixed behavior |
| --- | ---: | ---: |
| Close calls after 100 updates | 102 | 1 |
| Total closes after re-presentation | 104 | 2 |
| Real Settings / unrelated control close calls | 0 / 0 | 0 / 0 |

The original behavior fails the regression; all 10 focused guard tests pass with the fix. Run the opt-in native check with `CODEXBAR_SETTINGS_CLOSE_NATIVE_PROOF=1 make test-fast FILTER=PlaceholderSettingsWindowGuardTests`.

Maintainer validation on `a4c919819740459573b9485485badbc56e71c1d6`: `make check` passed with zero violations, and `make test` passed all 1,152 selections in 106 groups with zero failures, retries, or timeouts. Independent code review is clean. [Exact-head CI](https://github.com/steipete/CodexBar/actions/runs/35078019113) must finish successfully before merge. Contributor: @harjothkhara.

PR #3685, opened 2026-09-16
fix(antigravity): skip CSRF-gated agy spawn before print report

Niclassslua:
Known agy 1.2.2 and newer local servers reject tokenless calls, so CodexBar used to wait for a managed session that could never become ready before reaching the structured usage report. This skips that futile spawn after the bounded warm-reuse check. Older and unknown versions still use managed sessions; version-gated reports and selected-account identity restrictions remain intact.

The maintainer rewrite injects warm-reuse and spawn dependencies into the full-fetch tests, keeping even regressed tests away from shared live sessions. OAuth overrides are removed from version and report subprocesses. Includes release notes and provider documentation; contributor credit is preserved.

Validation: 132 focused Antigravity tests passed; `make check` passed in the isolated candidate; managed P2 reviews passed before commit and after synchronization with main. Maintainer tests used synthetic executables and injected state, with no real provider or credential probes. CI runs on the updated PR head.

The runtime comparisons below were supplied by the contributor, not rerun during maintainer validation.

## Proof: the CSRF requirement starts at 1.2.2

I downloaded official macOS arm64 builds from `storage.googleapis.com/antigravity-public` using the URLs in the historical Homebrew casks. Every SHA-256 matches its cask. Each binary ran in a PTY with an isolated `HOME`, and startup logs confirmed which language server version was running. Both ports got the identical tokenless request CodexBar sends:

```sh
curl -sk -X POST -H 'Content-Type: application/json' -H 'Connect-Protocol-Version: 1' -d '{}' \
  <scheme>://127.0.0.1:<port>/exa.language_server_pb.LanguageServerService/GetUserStatus
```

| agy (`--version` = LS log) | Homebrew cask commit | archive SHA-256 matches cask | HTTPS port | HTTP port |
|---|---|---|---|---|
| 1.1.28 | `044d9d0dfb3b` | yes | **200** `userStatus` | **200** `userStatus` |
| 1.2.0 | `517dfc7cb852` | yes | **200** `userStatus` | **200** `userStatus` |
| 1.2.1 | `854aa6a75e5b` | yes | **200** `userStatus` | **200** `userStatus` |
| 1.2.2 | `fa3d6d5dfe22` | yes | **401** `missing CSRF token` | **401** `missing CSRF token` |
| 1.2.3 | `10c6ac4957a1` | yes | **401** `missing CSRF token` | **401** `missing CSRF token` |
| 1.2.4 (installed, auto-updated) | `b3e603b02b8a` | n/a | **401** `missing CSRF token` | **401** `missing CSRF token` |

This matches @sudoHG's independent 1.2.1/1.2.2 comparison in #3586. Setting `ANTIGRAVITY_CSRF_TOKEN` when launching 1.2.4 doesn't help: the same request with that value in `X-Codeium-Csrf-Token` gets `401 invalid CSRF token`, so the server uses its own generated token.

## Proof: before/after with the real, signed-in agy 1.2.4

`codexbar usage --provider antigravity --source cli --json --log-level debug`, with `main` (`b75d765b6`) and this branch (before its conflict-only rebase) built from source and run alternately on the same machine and network:

| run | build | exit | total | managed `agy` spawned | time before `/usage` starts | `/usage` subprocess | result |
|---|---|---|---|---|---|---|---|
| 1 | main | 0 | 20.5 s | yes | 16 s | 4.0 s | `cli`, 4 windows |
| 1 | branch | 0 | **5.1 s** | no | 0 s | 3.9 s | `cli`, 4 windows |
| 2 | main | 0 | 20.2 s | yes | 15 s | 4.7 s | `cli`, 4 windows |
| 2 | branch | 0 | **6.4 s** | no | 1 s | 6.1 s | `cli`, 4 windows |
| 3 | main | 0 | 21.3 s | yes | 15 s | 5.7 s | `cli`, 4 windows |
| 3 | branch | 0 | **5.9 s** | no | 0 s | 5.6 s | `cli`, 4 windows |

Both builds return the same four quota-window IDs. On `main`, every request to the spawned session fails with `Antigravity session expired` (the 401 mapping) until the deadline. On the branch, the debug log shows `antigravity-cli-version` (48 ms) and then `Antigravity CLI HTTPS spawn skipped; agy local server requires a CSRF token`. No `agy` process was left over from any run.

## Remaining scope

This does not change agy's own report latency or its CSRF error message. Broader Antigravity authentication and account policy changes remain separate from this startup fix.


Niclassslua:
Both items from the review are done on `73f7f17bf`.

**1. Merge conflict.** I rebased onto `main` at `f13aaf15d`. The only conflict was `CHANGELOG.md`, and both entries are kept (the #3674 Settings entry first, then this Antigravity entry). The PR still changes the same 4 files, with no code changes from the rebase. Since then, `main` has moved to `2c994ef1c` (#3679). That commit only touches Grok sources, tests and docs, and `git merge-tree` shows no conflict with this branch.

**2. Validation on the reconciled branch.**
- `make check`: SwiftFormat clean, SwiftLint 0 violations in 2,306 files.
- `make test`: the only failure is `CLIServeWebUITests` › `shared costs and diagnostics survive account grouping without sharing credits`. It fails the same way on unmodified `main` (checked at `f13aaf15d`, and earlier at `b75d765b6`). The test expects `$2.00`, but `CLIServeWebUI+HTML.swift` formats with `Intl.NumberFormat(undefined, { style: "currency", currency: "USD" })`, which follows the system locale. The machine is `de_DE`, so it renders `2,00 $`. This is unrelated to this PR, and I left it unchanged.
- Full suite via `./Scripts/test.sh`, with a pass-through `--swift-command` wrapper that only adds `--skip CLIServeWebUITests` to test runs: **1,152 selections in 106 groups passed on the first pass, with 0 failed groups, 0 retries and 0 timeouts (825 s)**.
- One earlier full run of the rebased branch, done while a separate build was using the CPU, had two first-pass failures that passed on retry: `ClaudeCLISessionTests` › `overlapping profile captures serialize PTY ownership` and the existing `print timeout terminates its process` (1 s timeout). Both passed 10 out of 10 when run on their own on an idle machine, and the clean full run above had no retries.

steipete:
Landed in 7d7c7301850827f4469b147639995e02137e9667. Thanks @Niclassslua!

Known agy 1.2.2 and newer versions now reach the supported usage report without first waiting for a managed local server that rejects tokenless calls. Warm reuse still runs first; older/unknown versions retain managed startup; selected or injected Auto accounts still cannot borrow an identity-free report.

The maintainer rewrite injects warm-session and spawn dependencies into the regression tests, so a broken skip cannot reach shared live sessions. Validation passed 132 focused Antigravity tests, `make check`, managed P2 reviews, and every CI check on final head 748f256e6bb91dede24dd1e6897ea777216f0d60: https://github.com/steipete/CodexBar/actions/runs/35140709192. Maintainer tests used synthetic executables and state; the contributor's separately attributed runtime comparisons remain in the PR body.

Main was pulled and the merge verified. This removes CodexBar's futile readiness wait; agy's own report latency and broader CSRF diagnostics remain separate.

PR #3686, opened 2026-09-16
fix(settings): make the Homebrew update command copyable

### LINKED ISSUE #3682: You can't copy brew update line from settings
Image

harjothkhara:
Homebrew-managed installs showed an upgrade instruction in About that was difficult to copy. Keep native text selection and add a Copy button for the bare `brew upgrade --cask steipete/tap/codexbar` command, as requested in #3682.

An explicit manual-command capability supplies both the Homebrew message and the copied command. Generic unavailable builds have no manual command, and updater discovery and Sparkle behavior remain unchanged. The small production row can be verified without loading About's saved preferences.

Validation: two production Button-action tests passed through the existing deferred copy helper with an isolated writer; the native screenshot replay and 45 supporting clipboard/architecture tests passed. `make check` passed with zero violations. [Synthetic before/after proof](https://github.com/steipete/CodexBar/pull/3686#issuecomment-5706934145) is attached to this PR. Managed independent review is clean through P2. Full repository CI is the landing gate.

Fixes #3682. Thanks @harjothkhara for the original selection fix; contributor credit is preserved.


steipete:
The About update row now has a Copy button that copies exactly `brew upgrade --cask steipete/tap/codexbar`, while retaining native text selection.

Both images come from the isolated production row in a native grouped Form. The before image uses the original selection-only PR with a presentation-preserving row extraction; the after image uses the implemented button. They contain only synthetic update instructions.

The two action tests passed by invoking the real SwiftUI Button configuration through `onAppear`, then the existing deferred copy helper with an isolated writer. They verify the bare command and that a generic unavailable message cannot invent a Homebrew action. The screenshot replay and 45 supporting clipboard/architecture tests passed as well. No command was executed or general clipboard accessed during proof.

Before — selection-only proposal:
Before: selectable Homebrew instruction without a Copy button

After:
After: Homebrew instruction with a Copy button

steipete:
Landed as 094b18a2c5d7176420fca5e335f5a0529eecebf4 after all macOS/Linux CI checks passed in https://github.com/steipete/CodexBar/actions/runs/35170193178.

Validation included two production Button-action tests, the native before/after replay linked above, 45 supporting clipboard/architecture tests, `make check`, and a clean independent P2 review. The action test verifies the exact command through the existing deferred clipboard helper with an isolated writer. I consider that sufficient for this wiring change: the clipboard helper and updater detection are unchanged. The bot's request for a full installed-app copy-and-paste recording remains outside the evidence claimed here; no real settings or general clipboard were used.

Thanks @harjothkhara for the original fix.

PR #3683, opened 2026-09-16
fix(antigravity): complete per-source failure diagnostics

hhh2210:
Antigravity failures could hide which source failed, and an account-mismatch message claimed OAuth data was being used before OAuth succeeded. Report ordered source outcomes for auto-mode CLI failures and describe the account restriction accurately. The existing error-selection policy from #3701 remains authoritative.

Add optional strategy IDs and explicit outcomes to diagnostic exports, with legacy Codable compatibility in both directions. Terminal and exhausted pipeline failures produce one debug record using safe error categories; successful and cancelled runs stay quiet. The synthetic CLI fixture pins both its home and Gemini directory inside its own root.

Validation on `b4c6a91b3277e1690dab64aa89839e215f27b0e9`:

- `make test`: all 1,196 selections across 110 groups passed on the first attempt, with no retries or timeouts (941 seconds).
- Focused provider, export, architecture, and CLI coverage: 146 Swift Testing tests and 43 XCTest tests passed.
- `make check`: zero violations. Independent P0–P2 review: no actionable findings.
- Synthetic production CLI subprocess tests verify explicit-OAuth diagnostic serialization and safe terminal-failure logging. Ordered auto-mode summaries have separate helper coverage. No live account, provider request, browser import, or Keychain read is used. [Contributor's captured CLI output](https://github.com/hhh2210/CodexBar/blob/0ad0f8421c37207c803f58b8b17603f993515400/.github/pr-proof/antigravity-source-diagnostics.log) illustrates the same explicit-OAuth scenario.

[Current-head CI](https://github.com/steipete/CodexBar/actions/runs/35346772110) passed all checks. The combined five-fix integration tree also passed all 1,203 selections across 111 groups without retries or timeouts, with clean lint and independent review. The changelog entry is under 0.60.6 Unreleased.

Part of #3673; related to #3146 and #3662. Thanks @hhh2210 for the diagnostic fix and regression coverage.

Issue #3673, opened 2026-09-16
Antigravity: capability-based source adapters, honest auto errors, identity-aware selection, one pricing path


hhh2210:
## Problem

Antigravity is the highest-churn provider in CodexBar: **138 titled items (45 issues + 93 PRs) between January and September 2026**, and at least ten bug classes have been fixed two or more times. The `provider:antigravity` label was used four times and then abandoned, so the churn is not even measurable from labels today.

This is not one bug. It is a structural coupling problem: menu UI, history persistence, account selection, and cost estimation are all wired directly to whatever wire format we last reverse-engineered, across **four mutually inconsistent quota sources**, with no compatibility layer.

### 1. Four quota sources with incompatible, unversioned contracts

We probe: the desktop `language_server` (CSRF-gated HTTPS), the `agy` CLI (tokenless HTTPS → CSRF-gated in 1.2.2, #3586), identity-free `agy -p /usage` JSON (#3607), and Google OAuth (`fetchAvailableModels` reports `remainingFraction: 1` always, #1063; missing families + empty plan usage, #2427). Every upstream release shifts one of them and produces a new issue. The protocol is explicitly internal (`v1internal`, `exa.language_server_pb`), so this will continue.

### 2. The `auto` pipeline's last-error + identity guard hides working data

Reported errors get overwritten by less-authoritative sources' errors, and a selected OAuth account rejects identity-free CLI snapshots, leaving stale "Limits not available" residue from another account. #3146 has been "fixed" twice (#3161 claimed it in v0.55.0; #3595 only changed error text) and is **still open**. #3662 (re-added account shows no limits), #3649/#3650 (auto falls to OAuth while `agy` works) are the same class. This one is fully within our control — no upstream drift involved.

### 3. Schema drift + invented semantics

Quota history dropped durationless pool snapshots and inferred a five-hour window from missing `windowMinutes` (#2431 → #3619). The local SQLite/JSONL reader drifted across `agy` versions (timestamps: #3374/#3385/#3388/#3396). Two overlapping pricing paths now exist (#3412 models.dev vs #3619 dated API rates) on the same reader.

## Repeat-fix table (condensed)

| Bug class | Fixed in | Still open |
| --- | --- | --- |
| Loopback self-signed TLS | #692/#693, #733/#727 | — |
| CSRF token handling | #1060, #727, #3586 | workaround only |
| Process/bundle detection | #1049, #2836 | — |
[...]

Issue #3662, opened 2026-09-15
Antigravity "Limits not available" bug.


ozdemirmelihahmed:
I've been stuck with this error for 3 days. While using the Codexbar menu, I cannot see my Antigravity Limits. I tried removing and re-adding my account, but it still doesn't work.

sreejithraman:
### Workaround to restore your limits immediately

You can get your Antigravity limits displaying right away with a one-click settings change:

1. Open **CodexBar Settings** (`⌘,`).
2. Go to **Antigravity**.
3. Change **Usage source** from **Auto** to **Local API / agy CLI**.
*(Alternatively, if you use the terminal: `codexbar usage --provider antigravity --source cli`)*.

---

### Why this happens and why re-adding your account didn't fix it

1. **Google backend endpoint restrictions (HTTP 403)**:
   In **Auto** mode with a signed-in Google account, CodexBar queries Google's Cloud Code quota endpoint (`/v1internal:retrieveUserQuota`). For many personal/standard Google accounts, Google rejects quota inspection requests with **HTTP 403 Forbidden** (`Antigravity remote quota endpoint is not permitted for this account`). Re-adding your account only issues a new token for the same user, so Google's backend permissions remain unchanged.

2. **Empty rate windows**:
   Because Google confirms your account email and tier but refuses to return the quota buckets, CodexBar creates a valid authenticated snapshot with identity, but no rate windows. The menu bar then safely renders `Limits not available`.

3. **Account isolation in Auto mode**:
   The `agy` CLI can provide a structured usage report (`agy -p /usage`), but that report does not contain an account email. To prevent accidentally attributing local CLI quotas to a different signed-in Google account, **Auto** mode intentionally skips this anonymous fallback when an account is selected.

4. **Alternative solution**:
   Keeping the desktop **Antigravity.app** running and logged in also allows CodexBar to fetch the limits directly via the local app probe.

Chener:
Follow-up on the selected-account “Limits not available” case (#3662) and on PR https://github.com/steipete/CodexBar/pull/3708.

Facts as I understand them:
- CodexBar’s OAuth path queries Google’s personal-account quota API and gets 403, so the selected account has identity but no quota window.
- The agy CLI’s own privileged client can retrieve `RetrieveUserQuotaSummary`.
- That CLI report does not include an email, so Auto correctly refuses to attach it to a selected account (identity-aware; same concern as #3673).

[...]

PR #3650, opened 2026-09-14
Clarify Antigravity Auto account selection

Borisserz:
When Auto preserves a selected or injected Google account, it cannot accept an agy usage report that contains no account identity. Users could see no limits even though the explicit local source worked, while the source picker gave no explanation. The picker now explains that boundary and names **Local API / agy CLI**, including that it may use a different signed-in account.

The existing account guard, fetch order, and explicit-source behavior are preserved. Documentation explains the same choice. Descriptor construction is simplified by removing single-use binding/options locals; production Swift changes are net **−2 lines**.

Fixes #3649. Thanks @Borisserz for the report, original guidance proposal, and Settings proof.

Validation:
- `make test-fast 'FILTER=ProviderSettingsDescriptorTests|AntigravityCLIUsageReportTests'`: 67 tests across 2 suites passed, including selected/injected Auto rejection and explicit CLI behavior.
- `make check`: passed; zero SwiftLint violations across 2,287 files.
- Independent P2 review: no actionable findings.
- A Developer ID-signed, credential-isolated test host rendered the actual `ProviderSettingsPickerRowView` with synthetic settings. Before recreates the exact previous subtitle; after uses the current production descriptor. Both renders passed and were inspected. No live provider request or installed-app modification was involved.
- Full local `make test`: all 1,140 selections / 104 groups passed in 1,282.5 seconds. One CLI serve/snapshot/chart group exceeded its 180-second group timeout; all 12 isolated retries passed. An earlier run stopped in the unchanged SQLite payload benchmark under heavy host load and low disk space. After reclaiming completed-task build caches, the 100 MiB write passed at 7.4 seconds inside the completed full run, without changing the 30-second limit.
- [Final hosted CI](https://github.com/steipete/CodexBar/actions/runs/34868588731): all checks passed on the final revision, including both macOS shards and Linux builds. The final P2 review after syncing main found no actionable issues.

Native Settings row proof:

Before: source picker omits account-identity guidance

After: source picker explains account identity and the local source

PR #3689, opened 2026-09-16
fix(antigravity): honor CLI overrides and explain excluded reports

hhh2210:
The shared per-source diagnostics have landed in #3683 and are included in 0.61.0. This PR now contains only the remaining Antigravity override and identity-free report boundaries. Refs #3673, #3649, #3650, and #3662.

An unusable `ANTIGRAVITY_CLI_PATH` previously fell through to an ambient `agy`. A background usage refresh could therefore launch the CLI and trigger interactive login even when the explicit path was intended to disable discovery.

- Make only Antigravity's explicit override authoritative. An unusable override stops resolution before PATH, well-known locations, shell lookup, or alias lookup; an absent override retains normal discovery. Other providers keep the shared resolver's existing behavior.
- Explain why Auto excludes an identity-free `agy -p /usage` report when an account is selected or OAuth credentials are injected. Diagnose exports and debug logs retain the safe `identity_free_report_excluded` category. Explicit CLI mode keeps its report fallback.
- Preserve main's error precedence, including `notRunning` and `missingCSRFToken` placeholders. The merged #3683 pipeline, upstream proof-isolation fix, and released diagnostics note are retained unchanged.

**Compatibility decision for maintainers:** treating an unusable explicit Antigravity path as disabling ambient discovery is intentional and changes main's fallback behavior for that provider. This policy still needs maintainer acceptance; the unrelated providers' contracts are unchanged.

The latest main is merged. The two unshipped changes are documented in a new **Unreleased** section. Released changelog history exactly matches main, with #3683 correctly recorded under 0.61.0 and the stale duplicate under 0.60.5 removed.

## Validation

- Current integration: `make check`, `git diff --check`, exact source-anchor checks, and comparison of shared pipeline/CLI files with upstream main passed.
- Previous head `05d6b4993`: all nine [GitHub checks](https://github.com/steipete/CodexBar/actions/runs/35341197828) passed, including both macOS test shards and all Linux CLI builds.
- On that head, local focused suites passed 168 Swift Testing tests plus 43 XCTest tests. The full run completed all 1,189 selections in 110 groups on the first pass, with no retries or timeouts.
- Seven isolated production CLI scenarios passed: four unusable overrides made zero sentinel calls; an ambient-discovery positive control executed the sentinel; scoped Auto usage and diagnose omitted the identity-free report and exposed its safe category. [Sanitized evidence](https://github.com/hhh2210/CodexBar/blob/05d6b4993/.github/pr-proof/antigravity-identity-guard.log).

This update integrates main and fixes changelog placement; the guard/exclusion behavior is unchanged from that validated head. Local Swift/AppKit suites were not rerun for this integration. Current-head CI is pending after the push.


steipete:
Adopted the Antigravity-specific explicit-path guard and resolver regression coverage in #3847 (b275fcd7ab27). An unusable `ANTIGRAVITY_CLI_PATH` stops discovery before PATH, install locations, shell lookup, or aliases; removing the override restores normal discovery. Thanks @hhh2210. The extra diagnostic category is not included. All four invalid-override cases failed before the fix and pass afterward; 590 focused Swift tests, make check, independent review, and CI passed.

PR #3690, opened 2026-09-16
fix(keychain): recognize bundled CLI launches through symlinks

sreejithraman:
The documented external CLI symlink could make the bundled helper look like a standalone development executable, disabling its persistent credential-cache path. Use the operating system's running-executable identity for app detection, cache trust construction, and caller preflight.

The implementation deliberately does not grant trust to a mutable launch alias. Resolving such an alias while constructing an ACL could capture another application's identity if the link changed. Kernel-derived paths keep trust bound to the loaded executable, and identity lookup failure stays closed. Existing signature validation, disabled-access settings, ad-hoc handling, and no-prompt item-query gates remain in force.

Validation: 53 focused tests passed, including real signed-system-executable launches through direct, relative, chained, and parent-directory symlinks. Adversarial fixtures retarget an alias to another app before discovery and during trust construction; native in-memory Security objects still reject that replacement. Missing-image and standalone-binary controls, existing store tests, and Keychain prompt-safety audits passed. Independent P0–P2 review is clean after addressing the alias-retargeting finding; formatting/lint checks passed. CI on this rewritten head must complete before merge.

This proof covers process identity, path selection, and native signature checks. It does not claim a real Keychain-item roundtrip or a live provider-account probe. Thanks @sreejithraman for the report and original fix; contributor credit is preserved in the commit and changelog.


sreejithraman:
@clawsweeper re-review

steipete:
Landed in 151bc240ba0f00d20cbe09f0563293f21b3ae746. Thanks @sreejithraman!

Bundled CLI launches through installed symlinks now use the kernel-reported running image for cache-access classification and trust paths. The rewrite avoids trusting the mutable external alias: retargeting it before discovery or during path construction cannot admit another app's identity.

Validation: 53 focused tests passed, including signed executable launches through direct, relative, chained, and parent-directory aliases; retargeting regressions; and existing no-UI/prompt-safety audits. Tests used in-memory Security objects rather than real Keychain item operations, so this proves process/path/signature behavior rather than live cache-item interoperability. A test-expression lint failure was corrected, managed P2 reviews passed, and every CI check passed on final head b6c42b3e00d5edfa4765cc6fb231507bb7956c1c: https://github.com/steipete/CodexBar/actions/runs/35139978750.

Main was pulled and the merge verified. The changelog and CLI documentation include the change.

PR #3692, opened 2026-09-16
Fix the reporting date on shared usage snapshots

Chipagosfinest:
Shared usage cards used the chart's next-day boundary as their “Data through” date, so September 16 data could be labeled September 17. Formatting in the Mac's timezone could also shift a pinned reporting day.

Use the last included civil reporting day and carry the dashboard timezone into the formatter shared by PNGs and copied text. Chart bounds, tokens, and spend remain unchanged. Seven regression cases cover UTC, UTC+14, Los Angeles, daylight-saving transitions, and year-end.

Validation: 57 focused sharing and architecture tests passed; `make check` passed with zero violations. Fresh production-renderer before/after proof is [attached here](https://github.com/steipete/CodexBar/pull/3692#issuecomment-5706299496); both images use identical synthetic totals, and their published bytes match the inspected local PNGs. Generated proof is kept in PR attachments. Managed independent review is clean through P2. Full repository CI remains the landing gate.

Contributor implementation by @Chipagosfinest, with maintainer regression/proof updates and release-note credit preserved.


steipete:
Fresh maintainer proof: the production renderer now labels synthetic September 16 usage as **September 16**, preserving the same 10 tokens and $2 estimate. The before fixture recreates the previous builder's next-day boundary and current-timezone formatter inputs; the after fixture uses the corrected builder. Both images contain synthetic aggregate data only.

All 57 focused sharing and architecture tests passed, including seven reporting-date cases across UTC, UTC+14, Los Angeles, DST transitions, and year-end. The after PNG is byte-identical to the contributor's committed proof. Posting both images here so the PR can retain reviewable source and tests without a generated binary in the patch.

Before:
Before: identical September 16 data labeled September 17

After:
After: identical September 16 data labeled September 16

PR #3696, opened 2026-09-16
fix(openrouter): standardize pay-as-you-go balance and spend presentation

Chipagosfinest:
Uncapped OpenRouter keys expose useful spend and prepaid balance, but previously had no consistent pay-as-you-go summary. Show reported monthly key spend, lifetime key usage, or lifetime account usage with an explicit scope and period. Missing values remain unavailable; a reported zero remains zero.

Capped keys keep their existing single quota meter. Management-key counters are excluded from the key-spend summary, and prepaid balance comes only from the selected key's credits response. The shared menu model removes duplicate detail rows while their replacement summary is visible and restores them when it is hidden; raw CLI/JSON details remain intact. CLI text does not invent a zero-dollar budget for uncapped spending.

The rewrite removes the separate Swift test implementation of the JavaScript projection. Behavioral tests now exercise the real plugin in both JavaScript engines, through native models, CLI text, and JSON.

## Validation

- `make test`: 1,178 selections across 107 groups passed without failures, timeouts, or retries.
- `make check`: passed.
- Focused actual-plugin, quota, presentation, CLI, and architecture tests passed.
- Independent isolated review: no actionable P0–P2 findings.
- Developer ID-signed native proof rendered the shared menu in light and dark appearances.
- Live official OpenRouter API reads succeeded before and after through signed packaged CLIs. The new summary's monthly amount and optional balance matched the retained response details; uncapped output had no quota meter. Private billing data and credentials are omitted from the proof below.

## Before and after

Both images use the same synthetic response. Before renders without the newly added cost summary; after renders the final plugin projection.

| Before | After |
| --- | --- |
| Before: uncapped spend in raw details | After: scoped PAYG summary without duplicate detail rows |


Chipagosfinest:
@clawsweeper re-review

PR #3699, opened 2026-09-17
fix(antigravity): skip foreign databases without inventing empty history

urda:
Antigravity 1.2.3 writes `conversation_summaries.db` into a declared local-history root. That auxiliary database has no `gen_metadata` table. The old Boolean schema check treated a successfully completed inspection as a failure, so one unrelated file made valid neighboring history partial and withheld its usage.

Distinguish supported history, proven foreign databases, and unsupported or unreadable schemas. Skip foreign files without withholding valid history, and track recognized source presence independently of event count: unrelated databases alone leave history unavailable, while supported empty history remains complete. Undecodable schema names, unknown `gen_metadata` columns, empty or truncated schema walks, and failed inspections remain incomplete. WAL immutable fallback, bounded reads and handle ownership remain intact; discovering SQLite paths still prevents replacement with an older JSONL cache.

Validation: the added foreign-only regression reproduced four failed assertions on the original proposal. `make test-fast FILTER='Antigravity|ProviderArchitectureGatekeeperTests'` now passes 502 tests across 37 suites, including mixed valid/foreign files, foreign-only roots with a populated JSONL cache, supported empty sources in both orders, invalid UTF-8 schema names, schema drift, bounded walks and sidecar-less WAL. `make check` reports zero violations; independent P2 review of the full candidate is clean. Maintainer tests use synthetic files and the production reader/fetcher, with no live accounts or credentials.

The contributor also reported a full 107-group suite and a production CLI comparison over their real local history: 44 daily rows and 182,636,128 tokens with or without the summaries database; a drifted-schema control still withheld incomplete data. Those are contributor-reported results from the original proposal. Final macOS/Linux CI for the updated head is the landing gate.

Thanks @urda; the original fix and contributor credit are preserved.


steipete:
Landed as 7e3c7be10050a09f96ccbde152d7b85744b350b5 after all macOS/Linux CI checks passed: https://github.com/steipete/CodexBar/actions/runs/35176027099.

The reader now skips proven unrelated SQLite files beside valid history. The added source-presence guard keeps foreign-only roots unavailable and supported empty histories complete; undecodable schemas remain incomplete. SQLite still controls source selection, so a foreign file does not activate an older JSONL cache.

Validation: the foreign-only control reproduced four failures in the original proposal. `make test-fast FILTER='Antigravity|ProviderArchitectureGatekeeperTests'` passed 502 tests in 37 suites; `make check` and full-candidate independent P2 review passed. Maintainer proof used synthetic files and the production reader/fetcher. Thanks @urda for the original repair and the reported real-history comparison.

PR #3545, opened 2026-09-10
Add opt-in SSH costs for Claude and Codex

wicolian:
## Summary

- Adds a default-off **Fetch costs from SSH devices** toggle with up to eight SSH config aliases.
- Combines explicitly selected same-account SSH boxes into each provider's existing Today/history totals.
- Stacks SSH contribution into the existing daily bars in a separately configurable color, including SSH-only days.
- Keeps Claude and Codex separate, marks incomplete combinations partial, and avoids presenting Mac-only model breakdowns as combined truth.
- Generalizes `cost --remote` and `--summary-only` to Claude, Codex, and both.
- Does not use or require Agent Sessions.

## Safety

- Uses system SSH with BatchMode, a five-second connect timeout, existing host-key trust, and no interactive prompts.
- Transfers bounded numeric summaries and daily buckets only; no transcripts, paths, model names, session IDs, account identifiers, or credentials.
- Limits remote daily history to the requested window with a hard 365-day ceiling and a 64 KiB output cap.
- Account sameness is an explicit local choice; copied/resumed logs on multiple machines can therefore double-count.
- Throttles normal refreshes to once per 15 minutes and clears in-memory reports when disabled.

## Live UI proof

Previous exact-branch proof of the combined Claude chart and configurable SSH color:

Live combined local and SSH cost chart

Previous exact-branch settings proof with `sandbox` selected while Agent Sessions remains independent:

Live SSH cost settings

## Current update

- Merged current upstream `main` at `2b7816493`; current head is `03f937c29`.
- Fixed both actionable review findings: remote-only history now opens from the outer availability gate, and a missing local snapshot is explicitly partial.
- Wired selected SSH cost data into the normal provider cards, Overview totals, and compact stacked history bars.
- Combined snapshots intentionally omit Mac-only model/project/session breakdowns, so a local Sonnet label cannot be mistaken for the SSH box's Opus/Fable activity.
- Fresh live `sandbox` provider-native 90-day payload:
  - Codex: today $368.5050352 / 480,603,369 tokens; 90 days $778.5565562 / 772,862,499 tokens.
  - Claude: today $433.74387855 / 642,943,632 tokens; 90 days $5,189.458752 / 8,525,889,441 tokens.
  - Both reports were complete.
- SwiftFormat and `git diff --check` pass. The first current-head CI lint run found one parameter-count violation; `03f937c29` removes it and the replacement CI run is pending.
[...]

PR #3700, opened 2026-09-17
fix(sessions): preserve SSH username case

steipete:
SSH destination deduplication lowercased the entire `user@host` string, dropping distinct case-sensitive Unix accounts such as `user@host` and `USER@host`.

Preserve the username through the final `@` and normalize only the hostname for comparison. Original destination strings, first-occurrence ordering, and rejection of options, whitespace, and control characters remain intact. The regression fails on main and now retains both users while still removing hostname-only duplicates.

Validation: all 12 Tailscale/session tests pass, `make check` passes with zero violations, and managed independent review is clean through P2. Tests use strings and fixtures; no SSH connections were made. Full repository CI is the landing gate.

Extracted from #3538/#3545 with @ECuteri's contribution preserved. The SSH cost-aggregation proposals remain open separately.


steipete:
The username-case repair is now on main through #3700, merged as b88c78718da1372c341dae2cde6d53b830a1c813. Destinations such as `user@host` and `USER@host` stay distinct, while hostname-only case differences still deduplicate.

The regression reproduced the missing destination before the fix. `make test-fast FILTER='TailscaleSessionTests|RemoteSession'` passed 12 tests; `make check` and independent P2 review passed. All macOS and Linux CI checks succeeded at https://github.com/steipete/CodexBar/actions/runs/35170771260. These were sanitizer fixtures; no SSH connections were made.

Thanks @ECuteri. This fixes only destination identity handling; the broader remote cost-report proposals in #3538 and #3545 remain open.

PR #3701, opened 2026-09-17
fix(providers): honor resolved terminal fetch errors

steipete:
An unavailable final Antigravity source could replace a useful earlier API, transport, or sign-in error when it stopped fallback. The pipeline had already selected the right error through its provider resolver, but its terminal return discarded that result.

Use the resolved error for terminal failures, matching exhausted fallback chains. Cancellation remains authoritative; routing and per-attempt diagnostics still receive each original error, and later substantive failures retain their existing precedence. This extracts the bounded repair from https://github.com/steipete/CodexBar/pull/3683; that broader proposal remains open.

Validation: reproduced 13 failing assertions before the fix; 110 tests in ProviderFetchErrorTests, AntigravityCLIHTTPSFetchStrategyTests, and supporting provider coverage now pass. `make check` reports zero violations. Independent P2 review is clean. Tests use synthetic strategies and injected transports; no live accounts, cookie imports, or Keychain reads were used.

Thanks @hhh2210 for the original proposal.


steipete:
The terminal-error repair is now on main through #3701, merged as e56bb5b201830e1fb3354bddb3bb08fee9d836dc. An unavailable final source now preserves the error selected by the existing provider policy, while cancellation and later substantive failures keep their existing precedence. Each source still records its original diagnostic.

Validation: `make test-fast FILTER='ProviderFetchErrorTests|AntigravityCLIHTTPSFetchStrategyTests|ProviderArchitectureGatekeeperTests'` passed 110 tests after reproducing 13 failing assertions on the prior implementation. `make check`, independent P2 review, and every macOS/Linux CI check passed: https://github.com/steipete/CodexBar/actions/runs/35173592826. All provider inputs were synthetic.

Thanks @hhh2210. This extracts only the terminal resolver repair; the broader first-error/per-source diagnostic proposal in #3683 remains open.

PR #3702, opened 2026-09-17
fix(codex): redact System Account menu labels

steipete:
The Codex System Account submenu displayed full email addresses even with Hide Personal Info enabled. Its title builder used raw account names while the neighboring account switcher already applied privacy labels.

Use the existing stable account-number and workspace-label rules for this submenu when privacy is on. When privacy is off, preserve the original display names exactly, including Personal and workspace discriminators. The value-only menu builder preserves item order, checked state, promotion targets, blocking, and submenu visibility.

Validation: the regression reproduced three failed assertions before the fix. `make test-fast FILTER='CodexSystemAccountPrivacy|CodexAccountSwitcherRedaction|ProviderArchitectureGatekeeper'` passed 56 tests. Signed native before/after captures use the production item builder with synthetic accounts; they demonstrate the email leak and the corrected menu. No saved settings, credentials, provider calls, or account promotion actions were used. `make check` passed with zero violations; final independent P2 review is clean. Full macOS/Linux CI is the landing gate.

This extracts the privacy repair from https://github.com/steipete/CodexBar/pull/3550; the broader switching redesign remains open. Thanks @zenibako.


steipete:
Hide Personal Info now applies to the Codex System Account submenu. These captures show the same synthetic accounts through the production item builder and a native NSMenu in a Developer ID signed, isolated test host. The checked and enabled states are identical; no account actions were invoked.

Before, with Hide Personal Info on:
Before: System Account submenu exposes synthetic emails

After, with Hide Personal Info on:
After: System Account submenu uses stable private labels

The value-only tests also cover privacy off, duplicate workspace labels, embedded email-like workspace text, stable numbering, promotion targets, blocked actions, and the existing submenu visibility rules. This is synthetic native-menu proof; it does not claim a live-account switching run.

steipete:
The existing System Account submenu now honors Hide Personal Info on main through #3702, merged as 141ecf642a97d9ce4dadfd70102a790eb24578a5. It uses stable private account labels while keeping promotion targets and menu state intact; privacy off retains the original display names.

Validation: `make test-fast FILTER='CodexSystemAccountPrivacy|CodexAccountSwitcherRedaction|ProviderArchitectureGatekeeper'` passed 56 tests. Signed synthetic native before/after proof is at https://github.com/steipete/CodexBar/pull/3702#issuecomment-5707505346. `make check`, final independent P2 review, and every macOS/Linux check passed: https://github.com/steipete/CodexBar/actions/runs/35174630916. The only integration conflict was in the changelog; source and test blobs remained unchanged.

Thanks @zenibako. This repairs the existing submenu's privacy omission; the broader account-switching redesign in #3550 remains open.

PR #3703, opened 2026-09-17
fix(cursor): stop retries for unconfirmed cost credentials

steipete:
Cursor cost refresh could enter an endless sequence of forced requests after a successful fetch whose cookie remained unconfirmed in the app's current account scope. Publication correctly rejected the result, but treated every rejection as a scope change that required an immediate replacement.

Distinguish unchanged credential mismatches from real account or settings changes. Unchanged mismatches leave accepted usage and error state intact, clear only the matching fetch metadata, and wait for the next normal or manual refresh. Genuine changes still retry, and a fetch that confirms its own initially unresolved cookie can publish immediately. Capture the cost-settings revision so cookie confirmation cannot mask a timezone change during the request.

Validation: 23 regression assertions reproduced the loop and timezone race before the fix. `make test-fast FILTER='CursorCostRefreshRetryTests|UsageStoreCoverageTests|UsageStoreLocalLedgerPublicationTests|CursorImportedSessionScanningTests|ProviderArchitectureGatekeeperTests'` passes 109 tests. `make check` reports zero violations; independent P2 review is clean. Fixtures use dictionary-backed settings, a no-op Keychain policy, per-store fingerprint readers and injected loaders; no live account or Keychain access is involved. Full macOS/Linux CI is the landing gate.

Related to #3249. This fixes the demonstrated cost-refresh scheduling defect; the report's broader Keychain diagnosis and historical Claude-refresh starvation remain unproven, so the issue stays open.


steipete:
The demonstrated Cursor cost retry loop is now fixed on main by #3703, merged as 5cf8cee2b5df21b9f01099ccea4fef955ebf09cb. A result with still-unconfirmed credentials remains unpublished, but unchanged account state no longer schedules an endless chain of forced requests. Accepted usage remains intact, normal/manual refreshes can retry, and real account or settings changes still get a replacement request. The same change prevents cookie confirmation from masking a cost-timezone change.

Validation: 109 focused tests passed across the new scheduling regressions, existing publication controls, imported-session tests and architecture checks. `make check`, independent P2 review and all macOS/Linux CI checks passed: https://github.com/steipete/CodexBar/actions/runs/35175271140. Tests used injected results and isolated stores rather than live accounts.

This addresses the reproducible cost scheduling defect in #3249. Its broader Keychain diagnosis and historical Claude-refresh starvation remain unproven, so #3249 stays open for those remaining symptoms.

PR #3704, opened 2026-09-17
Lead the shared usage card with spend

Chipagosfinest:
Scope note: the singular-caption fix that started here landed separately in #3706, so this PR is now
purely the presentation change.

## Before / after

Sparse card — 1 subscription, no model history:

| Before | After |
|---|---|
| before sparse | after sparse |

Dense card — 5 subscriptions, 3 models, `~$123,456.78`:

| Before | After |
|---|---|
| before dense | after dense |

Multi-currency — 3 currencies:

| Before | After |
|---|---|
| before multi | after multi |

## What changed

| | Before | After |
|---|---|---|
| Hero | token count at 104pt; spend at 32pt on the right | spend at 76pt; tokens at 42pt supporting |
| Baselines | two independently-spaced columns, labels on different lines | one `HStack`, same label / value / detail rhythm |
| "local / aggregate" | stated 4x: badge, "native currencies kept separate", "Only aggregate usage…", footer | stated once, in the footer |
| Header badge | `LOCAL SNAPSHOT` (duplicated the footer) | removed |
| Empty models column | a sentence floating at the top of the column | a row-shaped empty state matching the real rows |
| Secondary currencies | a bare "+N more" pointer | explicit per-currency totals via `ShareStatsFormatting.spendCoverage` |

Card size stays `1200x630` — it is an OG image and the ratio is load-bearing.

## What the renders settled

**Dense fits.** The `286pt` rankings reservation is load-bearing. An earlier revision of this branch
removed it, and since SwiftUI does not clip an oversize frame, five provider rows plus three model
rows would have pushed the footer out of the image. Restored, alignment only.

**The hero no longer scales.** `~$123,456.78` renders unscaled at 76pt. At the original 104pt it
needed roughly 660-780pt against 654pt of available width (`1200 - 104 padding - 52 spacing - 390
right column`), so the headline would have silently shrunk on realistic values. The 104pt slot
previously held `compactCount` output, at most six characters, so it never exercised
`minimumScaleFactor`.

**Two of my own earlier choices were wrong, and the renders caught both.** Centring the rankings
split the sparse card's empty space and put a gap under the divider that reads as a fault; top
alignment leaves one contiguous region that reads as absent content. And replacing `LOCAL SNAPSHOT`
with the period label put `30 DAYS` directly above `EST. 30-DAY SPEND` — one redundancy traded for
another, so the badge is gone entirely.

## Method, and its limit

`ShareStatsCardView` is a pure function of `ShareStatsPayload`, so these were produced by compiling
the view verbatim alongside the payload types and `ShareStatsFormatting` from each revision and
rendering through `ImageRenderer` at the card's own size. Only `UsageFormatter.currencyString` and
`SpendDashboardSource.scanDays` are stubbed, neither of which affects layout. Method and payloads are
recorded in `docs/research/share-card-layout.md`.

**These are not production-bundle renders** — the app was not built for them. The view code is
identical to this branch so layout and typography are faithful, but if you want output from a signed
build before merging, say so and I will produce it.

## Tests

- `spend coverage preserves secondary currency when subscriptions overflow` — single, dual and
  triple-currency cases pinning the exact coverage string.
- The caption pluralisation test now comes from main via #3706; the duplicate here was dropped.
- All 400 `ProviderArchitectureGatekeeperTests` line anchors re-verified: zero stale.


Chipagosfinest:
@clawsweeper re-review

Chipagosfinest:
@clawsweeper re-review

Chipagosfinest:
@clawsweeper re-review

steipete:
The singular subscription caption correction has landed in #3706, with 58 focused tests, green macOS/Linux CI, and inspected before/after production renders. The card now says “1 subscription” for a single included subscription.

This PR remains open for the larger presentation proposal. Thanks for contributing the formatter and identifying the caption issue.

Chipagosfinest:
Cross-link: [#3747](https://github.com/steipete/CodexBar/pull/3747) keeps retained top-model families on the shared card when a sibling USD provider is unpriced (the empty **TOP MODELS** column on current 0.61.0). Independent of this spend-first layout. Verified against installed 0.61.0.

@clawsweeper this is a related ranking fix, not a substitute for the layout proof here.

Chipagosfinest:
Added production-path proof: `Tests/CodexBarTests/ShareStatsLayoutProductionRenderTests.swift` renders the sparse/dense/multi payloads through `ShareStatsRenderer.pngData` (`NSHostingView`) — the literal path `ShareStatsExporter.saveImage`/`copyImage` use in production — instead of the standalone `ImageRenderer` harness `docs/research/share-card-layout.md` already disclaimed as non-production. Each PNG is asserted non-trivial (`> 10_000` bytes) so the test fails if the frame ever renders blank/degenerate, and the resulting images are committed under `docs/research/share-card-layout/assets/production-*.png` plus linked from the doc.

@clawsweeper re-review — this should close the outstanding proof gap.

Chipagosfinest:
Fixed the broken image links — `docs/research/share-card-layout.md` was linking `assets/production-*.png`, which resolves relative to `docs/research/` (where the doc lives), not `docs/research/share-card-layout/` where the images actually landed. Corrected to `share-card-layout/assets/production-*.png` and verified all three resolve on disk.

@clawsweeper re-review

steipete:
Thanks @Chipagosfinest for the layout work and rendering evidence. Retaining the current headline layout and closing this proposal. The new helper loses the second currency’s coverage, and the headline selects the first currency even when its spend is unavailable. The singular-caption fix already landed in #3706. A future spend-first design should explicitly handle token-only data and unequal currency coverage.

PR #3706, opened 2026-09-17
fix(share): use singular subscription captions

steipete:
Shared usage cards now say “1 subscription” when exactly one subscription is included. The existing caption uses the contributor's small pluralization formatter; zero/multiple subscriptions and the currency-overflow caption retain their previous text. The same synthetic card was rendered before and after with identical totals, provider rows, and reporting date.

Validation: `make test-fast FILTER='ShareStatsTests|ShareStatsPeriodTests|ProviderArchitectureGatekeeperTests'` passed 58 tests, including the production PNG renderer and 0/1/2/12 subscription cases. `make check` reports zero violations, and independent P2 review is clean. Both attached images contain only the deterministic fixture from ShareStatsPeriodTests; the after render changes the singular caption.

This extracts the caption correction from #3704. The larger presentation proposal remains open. Thanks @Chipagosfinest for the contribution.

Before: one synthetic subscription uses the plural caption

After: the same synthetic card says one subscription


steipete:
Landed as 1239617059f74482bbc9d22c629bc4039f7262b0. The shared card now uses “1 subscription” for one included row; plural counts, currency overflow, layout, totals and reporting date retain their existing behavior.

Validation: 58 sharing/architecture tests passed, including 0/1/2/12 counts and the production PNG renderer; `make check` and independent P2 review passed. The inspected [before/after images in this PR](https://github.com/steipete/CodexBar/pull/3706) differ only within the caption, and both published PNGs match their local bytes. All macOS/Linux checks passed on head 90a7eecbf4695cb4daba4bca7fa763ada2efd413 in [CI run 35184697604](https://github.com/steipete/CodexBar/actions/runs/35184697604).

Thanks @Chipagosfinest. The broader #3704 presentation proposal remains open.

Issue #1721, opened 2026-06-23
Mistral: Subscription quota broke on latest CodexBar update (0.37.x)


lfmundim:
I describe this issue more thoroughly on #1689, but long story short the token passed to the method that retrieves Mistral usage quota from the API is now insufficient after some trimming of it. #1689 adds back in only whats necessary, keeping the security implemented during 0.37.x development

PR #3710, opened 2026-09-17
Show Mistral included API and Vibe Code allowances

JoPaMu:
Show Mistral's reported Included API and Vibe Code allowances alongside API spend and credits. Automatic menu-bar display keeps API cash spend; Included API and Monthly Plan can be selected independently. Labels and amount details use shared presentation hooks across cards, menus, widgets, and CLI. Missing reset dates never turn monetary details into reset clocks.

The subscription parser reads JSON records from the page's Flight stream, honors byte-framed text/binary records, and validates each allowance independently. Optional subscription failures preserve billing, credits, and the existing Vibe fallback. The cleanup removes unused wrappers, a permanently disabled replay test, and committed screenshot binaries.

Validation: `make test` passed 1,183 selections in 108 groups with no failures, retries, or timeouts. Final main integration passed focused Mistral, OpenRouter, CLI, and architecture checks plus `make check` and an independent isolated P0–P2 review. Signed native proof below renders synthetic billing and allowance data in the real shared card; light and dark appearances were inspected.

A live web fetch was attempted, but no usable Mistral session was available. Authenticated subscription-page compatibility remains unverified locally; parser and request/fallback behavior have controlled-response coverage.

| Before | After |
| --- | --- |
| Before: API spend and balance | After: distinct Included API and Monthly Plan allowances |

Related: #1721.

Final-head [CI 35332025424](https://github.com/steipete/CodexBar/actions/runs/35332025424) passed lint, Linux builds, and both macOS test shards.


JoPaMu:
@clawsweeper re-review

Both findings are addressed and the requested proof is in the PR body:

- **P2**: the `.mistral` special case is replaced by a shared `extraRateWindowUsesResetDescriptionAsDetail` presentation hook (used by Sub2API and Mistral). `StatusItemController+Animation.swift` is fingerprint-identical to main; the only gatekeeper edit is the sub2api detail cluster going from two references to one, because the special case that produced the second reference is gone. `ProviderArchitectureGatekeeperTests` passes.
- **P3**: `MistralProviderDescriptor` now supplies a `rateWindowLabeler`, covering both budget-present ("Included API") and budget-absent ("Balance") snapshots.
- **Real behavior proof**: freshly built app screenshots of the real `UsageMenuCardView` (light + dark), plus redacted output from the production fetch path (`fetchSubscriptionBudgets` → `attachSubscriptionBudgets`) replaying a captured authenticated subscription page.
- `make test` is green on the full sharded suite, and `swiftlint --strict` / `swiftformat` are clean. Xcode is installed now, so the earlier build limitation is gone.

PR #3712, opened 2026-09-17
Add Grok Bot layout token

yicone:
## Summary
- Add a reusable named-extra percentage layout token, initially for Cursor Grok Bot allowance.
- Store the full layout, provider overrides, and conditional library in new V4 keys while retaining V3 projections that 0.60.4 can decode.
- Keep explicit reset-window selections in V3; only the new named-extra token is omitted there, so an unchanged downgrade no longer falls back to V2 and loses reset choices.
- Keep V2/V1 projections for older releases and reconcile V4 against V3/V2/V1 so edits made by an older release still win.

## Validation
- `swift test --filter MenuBarLayoutTests` (64 tests)
- `swift test --filter MenuBarLayoutV4PersistenceTests` (3 tests)
- SwiftFormat 0.63.0 lint on all changed Swift files
- SwiftLint 0.65.1 `--strict --no-cache` (0 violations)
- The V4 regression suite decodes the V3 layout and override projections with a 0.60.4-style token surface; it also covers V3 edits winning on return-upgrade across the global layout, provider overrides, and conditional library, including a deliberate V3 library clear.
- Manual local UI verification of the Grok Bot palette token was completed before this compatibility fix. Rebuilt-app screenshot evidence is pending.


yicone:
@clawsweeper re-review

Addressed the V3 compatibility finding in commit `b07798bf0`: full-fidelity data now uses V4 keys, while V3 retains every pre-existing token including explicit reset windows and omits only the new named-extra token. Added downgrade/return-load regression coverage for global and provider layouts.

steipete:
Thanks @yicone. Your named-extra token and persistence work have been adopted in #3829 with co-author credit. V4 is needed because released 0.60.x token decoders reject unknown cases; the V3 projection preserves explicit reset selections. The implementation also hides unavailable allowances and refreshes when only Grok Bot changes. All 375 focused tests pass, with synthetic before/after proof in the PR. #3829 landed on main via #3830 and supersedes this implementation.

yicone:
@clawsweeper re-review

PR #3715, opened 2026-09-17
fix(codex): refresh selected account cards in open menus

### LINKED ISSUE #3709: Codex account switch leaves the usage card blank in segmented layout until a provider round-trip
### Environment

- CodexBar 0.60.3 (observed), macOS 15 (Darwin 25.6.0), Apple Silicon
- 3 Codex accounts: 1 `liveSystem` + 2 managed, added via the in-app **Add Account**
- Codex provider `source: oauth` in `~/.config/codexbar/config.json`
- `multiAccountMenuLayout` **unset** in `com.steipete.codexbar` → resolves to `.segmented`

### Symptom

With the segmented (tab strip) multi-account layout:

1. Select a different Codex account tab → the usage card stays empty.
2. Switch to the Claude provider tab, then back to that Codex account → that **one** account now renders its usage correctly.
3. Select a third Codex account tab → blank again.

So a provider round-trip repairs exactly one account at a time; an account-to-account switch never does.

### The data layer is healthy

```
codexbar usage --provider codex --all-accounts --format json
```

returns all three accounts with `dataConfidence: "exact"`, fresh `updatedAt`, `providerCost`, and 2–3 `extraRateWindows` each. Fetching per account works fine from the CLI, so this is a menu/publication problem, not an auth or network one.

### Likely cause

`~/Library/Application Support/CodexBar/codex-account-snapshots.json` only ever held **one** record (the `liveSystem` account) while the layout was segmented.

That lines up with `shouldFetchAllCodexVisibleAccounts()`:

https://github.com/steipete/CodexBar/blob/main/Sources/CodexBar/UsageStore%2BTokenAccounts.swift#L179-L185

```swift
func shouldFetchAllCodexVisibleAccounts() -> Bool {
    guard !self.shouldUseAmbientCodexPATForUsage() else { return false }
    let projection = self.freshCodexVisibleAccountProjectionForAccountRefresh()
    return self.settings.multiAccountMenuLayout == .stacked && projection.visibleAccounts.count > 1
}
```

With `.segmented` this returns `false`, so the fan-out never runs and no per-account usage is persisted for the non-active accounts. In that state a newly selected tab has nothing behind it, and the only way it can show numbers is if the switch path itself performs and publishes a live fetch — which is the part that appears not to land.

This reads as a gap rather than intended behaviour: the segmented layout exists precisely to switch between accounts, but the fan-out that makes per-account data available is gated on `.stacked`.

### Workaround (confirmed)

```
defaults write com.steipete.codexbar multiAccountMenuLayout stacked
```

(app quit before writing, then relaunched)

`codex-account-snapshots.json` went from 1 record to 3 on the very first refresh, all with `dataConfidence: "exact"`, and every account renders. This sidesteps the switch path entirely rather than fixing it.

### Notes

- Not re-tested on 0.60.4: the upgrade happened after the workaround was applied, so the segmented path was not exercised again on the newer build.
- Possibly related: #1527 and #3125 (tab switch showing stale/previous account data), #3168 (card repaints only on re-clicking the active tab), #3347 (stacked view showing one account at a time). #3550 notes the Claude and Codex switchers are still not unified.
- Suggestion: either allow the fan-out in segmented layout when more than one visible account exists, or have the account-switch path schedule the same revalidation that `menuWillOpen` triggers.

sandeep780049:
Fixes #3709. Selecting another Codex account could leave its usage card blank even after fetching succeeded because the refresh callback deferred the root-menu rebuild until tracking ended.

Schedule account data updates through the existing event-tracking/default run loop, preserving request coalescing, hosted submenus, and highlighted native commands. Recheck account and provider ownership before enqueueing and before rebuilding. Explicit interaction resumptions use the same tracking-safe path; ordinary plugin/manual refreshes retain their existing yielding task order. Fetching, credentials, and account activation stay in their existing layers.

Validation on `b12e07cdd8f82b955f0c756c27d3be6e9dda8424`: `make test` passed all 1,185 selections across 108 groups on the first attempt, without retries or timeouts; `make check`, 119 focused tests, and independent P0–P2 review passed. A freshly signed native test selects the attached B button in an actual tracking status menu, holds one synthetic fetch, then requires B and 17% in the attached card's screenshot/OCR before tracking returns. The baseline publishes that same usage but leaves the card blank. It uses contained fixtures and the real publication/ownership flow, without real provider access, scheduler overrides, manual rebuilds, or reparenting. Cleanup waits for the entire scoped refresh and widget task.

The images below were completely inspected and contain only synthetic data. [Final native proof and source/binary provenance](https://github.com/steipete/CodexBar/pull/3715#issuecomment-5729088433). [CI for the validated fix](https://github.com/steipete/CodexBar/actions/runs/35335617561) passed. The subsequent main integration at `befa45d8dabe4effd5809809f2720da85c8f098d` retains the fix and native harness unchanged, with clean `make check` and independent P0–P2 integration review. A combined integration run at `22f32150a3461d28e9cea58d74123680b830d894` also passed all 1,200 selections across 111 groups without retries or timeouts, including this unchanged menu fix, the Claude progress candidate, and current provider changes. [Current-head CI](https://github.com/steipete/CodexBar/actions/runs/35342771556) passed all checks.

Thanks @sandeep780049 for the original fix and investigation.

| Before: fetched usage remains blank | After: B usage appears while open |
| --- | --- |
| Baseline leaves the synthetic B usage card blank | Final candidate shows synthetic B usage at 17 percent |


sandeep780049:
@clawsweeper re-review

sandeep780049:
**Native macOS proof (revision 3) — request fulfilled**

Ran the account-switch flow on a real macOS 26 GitHub Actions runner (real AppKit, menu card rendering enabled, no test scheduler or rebuild mocks for the scheduling path): [workflow run](https://github.com/sandeep780049/CodexBar/actions/runs/35282417026).

**Transcript** (from [`proof/3709-artifacts`](https://github.com/sandeep780049/CodexBar/blob/proof/3709-artifacts/transcript.txt), PNGs `before-account-switch.png` / `after-account-switch.png` in the same branch):

```
before select: published email live@example.com
selected managed account; published email <nil>            ← card cleared during switch
pre-fetch drained; rebuilds 1
fetch resumed with managed@example.com snapshot (17%)
after fetch: published email managed@example.com           ← usage lands while menu is still open
hosted submenu still open after fetch: true                ← open chart submenu survives
parent rebuilds while submenu open: 0                      ← rebuild skipped while chart is open (preserved)
after submenu close: total parent rebuilds 2               ← deferred rebuild reconciles on chart close
```

Assertions in the run (all passing): new account's usage published into the still-open menu, hosted chart submenu preserved across the delayed rebuild, parent menu never closed, deferred parent rebuild lands after the submenu closes. Before/after renders of the actual menu card are captured as PNGs.

Also validated in the same macOS run: `StatusMenuCodexSwitcherTests` (incl. the new regression test) and `StatusMenuOpenRefreshTests` (incl. the new hosted-submenu reconciliation test) — 55 tests, 0 failures.

steipete:
Native before/after verification used a signed test host on macOS 27.0 (26A428), an actual tracking status menu, its attached B button, and exactly one held synthetic account fetch. The baseline publishes B's 17% usage but leaves the card blank; candidate `8b4832e665606d4de151569f151dba5de2be16e3` shows B and 17% before tracking returns. The harness waits for the complete scoped refresh and widget task before cleaning up.

Both complete captures were inspected and contain only synthetic data. No real provider or credential access, scheduler override, manual menu rebuild, or reparented card was used. Full-suite follow-up and CI are still in progress.

| Baseline after fetch | Candidate after fetch |
| --- | --- |
| Baseline still has a blank usage card | Candidate shows synthetic B account usage at 17 percent |

Image SHA-256: before `93ec39de09eaa6380be65c6df1a2a019cf119d42c53e1463bd0ae1d5fe2c4ff5`; after `81fe7fe0c2526f5396335cfddaf95941b3513acd46b322968b76d9f53431a572`.

steipete:
Final local validation passed on `b12e07cdd8f82b955f0c756c27d3be6e9dda8424`:

- `make test`: all 1,185 selections across 108 groups passed on the first attempt, with no retries or timeouts (1,123 seconds).
- `make check`: zero violations. Independent P0–P2 review is clean. The 119 focused menu, plugin, cleanup and architecture tests passed.
- Freshly signed native `StatusMenuCodexAccountSwitchNativeProofTests` passed on macOS 27.0 (26A428). The attached B button starts exactly one held synthetic fetch; the attached card shows B and 17% while the same menu tracks. The complete scoped refresh and widget task are drained before fixture cleanup. No real account/provider access or scheduler override is used.

The full-suite check exposed an ordering regression in an earlier draft: applying the tracking scheduler to every ordinary refresh could rebuild plugin cards before queued store observations settled. The final change limits that scheduler to account updates and explicit interaction resumptions, preserving the original ordinary refresh ordering. The unchanged plugin regression now passes. An adjacent cleanup test now awaits its owned sequence tasks instead of polling for an impossible load count.

[Baseline after-fetch capture](https://github.com/user-attachments/assets/8b0fa0c9-2dec-417c-9bb1-4d1fd626be11) remains unchanged. The new final-source capture below was fully inspected and contains only synthetic data. Its clean source tree is `7d978b48f52f1562be841e30c05e80ca8c708bc8`; signed test-binary SHA-256 is `5a7f95e736a87fe81abdeaa7865ef5d9f4fa5ac8a1b0e9eaa9292e9435a0ff0d`.

[Final-head CI](https://github.com/steipete/CodexBar/actions/runs/35335617561) is still pending; this PR has not been merged yet.

Final source: synthetic account B usage appears at 17 percent while the menu stays open

PR #3713, opened 2026-09-17
fix(sharing): retain gateway model families in shared output

Chipagosfinest:
Complete OpenRouter Activity history could show no model rows in shared cards or copied statistics: the producer retained `vendor/model` identifiers, but the sharing sanitizer rejected every slash before matching known public model families.

Accept one nonempty namespace after validating the entire input, then apply the existing regional-prefix and public-family matching. Shared output still contains only fixed labels such as GPT or Claude, with usage attributed to OpenRouter. Raw namespaces, private model suffixes, unknown families, malformed/deeper paths, and sensitive identifiers remain excluded. Unknown families still contribute to provider totals. The existing exclusion of incomplete currency-group rankings remains unchanged; #3714 is a separate policy proposal.

The regression matrix covers plain and namespaced families, case/spacing, private namespaces, regional names, short reasoning IDs, and the existing privacy/shape checks. A synthetic integration fixture runs through the bundled OpenRouter plugin in both JavaScript engines, the shared mapper, dashboard builder, share builder, copied-text formatter, and actual native card renderer.

Baseline: 20 failed regression assertions and no shared model rows. Candidate: 29 focused sharing/incomplete-history tests passed, followed by 101 expanded OpenRouter/sharing tests after main integration. The signed native after-capture was rerun on the integrated source. Both signed native captures below use identical synthetic data: 900 total tokens and $0.9375 total spend. After the fix, GPT combines 700 tokens/$0.75 and Claude retains 150 tokens/$0.125, all attributed to OpenRouter; the unknown family's 50 tokens/$0.0625 remain in the provider total. The 29/30 coverage reflects the dashboard's current-day window versus Activity's completed UTC days. No account credentials, real usage data, pasteboard changes, or live provider probes were used.

Validation on head `8a15e450d9c6903e42f2b7ed7d7a5568377feea7`: `make test` passed all 1,180 selections in 107 groups on the first attempt, with zero failures, retries, or timeouts. `make check` and independent P0–P2 review passed. [CI35316060384](https://github.com/steipete/CodexBar/actions/runs/35316060384) passed both macOS shards, both Linux glibc architectures, the musl build, and lint. The full run also verifies the updated exact architecture anchor after the sanitizer moved. Thanks @Chipagosfinest for the original fix.

| Before | After |
| --- | --- |
| Before: provider totals survive, but all gateway models are absent | After: public GPT and Claude families appear under OpenRouter |

PR #3717, opened 2026-09-17
Stop discarding OpenRouter spend history when reasoning tokens exceed completion tokens

Chipagosfinest:
OpenRouter Activity responses can report a reasoning count above the completion count. CodexBar rejected that relationship in both the bundled JavaScript provider and the shared Swift snapshot mapper. Main dropped the optional spend window; removing only the JavaScript check, as the initial patch did, still failed the entire provider refresh at the mapper.

Both layers now preserve the reported reasoning counter independently. Token totals remain **input + output**. The fix retains safe-integer checks, numeric and aggregate-overflow validation, date/row/duplicate checks, cost accounting, and account routing. It neither clamps reasoning nor adds it to token totals. The plugin interface and provider documentation state this contract without claiming that reasoning is universally inside or outside completion upstream.

Validation:

- Original main fails the new regressions on both QuickJS and JavaScriptCore. The original PR's CI also demonstrates the second mapper rejection: https://github.com/steipete/CodexBar/actions/runs/35269894575
- `make test-fast FILTER='OpenRouter|ProviderPlugin|SpendDashboard|ShareStats|CLIOutputTests'`: 459 tests passed.
- New fixtures cover multiple days/endpoints, duplicate Activity responses, zero completion, absent/null/zero reasoning, malformed reasoning after a valid row, and independent reasoning aggregate overflow. Valid quota and balance survive invalid history.
- `make check`: zero formatting/lint violations. Independent review found no actionable P0–P2 findings.
- Signed isolated native proof runs the bundled plugin and Swift mapper with synthetic HTTP responses, then renders the production menu card and dashboard section. Dashboard JSON export preserves 523 input, 400 output, **923 total**, and **441 reasoning** tokens, with $0.875 vendor-reported spend and three requests. CLI rendering retains the $60 balance and key quota. No live provider account, saved configuration, or credential store was used.

The synthetic fixture reports completed UTC days. The dashboard's 29/30 coverage is expected because its displayed window includes the current, not-yet-reported day. Ordinary usage JSON intentionally omits live cost history; the machine-readable proof uses the dashboard exporter.

Before and after, light appearance:

Before: valid key quota and balance survive but Activity history is unavailable

After: Activity spend and input-plus-output token totals are retained

<details>
<summary>Dark appearance and restored dashboard counters</summary>

Before: unavailable Activity history in dark appearance

After: restored Activity history in dark appearance

Restored dashboard: input and output sum to 923 while reasoning remains a separate 441

Restored dashboard counters in dark appearance

</details>

The [official Activity API schema](https://openrouter.ai/docs/api/api-reference/analytics/get-user-activity) defines all three counters separately and supplies no subset constraint. Namespaced model labels on shared cards remain the independent scope of #3713.


jhairabet-udr:
Corroborating reproduction on two machines with a valid management key — so this is not a
credential or network problem:

| Platform | CodexBar | Observed |
|---|---|---|
| macOS Tahoe 26.6.1 | 0.60.5 | no 30-day row anywhere in the popover |
| Ubuntu 26.04, glibc 2.43 | 0.60.5 | `codexbar usage --provider openrouter --no-color` omits the `Last 30 days` line entirely |

Evidence that the management key *is* being read, not ignored:

* Without the key the CLI prints `Last 30 days: Unavailable right now · Management API key not configured`.
* With `OPENROUTER_MANAGEMENT_API_KEY` present in the environment (and, separately, with the key stored via Settings → Providers → OpenRouter on macOS), that line **disappears** — so `managementKeyConfigured` is true and the activity path is taken — but no spend history appears in its place.
* No degradation text is printed either: the text path shows neither the figure nor `Request failed`, so the failure is completely silent from the CLI.
* A direct `GET https://openrouter.ai/api/v1/activity` with the same key returns HTTP 200 with 30 rows, i.e. the payload is reachable and well-formed at the HTTP layer.

Account context: recent traffic is reasoning-heavy (Gemini 3.7/3.8 Flash through OpenRouter), which fits the `reasoning_tokens > completion_tokens` rows described.

+1 for not letting a single row discard the entire window — retaining partial history with an explicit partial marker would be far more useful than an empty result — and for surfacing the parser's real message instead of `Request failed` (#3718).

steipete:
Landed in 4a7658d9544c0a8acc685f2a37430e3916a34c5b. The OpenRouter parser and shared cost-history mapper now both retain independent reasoning counts, including counts above completion tokens. Total tokens remain input plus output, and quota/balance survive independently.

Validation: `make check` passed with zero violations; 459 focused tests passed; full `make test` passed all 1,176 selections across 107 groups without retries or timeouts. [CI for the reviewed head](https://github.com/steipete/CodexBar/actions/runs/35296889030) passed, and the landed tree matches that tested candidate. The attached signed native menu/dashboard captures and production dashboard JSON/CLI-renderer checks use controlled synthetic responses; they do not claim a live-account refresh.

The separate report about successful history having no CLI row is being investigated independently; this change fixes counter rejection. Thanks @Chipagosfinest for the report and original repair.

PR #3722, opened 2026-09-17
Add a daily spend trend to the shared usage card

Chipagosfinest:
Stacked on #3704, which restructured the hero this builds on. Merge that first.

## Why

The card showed a total and a ranking but no movement. Reviewing how metric cards are built
elsewhere, pairing a headline value with its trend is close to universal — PatternFly's dashboard
card taxonomy defines a "trend card" as *a current value plus a sparkline of its change over time*,
and the component libraries that implement this pattern (Neon `MetricCard`, Aura `DataCard`, the
shadcn stats blocks) all take a value plus an optional series. A total with no shape is the one
card type none of them ship.

The data was already there. `SpendDashboardModel.CurrencyGroup.dailyPoints` carries per-provider
cost per day; it just never reached the payload.

## Before / after

| | Before (#3704) | After |
|---|---|---|
| sparse | before sparse | after sparse |
| dense | before dense | after dense |
| multi-currency | before multi | after multi |

## What it adds

`ShareStatsPayload.dailySpend: [Double]` — one value per covered day for the headline currency,
summed across every provider on that day, oldest first. Aggregate by construction: it exposes
nothing the subscription rows do not already show, and carries no provider, model or account
identity, so the card's `LOCAL · AGGREGATE ONLY` claim still holds.

The sparkline is a plain `Path` — no charting dependency — with a filled gradient and a 2pt stroke
in the existing accent. No axes, no gridlines, no labels on the series itself: the hero already
states the period and the total.

## Two deliberate constraints

**The vertical budget does not move.** The hero's right column was using roughly 96pt of its 140pt,
so the label and the 30pt sparkline fit in existing slack. The dense render above confirms the card
still fits 630pt with the footer clear — which matters because SwiftUI does not clip an oversize
frame.

**It is suppressed rather than faked.** `spendTrend` returns empty for a single point or a flat
series, so no sparkline is drawn when there is no shape in the data. Inventing one would imply
movement that is not there.

## One correction worth recording

The first revision put the sparkline directly beneath `TRACKED TOKENS` with no label, which read as
a *token* trend when it plots spend. It now carries a `DAILY SPEND` eyebrow. The render is what
caught it.

## Tests

`daily spend trend aggregates every provider per day in chronological order` builds a
`CurrencyGroup` whose `dailyPoints` are deliberately out of order, with two providers sharing the
middle day, and asserts `payload.dailySpend == [1, 5, 5]` through `ShareStatsBuilder.make` — the
real builder, not a reimplementation of its logic in the test.

All 398 `ProviderArchitectureGatekeeperTests` anchors re-verified; the `ShareStatsPayload.swift`
Mistral anchor shifts 198 → 203 because the new field sits above it, and is updated here.


Chipagosfinest:
@clawsweeper re-review

Chipagosfinest:
@clawsweeper re-review

Chipagosfinest:
Cross-link: [#3747](https://github.com/steipete/CodexBar/pull/3747) restores retained top-model rows on the OG card when the USD group is incomplete. Independent of the sparkline; this branch can stay stacked on the spend-first layout. Live-checked on Homebrew 0.61.0.

@clawsweeper related ranking fix on current main — not a replacement for this trend PR.

Chipagosfinest:
Rebased onto main and fixed the CI failure — `ProviderArchitectureGatekeeperTests` was failing because the daily-spend-trend commits shifted `ShareStatsPayload.swift`'s codex-snapshot construct from line 365 to 391 (the allowlist pins by line number, so any insertion above it needs a re-anchor). Fixed and re-verified: all 67 tests in `ProviderArchitectureGatekeeperTests`/`ShareStatsTests`/`ShareStatsPeriodTests` pass locally, `swiftformat`/`swiftlint --strict` clean.

Also resolved the merge conflicts from rebasing past the recent OpenRouter/share-card work (#3696, #3704, #3747-train) — `ShareStatsPayload`'s `hasPartialModels` (from main) and `dailySpend` (this PR) now coexist in the same initializer.

@clawsweeper re-review

steipete:
Thanks @Chipagosfinest for the trend work and fixtures. Closing this proposal in its current form. The latest helper drops daily partial-cost information and fills missing fallback days with zero without proving coverage. Isolated checks reproduce both behaviors. A future trend should preserve dates, gaps, currency, and completeness from the dashboard model. The per-provider completeness fix from #3817 is already on main.

PR #3744, opened 2026-09-18
Add copy action for update unavailable reason

elijahfriedman:
## Summary
- Add a copy button next to the updates-unavailable notice in the About preferences pane, so users can quickly copy messages like "Updates managed by Homebrew. Run: brew upgrade --cask steipete/tap/codexbar" without selecting text manually.

## Changes
- [PreferencesAboutPane.swift](Sources/CodexBar/PreferencesAboutPane.swift): added an icon-only, borderless copy button next to `updater.unavailableReason`, reusing the existing `doc.on.doc` copy pattern from `PreferencesDebugPane.swift`.

## Verification
- `swift build` succeeds
- `make check` succeeds

## Screenshots
Screenshot 2026-09-18 at 2 00 39 PM


elijahfriedman:
already added recently

PR #3745, opened 2026-09-18
Smooth status group expand/collapse animation

elijahfriedman:
## Summary
- Smooth out expanding/collapsing a group in Codex's status section: the reveal now animates instead of snapping, and the chevron rotates in place rather than swapping icons.
- Clip the status list's hosting view so the one-frame gap between SwiftUI applying the expand/collapse and the row being re-measured no longer bleeds into the menu item below it.

## Changes
- `Sources/CodexBar/StatusComponentsMenuView.swift`: wrap group expand/collapse in `withAnimation`, rotate a single chevron glyph instead of swapping `chevron.right`/`chevron.down`, and fade the revealed child rows in/out via `.transition(.opacity)`.
- `Sources/CodexBar/StatusItemController+HostedSubmenus.swift`: enable layer masking on the status components hosting view to prevent transient overflow from bleeding into the next menu row during re-measurement.

## Verification
- `swift build`
- `make check` (SwiftFormat + SwiftLint --strict, 0 violations)
- `make test`

## Screen recording

https://github.com/user-attachments/assets/42aad4e1-b9bc-4fa2-986a-eaea8bb0122c


steipete:
Adopted the rotating chevron in #3806 with your contributor credit. Child geometry remains immediate because the native host measures once on the next run-loop turn; opacity transitions and clipping were omitted without proof that they preserve taller content. Landed on main via #3823. Thanks @elijahfriedman!

PR #3743, opened 2026-09-18
Add DeepSeek and OpenRouter widgets

brzvsk:
DeepSeek and OpenRouter are now selectable in configurable widgets. Their credit balances use the existing menu-bar formatting; DeepSeek has no invented quota bar, while OpenRouter retains a real configured key limit alongside its balance. The optional snapshot field keeps older widget snapshots readable.

Small widgets now show their update age consistently. The age uses WidgetKit's native relative-date text: installed testing found that formatting a string inside `TimelineView(.everyMinute)` could freeze between reloads, so that implementation was replaced.

### Validation

- `make check` and independent P0–P2 review passed.
- `make test`: all 1,245 selections across 114 groups passed, with no failures, retries, or timeouts.
- 79 focused provider, snapshot, architecture, and widget-layout tests passed, including capped/uncapped balances and older JSON snapshots.
- Matching Developer ID signed app and WidgetKit extension built successfully; packaged CLI/resource/launch probes passed without access to the build checkout.
- Installed WidgetKit proof exercised the provider picker, DeepSeek without a quota bar, OpenRouter with and without a cap, and small/medium/large layouts. The existing widget survived the upgrade with its identity intact.
- Native age advanced from **11 min, 59 sec** to **14 min, 12 sec** over 133 seconds while the main app was stopped, the extension PID stayed unchanged, and snapshot bytes and modification time stayed unchanged.
- [CI run 35477115895](https://github.com/steipete/CodexBar/actions/runs/35477115895) passed all checks on `3c9d9ea57f0b63200449cc1e53db19896652e469`. A lint-tool download hit a connection reset before lint ran; rerunning the failed jobs passed.

All screenshots use synthetic data. The test widget was removed, original preferences/snapshot/registrations restored, and the installed app relaunched after verification.

### Provider picker before and after

| Shipped picker | Updated picker |
| --- | --- |
| Shipped widget picker without DeepSeek or OpenRouter | Updated picker includes DeepSeek and OpenRouter |

### Balance and quota states

| Uncapped OpenRouter | Capped OpenRouter |
| --- | --- |
| Uncapped OpenRouter shows a 60 dollar balance | Capped OpenRouter shows 75 percent credits left and a separate 60 dollar balance |

### DeepSeek and elapsed-time proof

| First observation | 133 seconds later |
| --- | --- |
| DeepSeek balance with age 11 minutes 59 seconds | Same DeepSeek balance with age 14 minutes 12 seconds |

Thanks @brzvsk for the provider widget support.

PR #3748, opened 2026-09-18
Show reported provider spend in CLI dashboards

Chipagosfinest:
OpenRouter Activity already supplies reported spend in its usage snapshot, but the CLI dashboard emitted `cost: null` because its producer only projected the separate local-cost result. Dashboard-v1 now uses a provider's compatible 30-day USD aggregate when no local cost payload exists. Existing local results retain precedence, including an unavailable local amount.

Provider history keeps its own calendar meaning. Completed UTC daily buckets never become local Today merely because their date string matches: `todayUSD` stays null, known zero stays zero, and incomplete-request counts remain attached to the reported window. Other currencies or history lengths remain unavailable in the fixed USD/30-day fields.

### Validation

- `make check` and independent P0–P2 review passed.
- All 1,247 selections across 114 full-suite groups passed with no failures, retries, or timeouts.
- 27 focused tests passed under `TZ=America/Los_Angeles`, including the UTC/local rollover, known zero, missing totals, incompatible currencies/windows, incomplete history, and local-cost precedence.
- Built, signed CLI before/after proof used identical no-network synthetic provider fixtures in task-owned copies of the helper and its resource bundle. Real-home reads/writes and all network access were denied for the one-shot command. Both exited successfully: baseline `cost: null`; candidate `{"todayUSD":null,"last30DaysUSD":12.5}`.
- The authenticated loopback HTTP route returned the same result, and the real Chrome dashboard rendered **Last 30 days $12.50** with Today omitted. The fixture makes no provider API requests and uses no real credentials.
- [CI run 35484141518](https://github.com/steipete/CodexBar/actions/runs/35484141518) passed all nine checks on `a8ee4c977986afd3c2bf086a78e9f4672035585b`, including macOS, Linux, and GitGuardian.

### Live dashboard before and after

The screenshots contain only synthetic data.

**Before**

Dashboard without reported Activity spend

**After**

Dashboard showing reported 30-day spend

Thanks @Chipagosfinest for identifying the missing dashboard projection.


Chipagosfinest:
Live 0.61.0 check that motivated this: `codexbar usage --provider openrouter` already prints `Last 30 days (UTC): $… (reported)` plus Activity token/request/model counts. `codexbar dashboard` left `cost: null` for the same provider because only the local cost-command providers were attached.

@clawsweeper re-review

Chipagosfinest:
Extended past the CLI dashboard: Overview now merges live `costUsage` when the spend publication omitted OpenRouter, and widgets persist Activity 30-day cost/tokens. OpenRouter is selectable in the widget gallery.

Overlap with [#3743](https://github.com/steipete/CodexBar/pull/3743): that PR shows **credit balance**; this one shows **reported Activity spend** on Overview and widget cost metrics.

@clawsweeper re-review

Issue #3736, opened 2026-09-18
Claude segmented accounts: chips silently activate credentials and stay disabled during post-switch refresh (0.60.4)


pjmuller:
### Summary

After upgrading 0.55.0 → 0.60.4, Claude's three claude-swap accounts changed from stacked cards to account chips. Clicking a chip looks like inspecting quota, but actually activates that account. The menu then disables all chips and can remain on `Details for account B` / `Loading…` while account A stays highlighted for over a minute.

### Environment

- CodexBar 0.60.4 (Homebrew), claude-swap 0.26.0
- macOS 26.6.2 (25G83), Apple Silicon
- Three existing accounts; merged provider menu; default Segmented multi-account layout
- Separate Claude CLI configuration homes are also used; unintended default-account activation is particularly disruptive here.

### Reproduction (use disposable/test accounts: this DOES switch credentials)

1. Configure three claude-swap accounts and select Settings → Menu → Multi-account layout → Segmented.
2. Open Claude in the merged menu. Account A is active, all chips enabled.
3. Click inactive account B once.
4. All three chips immediately become disabled. The card shows B's quota and `Details for B` / `Loading…`, while A remains highlighted.
5. In the observed run this persisted for more than a minute. The default Claude identity had already changed to B before the menu stopped loading.
6. After manual Refresh and reopening the menu, B was active and all chips enabled. Waiting alone did not clear it during the observed interval; Refresh vs elapsed completion was not isolated because further activation trials would alter real account state.

Expected: a clear distinction between inspecting quota and activating credentials; if chips intentionally activate, make that explicit and prevent the UI from appearing stuck after activation while ambient quota refresh completes. Show truthful pending/error state.

### Source findings

#3498 (released in 0.57.0, fixing #3382) made Claude Swap honor the default Segmented preference. `ClaudeSwapAccountSwitcherView.select` disables every button; `handleClaudeSwapAccountSelection` activates an inactive activatable account. `UsageStore.switchClaudeSwapAccount` leaves the transient switching state set until `await refreshProvider(.claude)` returns. The displayed card follows the pending target, while the highlighted chip follows adapter `isActive`.

These findings explain the observed state; the precise reason for the slow post-switch refresh has not been isolated. No relevant ClaudeSwap source changes exist between v0.60.4 and main at 6cd74d02ede6be70dd1a42ae22dd8676f749ff5c (0.60.6 unreleased); 0.60.5's changelog does not list a fix.

### Workaround verified

[...]

Issue #2731, opened 2026-08-07
Bug: Claude: `cswap` with two account breaks bar indicators


Meldiron:
On the screenshot below, notice I have Claude enabled, and using `cswap` I connected two accounts.

Notice both account have some limit used, but still got some tokens remaining.

Finally, notice the small icon indicator in the MacOS top menu shows both bars empty.

<img width="694" height="1576" alt="Image" src="https://github.com/user-attachments/assets/05cb1cf6-4894-4157-b672-109f7405d0fe" />

Issue #1785, opened 2026-06-29
Claude OAuth: usage history stays on the previous account after switching accounts (preferredAccountKey is never repointed)


kes02:
### Summary
With Claude configured via **OAuth** (Claude Code login; no manually-added token accounts), switching the active Claude account leaves the menu's plan-utilization (usage) history pinned to the **previous** account. Relaunching CodexBar does not fix it; only manually clearing the on-disk history does — after which fresh usage lands in `unscoped` rather than a per-account bucket.

### Environment
- CodexBar 0.37.2 (Homebrew), macOS 14+
- Claude provider = **OAuth** only (no token/cookie accounts configured), `codex` provider disabled

### Steps to reproduce
1. Use Claude via an OAuth account **A** for a while so plan-utilization history accumulates.
2. Switch the active Claude account to a different OAuth account **B** (e.g. `claude logout && claude login` into B, or a new subscription that re-creates the Claude Code keychain item).
3. Open the CodexBar menu.

**Expected:** usage reflects account **B** (the current account).
**Actual:** usage still reflects account **A**. On disk, `~/Library/Application Support/com.steipete.codexbar/history/claude.json` keeps `preferredAccountKey` pointing at A's bucket, and B never gets its own keyed bucket.

### What I found (code reading — please correct me if I'm wrong)
Claude **OAuth** usage snapshots carry no identity — `ClaudeUsageFetcher.mapOAuthUsage(...)` sets `accountEmail: nil` / `accountOrganization: nil`. In `UsageStore+PlanUtilization.resolvePlanUtilizationAccountKey(...)`, with no token-account key and no identity key, resolution falls to `stickyPlanUtilizationAccountKey`, which returns the persisted `preferredAccountKey` as long as it still exists in `accounts`. Nothing repoints it when the active OAuth account changes, and the credential-swap cleaner (`clearClaudeCredentialDerivedStateForCredentialSwapNow`) does not touch `planUtilizationHistory[.claude]`. So the old account's bucket stays preferred, and new OAuth samples are written into it (or fall to `unscoped` once it is removed).

This looks **distinct** from #848 (Claude **web/sessionKey** multi-account org-selection, fixed on `main`) — this is OAuth-only and lives in plan-utilization account keying.

### A possible signal for a fix
[...]

PR #3750, opened 2026-09-19
Bound the ambient wait when confirming a claude-swap switch

keepitmello:
Fixes the "stuck after switching" half of #3736.

## What happens

`UsageStore.switchClaudeSwapAccount` holds the transient switch state until
`await refreshProvider(.claude)` returns. That refresh reads Claude Code's own
ambient credential, and when that read has no answer it does not come back at
all — so the switch never completes:

- the card stays on `Details for B` / `Loading…` while chip A stays highlighted,
- and every later click is dropped, because `handleClaudeSwapAccountSelection`
  returns early while `claudeSwapTransientState.task != nil`.

The segmented control is then inert until relaunch. The issue reported the first
symptom; the second is why waiting it out never helped.

Observed with two accounts, where the adapter had already answered and the menu
still never moved:

```
11:32:15.256  chip clicked
11:32:15.962  adapter --switch-to 2   0.00s
11:32:16.035  adapter --list          0.02s   activeAccountNumber already 2
11:32:25      menu still "Loading…", chip still on the old account
```

It is not the adapter: the same thing happens with `cswap` and with a different
executable that speaks the same `--list` / `--switch-to` schema and answers in
0.02–0.04 s.

## The change

The ambient wait is bounded instead of removed. `ClaudeSwapSwitchReconciliationTests`
pins the serialization contract ("ambient completion keeps switching serialized
until the real adapter list finishes"), and detaching the refresh entirely breaks
it — and would likely reopen the "previous account's data after switching" class
of bugs (#1785, #2731). Past the bound, reconciliation continues on the adapter
list and the ambient snapshot lands whenever it completes.

5 s is a first proposal; the value is easy to move if you would rather have it
shorter, configurable, or expressed as a switch-phase deadline.

## Verified

| | before | after |
|---|---|---|
| first switch confirmed | not within 10 s | 2.0–2.5 s |
| 3 consecutive switches | 2nd and 3rd clicks dropped | all three confirmed |

- `swift build`
- `swift test --filter ClaudeSwap` — 182 tests in 21 suites, all passing
- `make check` — 0 violations in 2417 files
- Packaged bundle on macOS 26 (Apple Silicon), CodexBar 0.61.1, two claude-swap
  accounts, segmented layout: three consecutive chip switches confirmed, adapter
  invocations logged for each.


steipete:
Maintainer review of `b626085d5df470f8ac5c9815a374aec6427a04af`: the proposed five-second timeout does not actually bound the wait. `withTaskGroup` waits for every child before its scope exits; cancelling the child awaiting `ambient.value` does not finish the unstructured ambient task. I reproduced the same structure with a 100 ms timeout and two seconds of synthetic ambient work: it returned after 2.005 seconds. No provider or Keychain access was involved.

There is a second reconciliation concern when repairing the timeout. Current provider refreshes wait for predecessor requests before starting, and the Claude adapter-list refresh is scheduled only after that start. A second switch while the first ambient refresh remains suspended can therefore reach reconciliation without obtaining its own fresh adapter list.

Please add deterministic coverage for timeout expiry while ambient work is still held, a second switch before the first refresh finishes, fresh adapter/account identity, and rejection of late results. The fix needs to preserve the serialized account-reconciliation contract from #3736/#3740. Keeping this open for that repair; the existing two-second success path does not exercise timeout expiry.

steipete:
Adopted with a bounded-join rewrite in #3812. The timeout ends only the ambient wait, while adapter reconciliation and serialized credential transactions remain intact. Synthetic regressions cover a suspended ambient probe, a later switch queued behind it, retained switch errors, and late ambient results. Thanks @keepitmello!

PR #3756, opened 2026-09-19
Add TypeSafe provider

jangisaac-dev:
## Summary

Adds **TypeSafe** (typesafe.ai) as a provider: credit balance, billing-cycle spend, plan, and active credit
expirations from the TypeSafe console session. Off by default; Chrome cookies or a manual `Cookie:` header.

- TypeSafe's public API (`/v1/systemone`, `/v1/models`) has no balance/usage endpoint, so the bundled `typesafe.ts`
  plugin reads the console's read-only `getBillingOverviewResult` Next.js server action (the same call the
  Settings → Billing page makes).
- **Trust/fragility boundary, up front:** the server-action ID is a build hash that changes on deploy. The plugin
  recovers it from the billing page's static `/_next/static/chunks/*.js`, caches it for 12 h in one retained plugin
  runtime, and rediscovers once when the server answers `404` + `x-nextjs-action-not-found`. If it cannot be found
  the fetch fails soft as `parseFailure`; no other action is ever invoked.
- No usage percentage, quota, or reset date is inferred: credit expiry and `resetsInDays` are different concepts.
- Signed-out requests are followed (same-origin) to `/login`; that page is recognised by its `(auth)/login` route
  segment and classified as `authenticationExpired`.
- **Menu UI matches other balance-only providers** (Moonshot / Manus / OpenRouter): header plan row shows
  `Balance $x`; Billing rows show cycle spend, plan, and non-zero credits with expiry. With the inline cost summary
  enabled, the `.payAsYouGoSpend` card shows spend + balance and the duplicate Spent row is hidden.

### Cookie handling (reworked after review)

TypeSafe no longer goes through the generic `ScriptFetchStrategy` / `ProviderPluginCookieBroker` path. It uses a
dedicated `TypeSafeWebFetchStrategy` that follows the existing Replicate pattern and runs the same bundled plugin
with a pre-resolved Cookie header:

- **Destination-scoped import (P1).** `TypeSafeCookieImporter` queries Chrome with `.exact` domain matching for
  `console.typesafe.ai` and `typesafe.ai`, then applies RFC 6265 domain-match against the only destination,
  `console.typesafe.ai`: host-only cookies must be exactly `console.typesafe.ai`; domain cookies must be
  `.typesafe.ai` or `.console.typesafe.ai`. Look-alike hosts (`nottypesafe.ai`, `typesafe.ai.example.org`),
  host-only apex/sibling cookies (`typesafe.ai`, `www.typesafe.ai`) and sibling domain cookies (`.www.typesafe.ai`)
  are dropped before anything is cached or sent.
- **Expired cached session recovery (P2).** Automatic mode tries the cached header first; on
  `authenticationExpired` it clears only the entry it observed (`clearIfCurrent`), imports fresh Chrome sessions,
  tries each candidate with `ProviderCandidateRetryRunner`, and stores the winner with
  `storeIfObservationCurrent`. Non-auth failures keep the cache and never import; an unreadable Keychain cache
  stops before importing; pinned entries (`stopFallback`) never fall back.
- **No ambient cookie jar (P1).** TypeSafe requests no longer use `ProviderHTTPClient.shared`. The strategy owns
  `TypeSafeWebFetchStrategy.isolatedTransport`: an ephemeral `URLSessionConfiguration` with
  `httpCookieStorage = nil` and `httpShouldSetCookies = false`, wrapped in the existing
  `ProviderHTTPClient.redirectGuardedSession` (same shape as Replicate). The selected Cookie header is therefore the
  only credential on the wire: nothing from `HTTPCookieStorage.shared` is attached, `Set-Cookie` responses are
  never persisted, and requests without an explicit header (static chunks) carry no Cookie at all.
- **Refresh control.** The cookie-source picker now has the standard `ProviderCookieRefreshAction`
  (Refresh + status text), same as OpenCode. Anyone who ran a pre-review build of this branch can click it once to
  replace a header cached by the old substring-matched import.
- The earlier shared broker change (`operationPreservingAccessContext(forwarding:)`) is reverted: with TypeSafe on
  its own strategy, no bundled plugin resolves browser cookies through the broker anymore, so it had no user.

## Commands run

- `swift test --filter 'TypeSafeCookieStrategyTests|ProviderArchitectureGatekeeperTests'` — pass (53 tests; 12
  `TypeSafeCookieStrategyTests`)
- `make check` — pass (0 SwiftLint violations in 2457 files)
- `make test` on `3d8495b3d` (includes the latest `main` merge) — pass (113/113 groups first pass, 0 failed, 0 retries)

New `TypeSafeCookieStrategyTests` cover, at the outgoing-request boundary with the real bundled plugin and a stub
transport:
- mixed Chrome records → only `console.typesafe.ai`-scoped cookies reach the billing GET and server-action POST
  (none of the excluded values appear; static chunk requests carry no Cookie);
- a cached header answered with `307` is cleared, a fresh browser session is tried, succeeds, and is stored;
- manual mode never reads/imports/writes the cache; non-auth failures, unreadable cache, pinned cache,
  concurrent replacement and cancellation paths.

## Verification notes

- Response shape, action discovery, stale-ID 404 behaviour, and the login landing were verified against a real
  free-tier account in a browser session before this revision. Only a free-tier account was observed; paid/expired
  credit entries are parsed defensively.
- **Live run of this revision** (debug `CodexBarCLI` built from `7f7c32a34`; later commits only add macOS guards, real free-tier account, Chrome session;
  amounts redacted). `cookie refresh` runs user-initiated with an explicit Keychain retry (Chrome Safe Storage
  access allowed); it imported the destination-filtered Chrome cookies and validated them, then `usage` read the
  cached header without importing again:

  ```
  $ codexbar cookie refresh --provider typesafe --allow-keychain-prompt
  typesafe: ✅ Browser cookie refreshed.
  $ codexbar usage --provider typesafe --source web --no-color
  == TypeSafe (web) ==
  Spent (September 2026): $x.xx
  Plan: Free
  Credit: x.xx of 5, expires Oct 19
  Plan: Balance: $x.xx
  ```

  Before the refresh, plain `codexbar usage` reported `No TypeSafe session cookies found`. That is expected: CLI
  usage runs in the background context and never prompts, and the fresh debug binary had no Safe Storage grant yet.
- **Final-I/O cookie isolation proof** (real `URLSession`, loopback HTTP server that sets
  `server_set=prior-account` on `/set` and echoes the received `Cookie` header; `HTTPCookieStorage.shared`
  pre-seeded with `ambient_jar=prior-account` to stand in for a previous account). Run as a throwaway test against
  `3d8495b3d`:

  ```
  control, default-config ProviderHTTPClient   /set then /echo           -> cookie=ambient_jar=prior-account; server_set=prior-account
  TypeSafeWebFetchStrategy.isolatedTransport   /set (account-A header)
                                               /echo, no Cookie header   -> cookie=<none>
                                               /echo, account-B header   -> cookie=session=account-B
  ```

  The control shows the leak the review described (shared-jar cookie plus a persisted `Set-Cookie`). The TypeSafe
  transport sends neither, and after an account switch only the newly selected header reaches the server. Because the
  transport keeps no cookie state, expired-session recovery (covered by the `307` test above) cannot keep a stale
  credential alive: once the cached header is cleared, the next request carries only the freshly imported header.
  `production transport never reads or stores ambient cookies` pins the configuration in the suite.
### Native UI, freshly built bundle (`3d8495b3d`)

Built and launched with `./Scripts/compile_and_run.sh`, real free-tier Chrome session, no account name, e-mail, cookie
or key visible in either shot.

Menu, TypeSafe tab — cycle spend, plan, credit with expiry, and the header `Balance` row:

TypeSafe menu

Settings → TypeSafe — provider toggle, `web` source, the Refresh control next to it, the same billing values, and the
`Billing` usage item:

TypeSafe settings

- Fixtures use fictitious IDs only.


jangisaac-dev:
@clawsweeper re-review

jangisaac-dev:
@clawsweeper re-review

mohitsbsftester:
I compared this PR with the separate Jev request/token PR #3777. This PR should be the single TypeSafe integration: it already has the correct provider identity, hardened cookie-domain filtering, isolated transport, cache/session handling, and runtime proof.

The only capability missing here is the Usage dashboard data from `GET https://console.typesafe.ai/api/usage?granularity=day`, which returns daily request, input-token, and output-token buckets. I recommend extending this TypeSafe provider with those detail rows rather than landing a separate Jev provider.

The endpoint was verified in a signed-in TypeSafe console session, but the response contract and reporting-period semantics should be covered by a redacted fixture/runtime proof before merge. I am closing #3777 as superseded and keeping #3776 open for this consolidated implementation.

jangisaac-dev:
## Authenticated run against a live account

Built `CodexBarCLI` from this PR head (`d341820fc`) and ran it against a
signed-in TypeSafe account, then read the console minutes apart. Below,
`codexbar` abbreviates `/tmp/ts-pr/.build/debug/CodexBarCLI`, built with
`swift build --product CodexBarCLI` in a worktree at `d341820fc`. CLI output is
reproduced unchanged; the diagnostic log at the end is excerpted and reformatted.
Account identifiers and API key ids are redacted.

### Plugin output

```console
$ codexbar cookie refresh --provider typesafe --allow-keychain-prompt
typesafe: ✅ Browser cookie refreshed.

$ codexbar usage --provider typesafe
== TypeSafe (web) ==
Spent (September 2026): $0.62
Plan: Free
Credit: 4.37 of 5, expires Oct 19
Plan: Balance: $4.37
```

The last line is the balance surfaced through `identity.loginMethod`, which the
CLI renders under its `Plan` label — a display quirk of this path, not a parse
result.

### Console — `console.typesafe.ai/settings/billing`, 2026-09-21T05:53Z

Credit Balance: **$4.37**

| Granted (UTC) | Type | Amount | Remaining | Expires (UTC) |
| --- | --- | --- | --- | --- |
| Sep 19, 2026 | Monthly credit | $5.00 | $4.37 | Oct 19, 2026 |

### Displayed values

| Field | Console | Plugin | |
| --- | --- | --- | --- |
| Balance | $4.37 | $4.37 | ✅ |
| Credit amount | $5.00 | 5 | ✅ |
| Credit remaining | $4.37 | 4.37 | ✅ |
| Credit expiry | Oct 19, 2026 | Oct 19 | ✅ month/day only |
| Plan | — | Free | plugin-reported; the billing page shows no plan label |
| Cycle spend | — | $0.62 | not reconciled, see below |

These compare rendered values. The CLI reads the Chrome profile and the console
reading came from another browser; I did not separately verify that both resolve
to the same account and organization, so the matching figures are consistent
with one account rather than proof of it.

### One figure I could not reconcile

The billing page never prints `spent`, so there is no like-for-like field to
compare. Two indirect readings disagree by about a cent:

- Credit drawdown: `$4.37` displayed implies a true remaining in
  `[4.365, 4.375)`, so consumed is `(0.625, 0.635]`.
- The plugin displays `$0.62`, implying a raw `spent` in `[0.615, 0.625)`.

Those intervals do not overlap, so rounding alone does not explain it. The
console's own spend widget reads `$0.6343` for Sep 19–21, but that is a
client-side estimate (it prints `$0.042/MTok input · Free output` beside the
figure) and a different quantity, so it does not settle the question either.
The cause is unverified — `spent` and the credit balance may simply be updated
on different schedules. Flagging it rather than claiming a match.

### Why the server action, not a REST endpoint

Captured the console's own network traffic to check whether the action could be
avoided. `/settings/billing` issues exactly two data requests: `GET
/api/usage?granularity=day` and a `POST` to `/settings/billing` itself — the
same server action this plugin invokes.

Probed twelve candidate REST paths (`/api/billing`, `/api/billing/overview`,
`/api/credits`, `/api/balance`, `/api/account`, `/api/organization`,
`/api/spend`, `/api/subscription`, `/api/plan`, …); all return 404 except
`/api/me`, which carries no billing fields. The billing page's initial HTML
contains none of `Credit Balance`, `Monthly credit`, or the credit amounts, so
there is no server-rendered payload to parse instead.

The server action is not a shortcut around a public API — it is how the console
itself reads this data.

### What this exercises

The request path this PR adds, end to end with nothing stubbed: Chrome cookie
import, the dedicated transport, server-action id discovery from the billing
page's static `/_next/static/chunks/*.js`, and the `getBillingOverviewResult`
response parsed by the bundled `typesafe.ts`.

The isolation properties themselves — destination-scoped cookie filtering,
ambient cookie storage disabled, redirects refused — are asserted by this PR's
tests, not observed by this run.

One security property *was* observed directly. Without
`--allow-keychain-prompt`, the same binary refuses the `Chrome Safe Storage`
read and reports no session rather than raising a background prompt:

```
keychain-preflight: service=Chrome Safe Storage
  Keychain preflight requires interaction for the current process
browser-cookie-gate: browser=Chrome
  Skipping background Chromium cookie import to avoid a Keychain prompt
Error: No TypeSafe session cookies found.
```

---

Separately: I have the `/api/usage` work ready (daily request and token rows,
closes #3776), plus a one-line `statusLinkURL` for `status.typesafe.ai`. The
usage read is a plain `GET` — no server action, no chunk scan — and it fails
soft without touching the billing snapshot. Happy to push it here, or keep it
as a follow-up once this lands. Your call; I did not want to add scope to a
branch you are actively rebasing.

PR #3758, opened 2026-09-20
fix(providers): avoid crashes on oversized usage values

steipete:
Large numeric responses could terminate CodexBar while decoding Kimi or MiniMax fields, converting Chutes or MiniMax durations, or formatting Chutes and Perplexity quotas. Replace trapping integer conversions with checked conversions or whole-number formatting. Preserve rounding, truncation, usable percentages, and known-window defaults; omit unrepresentable optional fields and nonfinite credit descriptions.

Synthetic regressions cover integer boundaries, duration overflow, fractional compatibility, all three Perplexity credit pools, and generic versus known Chutes windows. Kimi and Chutes tests reproduced process traps before the fix; seven isolated source probes also reproduced the affected conversions. No live accounts or Keychain reads were used.

Linux CI caught an integer-zero C vararg in the initial Perplexity formatter. Both arguments now have explicit Double types, with portable coverage for zero, fractional rounding, negative zero, large counts, and empty pools.

Validation: 110 focused parser tests and 29 follow-up native, plugin-parity, and portable tests passed. The final full local suite passed all 1,248 selections in 114 groups with zero retries/timeouts. Strict checks and independent review through P2 passed. CI passed for the final commit: https://github.com/steipete/CodexBar/actions/runs/35485515721.

ClawSweeper's supplementary review produced no verdict because its input scan stopped on an unchanged MiniMax endpoint-userinfo rejection fixture. Maintainer inspection confirmed a reserved example host and synthetic userinfo; the numeric diff does not alter that fixture. The independent review above is the completed code review for this patch.


steipete:
Landed on main as d987dd59ad3424ece6ef2a2d9d5914e64ac0eef2.

Kimi, Chutes, MiniMax, and Perplexity now handle the reproduced oversized-value cases without integer-conversion traps, retaining usable quotas and established rounding behavior. Linux CI caught the initial zero-vararg formatting error; the corrected Double arguments and portable regression passed on both Linux architectures.

Validation: 110 focused parser tests, 29 follow-up native/plugin/portable tests, strict checks, independent review through P2, and all 1,248 local selections in 114 groups passed with zero retries/timeouts. Final CI: https://github.com/steipete/CodexBar/actions/runs/35485515721.

ClawSweeper's supplementary review remained unavailable because of an unchanged synthetic endpoint-userinfo fixture; it did not produce a code verdict. The separate independent review completed successfully.

Issue #3663, opened 2026-09-15
Switching Claude Code accounts, dont refresh cached cookie from browser, and dont show current account usage


miseas:
I use 2 CC accounts in my daily basis, but when I switch accounts, the app will still show me the usage of the account not being used currently because it is using a cached web cookie. 

So, please:
1 - Allow me to refresh the cached cookie manually in settings, or at least delete/clean it.
2- When selecting "switching account..." from the options. That doesn't seem to do anything today, as it still keeps showing the previous account data.

miseas:
This is reproducible with:
APP: CodexBar 0.60.4
OS: macOS
Claude Code subscription account, i dont have API

ehmo:
Live repro of the Web-mode scenario, with the version and selected source the review asked for, plus the mechanism.

## Environment

- CodexBar 0.60.5 (build 148), installed binary dated 17 Sep
- macOS 26.6.1 (25G76), Apple silicon
- Claude Code 2.1.277, Max plan, two accounts (account A and account B below)
- Claude settings: Usage source `auto`, `claudeSwapEnabled: false`, `claudeWebExtrasEnabled: false`

## What happened

Claude Code was signed in as account A. CodexBar showed account B's session and weekly numbers, taken from a cached Safari cookie through the `web` source. Nothing on the card indicated the numbers belonged to a different account.

Widget snapshot, same machine, no change other than the Claude source:

| Source in effect | secondary usedPercent | secondary resetsAt |
|---|---|---|
| `auto`, serving account B | 100 | 2026-09-21T00:00:00Z |
| `cli`, serving account A | 25 | 2026-09-24T16:00:00Z |

The weekly window read as exhausted while account A's real weekly usage was 25 percent.

## Why Auto selected Web here

The `oauth` step could not run, so `auto` fell through to `web`:

- `~/.claude/.credentials.json` does not exist. Claude Code 2.1.277 keeps credentials only in the Keychain item `Claude Code-credentials`.
- That item's decrypt ACL names exactly one trusted application:

```
entry 0:
    authorizations (6): decrypt derive export_clear export_wrapped mac sign
    description: Claude Code-credentials
    applications (1):
        0: /usr/bin/security (OK)
            requirement: identifier "com.apple.security" and anchor apple
```

[...]

PR #3760, opened 2026-09-20
fix(providers): preserve organization and account identity

steipete:
Devin browser import could combine an organization name from one storage record with an unrelated internal ID from another. Keep each name/ID candidate together, including nested objects and direct storage-key metadata, and prefer a complete matching candidate for explicit organization selections.

Claude CLI Web enrichment could also accept an organization display name as account identity when either email was missing. Require matching nonempty emails for CLI enrichment, retaining known-mismatch rejection and the documented OAuth organization-UUID match. Primary source selection and credential behavior are unchanged. This addresses an optional-enrichment edge case related to #3663 without claiming to resolve its broader Auto/Web account-selection report.

Validation: the new tests reproduced 13 failed assertions on the previous code; all 87 focused tests now pass, including existing auth and metadata compatibility coverage. `make check` and independent review through P2 pass. The full local suite passed all 1,246 selections across 114 groups with zero retries or timeouts. Exact-head macOS and Linux CI passed: https://github.com/steipete/CodexBar/actions/runs/35484995914. All proof uses synthetic records and account metadata.


steipete:
Landed on main as a812379b923ad8c01f7d1c8eadade8fc028a2b00.

Devin metadata now keeps each organization's name and internal ID together. Claude CLI Web enrichment requires matching emails; the established OAuth UUID path remains supported.

Validation: 87 focused tests, `make check`, independent review through P2, and all 1,246 local selections in 114 groups passed with zero retries/timeouts. Exact-head macOS/Linux CI passed: https://github.com/steipete/CodexBar/actions/runs/35484995914.

The broader Auto/Web primary-account report in #3663 remains open; this fixes the independently reproduced optional-enrichment case.

PR #3761, opened 2026-09-20
perf(workspaces): lower memory when loading Codex history

steipete:
Workspaces cache loading retained every encoded SQLite usage payload while also constructing the decoded history. Decode each row as SQLite yields it, retaining typed rows and separate physical row counts. Preserve ordering, pricing, malformed-row aggregate fallback, timezone checks, and incomplete coverage. Full/raw reads and scanner receipt paths keep their existing behavior.

Metadata and rows stay within one consistent read transaction, with filesystem reconciliation after COMMIT. Decoding lengthens that transaction and can defer WAL checkpoints until the read finishes; concurrent-write coverage verifies snapshot consistency. A developer-only StoreStress fixture command creates an isolated synthetic corpus without real account or session discovery.

Optimized macOS harness, three repetitions per mode and revision:

| Synthetic history | Previous peak | New peak |
| --- | ---: | ---: |
| 256,000 rows | 186.6 MiB | 64.9 MiB |
| 1,000,000 rows | 726.4 MiB | 231.1 MiB |

Every read returned the expected files and rows. The million-row full-read control stayed at 1,276 MiB, lean total CPU time remained about 4.6 seconds, and retained memory stayed around 227 MiB. This is a measured temporary allocation improvement; the broader idle-growth report in #3323 remains open.

Validation: 179 combined focused cache/store/config tests, `make check`, independent review through P2, and the final full local suite passed all 1,250 selections across 114 groups with no retries or timeouts. Exact-head macOS/Linux CI passed: https://github.com/steipete/CodexBar/actions/runs/35492452136. An initial CI failure exposed the config-watcher race repaired separately in #3763; this branch includes that landed fix, and its comparison contains only the memory improvement. Runtime and benchmark sources are unchanged from the measured candidate.


steipete:
Landed as 2e2e4ff2521947355b821d88e4358c1239c075b9.

Optimized macOS StoreStress measurements used isolated synthetic corpora, three repetitions per mode/revision, and exact file/row-count checks. Lean peak memory dropped from 186.6 to 64.9 MiB at 256,000 rows and from 726.4 to 231.1 MiB at one million rows. The million-row full-read control remained around 1,276 MiB, lean total CPU time around 4.6 seconds, and retained memory around 227 MiB. The generator and measured runtime sources are preserved unchanged in this PR. This establishes a temporary peak reduction; #3323 remains open for broader idle growth.

Validation: 179 focused cache/store/config tests; `make check`; independent review through P2 with no actionable findings; and `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 CODEXBAR_ALLOW_TEST_KEYCHAIN_ACCESS=0 make test` passed all 1,250 selections across 114 groups without retries or timeouts. Exact-head macOS/Linux CI passed: https://github.com/steipete/CodexBar/actions/runs/35492452136.

The first macOS CI failure exposed a config watcher registration race, fixed and landed separately in #3763. After synchronizing that landed prerequisite, the final source head was re-reviewed and fully tested. No real account probes, credential-store reads, or Keychain access were used for proof. Main was fast-forwarded and verified clean.

PR #3763, opened 2026-09-20
fix(config): keep file watching across atomic replacements

steipete:
An atomic config replacement could leave the file watcher attached to an unlinked inode. It read the changed path immediately after resuming its DispatchSource, before kernel registration was guaranteed, and callback-triggered replacements could then produce no further notification.

Reconcile from the registration handler, compare the open descriptor's device/inode with the current path after callbacks, and rearm when they differ. Ignore handlers from retired sources. Missing-file directory watching and app-write suppression keep their existing behavior.

A deterministic test replaces the file after opening its descriptor but before registration, then replaces it again inside the callback. The previous code observed only the first value and failed; the fix observes the second value and a subsequent in-place edit. The existing atomic-replacement test exposed this while validating #3761.

Validation: all 23 sync/settings tests passed, `make check` passed, and independent review through P2 found no actionable issues. The full local suite passed all 1,249 selections across 114 groups with no retries or timeouts. The two native registration/rearming regressions also passed ten consecutive runs. Exact-head macOS/Linux CI passed in run 35490735450. Proof uses contained synthetic files, with no app launch, CloudKit network, or Keychain access.


steipete:
Landed as f1a9b53a3b1bd339c0b4527b2828b61986741bd0.

The new deterministic native-file test failed against the prior watcher: replacing the file between descriptor open and source registration, then again inside the callback, left only the first change observed. The fix detects the second replacement and a subsequent in-place edit. Registration/rearming regressions passed ten consecutive runs.

Validation: 23 focused sync/settings tests; `make check`; independent review through P2 with no actionable findings; and `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 CODEXBAR_ALLOW_TEST_KEYCHAIN_ACCESS=0 make test` passed all 1,249 selections across 114 groups with no retries or timeouts. Exact-head macOS/Linux CI passed: https://github.com/steipete/CodexBar/actions/runs/35490735450.

Proof used contained synthetic files, with no app launch, CloudKit network, real account probes, or Keychain access. Main was fast-forwarded and verified clean. The local GitHub wrapper supports squash merges only, so the dependent memory branch is being updated to the resulting main and revalidated.

PR #3767, opened 2026-09-20
fix(claude): avoid expiry diagnostic conversion crashes

steipete:
A parsed Claude credential can contain an expiry whose millisecond or remaining-second value exceeds the integer range. Credential loading builds diagnostic metadata unconditionally, so these conversions could terminate the app before normal handling continued. Exact-source synthetic reproductions confirmed the trap at the integer boundary and with larger positive/negative expiries.

Make only the diagnostic conversions failable and report unrepresentable fields as `out_of_range`. Keep parsing, expiry comparisons, truncation/rounding for normal dates, refresh decisions, ownership, and credential-access policy unchanged. Missing expiry remains distinct as `nil`.

Portable regressions call the parser and diagnostic method directly with synthetic data, covering normal/fractional timestamps, integer boundaries, nonfinite arithmetic, and missing expiry. No credential-store, Keychain, real account, or network access is involved. All 46 focused tests, `make check`, independent tracked-diff review through P2, and the full local suite passed. The full run completed all 1,251 selections across 114 groups without retries or timeouts. Exact-head macOS/Linux CI passed: https://github.com/steipete/CodexBar/actions/runs/35493722825.


steipete:
Landed as c1e063d1e06b5455e5cfa739a74b2226be7e5b28.

Exact original production-model reproductions parsed synthetic expiry values at the integer boundary and at 1e30/-1e30, then trapped while constructing diagnostic metadata. The repaired conversions report out_of_range independently for unrepresentable fields. Normal millisecond truncation, remaining-second rounding, missing-expiry diagnostics, parsing, expiry comparisons, and authentication/refresh decisions retain their behavior.

Validation: `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 CODEXBAR_ALLOW_TEST_KEYCHAIN_ACCESS=0 make test-fast FILTER='ClaudeOAuthDiagnosticsTests|ClaudeOAuthTests|ClaudeOAuthCredentialHistoryLinuxTests'` passed 46 tests; `make check` passed; independent tracked-diff review through P2 found no actionable issues; and the full `make test` run with the same isolation flags passed all 1,251 selections across 114 groups without retries or timeouts. Exact-head macOS/Linux CI passed: https://github.com/steipete/CodexBar/actions/runs/35493722825.

The new regressions call the parser and diagnostics directly with synthetic data. No real credential stores, Keychain queries, provider accounts, or network calls were used. Main was fast-forwarded and verified clean.

PR #3775, opened 2026-09-20
fix(codex): preserve background task ownership

steipete:
A cancelled Codex credits refresh could finish after its replacement started for the same account and clear the replacement's task handle. History backfills had the same race after awaiting history loading, and early cancellation left their completed handle behind. This broke refresh coalescing and prevented later cancellation from reaching the current task.

Give credits and history-backfill tasks separate request tokens, keeping account keys solely for account identity. Cleanup only clears the task it owns; history cleanup runs in `defer` across every exit. Existing refresh/publication behavior remains intact. The history tests use an observation hook plus a gated synthetic history store, and the affected settings fixtures use in-memory defaults.

Validation: the three new regressions fail on the previous implementation with six failed expectations, then pass with all 37 tests in the affected suites. `make check` passes and independent review through P2 has no actionable findings. Tests use synthetic accounts and stores with Keychain access suppressed. No live provider or native UI probe is needed for these scheduling races.


steipete:
Landed in 8971410f9350b8f2c7230daf81e13f206a27844a.

Validation:
- The three new deterministic scheduling regressions fail before the fix with six failed expectations.
- `CODEXBAR_SUPPRESS_TEST_KEYCHAIN_ACCESS=1 CODEXBAR_ALLOW_TEST_KEYCHAIN_ACCESS=0 make test-fast FILTER='CodexBackgroundRefreshCoalescingTests|UsageStorePlanUtilizationAsyncLoadTests'` — 37 tests pass.
- The same isolated test command with `FILTER='ProviderArchitectureGatekeeperTests'` — 41 tests pass after realigning the unchanged UsageStore source anchors.
- `make check` passes, and final independent review through P2 has no actionable findings.
- [CI run 35502499524](https://github.com/steipete/CodexBar/actions/runs/35502499524) passes on reviewed head b9d081becc7c9848f0948ccb4e94c12d60974112, including both full macOS test shards, Linux x64/arm64 and musl builds, lint, security, and the aggregate gate.

The scheduling proof uses synthetic accounts, gated loaders, in-memory settings, and isolated history files. It exercises the production task owners without a live provider or native UI probe.

PR #3708, opened 2026-09-17
fix(antigravity): fall back to agy usage report when OAuth quota endpoints return 403

Chener:
## What Changed

- `AntigravityOAuthFetchStrategy` now re-runs `agy -p /usage` in Auto mode when the account-scoped OAuth fetch identifies a selected Google account but returns no model quotas (for example, `retrieveUserQuota` answering `403`), and only when the resolved local `agy` login email matches that account. The report is attributed to the selected account with the `cli` source label; a missing email, a mismatch, a missing binary, or a failed report keeps the OAuth identity-only result (`Limits not available`).
- Added `AntigravityCLIIdentityResolver`, which resolves the local `agy` login email from the `gemini`/`antigravity` generic-password Keychain item, else from the first email found in `~/.gemini/antigravity-cli/settings.json`, `auth.json`, or `jetski_state.pbtxt`.
- Updated the **Usage source** settings subtitle, `docs/antigravity.md`, and `CHANGELOG.md` to describe the identity-match rule, and added `AntigravityIdentityFallbackTests` covering the matched-identity fallback and the unmatched-identity `Limits not available` path with a stubbed OAuth transport and a fake `agy` binary.

## Risk Assessment

🚨 High: No new findings, but the code is unchanged since the previous round and the declined error-level defects remain in place (CodexBarTests target cannot compile because MockUsageFetcher conforms to a concrete struct and MockClaudeFetcher omits three protocol requirements; the unconditional `import Security` breaks the always-on Linux CLI build; the CLI fallback attributes an identity-free `agy` report to the selected account based on a guessed local file), so merging still requires the owner's explicit sign-off.

## Testing

[...]

Issue #3649, opened 2026-09-14
Antigravity: explain Auto fallback when selected account blocks CLI usage report


raviteja7748:
## Summary

Auto mode falls back to account-scoped OAuth with no quota windows for a selected Antigravity Google account, even though explicit CLI mode can read quotas through the structured `agy` usage report added in 0.60.2.

## Environment

- CodexBar 0.60.2
- Antigravity CLI (`agy`) 1.2.2
- macOS
- Antigravity.app closed
- Antigravity source set to Auto with a Google account selected

## Reproduction

1. Refresh with Auto. The result selects `source: "oauth"`, includes account identity, and has no usage windows. Diagnostics show the legacy CLI HTTPS request failing with `Antigravity session expired`; the OAuth quota endpoint is not permitted for this account.
2. Refresh with explicit CLI source. `codexbar usage --provider antigravity --source cli --format json` returns weekly and five-hour quota windows through the structured `agy` usage report, without launching the desktop app.

## Expected

Please keep the selected-account boundary. If Auto cannot accept the CLI report because it has no identity, show a clear explanation and point users to the CLI source, or consider a way to verify the `agy` account before accepting its report. Right now Auto shows no limits while explicit CLI mode works.

## Context

Related to #3586 and #3607. The 0.60.2 CLI recovery works in explicit CLI mode. This request is about Auto-mode account handling and user guidance, not accepting identity-free reports without validation.

steipete:
Fixed in #3650, merged as d3610d095eede0eb46b7f149a078ff95de57a638. The Usage source picker now explains why Auto skips agy reports without account identity for selected or injected Google accounts, and names Local API / agy CLI with a warning that it may use a different signed-in account. The selected-account guard and fetch behavior remain unchanged.

Verified with 67 focused tests, signed synthetic rendering of the actual Settings row, the completed full local suite, and green hosted CI. The PR includes the before/after images and detailed proof, including recovered test-run issues.

PR #3780, opened 2026-09-20
feat(antigravity): fetch usage per saved account via scoped agy

### LINKED ISSUE #3662: Antigravity "Limits not available" bug.

Sogl:
## Summary

Saved Antigravity Google accounts show **"Limits not available"** in Auto mode:
the OAuth quota endpoint 403s for many accounts, and the `agy` usage report is
rejected by the selected-account guard because the report carries no account
identity.

This PR makes the `agy` usage report **account-scoped**: for each saved account,
CodexBar stages that account's credentials into an isolated `HOME` and runs
`agy /usage` under that scope. The report therefore belongs to the requested
account by construction — no ambient login switching, no shared Keychain
mutation, no interactive prompts during background refresh.

- Fixes #3662
- Removes the limitation documented in #3649 / #3650 (Auto skipping `agy`
  reports for selected accounts) instead of working around it
- Broader alternative to #3708, which falls back to the ambient `agy` report
  only when the ambient `agy` login matches the selected account — this covers
  **every** saved account, including ones not signed in anywhere else
- Follows the identity-aware direction outlined in #3673; also addresses the
  missing model-family rows from #2427 by routing through `agy`

## How it works

- `AntigravityAgyCredentialScope` builds a per-account isolated home
  (`~/.codexbar/antigravity/accounts/<id>/home`), stages the account's stored
  OAuth credentials where `agy` reads them, and serializes concurrent runs
  behind a mutation lock so parallel account refreshes can't interleave. The
  lock is acquired nonblocking (`flock LOCK_NB` + suspending retry), so waiters
  stay cancellation-aware and never pin a cooperative worker.
- The OAuth client used to mint/refresh credentials resolves from the installed
  `agy` binary first, falling back to the Antigravity IDE client when `agy` is
  absent — so newly added accounts mint credentials the Code Assist surface
  accepts. Existing saved credentials keep working: the staged scope accepts
  both clients, and token refresh uses the client fields stored in each
  credential file.
- `agy` stderr is classified: eligibility/TOS rejections surface as
  `cliAccountNotEligible` with an actionable message instead of a silent
  "Limits not available"; auth-required maps to the existing re-login guidance.
- Without `agy` installed, behavior is unchanged: app-local/ide-local probes +
  account-scoped OAuth + offline fallback, plus the same eligibility
  diagnostics.

## Safety / platform notes

- Ambient `agy` and IDE sessions are never touched — the scope is a directory
  of staged files, not a login change. Removing an account deletes its
  directory and tombstones the scope, so queued or in-flight refreshes cannot
  restage credentials for a removed account.
- Scoped runs are serialized (one `agy` process at a time), so a stacked
  four-account refresh takes tens of seconds instead of ~5s.
- macOS-only (`#if canImport(Darwin)`); Linux keeps the existing OAuth path.

## Test plan

- `make check` — clean, 0 violations.
- New/updated tests: `AntigravityAgyCredentialScopeTests` (15 — including a
  real-`flock` cancellation/serialization test, staged-home isolation, scope
  removal + tombstoning), `AntigravityScopedHomeRemovalTests` (1),
  `AntigravityUsageReportFailureTests` (4), plus resolver and
  eligibility-mapping cases in `AntigravityOAuthCredentialsStoreTests` /
  `AntigravityRemoteUsageFetcherTests`.
- `make test` — green except two pre-existing environment-dependent failures
  unrelated to this diff (pixel-tolerance `CostHistoryChartMenuViewTests`,
  LaunchServices-dependent `BrowserDetectionTests`).
- Verified live on macOS 15.x + `agy` 1.2.7: four saved accounts return weekly
  quota concurrently in stacked layout; ambient `agy` login unchanged; no
  Keychain prompts during background refresh.


Sogl:
@clawsweeper re-review

Both blocking findings addressed in `85d81c944`:

**[P1] Staged credentials outlive account removal** — fixed. `SettingsStore.removeTokenAccount` now calls `AntigravityAgyScopedHomeLifecycle.removeScope(accountKey:)` with the saved account's UUID, which deletes `~/.codexbar/antigravity/accounts/<id>/` under the mutation lock and tombstones the account key. The tombstone is written before the deletion queues on the lock, so a fetch waiting on the lock re-checks under the lock and fails closed instead of restaging; an in-flight scoped run finishes first, then the delete lands.

- `AntigravityScopedHomeRemovalTests` — settings → coordinator wiring (removal requests scoped-home deletion by account UUID).
- `removeScope deletes staged home and blocks restaging` — directory removed; `prepare` and `withPreparedScope` both reject afterwards.
- `queued staging cannot restage a removed account` — a fetch queued behind another scoped run sees the tombstone under the lock and never restages.

**[P2] Blocking flock in cooperative pool** — fixed. `AntigravityAgyCredentialMutationLock` now acquires with `flock(fd, LOCK_EX | LOCK_NB)` and suspends via `Task.sleep` on `EWOULDBLOCK`, so waiters never pin a cooperative worker while the holder awaits subprocesses, and cancellation abandons the wait promptly.

- `real mutation lock releases cancelled waiters and reacquires` — exercises the production flock lock (not the fake lock): a cancelled waiter finishes immediately instead of hanging in a blocked syscall, and a later waiter acquires the lock after the holder releases.

**Account binding / isolation evidence** — `staged home exposes only the scoped account token` proves the scoped `HOME` contains exactly one file: the `agy` file-token payload minted from the selected account's stored grant. The child env pins `HOME`/`PWD`/`SSH_TTY`, so `agy` selects file token storage rather than the ambient keyring — ambient credentials are never on a path the scoped `agy` reads. The existing grant check (`prepare replaces staged token from a different grant`) additionally guarantees a stale staged grant from different credentials is always replaced before any subprocess runs.

**Expiry/refresh** — `prepare keeps fresher staged token and replaces staler one` covers the expired-grant path (agy's own refreshed token for the same grant is kept; an older injected credential never clobbers it), and refresh uses the client fields stored in each credential file (covered in `AntigravityRemoteUsageFetcherTests`).

Sogl:
@clawsweeper re-review

Both Revision-2 findings addressed in `a50abdf49`:

**[P1] Share account retirement across processes** — retirement is now persisted on the shared filesystem. `removeScope` writes a marker at `accounts/.retired/<sanitized-key>` *before* queueing the directory deletion on the mutation lock, and `isScopeRetired` (consulted on the fast path and re-checked under the lock) treats the marker as authoritative. A separate `codexbar` CLI process that captured the account before removal and acquires the shared `flock` afterwards now hits the marker and fails closed instead of restaging the grant. Additionally, each process reconciles once under the lock before its first scoped operation: staged homes whose markers already exist are deleted, so a removal interrupted by a crash still completes rather than leaving reusable tokens on disk.

**[P1] Legacy refresh fallback** — the shared resolver is split. `resolvedClient()` keeps the agy-first order and now serves only new logins (which persist the minting client into the credential fields). The fetcher's refresh fallback uses the new `legacyRefreshClient()` = environment override → installed-app client, byte-identical to the pre-change resolution, so existing grants minted by the IDE client keep refreshing with their issuer.

**Coverage added:**
- `queued staging in another coordinator cannot restage a removed account` — two coordinators sharing one directory and one real `flock` file; the second coordinator's fetch queues behind the first's locked operation, removal lands mid-wait, the queued fetch is rejected before any I/O.
- `removeScope blocks restaging from another coordinator sharing the directory` — a second coordinator with an empty in-memory tombstone is blocked by the persisted marker.
- `first scoped operation sweeps staged home left by interrupted removal` — marker present, directory leftover; the next scoped operation in another coordinator completes the deletion under the lock.
- `remote refresh of legacy grant without client fields uses fallback resolver` — expired grant without stored `client_id`/`client_secret` refreshes with the resolver-supplied (legacy app) client.
- `legacy refresh client resolves installed app client` / `is nil without installed app` — resolver unit coverage via injected application roots.

Verified live: removing all four saved accounts produced four `staged home removed` events plus four `.retired` markers; re-adding created fresh UUIDs whose staging and reports are unaffected by prior markers.

Sogl:
Revision 3 follow-up — commit `b83ee6c89`

## [P1] Preserve legacy spawning when scoped reports are unsupported — fixed

`fetchUsingWarmSession` no longer returns early when the account-scoped report
fails. The scoped print is still attempted first; on any non-cancellation
error (agy < 1.1.11, unrecognized version string, subprocess failure) it falls
through to the existing `spawnFetch` path, which keeps its own CSRF-version
gate, and its result is still identity-validated by
`AntigravitySelectedAccountGuard` in `fetchWithReportFallback`. Cancellation
still propagates without spawning.

Regression tests:
- `account scoped fetch failure falls back to ambient spawn`
- `account scoped fetch cancellation does not spawn`

## Production evidence (macOS, agy 1.2.7, this build)

Redacted excerpts. Emails appear only as `accountFingerprint` hashes; UUIDs
are CodexBar-internal account keys.

### Separate process rejects a retired account before provider I/O

A retirement marker was written to `accounts/.retired/8A0DF9AE-…` (simulating
a removal performed by another process), then the packaged CLI ran in its own
process: `codexbar usage --provider antigravity --all-accounts --log-level debug`.
The CLI's independent coordinator observed the persisted marker and refused
staging — no token file was written, no `agy` was spawned for that account:

    accountKey=1F468146-… agy scope: token staged → prepared → report succeeded
    accountKey=8A0DF9AE-… agy scope: staging skipped, account scope removed
    accountKey=8A0DF9AE-… Antigravity account-scoped report failed:
        account scope was removed
    accountKey=55C7A5E4-… agy scope: token staged → prepared → report succeeded

The same run also exercised the new fallback chain end-to-end: the blocked
scoped report fell through to the ambient spawn, the CSRF-version gate
rejected it on agy 1.2.7, the ambient print report stayed suppressed
(`account-scoped data required`), and the account was served by the OAuth
strategy instead.

A first CLI run additionally swept the stale staged home left behind:
`accounts/8A0DF9AE-…/` was removed by the separate process via the
reconciliation pass, while `.retired/8A0DF9AE-…` persisted.

### Scoped fetch per selected account

Ambient IDE was not running and the warm `agy` scan missed, so each selected
account fetched through its own staged HOME:

    Antigravity warm path missed; running account-scoped print report expectedAccount=ee765f6fc6
    agy scope: waiting for mutation lock accountKey=8A0DF9AE-…
    agy scope: token staged accountKey=8A0DF9AE-… action=kept-fresher-staged
        tokenFile=~/.codexbar/antigravity/accounts/8A0DF9AE-…/home/.gemini/antigravity-cli/antigravity-oauth-token
    agy scope: prepared accountKey=8A0DF9AE-… home=~/.codexbar/antigravity/accounts/8A0DF9AE-…/home
    agy version probe scope=account-scoped version=1.2.7
    agy scope: finished accountKey=8A0DF9AE-…
    Antigravity account-scoped report succeeded accountKey=8A0DF9AE-… usage=windows=2 extra=2

Switching between three saved accounts produced three distinct fingerprints
(`ee765f6fc6`, `b24b095681`, `2bd5bfb1bf`), each staging only its own token
file under `accounts/<uuid>/home/`.

### Removal cleanup + persistent tombstone

    Token account removed count=3 provider=antigravity
    agy scope: staged home removed accountKey=6C053EE3-…

    $ ls ~/.codexbar/antigravity/accounts/.retired/
    6C053EE3-82FB-4622-82FC-44409E6D93D1   ← durable cross-process marker

### Fresh login

    Token account added count=2 provider=antigravity
    agy scope: token staged accountKey=6C053EE3-… action=written credentialsExpiry=…
    Antigravity account-scoped report succeeded accountKey=6C053EE3-…

All four re-added accounts staged and fetched on first refresh after login.

### Expired legacy grant

No live trace: reproducing it requires a pre-PR grant whose access token
expires on demand. Covered by tests instead:

- `legacyRefreshClient` resolves env → installed IDE client (main's order)
- `remote refresh of legacy grant without client fields uses fallback resolver`
  — an expired grant with no stored `client_id`/`client_secret` refreshes via
  the IDE client, not the agy-first login resolver

Grants carrying stored client fields still pin to them; the fallback order is
unchanged from main for grants that lack them.

@clawsweeper re-review

Sogl:
## Account-isolation proof: ambient B vs scoped A (live, this build)

Setup: ambient `agy` (`~/.gemini/antigravity-cli/antigravity-oauth-token`,
untouched) is logged into account B (`ag***@gmail.com`), which is also one of
the saved accounts. CodexBar has three saved accounts: `ps***`, `pa***`,
`ag***` @gmail.com.

**1. Ambient run** — plain `agy -p "/usage" --output-format json` under the
default HOME (account B):

    Gemini Models    Weekly Limit Remaining   0%   reset 2026-09-23T10:04:56Z
    Claude and GPT   Weekly Limit Remaining   0%   reset 2026-09-23T14:58:41Z

**2. Scoped run** — `codexbar usage --provider antigravity --all-accounts`,
each saved account through its own staged HOME, ambient session untouched:

    ps***@gmail.com:  Gemini 0% (resets 2d 18h), Claude/GPT 0% (resets 4d 13h)
    pa***@gmail.com:  Gemini 0% (resets 2d 19h) — no Claude/GPT bucket at all
    ag***@gmail.com:  Gemini 0% (resets 2d 15h), Claude/GPT 0% (resets 2d 20h)
                      — identical to the ambient B report above

**3. Staged credential identity** — `id_token` email claims differ per staged
home (emails masked):

    accounts/1F468146-…/home/…/antigravity-oauth-token → ps***@gmail.com
    accounts/55C7A5E4-…/home/…/antigravity-oauth-token → ag***@gmail.com
    accounts/8A0DF9AE-…/home/…/antigravity-oauth-token → pa***@gmail.com

If `agy` had silently consumed ambient credentials, the `pa***` report would
have returned B's two-bucket shape; instead it produced a different quota
structure for a different account, and `ps***` returned distinct reset times.
Ambient B stayed authenticated and unchanged throughout.

@clawsweeper re-review

steipete:
Thanks @Sogl for the scoped-account investigation. I recommend closing this implementation without merging. At b83ee6c89dc4, staging uses `~/.codexbar` rather than Application Support; scoped failures can fall back to an ambient spawn; raw CLI output is logged; and new-login OAuth client selection relies on binary-order heuristics. The implementation also lacks the claimed macOS-only execution gate. With net +928 production lines and +2,217 total lines, repairing these boundaries exceeds this bounded lane. #3662 remains open for a separately scoped implementation with private staging, an isolated child environment, fail-closed credential handling, no raw-output logging, and platform-specific proof.

